PR #4, opened 2016-09-25
Add directory to place common code and move trivial functions into it

rleon:
This is patch series which adds directory to place common code. 

Highlights of this this series:
1. The min/max functions were moved into it.
2. OCRDMA list implementation was chosen as common code.
3. Libmlx5 was converted to use that list.h.
4. Remove container_of and offset local declarations.


jgunthorpe:
Superseded by https://github.com/linux-rdma/rdma-core/pull/5

PR #7, opened 2016-09-28
Add a .travis.yml file

jgunthorpe:
Necessary to use the Travis CI service.

First draft, TravisCI should run this automatically from the pull request.


jgunthorpe:
This is a reasonable starting point now.

Todo:
  - Use -Werror once my warning fixes land and we have a warning free build. Want to keep it that way.
 -  Use a matrix config and run through clang and the 32 bit build as well.

PR #9, opened 2016-09-29
ccan adjustments and COPYING update

jgunthorpe:
The copying stuff was discussed on the list, we should get the 'default' copyright updated soonish.
COPYING.md was updated to include the new ccan stuff

The two ccan changes are minor.


jgunthorpe:
@dledford this branch now needs some updating to reflect changes from #11 - If you apply that first I will fix this up.

PR #16, opened 2016-10-07
Final clang support

jgunthorpe:
This is the last batch of stuff to be able to build with clang 3.9 and turn it on in travis.

I need to look at the travis results from this PR, I'll add a comment when I'm happy with that and it is safe to merge.


jgunthorpe:
Travis looks good, sadly this bloats the APT download from 80MB to 132MB - that is so incredibly wasteful, but that just seems to be how Travis wants to work :(

PR #19, opened 2016-10-13
Sparse fixes from Christoph Hellwig

jgunthorpe:
These are just the first 15 that made it to the list, will update this branch with the rest when they are sent.


jgunthorpe:
I think this is done for now

PR #23, opened 2016-10-19
libqedr pull request

ramranix:
This is a pull request for libqedr, the userspace library for qedr.

Thanks,
Ram

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -139,6 +139,13 @@ M:	Devesh Sharma <Devesh.sharma@broadcom.com>
 S:	Supported
 F:	providers/ocrdma/
 
+QEDR USERSPACE PROVIDER (for qedr.ko)
+M:	Ram Amrani <Ram.Amrani@cavium.com>
+M:	Ariel Elior <Ariel.Elior@cavium.com>
+S:	Supported
+F:	providers/qedr/
+P:	Doug Ledford <dledford@redhat.com>
+
 RXE SOFT ROCEE USERSPACE PROVIDER (for rdma_rxe.ko)
 M:	Moni Shoua <monis@mellanox.com>
 S:	Supported
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -144,7 +144,6 @@ M:	Ram Amrani <Ram.Amrani@cavium.com>
 M:	Ariel Elior <Ariel.Elior@cavium.com>
 S:	Supported
 F:	providers/qedr/
-P:	Doug Ledford <dledford@redhat.com>
 
 RXE SOFT ROCEE USERSPACE PROVIDER (for rdma_rxe.ko)
 M:	Moni Shoua <monis@mellanox.com>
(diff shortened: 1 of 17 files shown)


MAINTAINERS
```diff
@@ -139,6 +139,13 @@ M:	Devesh Sharma <Devesh.sharma@broadcom.com>
 S:	Supported
 F:	providers/ocrdma/
 
+QEDR USERSPACE PROVIDER (for qedr.ko)
+M:	Ram Amrani <Ram.Amrani@cavium.com>
+M:	Ariel Elior <Ariel.Elior@cavium.com>
+S:	Supported
+F:	providers/qedr/
+P:	Doug Ledford <dledford@redhat.com>
```
jgunthorpe: Drop the P line

providers/qedr/qelr.h
```diff
@@ -0,0 +1,320 @@
...
+
+#define ROCE_WQE_ELEM_SIZE	sizeof(struct rdma_sq_sge)
+
+#define QELR_RESP_IMM (RDMA_CQE_RESPONDER_IMM_FLG_MASK <<	\
+			RDMA_CQE_RESPONDER_IMM_FLG_SHIFT)
+#define QELR_RESP_RDMA (RDMA_CQE_RESPONDER_RDMA_FLG_MASK <<	\
+			RDMA_CQE_RESPONDER_RDMA_FLG_SHIFT)
+#define QELR_RESP_RDMA_IMM (QELR_RESP_IMM | QELR_RESP_RDMA)
+
+#define round_up(_val, _x) \
```
jgunthorpe: Why two round ups?

providers/qedr/qelr_main.c
```diff
@@ -0,0 +1,286 @@
...
+
+	dev->ibv_dev.ops = qelr_dev_ops;
+
+	return &dev->ibv_dev;
+}
+
+static __attribute__ ((constructor))
+void qelr_register_driver(void)
+{
+	ibv_register_driver("qelr", qelr_driver_init);
```
jgunthorpe: Consider using verbs_register_driver

providers/qedr/qelr.h
ramranix: Funny, I spent time looking for a CCAN alternative when it seems none of these are even used in our code. I will remove both.

MAINTAINERS
ramranix: OK. I just did what saw most vendors did.

providers/qedr/qelr_main.c
ramranix: I see that verbs_register_driver is relevant for verb extensions.
We currently don't have verb extensions. So I wonder if it is relevant at ll.
When and where are verb extensions being used? Why?

Thanks,
Ram

jgunthorpe: Right it is needed to access the extension stuff.

One one hand, it would be nice if all drivers used the new API and we deprecated the old one, for maintainability.

On the other hand if you use the new API then users must use the newer libibverbs too.

Up to you

PR #32, opened 2016-11-03
libpvrdma pull request

aditr:
This is a pull request for the userspace library for the PVRDMA device.

Thanks,
Adit


buildlib/fixup-include/rdma-pvrdma-abi.h
```diff
@@ -0,0 +1,297 @@
...
+	PVRDMA_WC_WITH_IMM		= 1 << 1,
+	PVRDMA_WC_WITH_INVALIDATE	= 1 << 2,
+	PVRDMA_WC_IP_CSUM_OK		= 1 << 3,
+	PVRDMA_WC_WITH_SMAC		= 1 << 4,
+	PVRDMA_WC_WITH_VLAN		= 1 << 5,
+	PVRDMA_WC_FLAGS_MAX		= PVRDMA_WC_WITH_VLAN,
+};
+
+struct pvrdma_alloc_ucontext_resp {
+	struct ibv_get_context_resp	ibv_resp;
```
jgunthorpe: Any files under buildlib must be an exact byte for byte copy of the kernel header. I do not expect the accepted kernel header will include this member ?

providers/pvrdma/cq.c
```diff
@@ -0,0 +1,287 @@
...
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
jgunthorpe: If you are able to, please use the exact text of the default copyright as seen in other files.
(eg COPYING.BSD_MIT)

I would like to avoid license proliferation if it is easy to do.

providers/pvrdma/cq.c
```diff
@@ -0,0 +1,287 @@
...
+		if (pvrdma_poll_one(cq, &qp, wc + npolled) != CQ_OK)
+			break;
+	}
+
+	pthread_spin_unlock(&cq->lock);
+
+	return npolled;
+}
+
+void __pvrdma_cq_clean(struct pvrdma_cq *cq, uint32_t qpn)
```
jgunthorpe: These __ prefixes are not necessary. The providers are built in a way that prevents all external symbols from being exported. I would recommend dropping this for consistency.

providers/pvrdma/pvrdma.h
```diff
@@ -0,0 +1,347 @@
...
+#include "pvrdma_ring.h"
+
+#ifndef rmb
+#  define rmb() mb()
+#endif
+
+#ifndef wmb
+#  define wmb() mb()
+#endif
+
```
jgunthorpe: No ifdefs please, these are unconditionally defined

providers/pvrdma/pvrdma.h
```diff
@@ -0,0 +1,347 @@
...
+	   _a > _b ? _a : _b; })
+#endif
+
+#ifndef min
+#define min(a,b) \
+	({ typeof (a) _a = (a); \
+	   typeof (b) _b = (b); \
+	   _a < _b ? _a : _b; })
+#endif
+
```
jgunthorpe: No, you have to use the ccan type safe versions of these macros.

providers/pvrdma/pvrdma.h
```diff
@@ -0,0 +1,347 @@
...
+	while (val < size)
+		val <<= 1;
+
+	return val;
+}
+
+#define to_vxxx(xxx, type)						\
+	((struct pvrdma_##type *)					\
+	 ((void *) ib##xxx - offsetof(struct pvrdma_##type, ibv_##xxx)))
+
```
jgunthorpe: Please use container_of for this sort of stuff.

providers/pvrdma/pvrdma_ring.h
```diff
@@ -0,0 +1,136 @@
...
+#define __PVRDMA_RING_H__
+
+#include <linux/types.h>
+
+#define PVRDMA_INVALID_IDX	-1	/* Invalid index. */
+#define atomic_read(_x)		*(_x)
+#define atomic_set(_x, _y)	(*(_x) = (_y))
+
+typedef uint32_t atomic_t;
+
```
jgunthorpe: If this actually needs to be atomic then we need to get some kind of atomic system (eg C11's atomic stuff)

If this isn't supposed to be multi-threaded safe then please don't call this stuff atomic.

providers/pvrdma/qp.c
```diff
@@ -0,0 +1,505 @@
...
+
+#include <infiniband/arch.h>
+
+#include "pvrdma.h"
+
+int pvrdma_alloc_qp_buf(struct pvrdma_device *dev, struct ibv_qp_cap *cap,
+			enum ibv_qp_type type, struct pvrdma_qp *qp)
+{
+	qp->sq.wrid = malloc(qp->sq.wqe_cnt * sizeof(uint64_t));
+	if (!qp->sq.wrid)
```
jgunthorpe: multiplication and malloc like this is a known possibly security issue, can you use calloc? Same comment for all places this pattern appears..

providers/pvrdma/verbs.c
```diff
@@ -0,0 +1,234 @@
...
+
+static int is_ipv6_addr_v4mapped(const struct in6_addr *a)
+{
+	return ((a->s6_addr32[0] | a->s6_addr32[1]) |
+		(a->s6_addr32[2] ^ htonl(0x0000ffff))) == 0UL ||
+		/* IPv4 encoded multicast addresses */
+		(a->s6_addr32[0] == htonl(0xff0e0000) &&
+		((a->s6_addr32[1] |
+			(a->s6_addr32[2] ^ htonl(0x0000ffff))) == 0UL));
+}
```
jgunthorpe: There is a glibc macro for this test?

jgunthorpe:
Also @dledford  will not merge this until he merges the kernel driver.. But go ahead and keep this pull request current.Thanks

buildlib/fixup-include/rdma-pvrdma-abi.h
aditr: Sorry this isnt an exact copy of the kernel header. Without more of the UAPI changes I dont think we can completely have an exact copy here. I can move this back into the providers/pvrdma for now (or drop the last commit) so that rdma-core builds fine.

providers/pvrdma/cq.c
aditr: Thanks for talking a look! I'll drop the __ and rename this since we already have a pvrdma_cq_clean function below which uses this.

buildlib/fixup-include/rdma-pvrdma-abi.h
jgunthorpe: Put copies of the things you need to change and rely on the unchanged kernel header for all the enums and everything else that does not need to be changed.

providers/pvrdma/pvrdma_ring.h
jgunthorpe: Christoph Hellwig has a prototype patch to add some generic atomic stuff, if you need that then perhaps send him a note..

providers/pvrdma/pvrdma.h
aditr: Ok will remove them.

providers/pvrdma/pvrdma_ring.h
aditr: This doesnt have to be atomic since we hold a CQ/QP lock before calling the ring functions. I'll remove these.

providers/pvrdma/cq.c
aditr: On taking a closer look I realized that the text is the FreeBSD variant of the OpenIB.org license (COPYING.BSD_FB). I can update the COPYING.md to reflect this.

rleon:
Adit,
Do you need commit history or it can be squaeshed and merged as one commit?

In case you want commit history, the pull request should be rebased by you and fixup commit should be integrated in original commits.

In case you don't need commit history, these fixups can be incremental.

providers/pvrdma/cq.c
jgunthorpe: The majority of the code uses the MIT version, for consistency it is best to use that if possible. This is true in the kernel as well.

aditr:
Leon,
I don't need the commit history. I anyways thought Doug would squash everything as a single commit if its is easier. If not let me know and I can rebase.

rleon:
Thanks @aditr 
We will squash the commits.

aditr:
Thanks Jason! There was a comment about changing the name of the library to libvmw_pvrdma (in line with the module name). Do you want me to add that to this pull request or after this has been merged?

jgunthorpe:
You need to follow the name with whatever Doug merges into the kernel. I assume the kernel patchset will be reposted and renamed. Just force-push an update here.

dledford:
On Mon, 2016-11-21 at 11:12 -0800, Jason Gunthorpe wrote:
> You need to follow the name with whatever Doug merges into the
> kernel. I assume the kernel patchset will be reposted and renamed.
> Just force-push an update here.

I think my preference would be vmw_rdma and libvmw_rdma.  The pv in
pvrdma is redundant since vmw stands for vmware and that implies a pv
driver.

-- 
              GPG KeyID: 0E572FDD

aditr:
Hi Doug,

> I think my preference would be vmw_rdma and libvmw_rdma.  The pv in
> pvrdma is redundant since vmw stands for vmware and that implies a pv
> driver.

Based on our internal discussion, we would like to keep the name as vmw_pvrdma and libvmw_pvrdma. That is our convention with our upstream paravirtual-only drivers, 
for example, vmw_pvscsi. Also, the official name of the device included in our recent 
hypervisor release is PVRDMA so we would like to maintain consistency here.

PR #45, opened 2016-12-10
i40iw: Fixes and optimizations

tatyana-en:


rleon:
Applied all patches except 8 and 9

PR #47, opened 2016-12-12
SoftRoCE updates

bvanassche:


bvanassche:
I will leave out the "Port to Perl 6" patch.

PR #57, opened 2017-01-19
Debian fixes 2

rleon:


rleon:
@jgunthorpe @bdrung, please take a look.

Thanks

debian/libibverbs1.symbols
```diff
@@ -1,6 +1,7 @@
 libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.0@IBVERBS_1.0 1.1.6
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
+ IBVERBS_1.3@IBVERBS_1.3 1.1.12
  ibv_ack_async_event@IBVERBS_1.0 1.1.6
```
jgunthorpe: The last PR changed the package version string for libibverbs, this .symbols update was generated based on the old version string, so it needs to be redone.

Be careful here since Doug merged the stuff for release 13 now, ideally build the .symbols for the 12 official release and then another update for 13.

debian/rules
```diff
@@ -57,10 +57,10 @@ override_dh_makeshlibs:
...
 # FIXME: should we just have a rdma-core-dbg package ?
 override_dh_strip:
-	dh_strip -plibibcm1 --dbg-package=libibcm1-dbg
-	dh_strip -plibibumad3 --dbg-package=libibumad3-dbg
-	dh_strip -plibibverbs1 --dbg-package=libibverbs1-dbg
-	dh_strip -plibrdmacm1 --dbg-package=librdmacm1-dbg
+	dh_strip -plibibcm1 --dbgsym-migration='libibcm1-dbg (<< 12-1~)'
+	dh_strip -plibibumad3 --dbgsym-migration='libibumad3-dbg (<< 12-1~)'
+	dh_strip -plibibverbs1 --dbgsym-migration='libibverbs1-dbg (<< 12-1~)'
+	dh_strip -plibrdmacm1 --dbgsym-migration='librdmacm1-dbg (<< 12-1~)'
```
jgunthorpe: This seems like a good idea, but this breaks building on Jessie:

```
dh_strip -plibibcm1 --dbgsym-migration='libibcm1-dbg (<< 12-1~)'
dh_strip: unknown option; aborting
debian/rules:60: recipe for target 'override_dh_strip' failed
make[1]: Leaving directory '/home/jgg/src'
make[1]: *** [override_dh_strip] Error 25
debian/rules:4: recipe for target 'binary' failed
```

Which means at a minimum this patch should update buildlib/cbuild to remove jessie and include some other acceptable Debian distribution for testing..

debian/libibverbs1.symbols
talatb: OK, Thank you for the update, will fix and reupload.

debian/rules
talatb: OK,
I will debug, fix and update.

talatb: Hi Jason, 
I'm trying to solve the issue without removing jessie, by adding if define statements in the debian/rules file.
 It is working, but I had to leave the debug packages in the debian/control file since the control file doesn't have option for ifdefs.
We can solve it by making the control file auto generated by creating it during the build time. 
Please let me know your thoughts on this, How important it is to support Jessie?

jgunthorpe: If you want to have compat, I would just eliminate dbg package generation entirely on old systems.

talatb: perfect, i just came to the same conclusion. 
Many thanks.

dledford:
It's not always clear from the conversation threads, but have your issues been addressed Jason?

debian/rules
bdrung: I recommend to keep the debug packages and stay with debhelper 9 to still support jessie (and xenial). I see no need to rush to use the dbgsym packages and debhelper 10.

bdrung:
I am fine with following commits:
* debian: Add lsb-base dependency for ibacm, iwpmd and srptools
* debian: Add source function library for srptools daemon
* debian: Switch to dpkg-source 3.0 (quilt) format

I would defer these commits to a later point when jessie is EOL:
* debian: Update to debhelper 10
* debian: Switch to automatically created -dbgsym packages.

talatb:
I'm fine with @bdrung suggestion, although I already sent a patches to @rleon  that fix the installation issue in Jessie.
just to be clear, we do this fixes according to the package review that we get in debian mentors.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850145

bdrung:
Sometimes Debian mentors disagree. :wink:

I will do the full package review and become your sponsor.

debian/libibverbs1.symbols
```diff
@@ -1,6 +1,8 @@
 libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.0@IBVERBS_1.0 1.1.6
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
+ IBVERBS_1.3@IBVERBS_1.3 1.1.12
+ IBVERBS_1.4@IBVERBS_1.4 12
```
jgunthorpe: These version numbers are all still wrong.

jgunthorpe:
@dledford Not yet, the symbol version file is still wrong at least... I think in github I am supposed to switch the review to 'Approve Changes' once it is OK..

jgunthorpe:
@bdrung @talatb whatever you two decide is fine with me regarding jessie compat.

One option is to have talatb carry the dbgsyms patch when the package is submitted to unstable and keep upstream more compat.

dh 10 vs 9 doesn't seem to make a difference for this package??

debian/libibverbs1.symbols
talatb: Already sent patch to @rleon  that should update the symbols files.

bdrung:
The difference between dh 9 and dh 10 has no significant effect on this package (we could drop `--with systemd` from debian/rules with dh 10, since this is the default there). See https://manpages.debian.org/unstable/debhelper/debhelper.7.en.html for details about the difference.

debian/libibverbs1.symbols
rleon: I don't know what went wrong with my pull request.

I will issue new, clean pull request.

rleon: @talatb I checked it and my pull request has the same patches as you sent them to me.

I'll send two pull request now:
one with approved patches by @bdrung 
* debian: Add lsb-base dependency for ibacm, iwpmd and srptools
* debian: Add source function library for srptools daemon
* debian: Switch to dpkg-source 3.0 (quilt) format

and the second with the rest patches

rleon:
@bdrung, @jgunthorpe, @dledford and @talatb 
https://github.com/linux-rdma/rdma-core/pull/66 with three approved patches

PR #60, opened 2017-01-24
A couple of fixes

ramranix:
libqedr: fix SQ/RQ wrap bug
	libqedr: allow post send/recv if QP is in error state

Thanks,
Ram


ramranix:
OK. Thanks.

PR #61, opened 2017-01-25
Timestamp support in mlx4 provider

yishaih:
This series implements the libibverbs 'Completion timestamping' API
in the mlx4 provider.


yishaih:
Merging the TS series used by the mlx4 provider.

PR #64, opened 2017-01-26
ibverbs: Define barriers for MIPS

bdrung:
Add barriers for MIPS and avoid build failures (FTBFS) in those
architectures where barriers haven't been defined. The warning message
was the behavior in previous releases of libibverbs.

This patch is part of the libibverbs 1.2.1-2 Debian package.

### DIFF
--- a/libibverbs/arch.h
+++ b/libibverbs/arch.h
@@ -127,9 +127,16 @@ static inline uint64_t ntohll(uint64_t x) { return be64toh(x); }
 #define wmb()	{ asm volatile("dsb st" ::: "memory"); }
 #define wc_wmb() wmb()
 
+#elif defined(__mips__)
+
+#define mb()   { asm volatile("sync" ::: "memory"); }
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Since fc645ff22e54361f809acc38d8f8e40035eda517 the build will work on any arch without barriers.

Did you see build failures on MIPS with that patch?

This patch is not really right because these barriers are supposed to provide DMA cache coherence, AFAIK no MIPS processor can do that, so it should follow the approach we used in ARM32 and have no barriers.

bdrung:
I haven't tried to build the package on MIPS yet, since I still do the package review (on amd64) and I don't own a MIPS system (but I could get access to a Debian porter box). libibverbs 1.2.1-2 shipped the patch and therefore I forwarded it to not accidentally lose it.

Let's assume that commit fc645ff will build on mips and close this pull request. I will open a new issue in case the package fails to build on mips (once the Debian package is ready).

PR #65, opened 2017-01-26
Make the provider facing libibverbs interface private

jgunthorpe:
The first part is to remove the header files that define the
prototypes for these symbols from the set of packaged headers.
This ensures that nothing can compile and use these symbols.

Next we move the symbols into a new symbol version stanza only
for the private ABI. This breaks every existing out of-tree provider,
but the earlier change to ibv_cmd_create_ah already did that.

There are a few symbols that are still private by virtue of not being
in public headers, but these are used internally by the other libraries.
For distribution sanity continue to treat them as public ABI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 set(PACKAGE_NAME "RDMA")
 
 # See Documentation/versioning.md
-set(PACKAGE_VERSION "12")
+set(PACKAGE_VERSION "13")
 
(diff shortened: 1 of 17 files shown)


jgunthorpe:
This now depends on and includes #70 as well

jgunthorpe:
Observe the discussion here ofiwg/libfabric#2684 we may want to preserve a compat ibv_register_driver symbol for libfabric usnic.

PR #67, opened 2017-01-29
Debian fixes approved

rleon:


rleon:
Travis CI PR passed, because it checks these actual commit, while Travis CI push checks together with my master which have different .travis.yml file.

PR #68, opened 2017-01-29
Debian fixes last

rleon:


debian/libibverbs1.symbols
```diff
@@ -1,6 +1,8 @@
 libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.0@IBVERBS_1.0 1.1.6
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
+ IBVERBS_1.3@IBVERBS_1.3 1.1.12
```
jgunthorpe: These are all still wrong.

rleon: @talatb I double and triple checked it and this patch exactly as you forwarded it to me.

jgunthorpe: FWIW, #65 will eliminate the need for this patch...

debian/libibverbs1.symbols
```diff
@@ -15,27 +17,33 @@ libibverbs.so.1 libibverbs1 #MINVER#
  ibv_cmd_alloc_pd@IBVERBS_1.0 1.1.6
  ibv_cmd_attach_mcast@IBVERBS_1.0 1.1.6
  ibv_cmd_close_xrcd@IBVERBS_1.1 1.1.8
- ibv_cmd_create_ah@IBVERBS_1.0 1.1.6
+ ibv_cmd_create_ah@IBVERBS_1.4 12
```
bdrung: This ABI breakage needs to be discussed and decided (see mailing list).

rleon:
Let's drop this series completely and wait till ABI discussion will be closed.
@bdrung, @jgunthorpe and @talatb what do you think?

bdrung:
Yes, I agree to wait for the ABI discussion to be closed before changing anything.

jgunthorpe:
The init.d script patch seems fine...

bdrung:
Right. Since both init scripts already depend on $network, adding a dependency to $remote_fs is safe.

bdrung:
Result from my side:

* debian: Update libibverbs symbols file -> wait for ABI discussion results
* debian: Update to debhelper 10 -> wait for end-of-life of jessie
* debian: Fix init.d script missing dependency on remote_fs -> accept
* debian: Switch to automatically created -dbgsym packages. -> wait for end-of-life of jessie

rleon:
@bdrung, I'll cherry-pick that init.d patch.
This pull request is closed.

Thanks

PR #70, opened 2017-01-30
Updates for versioning

jgunthorpe:
Teach travis to check that the shared library versions and their internal symvers match our internal scheme.

This is to prevent recent mistakes in missing the cmakefile update when changing the map file.

Also update the package version to 13.

### DIFF
--- a/buildlib/travis-build
+++ b/buildlib/travis-build
@@ -10,6 +10,7 @@ cd build
 # The goal is warning free compile on latest gcc.
 CC=gcc-6 CFLAGS=-Werror cmake -GNinja ..
 ninja
+../buildlib/check-build --src ..
 
 # .. and latest clang
 cd ../build-clang
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 set(PACKAGE_NAME "RDMA")
 
 # See Documentation/versioning.md
-set(PACKAGE_VERSION "12")
+set(PACKAGE_VERSION "13")
 
 #-------------------------
 # Basic standard paths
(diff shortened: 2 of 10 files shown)


rleon:
Why do we have both rdma-core.spec and redhat/rdma-core.spec?

jgunthorpe:
redhat/rdma-core.spec is used to build for rehl, centos and fedora. ./rdma-core.spec is used for everything else (eg suse).

I'd like to see proper suse packaging and then drop the ./rdma-core.spec...

rleon:
@jgunthorpe I have no clue who and when will do it :)

buildlib/travis-build
```diff
@@ -10,6 +10,7 @@ cd build
 # The goal is warning free compile on latest gcc.
 CC=gcc-6 CFLAGS=-Werror cmake -GNinja ..
 ninja
+../buildlib/check-build --src ..
```
rleon: Jason, can you please put this call all into .travis.yml?

It will give us clean structure of subscripts for travis.

Thanks

jgunthorpe: Not sure I understand? The script has to be run at the right time and from the right cwd which won't trivially happen if it is put someplace else... Seems frail to move it.

rleon: Something like that
```
diff --git a/.travis.yml b/.travis.yml
index c4fa1985..dae54a1b 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,3 +37,4 @@ addons:
 
 script:
   - buildlib/travis-build
+  - buildlib//check-build --src .
```

rleon: @jgunthorpe , i merged the code as is. it can be changed if needed in followup patch.Anyway it looks like the patches to build infrastructure are coming from two of us only.

PR #71, opened 2017-01-31
Debian packaging changes from libibverbs 1.2.1-2

bdrung:
A full review of the packaging differences between libibverbs 1.2.1-2 from Debian and the combined rdma-core package revealed some changes. These commits contain the changes that should be carried over to rdma-core.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -1,5 +1,7 @@
 #!/usr/bin/make -f
 
+export DEB_BUILD_MAINT_OPTIONS=hardening=+all
+
 %:
 	dh $@ --with systemd --builddirectory=build-deb
 
--- a/debian/control
+++ b/debian/control
@@ -15,6 +15,7 @@ Build-Depends: cmake (>= 2.8.11),
 Standards-Version: 3.9.8
 Vcs-Git: https://github.com/linux-rdma/rdma-core.git
 Vcs-Browser: https://github.com/linux-rdma/rdma-core
+Homepage: https://github.com/linux-rdma/rdma-core
 
 Package: ibacm
 Architecture: any
--- a/debian/control
+++ b/debian/control
@@ -29,7 +29,6 @@ Description: InfiniBand Communication Manager Assistant (ACM)
  A primary user of the ibacm service is the librdmacm library.
 
 Package: ibverbs-utils
-Section: net
 Architecture: linux-any
 Depends: ${misc:Depends}, ${shlibs:Depends}
 Description: Examples for the libibverbs library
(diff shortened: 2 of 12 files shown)


debian/libibverbs-dev.links
```diff
@@ -0,0 +1,16 @@
+usr/share/man/man3/ibv_alloc_pd.3		usr/share/man/man3/ibv_dealloc_pd.3
```
jgunthorpe: Why is the .links file necessary? The links are installed by cmake and dh does the right thing:

```
$ dpkg-deb -c ../libibverbs-dev_12-1_amd64.deb | grep -i alloc
-rw-r--r-- root/root       612 2017-01-31 09:27 ./usr/share/man/man3/ibv_alloc_pd.3.gz
-rw-r--r-- root/root       995 2017-01-31 09:27 ./usr/share/man/man3/ibv_alloc_mw.3.gz
lrwxrwxrwx root/root         0 2017-01-31 09:28 ./usr/share/man/man3/ibv_dealloc_mw.3.gz -> ibv_alloc_mw.3.gz
lrwxrwxrwx root/root         0 2017-01-31 09:28 ./usr/share/man/man3/ibv_dealloc_pd.3.gz -> ibv_alloc_pd.3.gz
```

If any are missing please update libibverbs/man/CMakeLists.

debian/rules
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/make -f
 
+export DEB_BUILD_MAINT_OPTIONS=hardening=+all
+
 %:
```
jgunthorpe: I'm nervous about this - libibverbs is a performance library, adding security check overhead is probably not desired. Do you know how much overhead there is?

bdrung: No, I don't know how much performance impact it has. Can somebody can measuring that?

debian/libibverbs-dev.links
bdrung: Yes, you are right. My bad. I checked that all symlinks are specified in libibverbs/man/CMakeLists. Thus I can drop that commit.

debian/rules
jgunthorpe: Well, looking at the list:
- HARDENING_FORMAT, we are already compile with -Wformat=2, so this does nothing
- FORTIFY - This  probably does nothing on this tree, but is performance-neutral
- STACKPROTECTOR - this has a runtime cost
- PIE - We don't have any performance critical executables, and libraries are already PIC, so this is neutral - however we should check that this flag is passed around properly in cmake, it is a bit tricky..
- RELRO - No performance cost for us
- BINDNOW - No performance cost for us

I looked into it some more and FC25 is enabling -fstack-protector-strong by default now. As is Centos 7 - so I think it is safe for debian to do it as well. If someone determines it is a problem we will have to handle it globally...

Can you check that pie worked properly?

jgunthorpe: pie is not really passed into cmake properly, it is applying to shared libraries too.. But that seems to be OK. or at least if it isn't Debian will have to fix debhelper or the compiler...

PR #72, opened 2017-01-31
Updates for GCC 7

jgunthorpe:
Hopefully we can enable this on travis soon..

gcc 7 has a few new warning classes that hit interesting things.

### DIFF
--- a/util/compiler.h
+++ b/util/compiler.h
@@ -37,4 +37,11 @@
 #define ALWAYS_INLINE
 #endif
 
+/* Use to mark fall through on switch statements as desired. */
+#if __GNUC__ >= 7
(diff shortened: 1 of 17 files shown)


util/compiler.h
```diff
@@ -37,4 +37,11 @@
 #define ALWAYS_INLINE
 #endif
 
+/* Use to mark fall through on switch statements as desired. */
+#if __GNUC__ >= 7
```
rleon: I think that clang supports it too.
https://clang.llvm.org/docs/AttributeReference.html

jgunthorpe: Confusingly clang only supports this in C++ mode.

PR #74, opened 2017-02-05
[v2] ibverbs: remove unused ibv command declarations

ramranix:
Remove the IBV command declarations ibv_cmd_query_gid and
ibv_cmd_query_pkey as they are not implemented.

### DIFF
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -1,5 +1,5 @@
 Name: rdma-core
-Version: 12
+Version: 13
 Release: 1%{?dist}
 Summary: RDMA core userspace libraries and daemons
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 set(PACKAGE_NAME "RDMA")
 
 # See Documentation/versioning.md
-set(PACKAGE_VERSION "12")
+set(PACKAGE_VERSION "13")
 
 #-------------------------
 # Basic standard paths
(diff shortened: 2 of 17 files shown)


redhat/rdma-core.spec
```diff
@@ -118,6 +118,7 @@ Device-specific plug-in ibverbs userspace drivers are included:
 - libnes: NetEffect RNIC
 - libocrdma: Emulex OneConnect RDMA/RoCE Device
 - librxe: A software implementation of the RoCE protocol
+- libqedr: QLogic QL4xxx RoCE HCA
```
rleon: Please go one step further and add all missing drivers.

Thanks

jgunthorpe: And keep the list sorted please

ramranix:
I'll place libqedr in alphabetical location.
But I will leave the other vendors to add their own lines as I don't know
what is their supported HW.

Thanks,
Ram

rleon:
You can find it easily by checking their kernel code and looking in their cover letters.

Please invest time and stop throwing this shit on me and Jason. We are not alone here.

ramranix:
I appreciate your work and don't take it for granted.
I'm glad to help and take part in the effort and I'm sure that my
involvement will increase over time.
I'm not "throwing shit around" and would appreciate a lower tone.

The description we entered for qedr is not the one that appears in our
cover letter and that is why I reckoned that hns/vmw_pcrdma would like to
add it themselves.

Still, I understand that you want to lower overhead. So I'll grab what I
can from their cover letters.

Thanks,
Ram

PR #76, opened 2017-02-06
remove unused declarations; add providers to redhat's spec

ramranix:
Please pull the following commits:
1. Remove unused declarations
2. Add providers to redhat's rdma-core.spec - libqedr, libhns and libvmw_pvrdma

Thanks,
Ram


rleon:
Thanks

PR #81, opened 2017-02-16
Run a debian package build as part of travis

jgunthorpe:
Requested by @rleon. This will help detect obvious missed things in patches - eg symbol version changes, new header files, shlibs, etc.

@talatb @bdrung

### DIFF
--- a/buildlib/travis-build
+++ b/buildlib/travis-build
@@ -5,19 +5,15 @@ set -e
 # Echo all commands to Travis log
 set -x
 
-mkdir build build-clang build32 build-no-dma
-cd build
-# The goal is warning free compile on latest gcc.
-CC=gcc-6 CFLAGS=-Werror cmake -GNinja ..
+mkdir build-clang build32
+
+# Build with latest clang first
+cd build-clang
+CC=clang-3.9 CFLAGS=-Werror cmake -GNinja ..
 ninja
 ../buildlib/check-build --src ..
 
-# .. and latest clang
-cd ../build-clang
-CC=clang-3.9 CFLAGS=-Werror cmake -GNinja ..
-ninja
-
-# 32 bit build
+# 32 bit build to check format strings/etc
 cd ../build32
 # travis's trusty is not configured in a way that enables all 32 bit
 # packages. We could fix this with some sudo stuff.. For now turn off libnl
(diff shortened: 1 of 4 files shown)


debian/rules
```diff
@@ -69,3 +69,5 @@ override_dh_strip:
 # Upstream encourages the use of 'build' as the developer build output
 # directory, allow that directory to be present and still allow dh to work.
 .PHONY: build
+build:
```
bdrung: Why?

buildlib/travis-build
```diff
@@ -26,8 +22,23 @@ ninja
...
+
+# When running cmake through debian/rules it is hard to set -Werror,
+# instead force it on by changing the CMakeLists.txt
 cd ..
+cp util/udma_barrier.h.old util/udma_barrier.h
+echo 'set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")' >> buildlib/RDMA_EnableCStd.cmake
+sed -i -e 's/-DCMAKE_BUILD_TYPE=Release//g' debian/rules
+sed -i -e 's/ninja \(.*\)-v/ninja \1/g' debian/rules
+
+CC=gcc-6 debian/rules build
```
bdrung: Please use the normal utilities, i.e. dpkg-buildpackage from dpkg-dev:
```
dpkg-buildpackage -uc -us -b
```

buildlib/travis-build
```diff
@@ -26,8 +22,23 @@ ninja
...
+# good clue if patches are changing packaging related things, the RPM stuff
+# will have to be audited by hand.
+
+# When running cmake through debian/rules it is hard to set -Werror,
+# instead force it on by changing the CMakeLists.txt
 cd ..
+cp util/udma_barrier.h.old util/udma_barrier.h
+echo 'set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")' >> buildlib/RDMA_EnableCStd.cmake
+sed -i -e 's/-DCMAKE_BUILD_TYPE=Release//g' debian/rules
+sed -i -e 's/ninja \(.*\)-v/ninja \1/g' debian/rules
```
bdrung: Why do you adjust the ninja call?

debian/rules
jgunthorpe: The phony line causes the pattern rule at the top to not match 'build' so 'make -f debian/rules build' did nothing. Adding an empty edge causes the pattern rule to match again.

Weird quirk of make.

buildlib/travis-build
jgunthorpe: So the travis log output from the package build matches the rest of the log output, we do not need to log command lines when running travis as anyone can reproduce the command lines locally.

buildlib/travis-build
jgunthorpe: Hm, I ran it that way because I thought dpkg-buildpackage scrubbed the environment and would eat my CC=, but that is not true.

The downside is that dpkg-buildpackage does some unnecessary work for the travis environment, do you think it is worth changing?

debian/rules
bdrung: Oh, you are right. I wasn't aware that the phony target do not consider implicit rules (like pattern rules). Then I recommend to move the .PHONY and build target up (just below the pattern rule).

buildlib/travis-build
bdrung: What kind of unnecessary work?

jgunthorpe: IIRC it does a few extra clean steps, it does some other dpkg commands like genchangelog. I didn't benchmark it in travis to try and see, that is a bit painful.

PR #82, opened 2017-02-19
libqedr: Refactor post_send

ramranix:
Refactor post send by creating a function that posts a single WR
that is invoked from a loop, early all WRs validity checks so that
WR posting won't even start if it is invalid, and move the edpm
structure outside of the QP structure as a preparation for dpm.

Signed-off-by: Ariel Elior <Ariel.Elior@cavium.com>

### DIFF
--- a/buildlib/check-build
+++ b/buildlib/check-build
@@ -146,7 +146,7 @@ def test_public_headers(args):
     with private_tmp() as tmpd:
         with open(os.path.join(tmpd,"build.ninja"),"wt") as F:
             print >> F,"rule comp";
-            print >> F," command = cc -c -I %s $in -o $out"%(incdir);
+            print >> F," command = cc -Werror -c -I %s $in -o $out"%(incdir);
             print >> F," description=Header check for $in";
             count = 0;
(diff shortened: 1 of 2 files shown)


providers/qedr/qelr_verbs.c
```diff
@@ -1127,252 +1130,273 @@ static void doorbell_edpm_qp(struct qelr_qp *qp)
...
-		if ((qelr_chain_get_elem_left_u32(&qp->sq.chain) <
-					QELR_MAX_SQ_WQE_SIZE) ||
-		     (wr->num_sge > qp->sq.max_sges)) {
-			status = -ENOMEM;
-			*bad_wr = wr;
-			break;
-		}
+	if ((wr->send_flags & IBV_SEND_INLINE) &&
+	    (data_size > ROCE_REQ_MAX_INLINE_DATA_SIZE)) {
+		DP_ERR(stderr, "Too much inline data in WR: %d\n", data_size);
```
rleon: I think that you should print the error to cxt->dbg_fp and not to stderr

PR #87, opened 2017-02-28
Cleanup ibv_register_driver

jgunthorpe:
Now that the provider API is private there is no reason to have two different ways to register a driver. Consolidate everything into verbs_register_driver.

This approach is simplified by not also forcing a change to the new context allocation protocol, which is a bit more complex.


rosenbaumalex:
nice and quick move to verbs_device
:+1:

PR #91, opened 2017-03-09
Fix an endianness bug in srp_daemon

bvanassche:
Additionally, make it possible for sparse to detect and report endianness bugs when building with sparse and -D__CHECK_ENDIAN__.


libibverbs/examples/devinfo.c
```diff
@@ -592,7 +592,7 @@ int main(int argc, char *argv[])
 			{ .name = "ib-port",  .has_arg = 1, .val = 'i' },
 			{ .name = "list",     .has_arg = 0, .val = 'l' },
 			{ .name = "verbose",  .has_arg = 0, .val = 'v' },
-			{ 0, 0, 0, 0}
+			{ .name = NULL, 0, 0, 0 }
 		};
```
jgunthorpe: I'm fine with it like this but simply '{}' would be fine in both places too.

librdmacm/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ publish_headers(rdma
   )
 publish_headers(infiniband
   ib.h
+  endian.h
   )
```
jgunthorpe: I don't think a new public header is a very good idea.. Is it *really* necessary?

eg it seems like a really bad idea to put common names like  be16_to_cpu into a  public header.

libibumad/umad.h
```diff
@@ -225,14 +224,14 @@ static inline void umad_free(void *umad)
...
 /* Users should use the glibc functions directly, not these wrappers */
-static inline __attribute__((deprecated)) uint64_t htonll(uint64_t x) { return htobe64(x); }
-static inline __attribute__((deprecated)) uint64_t ntohll(uint64_t x) { return be64toh(x); }
-#define htonll htonll
-#define ntohll ntohll
+static inline __attribute__((deprecated)) uint64_t cpu_to_be32l(uint64_t x) { return cpu_to_be64(x); }
+static inline __attribute__((deprecated)) uint64_t be32_to_cpul(uint64_t x) { return be64_to_cpu(x); }
+#define cpu_to_be32l cpu_to_be32l
+#define be32_to_cpul be32_to_cpul
 #endif
```
jgunthorpe: These have to stay the way they were... They are compat stuff

libibcm/cm.h
```diff
@@ -325,8 +325,8 @@ struct ib_cm_attr_param {
 int ib_cm_attr_id(struct ib_cm_id *cm_id,
 		  struct ib_cm_attr_param *param);
 
-#define IB_CM_ASSIGN_SERVICE_ID_MASK htobe64(0xFF00000000000000ULL)
-#define IB_CM_ASSIGN_SERVICE_ID      htobe64(0x0200000000000000ULL)
+#define IB_CM_ASSIGN_SERVICE_ID_MASK cpu_to_be64(0xFF00000000000000ULL)
+#define IB_CM_ASSIGN_SERVICE_ID      cpu_to_be64(0x0200000000000000ULL)
```
jgunthorpe: I'm really not comfortable with using such common names in public header files.

Is there some sneaky/hacky way you could teach sparse about the glibc macros? Maybe force sparse to include your own endian.h instead of glibc's with the proper annotations?

libibverbs/examples/devinfo.c
bvanassche: Just changed the new code into "{}".

librdmacm/CMakeLists.txt
bvanassche: I have reworked <librdmacm/endian.h> such that it redefines the glibc macros instead of introducing new conversion functions.

libibumad/umad.h
bvanassche: That was unintended. Should have been fixed in the latest version of this patch series.

libibcm/cm.h
bvanassche: I have reworked <infiniband/endian.h> such that it redefines the glibc macros instead of introducing new conversion functions. Additionally, all #include <infiniband/endian.h> directives have been changed into #include <endian.h>.

jgunthorpe:
Looks good Bart, I made some changes for you - if you want you can inline them into your series as appropriate:

jgunthorpe@ba5d3c5f1b37c0f9d3ea1db97e40ef1b587fd312

Mostly makes sparse function automatically via cmake stuffery

PR #95, opened 2017-03-13
Barriers

yishaih:
This series comes to optimize both mlx4 and mlx5 as some follow-up of previous
barrier series that was already merged.


jgunthorpe:
Please replace patch 1 with the version I sent to the list, otherwise looks fine. Thanks for looking at this.

PR #97, opened 2017-03-14
Sparse bugs

jgunthorpe:
Now that Bart got sparse working these little bugs were detected by sparse.

They change behaviour so I think Sean/etc should look at rdma cm bits.

### DIFF
--- a/ibacm/prov/acmp/src/acmp.c
+++ b/ibacm/prov/acmp/src/acmp.c
@@ -656,7 +656,7 @@ acmp_record_mc_av(struct acmp_port *port, struct ib_mc_member_rec *mc_rec,
 	dest->path.dgid = mc_rec->mgid;
 	dest->path.sgid = mc_rec->port_gid;
 	dest->path.dlid = mc_rec->mlid;
-	dest->path.slid = htobe16(port->lid) | port->sa_dest.av.src_path_bits;
+	dest->path.slid = htobe16(port->lid | port->sa_dest.av.src_path_bits);
 	dest->path.flowlabel_hoplimit = htobe32(sl_flow_hop & 0xFFFFFFF);
 	dest->path.tclass = mc_rec->tclass;
 	dest->path.reversible_numpath = IBV_PATH_RECORD_REVERSIBLE | 1;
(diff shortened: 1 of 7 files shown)


jgunthorpe:
@shefty, can't put you on the review list for some reason.. Please take a look Thnks

shefty:
I may need to be added to at least the reader list for the group to be added to any list.

ibacm/prov/acmp/src/acmp.c
```diff
@@ -2419,8 +2419,8 @@ static struct acmp_port *acmp_get_port(struct acm_endpoint *endpoint)
 	struct acmp_device *dev;
 
 	acm_log(1, "dev 0x%" PRIx64 " port %d pkey 0x%x\n",
-		endpoint->port->dev->dev_guid, endpoint->port->port_num,
-		endpoint->pkey);
+		be64toh(endpoint->port->dev->dev_guid),
```
weiny2: same comment above...  It is frustrating that ibv_get_device_guid returns uint64_t instead of a be64 variant.

ibacm/prov/acmp/src/acmp.c
```diff
@@ -656,7 +656,7 @@ acmp_record_mc_av(struct acmp_port *port, struct ib_mc_member_rec *mc_rec,
 	dest->path.dgid = mc_rec->mgid;
 	dest->path.sgid = mc_rec->port_gid;
 	dest->path.dlid = mc_rec->mlid;
-	dest->path.slid = htobe16(port->lid) | port->sa_dest.av.src_path_bits;
+	dest->path.slid = htobe16(port->lid | port->sa_dest.av.src_path_bits);
```
jgunthorpe: I have more patches that add annotations, this series is just the bug fixes that change code behaviour.

ibacm/prov/acmp/src/acmp.c
```diff
@@ -2458,7 +2459,7 @@ static uint16_t acmp_get_pkey_index(struct acm_endpoint *endpoint)
 
 	for (i = 0, ret = 0; !ret; i++) {
 		ret = ibv_query_pkey(port->dev->verbs, port->port_num, i, &pkey);
-		if (!ret && endpoint->pkey == pkey)
+		if (!ret && endpoint->pkey == be16toh(pkey))
```
shefty: gurgle - how did this ever work?  i guess only for pkey index 0?

jgunthorpe: This is the one patch I was most uncertain about.. There are lots of places in acm that store be and le values in the same struct member. However, I could not find any place that set endpoint->pkey to a BE value...

dledford:
Merging now so I can test another -rc release.

PR #98, opened 2017-03-14
Improve endianness annotations for sparse

bvanassche:


jgunthorpe:
@bvanassche FYI, here is  a lot of my stuff on sparse..

https://github.com/jgunthorpe/rdma-plumbing/commits/sparse

buildlib/fixup-include/netinet-in.h
```diff
@@ -0,0 +1,74 @@
...
+
+#undef INADDR_ANY
+#define INADDR_ANY		htobe32(0)
+#undef INADDR_BROADCAST
+#define INADDR_BROADCAST	htobe32(0xffffffffU)
+#undef INADDR_NONE
+#define INADDR_NONE		htobe32(0xffffffffU)
+#undef INADDR_LOOPBACK
+#define INADDR_LOOPBACK		0x7f000001 /* Inet 127.0.0.1.  */
+
```
jgunthorpe: This one is BE as well

I was also mulling over sed scripting /usr/include/net/in.h instead of providing a wrapper, it is hard to be complete with the wrapper approach..

bvanassche: OK, I will wait until your changes are upstream.

jgunthorpe:
@bvanassche  I updated my branch to use a different scheme for fixing glibc headers... What do you think?

https://github.com/jgunthorpe/rdma-plumbing/commits/sparse

bvanassche:
The approach using #include_next is more robust because the slightest change in a glibc header will cause a patch against such a header no longer to apply. I think we should try to minimize the effort it takes to maintain fixup header files.

jgunthorpe:
Well, it is fairly robust for endian.h, but very frail for netinet, and impossible for things like the pthread header and others. It is turning out there are lots of different little adjustments required. :(

I'm not sure how much trouble this will be, patch is fairly good about handling fuzz and I don't know how much churn there is in these headers across glibc versions.

One answer is that we don't really support sparse except as part of a cbuild docker container, which would have a known version of glibc, so all the diffs apply..

bvanassche:
Only supporting the glibc version of the cbuild docker container would be inconvenient. I expect that developers will want to run sparse against the glibc version installed on their development system.

rleon:
@bvanassche is right, I'm one of these developers, because I run sparse and smatch for the kernel patches, so I would like to be able to run the same setup for rdma-core.

jgunthorpe:
I think pull #101 encompasses everything in this PR.

jgunthorpe:
Sorry I meant PR's #100  and #101 together cover this.

PR #100, opened 2017-03-15
Make sparse work from travis

jgunthorpe:
Don't merge this. This is a placeholder to test travis running this branch. I suggest we go ahead after the next release.


shefty:
The changes related to the librdamcm posted to the mailing list look good. 👍

libibumad/umad.h
```diff
@@ -183,7 +183,7 @@ int umad_status(void *umad);
 ib_mad_addr_t *umad_get_mad_addr(void *umad);
 int umad_set_grh_net(void *umad, void *mad_addr);
 int umad_set_grh(void *umad, void *mad_addr);
-int umad_set_addr_net(void *umad, int dlid, int dqp, int sl, int qkey);
+int umad_set_addr_net(void *umad, __be16 dlid, __be32 dqp, int sl, __be32 qkey);
```
weiny2: Is this going to potentially break apps using this interface?

jgunthorpe: There is  a slight chance, it changes the underlying types a tiny bit, notably swapping in a short..

However, I am not aware of any architecture that would generate different machine code for this case..

If there is doubt we can symbol version it..

weiny2: Ok.

opensm uses this 3 times.  All 3 should be fine.

OPA management code does not use this and neither does the infiniband-diags...

So we are probably safe.

weiny2:
looks good.

buildlib/sparse-include/23/endian.h.diff
```diff
@@ -0,0 +1,57 @@
...
+@@ -59,38 +59,23 @@
+ /* Conversion interfaces.  */
+ # include <bits/byteswap.h>
+ 
+-# if __BYTE_ORDER == __LITTLE_ENDIAN
+-#  define htobe16(x) __bswap_16 (x)
+-#  define htole16(x) (x)
+-#  define be16toh(x) __bswap_16 (x)
+-#  define le16toh(x) (x)
++// Simple vesion for sparse with annotations
```
rleon: The comment is not in C format.

jgunthorpe: C99 added // comments, we are using C11... But I'll switch it.

jgunthorpe:
The goal here is to make sparse run automatically from Travis. We have a lot of control over the Travis environment, so this scheme is unlikely to randomly break inside travis as the glibc will always be the same.

sparse in user space turns out to be very hard because sparse and glibc are just not compatible. So we *need* to change libc headers and no matter what approach is used it will be inherently frail.

So.. any solution is going to be compromised in some way or another.. I propose we focus on having it run inside Travis and a few distros that people are interested in.

Sparse clearly is finding bugs, so it is more valuable to have *something* working even if it isn't everything you hoped for.

rleon:
@jgunthorpe, I'm fine as long as it works :)

PR #102, opened 2017-03-21
rdma-ndd: Use NULL instead of 0 where a pointer is expected

bvanassche:
Signed-off-by: Bart Van Assche <bart.vanassche@sandisk.com>


dledford:
I think after Jason's patches this is now just some whitespace cleanup.  Merging.

PR #103, opened 2017-03-21
Vendor cq

yishaih:
Uploading a new series based on the discussion that was on the list re the verbs patch.

### DIFF
--- a/libibverbs/device.c
+++ b/libibverbs/device.c
@@ -132,6 +132,26 @@ __be64 __ibv_get_device_guid(struct ibv_device *device)
 }
 default_symver(__ibv_get_device_guid, ibv_get_device_guid);
 
+void verbs_init_cq(struct ibv_cq *cq, struct ibv_context *context,
+		       struct ibv_comp_channel *channel,
+		       void *cq_context)
+{
+	cq->context		   = context;
+	cq->channel		   = channel;
+
+	if (cq->channel) {
+		pthread_mutex_lock(&context->mutex);
+		++cq->channel->refcnt;
+		pthread_mutex_unlock(&context->mutex);
+	}
+
+	cq->cq_context		   = cq_context;
+	cq->comp_events_completed  = 0;
+	cq->async_events_completed = 0;
+	pthread_mutex_init(&cq->mutex, NULL);
+	pthread_cond_init(&cq->cond, NULL);
+}
+
 static struct ibv_cq_ex *
 __lib_ibv_create_cq_ex(struct ibv_context *context,
 		       struct ibv_cq_init_attr_ex *cq_attr)
@@ -144,23 +164,11 @@ __lib_ibv_create_cq_ex(struct ibv_context *context,
 		return NULL;
 	}
 
-	pthread_mutex_lock(&context->mutex);
-
 	cq = vctx->priv->create_cq_ex(context, cq_attr);
 
-	if (cq) {
-		cq->context		   = context;
-		cq->channel		   = cq_attr->channel;
-		if (cq->channel)
-			++cq->channel->refcnt;
-		cq->cq_context		   = cq_attr->cq_context;
-		cq->comp_events_completed  = 0;
-		cq->async_events_completed = 0;
-		pthread_mutex_init(&cq->mutex, NULL);
-		pthread_cond_init(&cq->cond, NULL);
-	}
-
-	pthread_mutex_unlock(&context->mutex);
+	if (cq)
+		verbs_init_cq(ibv_cq_ex_to_cq(cq), context,
+			        cq_attr->channel, cq_attr->cq_context);
 
 	return cq;
 }
--- a/libibverbs/libibverbs.map
+++ b/libibverbs/libibverbs.map
@@ -86,7 +86,7 @@ IBVERBS_1.1 {
 
 /* If any symbols in this stanza change ABI then the entire staza gets a new symbol
    version. Also see the private_symver() macro */
-IBVERBS_PRIVATE_13 {
+IBVERBS_PRIVATE_14 {
 	global:
 		/* These historical symbols are now private to libibverbs */
 		ibv_cmd_alloc_mw;
@@ -138,4 +138,5 @@ IBVERBS_PRIVATE_13 {
 		ibv_query_gid_type;
 		ibv_register_driver;
 		verbs_register_driver;
+		verbs_init_cq;
 };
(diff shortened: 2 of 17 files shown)


libibverbs/device.c
```diff
@@ -120,8 +120,9 @@ uint64_t __ibv_get_device_guid(struct ibv_device *device)
 }
 default_symver(__ibv_get_device_guid, ibv_get_device_guid);
 
-struct ibv_cq_ex *__lib_ibv_create_cq_ex(struct ibv_context *context,
-					 struct ibv_cq_init_attr_ex *cq_attr)
+struct ibv_cq_ex *ibv_vendor_create_cq_ex(struct ibv_context *context,
+					 struct ibv_cq_init_attr_ex *cq_attr,
+					 void *vendor_data)
 {
```
jgunthorpe: Stop using the word vendor, these are for providers or drivers..

libibverbs/libibverbs.map
```diff
@@ -139,3 +139,8 @@ IBVERBS_PRIVATE_13 {
 		ibv_register_driver;
 		verbs_register_driver;
 };
+
+IBVERBS_PRIVATE_14 {
+	global:
+		ibv_vendor_create_cq_ex;
+} IBVERBS_PRIVATE_13;
```
jgunthorpe: No, there should be only one PRIVATE stanza, bump the whole thing to 14.

providers/mlx5/libmlx5.map
```diff
@@ -6,3 +6,8 @@ MLX5_1.0 {
 		mlx5dv_init_obj;
 	local: *;
 };
+
+MLX5_1.1 {
+	global:
+		mlx5dv_create_cq_ex;
+} MLX5_1.0;
```
jgunthorpe: It would be better to follow the convection of the rest of the library and call this MLX5_1.14

1.0 should have be 1.13

And make sure the library symlink has  1.14 in the name

libibverbs/verbs.h
```diff
@@ -1505,6 +1505,10 @@ static inline struct verbs_context *verbs_get_ctx(struct ibv_context *ctx)
 	if (vctx && (vctx->sz >= sizeof(*vctx) - offsetof(struct verbs_context, op))) \
 		vctx->op = ptr; })
 
+#define verbs_set_ctx_private_op(_vctx, op, ptr)({ \
+	struct verbs_context *vctx = _vctx; \
+	vctx->priv->op = ptr; })
+
 /**
```
jgunthorpe: Now that the providers are providers are private we can get rid of this goofy system. Just write the op directly.

providers/mlx5/mlx5.c
```diff
@@ -609,16 +610,35 @@ static int mlx5_map_internal_clock(struct mlx5_device *mdev,
 	return 0;
 }
 
+struct ibv_cq_ex *mlx5dv_create_cq_ex(struct ibv_context *context,
+				      struct ibv_cq_init_attr_ex *cq_attr,
+				      struct mlx5dv_cq_init_attr_ex *mlx5_cq_attr)
+{
+	return ibv_vendor_create_cq_ex(context, cq_attr, mlx5_cq_attr);
+}
```
jgunthorpe: This seems really silly. Just add a new provider private rdma_init_cq() that contains the init stuff.

Since this is a performance path the driver should tend to its own locking, thus we don't need the strange wrapper.

providers/mlx5/mlx5dv.h
```diff
@@ -54,22 +54,47 @@ enum {
...
+	MLX5DV_CREATE_CQ_MASK_RESERVED          = 1 << 1,
+};
+
+struct mlx5dv_cq_init_attr_ex {
+	uint64_t comp_mask; /* Use enum mlx5dv_create_cq_vendor_data_mask */
+	uint8_t cqe_comp_res_format; /* Use enum mlx5dv_cqe_comp_res_format */
 };
 
+struct ibv_cq_ex *mlx5dv_create_cq_ex(struct ibv_context *context,
+				      struct ibv_cq_init_attr_ex *cq_attr,
```
jgunthorpe: You can probably drop the _ex in all of this. We only need to use _ex if it conflicted with a historical name.

jgunthorpe:
And you'll have to rebase it.

providers/mlx5/mlx5.c
yishaih: Looks like you missed the point here, without calling libibverbs it won't have the chance to activate any private API upon this flow, currently there is some pre and post stuff that is done in libibverbs, the idea is that all of this code won't be duplicated to all providers, in addition any future support that may be added as of resource tracking per context will be in one place as well.

By using that wrapper the input private attr (i.e mlx5_cq_attr) is passed all the way from application to the driver without any need to save it locally on some context and the shared code of libibverbs is used, as of create CQ is not considered to be a performance path we see it as the  preferred way to go.

libibverbs/verbs.h
yishaih: The idea was to supply a macro that wrap it but I'm fine with explicit setting as well.

providers/mlx5/mlx5.c
jgunthorpe: I get the point, I just think calling some simple helpers is clearer than this indirection through yet another function pointer. There is no 'pre' step for mlx5 - it doesn't look like it needs that lock and it duplicates the wc_flags check already

```
cq = mlx5_create_cq(...);
if (cq)
    verbs_init_cq(cq);
return cq;
```

yishaih: Basically I'm fine with some helpers, however it's not only a post helper. Why the lock that is taken in libibverbs before calling the provider is not needed here ? (i.e. pthread_mutex_lock(&context->mutex);) this locks currently comes to protect from parallel call to ibv_destroy_comp_channel which is still applicable here and in the future may be used for other purposes. If you think that having 2 helpers is better than calling one API that will callback the provider I can go with that as well.

jgunthorpe: I said the lock didn't look needed because there didn't seem to be any reason for it to cover the driver callback.

It looks like the lock only needs to cover refcnt, so you should revise it and just have the one helper function.

Obviously the lock cannot protect against racing create with ibv_destroy_comp_channel, as destroy could happen immediately before the lock is acquired and the channel pointer will be freed. It is an error for the user to allow this to happen.

Ideally refcnt would become a 'kref' like thing using atomics and a lot of this ugly mess would go away including the lock and EBUSY failure...

jgunthorpe:
There is still the one comment on the mailing list for the last patch

jgunthorpe:
And the various warning messages from travis's debian package build need to be fixed too.

PR #104, opened 2017-03-21
Enable endianness checking for GIDs

bvanassche:
Additionally, simplify rs_host_is_net().

### DIFF
--- a/libibumad/umad.c
+++ b/libibumad/umad.c
@@ -736,7 +736,7 @@ int umad_set_grh(void *umad, void *mad_addr)
 
 	if (mad_addr) {
 		mad->addr.grh_present = 1;
-		memcpy(mad->addr.gid, addr->gid, 16);
+		mad->addr.ib_gid = addr->ib_gid;
 		/* The definition for umad_set_grh requires that the input be
 		 * in host order */
 		mad->addr.flow_label = htobe32((__force uint32_t)addr->flow_label);
--- a/libibumad/umad.c
+++ b/libibumad/umad.c
@@ -125,7 +125,7 @@ static int check_for_digit_name(const struct dirent *dent)
 static int get_port(const char *ca_name, const char *dir, int portnum, umad_port_t * port)
 {
 	char port_dir[256];
-	uint8_t gid[16];
+	union umad_gid gid;
 	struct dirent **namelist = NULL;
 	int i, len, num_pkeys = 0;
 	uint32_t capmask;
@@ -162,11 +162,11 @@ static int get_port(const char *ca_name, const char *dir, int portnum, umad_port
 
 	port->capmask = htobe32(capmask);
 
-	if (sys_read_gid(port_dir, SYS_PORT_GID, gid) < 0)
+	if (sys_read_gid(port_dir, SYS_PORT_GID, &gid) < 0)
 		goto clean;
 
-	memcpy(&port->gid_prefix, gid, sizeof port->gid_prefix);
-	memcpy(&port->port_guid, gid + 8, sizeof port->port_guid);
+	port->gid_prefix = gid.global.subnet_prefix;
+	port->port_guid = gid.global.interface_id;
 
 	snprintf(port_dir + len, sizeof(port_dir) - len, "/pkeys");
 	num_pkeys = scandir(port_dir, &namelist, check_for_digit_name, NULL);
@@ -569,7 +569,7 @@ int umad_get_cas_names(char cas[][UMAD_CA_NAME_LEN], int max)
 	return j;
 }
 
-int umad_get_ca_portguids(const char *ca_name, uint64_t * portguids, int max)
+int umad_get_ca_portguids(const char *ca_name, __be64 *portguids, int max)
 {
 	umad_ca_t ca;
 	int ports = 0, i;
@@ -588,8 +588,8 @@ int umad_get_ca_portguids(const char *ca_name, uint64_t * portguids, int max)
 		}
 
 		for (i = 0; i <= ca.numports; i++)
-			portguids[ports++] =
-			    ca.ports[i] ? ca.ports[i]->port_guid : 0;
+			portguids[ports++] = ca.ports[i] ?
+				ca.ports[i]->port_guid : htobe64(0);
 	}
 
 	release_ca(&ca);
--- a/libibumad/umad.h
+++ b/libibumad/umad.h
@@ -157,8 +157,8 @@ typedef struct umad_port {
 	unsigned phys_state;
 	unsigned rate;
 	__be32 capmask;
-	uint64_t gid_prefix;
-	uint64_t port_guid;
+	__be64	 gid_prefix;
+	__be64	 port_guid;
 	unsigned pkeys_size;
 	uint16_t *pkeys;
 	char link_layer[UMAD_CA_NAME_LEN];
@@ -180,7 +180,7 @@ int umad_init(void);
 int umad_done(void);
 
 int umad_get_cas_names(char cas[][UMAD_CA_NAME_LEN], int max);
-int umad_get_ca_portguids(const char *ca_name, uint64_t * portguids, int max);
+int umad_get_ca_portguids(const char *ca_name, __be64 *portguids, int max);
 
 int umad_get_ca(const char *ca_name, umad_ca_t * ca);
 int umad_release_ca(umad_ca_t * ca);
--- a/libibumad/umad.h
+++ b/libibumad/umad.h
@@ -54,6 +54,19 @@ typedef __be16 __attribute__((deprecated)) be16_t;
 typedef __be32 __attribute__((deprecated)) be32_t;
 typedef __be64 __attribute__((deprecated)) be64_t;
 
+/*
+ * A GID data structure that may be used in definitions of on-the-wire data
+ * structures. Do not cast umad_gid pointers to ibv_gid pointers because the
+ * alignment of these two data structures is different.
+ */
+union umad_gid {
+	uint8_t		raw[16];
+	struct {
+		__be64	subnet_prefix;
+		__be64	interface_id;
+	} global;
+} __attribute__((aligned(4))) __attribute__((packed));
+
 #define UMAD_MAX_DEVICES 32
 #define UMAD_ANY_PORT	0
 typedef struct ib_mad_addr {
@@ -66,7 +79,10 @@ typedef struct ib_mad_addr {
 	uint8_t gid_index;
 	uint8_t hop_limit;
 	uint8_t traffic_class;
-	uint8_t gid[16]; /* network-byte order */
+	union {
+		uint8_t		gid[16]; /* network-byte order */
+		union umad_gid	ib_gid;
+	};
 	__be32 flow_label;
 	uint16_t pkey_index;
 	uint8_t reserved[6];
--- a/srp_daemon/srp_daemon.c
+++ b/srp_daemon/srp_daemon.c
@@ -77,7 +77,8 @@
 
 enum log_dest { log_to_syslog, log_to_stderr };
 
-static int get_lid(struct umad_resources *umad_res, ib_gid_t *gid, uint16_t *lid);
+static int get_lid(struct umad_resources *umad_res, union umad_gid *gid,
+		   uint16_t *lid);
 
 static const int   node_table_response_size = 1 << 18;
 static const char *sysfs_path = "/sys";
@@ -405,7 +406,7 @@ static int add_non_exist_target(struct target_details *target)
 	struct dirent *subdir;
 	char *subdir_name_ptr;
 	int prefix_len;
-	ib_gid_t dgid_val;
+	union umad_gid dgid_val;
 	char target_config_str[255];
 	int len;
 	int not_connected = 1;
@@ -458,9 +459,10 @@ static int add_non_exist_target(struct target_details *target)
 					      dgid_val.raw))
 				continue;
 		}
-		if (htobe64(target->subnet_prefix) != dgid_val.unicast.prefix)
+		if (htobe64(target->subnet_prefix) !=
+		    dgid_val.global.subnet_prefix)
 			continue;
-		if (htobe64(target->h_guid) != dgid_val.unicast.interface_id)
+		if (htobe64(target->h_guid) != dgid_val.global.interface_id)
 			continue;
 
 		/* If there is no local_ib_device in the scsi host dir (old kernel module), assumes it is equal */
@@ -2029,7 +2031,7 @@ int main(int argc, char *argv[])
 	struct resources       *res;
 	uint16_t 		lid;
 	uint16_t 		pkey;
-	ib_gid_t 		gid;
+	union umad_gid 		gid;
 	struct target_details  *target;
 	struct sigaction	sa;
 	int			subscribed;
@@ -2330,7 +2332,8 @@ static int recalc(struct resources *res)
 	return ret;
 }
 
-static int get_lid(struct umad_resources *umad_res, ib_gid_t *gid, uint16_t *lid)
+static int get_lid(struct umad_resources *umad_res, union umad_gid *gid,
+		   uint16_t *lid)
 {
 	srp_ib_user_mad_t		out_mad, in_mad;
 	struct srp_dm_rmpp_sa_mad 	*in_sa_mad  = get_data_ptr(in_mad);
--- a/srp_daemon/srp_daemon.c
+++ b/srp_daemon/srp_daemon.c
@@ -2036,6 +2036,21 @@ int main(int argc, char *argv[])
 	int			lockfd = -1;
 	int			received_signal = 0;
 
+#ifndef __CHECKER__
+	/*
+	 * Hide these checks for sparse because these checks fail with
+	 * older versions of sparse.
+	 */
+	STATIC_ASSERT(sizeof(ib_sa_mad_t) == 256);
+	STATIC_ASSERT(sizeof(ib_path_rec_t) == 64);
+	STATIC_ASSERT(sizeof(ib_mad_t) == 24);
+	STATIC_ASSERT(sizeof(ib_inform_info_t) == 36);
+	STATIC_ASSERT(sizeof(ib_mad_notice_attr_t) == 80);
+	STATIC_ASSERT(offsetof(ib_mad_notice_attr_t,
+			       data_details.ntc_64_67.gid) == 16);
+
+	STATIC_ASSERT(__alignof__(union umad_gid) == 4);
+#endif
 	STATIC_ASSERT(sizeof(struct srp_dm_mad) == 256);
 	STATIC_ASSERT(sizeof(struct srp_dm_rmpp_sa_mad) == 256);
 	STATIC_ASSERT(sizeof(struct srp_sa_node_rec) == 108);
--- a/srp_daemon/srp_daemon.h
+++ b/srp_daemon/srp_daemon.h
@@ -368,7 +368,7 @@ struct sync_resources {
 	struct {
 		uint16_t lid;
 		uint16_t pkey;
-		ib_gid_t gid;
+		union umad_gid gid;
 	} tasks[SIZE_OF_TASKS_LIST];
 	pthread_mutex_t mutex;
 	struct target_details *retry_tasks_head;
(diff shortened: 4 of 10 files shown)


libibumad/umad.h
```diff
@@ -54,6 +54,21 @@ typedef __be16 __attribute__((deprecated)) be16_t;
...
+		__be64	prefix;
+		__be64	interface_id;
+	} unicast;
+
+	struct {
+		uint8_t	header[2];
+		uint8_t	raw_group_id[14];
+	} multicast;
+} ib_gid_t;
+
```
jgunthorpe: This drops the packed on this, doesn't that cause a problem in at least union _data_details / _ntc_64_67 which has it as an unaligned member?

libibumad/umad.c
```diff
@@ -162,11 +162,11 @@ static int get_port(const char *ca_name, const char *dir, int portnum, umad_port
...
 	port->capmask = htobe32(capmask);
 
-	if (sys_read_gid(port_dir, SYS_PORT_GID, gid) < 0)
+	if (sys_read_gid(port_dir, SYS_PORT_GID, &gid) < 0)
 		goto clean;
 
-	memcpy(&port->gid_prefix, gid, sizeof port->gid_prefix);
-	memcpy(&port->port_guid, gid + 8, sizeof port->port_guid);
+	port->gid_prefix = gid.unicast.prefix;
+	port->port_guid = gid.unicast.interface_id;
```
jgunthorpe: Trvavis says umad_port_t gid_prefix and port_guid need to be __be64 too..

bvanassche: I will update this.

libibumad/umad.h
bvanassche: A compile-time check I added shows that the size of ib_mad_notice_attr_t has not changed.

libibumad/umad.h
```diff
@@ -54,6 +54,21 @@ typedef __be16 __attribute__((deprecated)) be16_t;
...
+	struct {
+		__be64	prefix;
+		__be64	interface_id;
+	} unicast;
+
+	struct {
+		uint8_t	header[2];
+		uint8_t	raw_group_id[14];
+	} multicast;
+} ib_gid_t;
```
bvanassche: Good point. Should a configure test be added first to opensm?

libibumad/umad.h
```diff
@@ -81,7 +81,10 @@ typedef struct ib_mad_addr {
 	uint8_t gid_index;
 	uint8_t hop_limit;
 	uint8_t traffic_class;
-	uint8_t gid[16]; /* network-byte order */
+	union {
+		uint8_t	 gid[16]; /* network-byte order */
+		ib_gid_t ib_gid;
```
bvanassche: My concern is that that would break existing users of ib_mad_addr_t that access the GID through the "gid" member name.

libibumad/umad.h
weiny2: Yea I think so...  Can you cc someone on GitHub with this review?

I forwarded the email from GitHub to Hal.

libibumad/umad.h
weiny2: Ah...  yea I missed that this is in the public header...  I wonder if we can deprecate that access somehow?

libibumad/umad.h
jgunthorpe: The ib_mad_notice_attr_t has the PACKED attribute set which does eliminate much of the padding, so the overall size will not change.

However, the ib_gid_t's in that structure are not 8 byte aligned:

```
_Static_assert(__alignof__(ib_gid_t) == 8, "err");
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_64_67.gid) % 8 == 0, "err");
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_257_258.gid1) % 8 == 0, "err");
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_257_258.gid2) % 8 == 0, "err");
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_259.gid1) % 8 == 0, "err");
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_259.gid2) % 8 == 0, "err");
```

Yields compile failures.

The compiler may generate correct code for

```
ib_gid_t tmp = notice->data_details.ntc_259.gid2;
```

Since it can see that gid2 has a strange alignment.

However, there is now a subtle trap, where this will alignment trap on some arches:

```
extern void some_function(const ib_gid_t *arg);
some_function(&notice->data_details.ntc_259.gid2);
```

As the compiler in some_function() will assume that 'arg' has 8 byte alignment.

Generally speaking it is a dangerous idea to have structs off their defined alignment.

Within IB mads, GIDs are only aligned to at most 4 bytes, so a gid struct designed for use with mads should be set to packed or to align 4.

hnrose: The overall size may not have changed but I don't think that's sufficient. All the structs inside the data details union need to be properly located and I'm not sure they are without the packing. As Jason pointed out, some are not naturally aligned.

hnrose: The overall size may not have changed but I don't think that's sufficient. All the structs inside the data details union need to be properly located and I'm not sure they are without the packing. As Jason pointed out, some are not naturally aligned.

libibumad/umad.h
hnrose: What would OpenSM configure test do ? Check for (some member in) ib_gid_t in umad.h (via AC_CHECK_MEMBER ?) and if present use that rather than the one in ib_types.h ?

This is going to require release of OpenSM to go with rdma-core which supports this.

It may also affect other tools which include ib_types.h (infiniband-diags).

weiny2: Yes something like that.

I'm starting to think that what we should do is call this something other than ib_gid_t.  Deprecate ib_gid_t in all headers and then eventually remove ib_gid_t in favor of the new type.

I'm terrible at names but "rdma_gid_t"?

shefty: Why is a typedef needed?

libibumad/umad.c
```diff
@@ -569,7 +569,7 @@ int umad_get_cas_names(char cas[][UMAD_CA_NAME_LEN], int max)
 	return j;
 }
 
-int umad_get_ca_portguids(const char *ca_name, uint64_t * portguids, int max)
+int umad_get_ca_portguids(const char *ca_name, __be64 *portguids, int max)
```
hnrose: libibumad/man/umad_get_ca_portguids.3 should be updated to go along with this.

libibumad/umad.h
bvanassche: @shefty: I will drop the typedef.
@weiny2: The kernel header files define struct ib_gid in <rdma/ib_verbs.h> and struct rxe_gid in <uapi/rdma/rdma_user_rxe.h>. I just noticed that there is already a struct ibv_gid definition in libibverbs/verbs.h. I will see whether I can reuse that definition.

jgunthorpe: struct ibv_gid is defined to have aligned(8). It probably isn't safe to change it to packed.

weiny2: Bart,

That sounds like a really good plan...  libibumad and others should migrate to that definition.  I'll put it on my list for the diags and libibmad.  They don't really use ib_gid_t that much though.

I still think opensm should stop exporting this as well.  Hal do you agree?  There are probably other types which could be migrated to the core as well.  I just don't have time right now to do an audit.

hnrose: Jason,

So ibv_gid is different from what's needed here if it's going to include use by OpenSM, right ?

hnrose: Ira,

While it's OK to migrate certain types to core in Linux, OpenSM is supported on more than Linux and some of these platforms do not use libibumad as it's vendor layer (interface to IB port). OpenSM will need to accommodate when the types are in rdma-core and when they're not.

srp_daemon/srp_ib_types.h
```diff
@@ -344,8 +326,8 @@ typedef struct _ib_mad_notice_attr		// Total Size calc  Accumulated
 			uint8_t		sl;		// 1
 			__be32		qp1;		// 4
 			__be32		qp2;		// 4
-			ib_gid_t	gid1;		// 16
-			ib_gid_t	gid2;		// 16
+			union ibv_gid	gid1;		// 16
+			union ibv_gid	gid2;		// 16
 		} PACK_SUFFIX ntc_257_258;
```
hnrose: This layout prior to change does not match IB spec for this trap. Is it used by SRP ?

In the spec, I don't think the GIDs are 8 byte aligned. I think they're 4 byte aligned.

srp_daemon/srp_ib_types.h
```diff
@@ -358,15 +340,15 @@ typedef struct _ib_mad_notice_attr		// Total Size calc  Accumulated
 			__be32		qp1;		// 4
 			uint8_t		qp2_msb;	// 1
 			__be16		qp2_lsb;	// 2
-			ib_gid_t	gid1;		// 16
-			ib_gid_t	gid2;		// 16
+			union ibv_gid	gid1;		// 16
+			union ibv_gid	gid2;		// 16
 			__be16		sw_lid;		// 2
 			uint8_t		port_no;	// 1
 		} PACK_SUFFIX ntc_259;
```
hnrose: This layout prior to change does not match IB spec for this trap. Is it used by SRP ?

In the spec, I don't think the GIDs are 8 byte aligned.

srp_daemon/srp_ib_types.h
hnrose: Looks like only ntc_64_57 and ntc_144 are used by SRP daemon so maybe it's best to just remove this.

srp_daemon/srp_ib_types.h
hnrose: Looks like only ntc_64_57 and ntc_144 are used by SRP daemon so maybe it's best to just remove this.

libibumad/umad.c
hnrose: My bad; it's the very first thing. Sorry,

libibumad/umad.h
```diff
@@ -39,6 +39,7 @@
 #include <stdlib.h>
 #include <arpa/inet.h>
 #include <linux/types.h> /* __be16, __be32 and __be64 */
+#include <infiniband/verbs.h> /* union ibv_gid */
```
hnrose: I have mixed feelings about this. Up to now, umad was not dependent on verbs and on some management machines, verbs was not installed. Now both are in rdma-core so they're together.

If this is to be done, shouldn't there be configure check for this ?

bvanassche: I'm not sure we want to make the output of any configure process visible to libibumad users. Anyway, I will remove the #include directive.

srp_daemon/srp_ib_types.h
bvanassche: I will remove the unused notice types.

srp_daemon/srp_daemon.c
```diff
@@ -77,7 +77,8 @@
 
 enum log_dest { log_to_syslog, log_to_stderr };
 
-static int get_lid(struct umad_resources *umad_res, ib_gid_t *gid, uint16_t *lid);
+static int get_lid(struct umad_resources *umad_res, union ibv_gid *gid,
```
hnrose: Should this be ibu_gid rather than ibv_gid ?

srp_daemon/srp_daemon.c
```diff
@@ -405,7 +406,7 @@ static int add_non_exist_target(struct target_details *target)
 	struct dirent *subdir;
 	char *subdir_name_ptr;
 	int prefix_len;
-	ib_gid_t dgid_val;
+	union ibv_gid dgid_val;
```
hnrose: ibu_gid ?

srp_daemon/srp_daemon.c
```diff
@@ -2029,13 +2031,25 @@ int main(int argc, char *argv[])
 	struct resources       *res;
 	uint16_t 		lid;
 	uint16_t 		pkey;
-	ib_gid_t 		gid;
+	union ibv_gid 		gid;
```
hnrose: ibu_gid ?

srp_daemon/srp_daemon.c
```diff
@@ -2315,7 +2329,8 @@ static int recalc(struct resources *res)
 	return ret;
 }
 
-static int get_lid(struct umad_resources *umad_res, ib_gid_t *gid, uint16_t *lid)
+static int get_lid(struct umad_resources *umad_res, union ibv_gid *gid,
```
hnrose: ibu_gid ?

srp_daemon/srp_daemon.h
```diff
@@ -368,7 +368,7 @@ struct sync_resources {
 	struct {
 		uint16_t lid;
 		uint16_t pkey;
-		ib_gid_t gid;
+		union ibv_gid gid;
```
hnrose: ibu_gid ?

srp_daemon/srp_daemon.h
```diff
@@ -418,14 +418,15 @@ int create_trap_resources(struct ud_resources *ud_res);
 int register_to_traps(struct resources *res, int subscribe);
 uint16_t get_port_lid(struct ibv_context *ib_ctx, int port_num);
 int create_ah(struct ud_resources *ud_res);
-void push_gid_to_list(struct sync_resources *res, ib_gid_t *gid, uint16_t pkey);
+void push_gid_to_list(struct sync_resources *res, union ibv_gid *gid,
```
hnrose: ibu_gid ?

srp_daemon/srp_daemon.h
```diff
@@ -418,14 +418,15 @@ int create_trap_resources(struct ud_resources *ud_res);
...
+		      uint16_t pkey);
 void push_lid_to_list(struct sync_resources *res, uint16_t lid, uint16_t pkey);
 struct target_details *pop_from_retry_list(struct sync_resources *res);
 void push_to_retry_list(struct sync_resources *res,
 			struct target_details *target);
 int retry_list_is_empty(struct sync_resources *res);
 void clear_traps_list(struct sync_resources *res);
-int pop_from_list(struct sync_resources *res, uint16_t *lid, ib_gid_t *gid,
+int pop_from_list(struct sync_resources *res, uint16_t *lid, union ibv_gid *gid,
```
hnrose: ibu_gid ?

srp_daemon/srp_ib_types.h
```diff
@@ -39,6 +39,7 @@
 #include <endian.h>
 #include <stdint.h>
 #include <linux/types.h> /* __be16, __be32 and __be64 */
+#include <infiniband/verbs.h> /* union ibv_gid */
```
hnrose: Not needed if use ibu_gid rather than ibv_gid ?

srp_daemon/srp_ib_types.h
```diff
@@ -155,8 +137,8 @@ enum {
 typedef struct _ib_path_rec
 {
 	uint8_t		resv0[8];
-	ib_gid_t	dgid;
-	ib_gid_t	sgid;
+	union ibv_gid	dgid;
+	union ibv_gid	sgid;
```
hnrose: ibu_gid ?

srp_daemon/srp_ib_types.h
```diff
@@ -229,7 +211,7 @@ ib_mad_init_new(ib_mad_t* const	p_mad,
 
 typedef struct _ib_inform_info
 {
-	ib_gid_t	gid;
+	union ibv_gid	gid;
```
hnrose: ibu_gid ?

srp_daemon/srp_ib_types.h
```diff
@@ -294,47 +276,16 @@ typedef struct _ib_mad_notice_attr		// Total Size calc  Accumulated
 		struct _ntc_64_67
 		{
 			uint8_t		res[6];
-			ib_gid_t	gid;	// the Node or Multicast Group that came in/out
+			union ibv_gid	gid;	// the Node or Multicast Group that came in/out
```
hnrose: ibu_gid ?

srp_daemon/srp_ib_types.h
```diff
@@ -379,9 +314,9 @@ typedef struct _ib_mad_notice_attr		// Total Size calc  Accumulated
 *
 * SYNOPSIS
 */
-static inline __be64 ib_gid_get_guid(const ib_gid_t *const p_gid)
+static inline __be64 ib_gid_get_guid(const union ibv_gid *const p_gid)
```
hnrose: ibu_gid ?

srp_daemon/srp_sync.c
```diff
@@ -202,7 +203,7 @@ void clear_traps_list(struct sync_resources *res)
 
 
 /* assumes that res->mutex is locked !!! */
-int pop_from_list(struct sync_resources *res, uint16_t *lid, ib_gid_t *gid,
+int pop_from_list(struct sync_resources *res, uint16_t *lid, union ibv_gid *gid,
```
hnrose: ibu_gid ?

srp_daemon/srp_sync.c
```diff
@@ -117,7 +117,8 @@ void sync_resources_cleanup(struct sync_resources *res)
 	pthread_mutex_destroy(&res->mutex);
 }
 
-void push_gid_to_list(struct sync_resources *res, ib_gid_t *gid, uint16_t pkey)
+void push_gid_to_list(struct sync_resources *res, union ibv_gid *gid,
```
hnrose: ibu_gid ?

jgunthorpe:
This is still not OK with alignment:

```
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_257_258.gid1) % 8 == 0, "err");
_Static_assert((uintptr_t)(&((ib_mad_notice_attr_t *)0)->data_details.ntc_257_258.gid2) % 8 == 0, "err");
```

Both fail. 'union ibv_gid' must be 8 byte aligned.

I recommend you make ibu_gid the version of the struct with 1 byte alignment and use it in these MAD related places. But, it is an error to cast from ibu_gid to ibv_gid...

hnrose:
@jgunthorpe 
Isn't data_details.ntc_257_258 gone in latest version ?
Only gid remaining is in data_details.ntc_64_67.

Suggested using ibu_gid rather than ibv_gid.

jgunthorpe:
@hnrose, I pulled and got 6aa7de806c41f6 which has it.

But even if we drop it, an 8 byte aligned ibu_gid is not going to work for opensm either, or anything else working with MADs.

hnrose:
@jgunthorpe 
Agreed that 8 byte alignment doesn't cover all the use cases but I think 4 byte alignment is sufficient but don't know if that saves anything over 1 byte alignment which is clearly the safest.

srp_daemon/srp_daemon.c
```diff
@@ -2036,6 +2036,18 @@ int main(int argc, char *argv[])
...
+	/*
+	 * Hide these checks for sparse because sparse gets most of them wrong.
+	 */
+	STATIC_ASSERT(sizeof(ib_sa_mad_t) == 256);
+	STATIC_ASSERT(sizeof(ib_path_rec_t) == 64);
+	STATIC_ASSERT(sizeof(ib_mad_t) == 24);
+	STATIC_ASSERT(sizeof(ib_inform_info_t) == 36);
+	STATIC_ASSERT(sizeof(ib_mad_notice_attr_t) == 80);
+	STATIC_ASSERT(offsetof(ib_mad_notice_attr_t,
+			       data_details.ntc_257_258.gid2) == 45);
```
hnrose: Shouldn't this one be removed now that ntc_257_258 is gone ?

Maybe replace with check of offset of ntc_64_67.gid ?

jgunthorpe:
@hnrose, yes, like this:

```
/* Storage for a GID that is aligned to 4 bytes. This is suitable for use with
 * MAD packet layouts. Despite the same layout, this cannot be casted to an
 * ibv_gid, as it has a 8 byte alignment. */
union ibu_gid {
	uint8_t		raw[16];
	struct {
		__be64	subnet_prefix;
		__be64	interface_id;
	} global;
} __attribute__ ((aligned (4))) __attribute__((packed));

_Static_assert(__alignof__(union ibu_gid) == 4,"Err");
```

srp_daemon/srp_daemon.c
bvanassche: ntc_257_258 is still present and is still used.

srp_daemon/srp_ib_types.h
bvanassche: OK, I will change ibv_gid into ibu_gid in srp_daemon.

srp_daemon/srp_ib_types.h
bvanassche: I will update the #include directive.

srp_daemon/srp_daemon.c
hnrose: I missed where it's used. Would you point me at it ? I see the other notices in srp_handle_traps.c but not this one.

If it needs to be preserved, it's wrong/not according to IBA spec. It was lifted from old ib_types.h and needs fixing. It's why offset of gid there is mistakenly 45!

hnrose:
@weiny2 Naming scheme for umad uses umad_ prefix so umad_gid ?

bvanassche:
I will rename ibu_gid into umad_gid.

srp_daemon/srp_daemon.c
bvanassche: @hnrose : You are right, ntc_257_258 is not used. I will remove it.

weiny2:
Looks good to me!  Thanks Bart!

hnrose:
:+1: 

Thanks Bart

PR #105, opened 2017-03-22
libibumad/man/umad_set_addr_net.3: Update function signature to be consistent with umad.h

hnrose:
Consistent with change in commit 8a342f390e7af71bba0d8794d741453604647e98
umad: Add missing be annotation


rleon:
Thanks a lot.

PR #106, opened 2017-03-23
Fixes from the mailing list discussions

jgunthorpe:
This fixes the build problems @jarodwilson identified and makes it more obvious if a similar failure class happens in future.

This is also the patch to get sparse running on tumbleweed that @bvanassche tested.

As well as a few other minor things.


bvanassche:
This makes building with sparse against glibc 2.25 work again. Thanks!

rleon:
Thank you a lot.

PR #108, opened 2017-03-25
Eliminate ClassPortInfo duplication between libibumad and srp_daemon

hnrose:
Use umad_class_port_info rather than srp_class_port_info in SRP daemon.

In umad_class_port_info, use umad_gid unions and deprecate original [redir trap]_gid fields.

Follow on to Bart's work in PR #104

### DIFF
--- a/libibumad/tests/umad_compile_test.c
+++ b/libibumad/tests/umad_compile_test.c
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2017 Mellanox Technologies LTD. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * OpenIB.org BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+#include <config.h>
+
+#include <stddef.h>
+#include <endian.h>
+#include <ccan/build_assert.h>
+#include <infiniband/umad.h>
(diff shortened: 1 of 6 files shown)


srp_daemon/srp_daemon.c
```diff
@@ -2052,12 +2053,16 @@ int main(int argc, char *argv[])
...
+			       trapgid) == 40);
 #endif
 	STATIC_ASSERT(sizeof(struct srp_dm_mad) == 256);
 	STATIC_ASSERT(sizeof(struct srp_dm_rmpp_sa_mad) == 256);
 	STATIC_ASSERT(sizeof(struct srp_sa_node_rec) == 108);
 	STATIC_ASSERT(sizeof(struct srp_sa_port_info_rec) == 58);
-	STATIC_ASSERT(sizeof(struct srp_class_port_info) == 72);
+	STATIC_ASSERT(sizeof(struct umad_class_port_info) == 72);
 	STATIC_ASSERT(sizeof(struct srp_dm_iou_info) == 132);
 	STATIC_ASSERT(sizeof(struct srp_dm_ioc_prof) == 128);
```
jgunthorpe: I would suggest putting these things that apply to umad structs somewhere in umad, eg a compile_test.c file that is always built.

hnrose:
@bvanassche , does this PR look OK to you ? Thanks.

libibumad/tests/umad_compile_test.c
```diff
@@ -0,0 +1,52 @@
...
+#include <endian.h>
+#include <infiniband/umad.h>
+#include <infiniband/umad_types.h>
+#include <infiniband/umad_sm.h>
+#include <infiniband/umad_sa.h>
+#include <infiniband/umad_cm.h>
+
+int main(int argc, char *argv[])
+{
+	assert(__alignof__(union umad_gid) == 4);
```
bvanassche: Why are you using assert() instead of static_assert()? See also http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf.

bvanassche:
Hello Hal,

Have you noticed that I posted two review comments?

Thanks,

Bart.

On 03/27/2017 12:25 PM, hnrose wrote:

@bvanassche<https://github.com/bvanassche> , does this PR look OK to you ? Thanks.

hnrose:
@bvanassche ,
Added patches to address both comments.
Thanks.
-- Hal

bvanassche:
Hi Bart,

On 3/27/2017 4:29 PM, Bart Van Assche wrote:
> Hello Hal,
> 
> Have you noticed that I posted two review comments?

Yes, I just added 2 new patches to address those comments.

Thanks.

-- Hal

libibumad/tests/umad_compile_test.c
```diff
@@ -0,0 +1,65 @@
...
+#include <infiniband/umad_cm.h>
+
+#ifdef __cplusplus
+template <bool b> struct vki_static_assert { int m_bitfield:(2*b-1); };
+#define STATIC_ASSERT(expr) \
+	(void)(sizeof(vki_static_assert<(expr)>) - sizeof(int))
+#else
+#define STATIC_ASSERT(expr) (void)(sizeof(struct { int:-!(expr); }))
+#endif
+
```
jgunthorpe: We have ccan/build_assert.h - everythin should use it instead of open coding.

libibumad/tests/umad_compile_test.c
```diff
@@ -0,0 +1,65 @@
...
+int main(int argc, char *argv[])
+{
+#ifndef __CHECKER__
```
jgunthorpe: There is no need for a main() or an executable really, just put the static asserts in the global scope (they do not create anyhting) and link the resulting empty .o file into the shared library.

But this is OK too, just a bit strange.

jgunthorpe:
You can rebase and squash some of these commits and then force push your branch, it is getting a bit long doing and redoing the same stuff. 

It probably also make sense to just disab;e BUILD_ASSERT when using sparse globally, since sparse doesn't work right..  That would fix your travis failure.

hnrose:
I squashed the changes down.

Should srp_daemon also be converted to use ccan/build_assert.h rather than STATIC_ASSERT ?

hnrose:
@jgunthorpe ,
I squashed the changes down.

Should srp_daemon also be converted to use ccan/build_assert.h rather than STATIC_ASSERT ?

jgunthorpe:
Yes, someday srp should change to. Ideally we would use the standardized C11 name for this function.

PR #113, opened 2017-04-04
ARM 64bit build on travis

shamisp:
This is a follow up on the conversation with Leon during OFA workshop.
The patch uses Linaro cross-compilation tool-chain to build aarch64 rdma-core.
In the process, I also discovered an issue in mthca provider which caused travis failure on aarch64.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,11 +35,18 @@ addons:
       - python
       - valgrind
       - sparse
+      - wget
 
       # 32 bit support packages
       - gcc-multilib
       - lib32gcc-6-dev
 
+before_script:
+  - export LATEST_GCC_LINARO_URL=`wget -qO - https://releases.linaro.org/components/toolchain/binaries/latest/aarch64-linux-gnu/ | grep -o '<a href=['"'"'"].*gcc-linaro-.*x86_64_aarch64-linux-gnu.tar.xz['"'"'"]'  |  sed -e 's/^<a href=["'"'"']//' -e 's/["'"'"']$//'`
+  - export LATEST_GCC_LINARO_TAR=`basename $LATEST_GCC_LINARO_URL`
+  - wget -q http://releases.linaro.org/$LATEST_GCC_LINARO_URL
+  - mkdir $HOME/aarch64 && tar xf $LATEST_GCC_LINARO_TAR -C $HOME/aarch64 --strip 1
+  - rm $LATEST_GCC_LINARO_TAR
 script:
   - buildlib/travis-build
   - buildlib/github-release
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -244,6 +244,12 @@ class travis(APTEnvironment):
         cmds.append("RUN " + " && ".join(scmds));
         return cmds;
 
+    def get_before_script(self):
+        """Return a list of commands to run from before_script"""
+        cmds = ["RUN useradd -ms /bin/bash travis && \\"
+                "su -l -c %s"%(pipes.quote(" && ".join(self.yaml["before_script"]))) + " travis"];
+        return cmds
+
     def get_docker_file(self):
         # First this to get apt-add-repository
         self.pkgs = {"software-properties-common"}
@@ -255,6 +261,9 @@ class travis(APTEnvironment):
         # Package list from the travis.yml
         res.lines.append("RUN apt-get update && apt-get install -y --no-install-recommends %s"%(
             " ".join(sorted(self.yaml["addons"]["apt"]["packages"]))));
+
+        # Adding before_script commands
+        res.lines.extend(self.get_before_script())
 
         return res;
 
@@ -534,17 +543,18 @@ def run_travis_build(args,env):
                                    "--git-dir",os.path.join(opwd,".git"),
                                    "reset","--hard","HEAD"]);
 
-        home = os.path.join(os.path.sep,"home",os.getenv("LOGNAME"));
+        home = os.path.join(os.path.sep,"home","travis");
+        home_build = os.path.join(os.path.sep,home,"build");
 
         opts = [
             "run",
             "--read-only",
             "--rm=true",
-            "-v","%s:%s"%(tmpdir,home),
-            "-w",os.path.join(home,"src"),
-            "-u",str(os.getuid()),
+            "-v","%s:%s"%(tmpdir, home_build),
+            "-w",os.path.join(home_build,"src"),
+            "-u","travis",
             "-e","HOME=%s"%(home),
-            "-e","TMPDIR=%s"%(os.path.join(home,"tmp")),
+            "-e","TMPDIR=%s"%(os.path.join(home_build,"tmp")),
         ];
 
         # Load the commands from the travis file
@@ -564,7 +574,7 @@ def run_travis_build(args,env):
         if args.run_shell:
             opts.append("/bin/bash");
         else:
-            opts.extend(["/bin/bash",os.path.join(home,"go.sh")]);
+            opts.extend(["/bin/bash",os.path.join(home_build,"go.sh")]);
 
         docker_cmd(args,*opts);
 
--- a/buildlib/travis-build
+++ b/buildlib/travis-build
@@ -5,7 +5,7 @@ set -e
 # Echo all commands to Travis log
 set -x
 
-mkdir build-clang build32 build-sparse
+mkdir build-clang build32 build-sparse build-aarch64
 
 # Build with latest clang first
 cd build-clang
@@ -18,6 +18,11 @@ cd ../build32
 # travis's trusty is not configured in a way that enables all 32 bit
 # packages. We could fix this with some sudo stuff.. For now turn off libnl
 CC=gcc-6 CFLAGS="-Werror -m32" cmake -GNinja .. -DENABLE_RESOLVE_NEIGH=0
+ninja
+
+# aarch64 build to check compilation on ARM 64bit platform
+cd ../build-aarch64
+CC=$HOME/aarch64/bin/aarch64-linux-gnu-gcc CFLAGS="-Werror -Wno-maybe-uninitialized" cmake -GNinja .. -DENABLE_RESOLVE_NEIGH=0
 ninja
 
 # Run sparse on the subdirectories which are sparse clean


.travis.yml
```diff
@@ -40,6 +40,10 @@ addons:
       - gcc-multilib
       - lib32gcc-6-dev
 
+before_script:
+  - wget http://releases.linaro.org/components/toolchain/binaries/6.3-2017.02/aarch64-linux-gnu/gcc-linaro-6.3.1-2017.02-x86_64_aarch64-linux-gnu.tar.xz
```
rleon: Thank you a lot for your work.
It added less than 2 minutes to travis check which is pretty impressive.

I wonder if it is possible to use http://releases.linaro.org/components/toolchain/binaries/latest/ link and hardcoded version.

jgunthorpe: Well, it doubles the amount downloaded and increases the run time from ~4 mins to ~6 mins..

Maybe this would be better suited to run in cbuild? Someone can run the cbuild tests during the release cycle..

util/compiler.h
```diff
@@ -5,11 +5,13 @@
...
-   certainly too complex for a human to understand.
+   This is only enabled for old compilers and non-x86 architectures. gcc 6.x
+   and beyond have excellent static flow analysis for x86 platforms. If code
+   solicits a warning from 6.x it is almost certainly too complex for a human to
+   understand.
 */
-#if __GNUC__ >= 6 || defined(__clang__)
+#if (__GNUC__ >= 6 && !defined(__aarch64__)) || \
+    defined(__clang__)
 #define uninitialized_var(x) x
```
jgunthorpe: What warning do you get? Is this a bug in linaro gcc or does gcc really have different flow analysis for different targets?

.travis.yml
rleon: @jgunthorpe 6 minutes is good enough and it is better to run for every pull request. Anyway, we are not merging pull request immediately. The travis limit is 50 minutes.

shamisp: @rleon I intentionally used an explicit reference for the toolchain version in order to avoid potential issue introduced by updates from Linaro side. Also they don't have version file under the latest, so I would have to run curl + awk/sed to extract the name/version.

But overall, I don't have a strong opinion about this. If maintainers prefer to go with latest, I can update the script.

util/compiler.h
shamisp: Please see below:
```
../providers/mthca/qp.c: In function ‘mthca_arbel_post_send’:
../providers/mthca/qp.c:448:29: error: ‘f0’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  uint32_t uninitialized_var(f0);
                             ^~
../providers/mthca/qp.c:449:29: error: ‘op0’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  uint32_t uninitialized_var(op0);
                             ^~~
cc1: all warnings being treated as errors
[44/92] Building C object providers/mlx5/CMakeFiles/mlx5.dir/verbs.c.o
```

I can check with compiler people but my guess is that the static analysis flow is not the same.

.travis.yml
rleon: Let's wait for other opinions, but I personally prefer to go for the latest and see if it is stable. We can always move to specific version.

util/compiler.h
bvanassche: That should be documented either in the code or in the commit message because otherwise anyone who sees " && !defined(__aarch64__)" will wonder why it's there ...

shamisp: @bvanassche - sure. I will fix.

.travis.yml
shamisp: Not a problem. It may looks something like this:
```
linaro_latest=`wget -qO - https://releases.linaro.org/components/toolchain/binaries/latest/aarch64-linux-gnu/ | grep -o '<a href=['"'"'"].*gcc-linaro-.*x86_64_aarch64-linux-gnu.tar.xz['"'"'"]'  |  sed -e 's/^<a href=["'"'"']//' -e 's/["'"'"']$//'`
```

util/compiler.h
jgunthorpe: @shamisp - certainly this case has been fixed for a long time in x86. It is probably worth referring the Linaro compiler team to this example.. It would be interesting to know of upstream gcc has this problem too.

I'd rather you use -Wno-maybe-uninitialized in the travis wrapper than globally modify compiler.h because I assume this will eventually get fixed in gcc.

jgunthorpe:
This looks like it breaks 'cbuild pkg travis'.

A fix is to teach cbuild to run the before_script during image setup, this would get the arm compiler into the permanent docker image.

rleon:
@jgunthorpe in such case "latest" compiler won't be always "latest". We are not rebuilding images very often. Is it possible to put all these "before_script" instructions in buildlib/travsi-build, so it will be updated every time we run cbuild travis?

jgunthorpe:
@rleon, that is true for all the compilers from APT as well, no reason for this arm compiler to be different.

In my view, 'cbuild pkg travis' is far more useful if it runs quickly than if it takes 6 mins like on travis.com - this is why I prefer to shift as much as possible into the image build step, so it can be cached.

This is particularly painful for me since this download is done with wget and does not get cached by my existing apt/yum/zypper caching infrastructure.

rleon:
@jgunthorpe I got your point and the "funny" part of it that I forgot my use case for these docker images - offline work. It won't be possible with my proposal.

@shamisp so to summarize:
1. There is need to update the commit to use latest GCC compiler and drop commit with fix to mthca.
2. Need to run -Wno-maybe-uninitialized for ARM64 builds.
3. Need teach cbuild travis to handle "before_script" directive during image creation.

Thanks

shamisp:
@rleon sounds reasonable

shamisp:
@rleon @jgunthorpe please take a look.

buildlib/travis-build
```diff
@@ -20,6 +20,16 @@ cd ../build32
 CC=gcc-6 CFLAGS="-Werror -m32" cmake -GNinja .. -DENABLE_RESOLVE_NEIGH=0
 ninja
 
+# aarch64 build to check compilation on ARM 64bit platform
+cd ../build-aarch64
+if ! (which aarch64-linux-gnu-gcc &> /dev/null); then
+   AARCH64_PATH=`find / -type d -name 'gcc-linaro-*-x86_64_aarch64-linux-gnu' -print -quit 2>/dev/null | xargs echo`
+   export PATH=$PATH:${AARCH64_PATH}/bin
```
jgunthorpe: That find is really horrible.. Surely we can find some stable path in the travis environment to install some compiler symlinks into???

shamisp: @jgunthorpe it is not that horrible actually. 

The find is only invoked in the docker environment and even than since the gcc-linaro is unpacked in /, it find it immediately. On the other hand if one day you shuffle the build environment the script will continue to work.

jgunthorpe: Hmm. cbuild should emulate a bit more of real travis, setup HOME=/home/travis and then have before_script extract the compiler to $HOME/aarch64 and set things so that $HOME/aarch64/bin/ has the compiler binaries. 

Then just refer to $HOME/aarch64 in the build step.

Less risk that unpacking a tarfile from the internet in to the container's / will cause problems!

shamisp: @jgunthorpe Maybe I'm doing something wrong but as I see, the environment is a bit different.

if I run ```./buildlib/cbuild build-images travis``` it "sudo" and creates the image under root, while the root home is set to "/".

When I run ```./buildlib/cbuild pkg travis``` it runs everything under my current shell user (not root or travis), so the home is quite different.

Travis behavior is a bit different since it runs everything under ~travis so all the files are download to the "right" home. Since during the image creation cbuild I don't know what user is used, I just used default root home "/"

jgunthorpe: Ah, right, but there is no reason for that, it was just copy and pasted from the rpm/deb flow where it does make sense.

Add this hunk, and setup HOME=/home/travis before doing the RUN of the before_script and it will more closely match the real travis.org environment.

```
diff --git a/buildlib/cbuild b/buildlib/cbuild
index 6ccf07a4a722f5..6bb7353401a3e9 100755
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -534,7 +534,7 @@ def run_travis_build(args,env):
                                    "--git-dir",os.path.join(opwd,".git"),
                                    "reset","--hard","HEAD"]);
 
-        home = os.path.join(os.path.sep,"home",os.getenv("LOGNAME"));
+        home = os.path.join(os.path.sep,"home","travis");
 
         opts = [
             "run",
```

shamisp:
@jgunthorpe please take a look.

.travis.yml
```diff
@@ -35,11 +35,18 @@ addons:
       - python
       - valgrind
       - sparse
+      - wget
```
jgunthorpe: Can we use curl instead? Should be pre-installed?

.travis.yml
```diff
@@ -35,11 +35,18 @@ addons:
...
       - gcc-multilib
       - lib32gcc-6-dev
 
+before_script:
+  - export LATEST_GCC_LINARO_URL=`wget -qO - https://releases.linaro.org/components/toolchain/binaries/latest/aarch64-linux-gnu/ | grep -o '<a href=['"'"'"].*gcc-linaro-.*x86_64_aarch64-linux-gnu.tar.xz['"'"'"]'  |  sed -e 's/^<a href=["'"'"']//' -e 's/["'"'"']$//'`
+  - export LATEST_GCC_LINARO_TAR=`basename $LATEST_GCC_LINARO_URL`
+  - wget -q http://releases.linaro.org/$LATEST_GCC_LINARO_URL
+  - mkdir aarch64 && tar xf $LATEST_GCC_LINARO_TAR -C aarch64 --strip 1
+  - export PATH=$PATH:$PWD/aarch64/bin
 script:
```
jgunthorpe: No need for the path here, since it is below right?

.travis.yml
```diff
@@ -35,11 +35,18 @@ addons:
...
       # 32 bit support packages
       - gcc-multilib
       - lib32gcc-6-dev
 
+before_script:
+  - export LATEST_GCC_LINARO_URL=`wget -qO - https://releases.linaro.org/components/toolchain/binaries/latest/aarch64-linux-gnu/ | grep -o '<a href=['"'"'"].*gcc-linaro-.*x86_64_aarch64-linux-gnu.tar.xz['"'"'"]'  |  sed -e 's/^<a href=["'"'"']//' -e 's/["'"'"']$//'`
+  - export LATEST_GCC_LINARO_TAR=`basename $LATEST_GCC_LINARO_URL`
+  - wget -q http://releases.linaro.org/$LATEST_GCC_LINARO_URL
+  - mkdir aarch64 && tar xf $LATEST_GCC_LINARO_TAR -C aarch64 --strip 1
+  - export PATH=$PATH:$PWD/aarch64/bin
```
jgunthorpe: If we change this to 

```
mkdir $HOME/aarch64 && tar xf $LATEST_GCC_LINARO_TAR -C $HOME/aarch64 --strip 1
```

It will be a bit more robust against travis changes

buildlib/cbuild
```diff
@@ -244,6 +244,13 @@ class travis(APTEnvironment):
         cmds.append("RUN " + " && ".join(scmds));
         return cmds;
 
+    def get_before_script(self):
+        """Return a list of commands to run from before_script"""
+        cmds = ["RUN useradd -ms /bin/bash travis"]
+        cmds.append("RUN cd /home/travis && mkdir build && " + " && ".join(self.yaml["before_script"]));
+        cmds.append("RUN chown travis /home/travis/build");
```
jgunthorpe: Would this work?

```
cmds = ["RUN useradd -ms /bin/bash travis  && \\ "
        "su -l -c %s"%(pipes.quote(" && ".join(self.yaml["before_script"]))];
```
Eg run the before_script as the travis user as though run from a login shell

buildlib/travis-build
```diff
@@ -20,6 +20,12 @@ cd ../build32
 CC=gcc-6 CFLAGS="-Werror -m32" cmake -GNinja .. -DENABLE_RESOLVE_NEIGH=0
 ninja
 
+# aarch64 build to check compilation on ARM 64bit platform
+cd ../build-aarch64
+export PATH=$PATH:/home/travis/aarch64/bin
+CC=aarch64-linux-gnu-gcc CFLAGS="-Werror -Wno-maybe-uninitialized" cmake -GNinja .. -DENABLE_RESOLVE_NEIGH=0
```
jgunthorpe: This can probably just be:

```
CC=$HOME/aarch64/bin/aarch64-linux-gnu-gcc CFLAGS="-Werror -Wno-maybe-uninitialized" cmake -GNinja .. 
```

At the very least PATH should use $HOME

.travis.yml
jgunthorpe: .. also rm the download tar file, no need to carry that when using docker

.travis.yml
shamisp: I prefer the command line of wget. It is by default installed in travis image. I added wget for docker. So technically it does not add extra time.

.travis.yml
shamisp: Probably I can avoid it.

.travis.yml
shamisp: ok

.travis.yml
jgunthorpe: OK

buildlib/cbuild
shamisp: interesting idea :)

shamisp:
@jgunthorpe please see updated version of the code. Thanks for the comments. I think I addressed all the issues. THanks !

PR #116, opened 2017-04-12
srp_daemon/srp_daemon.c: Don't rely on attribute offset in get_shared_pkeys

hnrose:
get_shared_pkeys has been using SubAdmGet rather than SubAdmGetTable since
commit 2ad09524931dbf98d412e1912c1bdbf22f8ac81d
srp_daemon: Work around SM bug over non-default P_Key support

so RMPP is no longer used in response so it's not safe to
rely on AttributeOffset field. Good MAD status is sufficient
to say that valid PathRecord was returned.

Found-by: Honggang LI <honli@redhat.com>
using embedded subnet manager running on an Intel True Scale
Edge Switch 12300.

This has been broken since srptools-1.0.1 which was first
release containing commit mentioned above.

Also, added trivial patches for:
srp_daemon/srp_daemon.c: Eliminate some unneeded code in get_shared_pkeys
srp_daemon: Use consistent format when printing LID

### DIFF
--- a/srp_daemon/srp_daemon.c
+++ b/srp_daemon/srp_daemon.c
@@ -632,8 +632,8 @@ recv:
 			ret = umad_status(in_mad);
 			if (ret) {
 				pr_err(
-					"bad MAD status (%u) from lid %d\n",
-					ret, (uint16_t) be16toh(out_mad->hdr.addr.lid));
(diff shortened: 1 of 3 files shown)


hnrose:
Just updated the commit to say srptools-1.0.1 rather than 1.0.4.

Thanks

PR #118, opened 2017-04-18
Debian packaging changes from libibumad review

bdrung:
Here are some Debian packaging changes that I found while reviewing the changes from libibumad 	1.3.10.2-2 to rdma-core.


rleon:
@bdrung, we promised to be fast with PR for packaging, so we are :)

PR #126, opened 2017-05-08
rdma-core: define link speed and link width enums

dsharma283:
Currently application either need to define its own enum
values for link-speed and link-width or use hard coded
values. This patch inharits the enums from kernel space
ib_verbs.h and puts into verbs.h for ease of programming.

Signed-off-by: Devesh Sharma <devesh.sharma@broadcom.com>


dsharma283:
Will send out a V1.

PR #129, opened 2017-05-09
Implement "status" command for Debian srptools init script

bvanassche:
As discussed on the linux-rdma mailing list, add support for querying the status of the srp_daemon(s) to the srptools init script.

### DIFF
--- a/debian/srptools.init
+++ b/debian/srptools.init
@@ -24,6 +24,10 @@ LOG_DEFAULT=/var/log/srp_daemon.log
 
 [ -f /etc/default/srptools ] &&  . /etc/default/srptools
 
+max() {
+    echo $(($1 > $2 ? $1 : $2))
+}
+
 run_daemon() {
     local STATUS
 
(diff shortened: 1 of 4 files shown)


debian/srptools.init
```diff
@@ -9,83 +9,126 @@
...
-if [ "$PORTS" = "ALL" ]  ; then
-    for HCA_ID in `/bin/ls -1 ${IBDIR}`
-      do
-      for PORT in `/bin/ls -1 ${IBDIR}/${HCA_ID}/ports/`
-        do
-        run_daemon
-      done
-    done
-fi
+    STATUS=$(<"${IBDIR}/${HCA_ID}/ports/${PORT}/state")
```
jgunthorpe: Failing to start the daemon due to port state is racey and will lead to boot up failures. Ultimately srp_daemon itself has to monitor the port state...

bvanassche: The scope of this pull request is to add a status command to the Debian init script. The code your comment applies to is not new code but rework of existing code. In my opinion your comment falls outside the scope of this pull request. I will address port state monitoring separately.

jgunthorpe: Seems reasonable

jgunthorpe:
Our build process sets the execute bits during install so we could drop +x them from the git tree if desired.

bvanassche:
Hello Doug, I have dropped the patch that sets executable mode for the Debian init script from this series.

PR #139, opened 2017-06-02
Fix an inaccuracies about return value of rdma_getaddrinfo in manpage

gladkovdmitry17:
Added definition of possible error codes (the same as for getaddrinfo).

Signed-off-by: Dmitry Gladkov <dmitry.gladkov@intel.com>

### DIFF
--- a/librdmacm/rdma_cma.h
+++ b/librdmacm/rdma_cma.h
@@ -35,6 +35,7 @@
 #define RDMA_CMA_H
 
 #include <netinet/in.h>
+#include <netdb.h>
 #include <sys/socket.h>
 #include <infiniband/verbs.h>
 #include <infiniband/sa.h>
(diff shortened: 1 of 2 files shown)


librdmacm/rdma_cma.h
```diff
@@ -85,6 +86,32 @@ enum rdma_port_space {
...
+# define RAI_NOTCANCELED	EAI_NOTCANCELED /* Request not canceled */
+# define RAI_ALLDONE		EAI_ALLDONE	/* All requests done */
+# define RAI_INTR		EAI_INTR	/* Interrupted by a signal */
+# define RAI_IDN_ENCODE		EAI_IDN_ENCODE	/* IDN encoding failed */
+#endif
+/*
+ * rdma_getaddrinfo returns -1 (EAI_BADFLAGS in terms
+ * of getaddrinfo) and set system error in `errno'
+ */
+#define RAI_SYSTEM		EAI_BADFLAGS	/* System error returned in `errno' */
```
shefty: Do we need to define this explicitly as -1, rather than assuming the underlying definition is -1?

rleon:
@gladkovdmitry17 please post this patch to the linux-rdma mailing list. the rdma_cma.h will be published and I want to be sure that it will get right exposure.

Thanks

librdmacm/rdma_cma.h
gladkovdmitry17: Maybe something like that:
```
#if EAI_BADFLAGS != -1
#define RAI_SYSTEM -1
#else
#define RAI_SYSTEM EAI_BADFLAGS
#endif
```
And do this for all definitions (but some numeric equivalent instead of '-1').

gladkovdmitry17:
@rleon Sure, will do that today.

librdmacm/rdma_cma.h
rleon: @gladkovdmitry17 it will look awful

gladkovdmitry17: @rleon yes, I agree, but only one possible way is just define RAI_SYSTEM to -1. The drawback of this solution is if one of the underlying definition will be '-1', it will be look as a RAI_SYSTEM also.

rleon:
We are waiting for patches in ML.

Thanks

PR #140, opened 2017-06-05
Fix configuration and building with uClibc

abrodkin:
In uClibc some flags are not defined by default.
And that's the case for O_CLOEXEC for example, see
https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/tree/libc/sysdeps/linux/arm/bits/fcntl.h#n52

Its definition only happens if _GNU_SOURCE is defined.
Thus users of O_CLOEXEC must define _GNU_SOURCE before inclusion of
related headers.

This solves configuration issue:
-------------------------->8---------------------
-- Performing Test HAS_CLOEXEC
-- Performing Test HAS_CLOEXEC - Failed
CMake Error at CMakeLists.txt:198 (message):
  O_CLOEXEC/SOCK_CLOEXEC/fopen(..,"e") support is required but not found

-- Configuring incomplete, errors occurred!
-------------------------->8---------------------

And later compilation issue:
-------------------------->8---------------------
.../librdmacm/cma.c: In function 'rdma_create_event_channel':
.../librdmacm/cma.c:369:57: error: 'O_CLOEXEC' undeclared (first use in this function)
  channel->fd = open("/dev/infiniband/rdma_cm", O_RDWR | O_CLOEXEC);
                                                         ^~~~~~~~~
-------------------------->8---------------------

Signed-off-by: Alexey Brodkin <abrodkin@synopsys.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,6 +185,7 @@ RDMA_AddOptCFlag(NO_STRICT_ALIASING_FLAGS HAVE_NO_STRICT_ALIASING
 # supported so it really doesn't work right if this isn't available. Thus hard
 # require it.
 CHECK_C_SOURCE_COMPILES("
+ #define _GNU_SOURCE
  #include <sys/types.h>
  #include <sys/stat.h>
  #include <sys/socket.h>
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -30,6 +30,7 @@
  * SOFTWARE.
  */
 
+#define _GNU_SOURCE
 #include <config.h>
 
 #include <stdlib.h>


rleon:
Thanks

jgunthorpe:
This is not the right fix.

O_CLOEXEC is defined by POSIX so  ulibc is broken to require _GNU_SOURCE.

We do not set _GNU_SOURCE in very many files, so this only works by luck.

You need to directly detect this broken case and force _GNU_SOURCE on for all files. File a bug against uclibc as well to fix their defines.

abrodkin:
@jgunthorpe Hm, right. I was under impression O_CLOEXEC is yet another glibc's extension to POSIX but in reality it's a part of POSIX 2008 update. Which means uClibc indeed requires update in terms of moving O_CLOEXEC and friends from `#ifdef  __USE_GNU` to `#ifdef __USE_XOPEN2K8`.


I hope we'll get it fixed in uClibc soonish.


Adding uClibc-ng's maintainer Waldemar here so he's in the loop @wbx-github.

wbx-github:
Alex, will you provide a patch which fixes it similar to
glibc ./sysdeps/unix/sysv/linux/bits/fcntl-linux.h  for all architectures in uClibc-ng?

jgunthorpe:
@abrodkin In the mean time can you revise this solution to directly check if _GNU_SOURCE is needed and add it to the defines in cmake and largely revert the changes here? That will be more robust for future rdma-core changes.. Thanks

abrodkin:
@wbx-github I'll try to :)
@jgunthorpe Not sure I understood your thought here. What would be that condition when _GNU_SOURCE is required? Could you please elaborate on that a little bit?

jgunthorpe:
@abrodkin, sort of like this, untested
```
index 9ca0dcdbf65747..4942f057157bc7 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,7 +185,6 @@ RDMA_AddOptCFlag(NO_STRICT_ALIASING_FLAGS HAVE_NO_STRICT_ALIASING
 # supported so it really doesn't work right if this isn't available. Thus hard
 # require it.
 CHECK_C_SOURCE_COMPILES("
- #define _GNU_SOURCE
  #include <sys/types.h>
  #include <sys/stat.h>
  #include <sys/socket.h>
@@ -195,6 +194,26 @@ CHECK_C_SOURCE_COMPILES("
     socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0);
     return 0;
  }" HAS_CLOEXEC)
+
+if (NOT HAS_CLOEXEC)
+  # At least uclibc wrongly hides this POSIX constant behind _GNU_SOURCE
+  CHECK_C_SOURCE_COMPILES("
+ #define _GNU_SOURCE)
+ #include <sys/types.h>
+ #include <sys/stat.h>
+ #include <sys/socket.h>
+ #include <fcntl.h>
+ int main(int argc,const char *argv[]) {
+    open(\".\",O_RDONLY | O_CLOEXEC);
+    socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0);
+    return 0;
+ }" HAS_CLOEXEC_GNU_SOURCE)
+  if (HAS_CLOEXEC_GNU_SOURCE)
+    set(HAS_CLOEXEC 1)
+    add_definitions("-D_GNU_SOURCE")
+  endif()
+endif()
+
 if (NOT HAS_CLOEXEC)
   message(FATAL_ERROR "O_CLOEXEC/SOCK_CLOEXEC/fopen(..,\"e\") support is required but not found")
 endif()
diff --git a/librdmacm/cma.c b/librdmacm/cma.c
index eefb307d82ce57..7aaf45f2c90005 100644
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -30,7 +30,6 @@
  * SOFTWARE.
  */
 
-#define _GNU_SOURCE
 #include <config.h>
 
 #include <stdlib.h>
```

rleon:
@jgunthorpe Do you want me to revert it?

jgunthorpe:
Lets give @abrodkin a few days to see if the above patch can get sorted out.

rleon:
no problem

jgunthorpe:
@abrodkin BTW, uclibc does not work with rdma-core because symbol versions don't work properly through its tool chain..

abrodkin:
@jgunthorpe I'm not sure if I understood your thought. Care to explain in more details?

jgunthorpe:
@abrodkin Do you have questions about the patch I sent above??

abrodkin:
@jgunthorpe Nope that's super clear, thanks for that. I'm about to implement it and once done will send a pull-request.

jgunthorpe:
@abrodkin You are asking about uclib then? I helped someone else get ulibc to work and the conclusion was that it does not support symbol versions in its dynamic linker. Since rdma-core is always build with symbol versions turned on, this resulted in randomly mis-linking to the wrong verbs symbol at runtime, and resulting crashes.

I feel like this is a bug someplace in the uclibc world - as I would expect symver enabled libraries to be backwards compatible with non-symver dynamic linkers..

abrodkin:
@jgunthorpe Hm, that's an interesting point. Indeed I only did build testing and once got rdma-core built I passed my patch to another interested person who I supposed was going to really use it. Let me check with him if stuff really works for him in run-time.

abrodkin:
@jgunthorpe Ok your patch really helps. But now we're starting to see messages like:
```
[ 65%] Building C object srp_daemon/CMakeFiles/srp_daemon.dir/srp_daemon.c.o
/home/abrodkin/Projects/sources/git/rdma-core/srp_daemon/srp_daemon.c:40:0: warning: "_GNU_SOURCE" redefined
 #define _GNU_SOURCE
 <command-line>:0:0: note: this is the location of the previous definition

...

[ 72%] Building C object librdmacm/CMakeFiles/rspreload.dir/preload.c.o
/home/abrodkin/Projects/sources/git/rdma-core/librdmacm/preload.c:33:0: warning: "_GNU_SOURCE" redefined
 #define _GNU_SOURCE
 
<command-line>:0:0: note: this is the location of the previous definition
...
```

BTW commit 400122ef15bd "Be explicit about _GNU_SOURCE" exactly moves `_GNU_SOURCE` from makefiles to sources.

So at this point I'm a bit confused. If we act in the spirit of mentioned commit my existing solution IMHO looks better.

Any thoughts?

vennila-21:
@jgunthorpe I was facing the exact same problem with the symbol versions and uclibc dynamic linking. I got around the problem by building rdma core statically and linking the required libraries statically with my application. How exactly did you solve this problem?

Also I build rdma core with -DENABLE_STATIC=1, I see the following error. I got around this  by not building the mlx5 provider. But I am sure there must be fix for this.

CMake Error at buildlib/rdma_functions.cmake:130 (add_library):
  add_library cannot create target "mlx5" because another target with the
  same name already exists.  The existing target is a static library created
  in source directory
  "/nfs/sc/disks/fabric_work/vmegavan/rdma-next/output/build/librdmacore-14/providers/mlx5".
  See documentation for policy CMP0002 for more details.
  providers/mlx5/CMakeLists.txt:13 (rdma_shared_provider)


CMake Error at buildlib/rdma_functions.cmake:276 (get_property):
  get_property could not find TARGET mlx5-rdmav2.  Perhaps it has not yet
  been created.
  CMakeLists.txt:414 (rdma_finalize_libs)


CMake Warning (dev) at buildlib/rdma_functions.cmake:278 (get_target_property):
  Policy CMP0045 is not set: Error on non-existent target in
  get_target_property.  Run "cmake --help-policy CMP0045" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  get_target_property() called with non-existent target "mlx5-rdmav2".
  CMakeLists.txt:414 (rdma_finalize_libs)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at buildlib/rdma_functions.cmake:284 (get_property):
  get_property could not find TARGET mlx5-rdmav2.  Perhaps it has not yet
  been created.
  CMakeLists.txt:414 (rdma_finalize_libs)

jgunthorpe:
@abrodkin oops, my mistake, should read:

```
add_definitions("-D_GNU_SOURCE=")
```

As without the equals GCC will define it to '1'

jgunthorpe:
@vennila-21 it is necessary to remove all compat defintions from the source code, eg every

```
symver(__ibv_destroy_ah_1_0, ibv_destroy_ah, IBVERBS_1.0);
```

And possibly other cases.

Regarding the static linking failure, that is a bug for sure, we do not really test static linking as it is a fairly nonfunctional - how did you get the provider to work with static linking?

vennila-21:
@jgunthorpe I got around the problem by not building the mlx5 provider. (removed the add_subdirectory(providers/mlx5) from the CMakeLists.txt). But I am looking for the correct fix.  I occasionally run into this error even without -DENABLE_STATIC=1.

Also while building statically, the constructor that registers the provider gets optimized out. eg: in case of the hfi1 provider, 

 static __attribute__((constructor)) void hfi1_register_driver(void)
{
        verbs_register_driver("hfi1verbs", hfi1_driver_init);
}

So I had to use the --whole-archive option while linking the library to include all the symbols.

Just in case someone is trying a static build, that might help.

And as far as removing 
symver(__ibv_destroy_ah_1_0, ibv_destroy_ah, IBVERBS_1.0);
we don't want to make any changes to the source files as the source tarball gets downloaded and built by buildroot toolchain and we want it to work as is.

jgunthorpe:
@vennila-21 see #142 for the static library thing..

To properly support uclibc someone will have to fix symver support in ulibc's toolchain - this does not necessary require full symver support in the dynamic linker, but something to cause it match the default (@@) symbol instead of a random symbol would be enough.

Alternatively, someone would have to send a patch for rdma-core to disable all non-deafult symbols during the build, controlled by cmake -DDISABLE_COMPAT=1

vennila-21:
@jgunthorpe thank you for the quick fix. I will try it out.

abrodkin:
@jgunthorpe Do you think if `cmake -DDISABLE_COMPAT=1` approach is fine for any uClibc build?
I mean if uClibc is detected by defined __UCLIBC__ we simply disable compatibility options in rdma-core.
That should be simple and I may do it then shortly.

abrodkin:
@jgunthorpe Yep, solution with `add_definitions("-D_GNU_SOURCE=")` works perfectly fine.
Preparing a patch now and will send a pull-request in a moment.

jgunthorpe:
@abrodkin Well, it could be a place to start.

I am wondering if the symver stuff is a bit broken, ideally we should totally remove all symver things for uclibc *and* static libraries.. Lets see what patches look like I guess.

abrodkin:
And just FYI Waldemar took an action and updated uClibc-ng such that `O_CLOEXEC` and some others are now guarded by `__USE_XOPEN2K8` instead of former `__USE_GNU`. See https://cgit.openadk.org/cgi/cgit/uclibc-ng.git/commit/?id=197a82dd6d478b98cfe499b31cf6e2cfde682314
I assume the next release of uClibc-ng which is going to happen this weekend will have that fix.

PR #144, opened 2017-06-09
Common MMIO accessors 

jgunthorpe:
This is the version posted on the mailing list with comments applied, I was sitting on it until the next release came out due to the higher risk. 

Let's give Mellanox some time to check it out internally again. @rleon @yishaih


rleon:
Thanks

providers/mlx5/qp.c
```diff
@@ -942,8 +942,7 @@ static inline int _mlx5_post_send(struct ibv_qp *ibqp, struct ibv_send_wr *wr,
 			mlx5_bf_copy(bf->reg + bf->offset, (unsigned long long *)ctrl,
 				     align(size * 16, 64), qp);
 		else
-			mlx5_write64((__be32 *)ctrl, bf->reg + bf->offset,
-				     &ctx->lock32);
+			mmio_write64_be(bf->reg + bf->offset, *(__be64 *)ctrl);
```
yishaih: Didn't you miss here the endian handling that was previously done as part of mlx5_write64() ->MLX5_PAIR_TO_64 ? looks like mmio_write64_be just store the value.

providers/mlx4/qp.c
```diff
@@ -38,11 +38,12 @@
 #include <pthread.h>
 #include <string.h>
 #include <errno.h>
+#include <util/mmio.h>
 #include <util/compiler.h>
 
 #include "mlx4.h"
-#include "doorbell.h"
 #include "wqe.h"
+#include "mmio.h"
```
yishaih: Don't code use <util/mmio.h> ? (see above line 41)

yishaih:
Yes, this series is sensitive and must pass internal performance/regression testing before can be approved.
Please see few comments internally, in addition, please send to the list as well to let other people see,

jgunthorpe:
@yishaih It has been on the mailing list for months, v2 hasn't got any notable changes..

providers/mlx5/qp.c
jgunthorpe: No, it is not missed.

It is very obfuscated, but MLX5_PAIR_TO_64  is exactly equivalent to *(uint64 t*)val, eg a native 64 bit load. This means that mlx5_write64() is identical to mmio_write64_be(). 

Eg consider this program on a LE machine:

```
#include <stdint.h>
#include <assert.h>

int main(int argc,const char *argv[])
{
	uint8_t bytes[8] = {0,1,2,3,4,5,6,7};
	uint32_t *bytes32 = bytes;

	assert(*(uint64_t *)bytes == 0x0706050403020100);
	assert(bytes32[1] == 0x07060504);
	assert(bytes32[0] == 0x03020100);
	assert(((uint64_t) bytes32[1] << 32 | bytes32[0]) == 0x0706050403020100);
}
```

And this one on a BE machine:

```
int main(int argc,const char *argv[])
{
	uint8_t bytes[8] = {0,1,2,3,4,5,6,7};
	uint32_t *bytes32 = bytes;

	assert(*(uint64_t *)bytes == 0x0001020304050607);
	assert(bytes32[1] == 0x04050607);
	assert(bytes32[0] == 0x00010203)
	assert(((uint64_t) bytes32[0] << 32 | bytes32[1]) == 0x0001020304050607);
}
```

The native load and the macro produce identical results.

providers/mlx4/qp.c
jgunthorpe: No, at this particular point in the patch series the special mlx4 mmio.h still has required content, both are required to be included. At the memcpy patch mlx4/mmio.h is removed.

rleon:
@yishaih please say go/no go till EOW. It is washed enough time here and the default is "go".

Thanks

yishaih:
We can go with that series, thanks.

PR #145, opened 2017-06-12
Adding rdma service

rleon:


rleon:
@jarodwilson @bdrung @jgunthorpe @dledford please take a look.

Thanks

rleon:
file problem creating directory: /etc/rdma/

talatb:
my mistake, for some reason i didn't see this error on my setup, maybe need to add "etc/rdma/" in debian/rdma-core.install file.

talatb:
Need to remove the MAKE_DIRECTORY command in the CMake file.

jgunthorpe:
I would rather we go piece by piece and not just take everything in one lump. A lot of this is obsolete and should be deleted, or tied very closely to RH and should not be in Debian..

jarodwilson:
Yeah, what Jason said. Some of these were necessary several years ago, on very specific hardware (i.e., the mtrr mess on certain AMD setups), and not something you really want to start shipping just now. The mtrr one might be something to consider removing from the tree entirely, or at least nestling away somewhere that it's clear "you probably don't want this, but here it is if you really do". Doug would have to comment on the sriov bits, I've never actually used them, but they do appear potentially useful still in certain environments. The udev rules might be best to try to work into the default upstream udev rules. Otherwise, I don't really see a problem with all distros sharing anything that is still useful.

talatb:
I think so, we need to check who's using this files.
@jgunthorpe @jarodwilson could you please suggest the list of files that you think that should be removed?
 In addition we can send mail to the mailing list on that

jgunthorpe:
AFAIK the mtrr stuff is done by the kernel these days and the AMD 8131 stuff was only relevant to ipath which is removed from tree.

As I said on the list, I think we need to address some of these things by patching the kernel:

rdma.conf - Unclear if this applies to debian or not
rdma.cxgb3.sys.modprobe - If we are going to autoload iwarp then the kernel should do it, not modprobe
rdma.cxgb4.sys.modprobe
rdma.fixup-mtrr.awk - I think the modern kernel can rework the mtrrs, so this is obsolete
rdma.ifdown-ib - Pretty sure this stuff is RH specific
rdma.ifup-ib
rdma.kernel-init - Module loading needs to be thought about more carefully, eg we do not need to load IB only modules on !IB, this script is also not really hot plug safe.
rdma.mlx4-setup.sh - We certainly need something like. I'm hoping @rleon's rdma-tool will provide a common interface to setting the port wire protocol so this should be refactored into something device generic not mlx4
rdma.mlx4.conf 
rdma.mlx4.sys.modprobe - Same comment as for iwarp module loading
rdma.mlx4.user.modprobe - Seems reasonable to include as is
rdma.modules-setup.sh - Debian does not use dracut
rdma.service 
rdma.sriov-init - Not really sure what these do..
rdma.sriov-vfs
rdma.udev-ipoib-naming.rules - Seems fine to include as is
rdma.udev-rules

rleon:
@jgunthorpe re: rdma.mlx4-setup.sh
I briefly looked on it and it looks like devlink is already doing it
https://www.systutorials.com/docs/linux/man/8-devlink-port/
or should do 
devlink port set eth ...

dledford:
@rleon I thought devlink was for mlx5 ports only...

@jgunthorpe 
> rdma.conf - Unclear if this applies to debian or not

If you keep the init script, then this applies as the init script sources it.  If you use the systemd service, then the script started by the service also references it.

> rdma.cxgb3.sys.modprobe - If we are going to autoload iwarp then the kernel should do it, not
> modprobe

Arguable.  This file was originally part of libcxgb3, so you got iWARP support on the device if you have the corresponding user space library installed, but not if you didn't.  That made support an all or nothing affair with no ambiguous situations where the kernel supported iWARP connections but user space did not.  When libcxgb3 was absorbed into the rdma-core package, this came with it, but it still maintains the same all or nothing behavior.  So it depends on whether you want to answer user questions about "Why can the kernel create iWARP connections but program <blah> can not?"

> rdma.cxgb4.sys.modprobe

Ditto.

> rdma.fixup-mtrr.awk - I think the modern kernel can rework the mtrrs, so this is obsolete

Indeed.  When I wrote this the kernel did not ever change the MTRRs but left them as it found them from the BIOS.  Some time later the kernel started rewriting them to be correct, and that's when this became obsolete.  That's been long enough ago now that this can safely be removed.

> rdma.ifdown-ib - Pretty sure this stuff is RH specific
> rdma.ifup-ib

Yes, these both are.  However, if you are using sysv init scripts, these complement the ifdown-eth and ifup-eth scripts found in the sysv initscripts package.  I don't know what debian uses, so no clue if they should stay or not, but if debian uses sysv initscripts and you don't take them, then you likely need to update the ifup-eth and ifdown-eth scripts to add the features present in these scripts (pkeys mainly).

> rdma.kernel-init - Module loading needs to be thought about more carefully, eg we do not need to
> load IB only modules on !IB, this script is also not really hot plug safe.

This is the old sysv init script.  It predates hotplug, and really predates !IB.  But, if you use sysv initscripts, it's what you need.  I agree it could use some updating though.

> rdma.mlx4-setup.sh - We certainly need something like. I'm hoping @rleon's rdma-tool will provide a
> common interface to setting the port wire protocol so this should be refactored into something device
> generic not mlx4

Right.  This is entirely hardware specific and only should be kept as a stop gap measure at best if not dropped entirely.  Right now it only deals with port types.  For mlx5 based hardware, the preferred way to set the port types is using the mstconfig tool to set the port type in the card's firmware so that no runtime setup is needed.  As such, this might be a relic we could get rid of if the mstconfig tool can be used on mlx4 devices too.

> rdma.mlx4.conf

Only needed if we are doing runtime setting of port types instead of firmware setting of port types.

> rdma.mlx4.sys.modprobe - Same comment as for iwarp module loading
> rdma.mlx4.user.modprobe - Seems reasonable to include as is
> rdma.modules-setup.sh - Debian does not use dracut

Yep...

> rdma.service

Systemd service file, I assume debian doesn't need this?

> rdma.sriov-init - Not really sure what these do..
> rdma.sriov-vfs

These two are a holdover from when the upstream virt stuff could not manage mlx4 SRIOV VFs.  I can't say for certain that the virt stuff handles it properly today, but it's getting there if not there already.  You needed the init script to manage the setting of gids and pkeys on each VF, and the other file to configure what those gids and pkeys on each vf should be.  Of note is the fact that if you need these, then your guests are tied to specific VFs if you want them to always have the right gids/pkeys.  And migration is not supported with this method of setup.  This was really only to get SRIOV working when it wasn't really supported by qemu/kvm/libvirt on mlx4 devices.  Once these fully support managing the gids/pkeys out of a pool of VFs on their own, these files become totally obsolete.

> rdma.udev-ipoib-naming.rules - Seems fine to include as is
> rdma.udev-rules

jgunthorpe:
Right, so I think we are back at what I said to start, this migration needs to go piece by piece so we can understand and revise each component as necessary.

bdrung:
Before going to deep into reviewing the complete merge proposal, I second to migrate the service piece by piece, starting with the basic support.

dledford:
Then lets close this pull request and do a totally new one for each piece.

rleon:
@dledford 
My point was to show that such tool already exists.

talatb:
> @rleon I thought devlink was for mlx5 ports only...
> 
> @jgunthorpe
> 
> rdma.conf - Unclear if this applies to debian or not
> If you keep the init script, then this applies as the init script sources it. If you use the systemd service, then the script started by the service also references it.

This conf file is needed by the rdma service and this commit introduce the rdma service to all the distros,
currently RedHat and SUSE install the rdma service, i think that it is needed to all the distros.

> 
> rdma.cxgb3.sys.modprobe - If we are going to autoload iwarp then the kernel should do it, not
> modprobe
> Arguable. This file was originally part of libcxgb3, so you got iWARP support on the device if you have the corresponding user space library installed, but not if you didn't. That made support an all or nothing affair with no ambiguous situations where the kernel supported iWARP connections but user space did not. When libcxgb3 was absorbed into the rdma-core package, this came with it, but it still maintains the same all or nothing behavior. So it depends on whether you want to answer user questions about "Why can the kernel create iWARP connections but program can not?"
> 
> rdma.cxgb4.sys.modprobe
> Ditto.
> 
> rdma.fixup-mtrr.awk - I think the modern kernel can rework the mtrrs, so this is obsolete
> Indeed. When I wrote this the kernel did not ever change the MTRRs but left them as it found them from the BIOS. Some time later the kernel started rewriting them to be correct, and that's when this became obsolete. That's been long enough ago now that this can safely be removed.

Agreed. will remove the file and delete the lines that use it (rdma.service).
> 
> rdma.ifdown-ib - Pretty sure this stuff is RH specific
> rdma.ifup-ib
> Yes, these both are. However, if you are using sysv init scripts, these complement the ifdown-eth and ifup-eth scripts found in the sysv initscripts package. I don't know what debian uses, so no clue if they should stay or not, but if debian uses sysv initscripts and you don't take them, then you likely need to update the ifup-eth and ifdown-eth scripts to add the features present in these scripts (pkeys mainly).
> 
The commit didn't touch these files at all, it keeps them in redhat directory.

> rdma.kernel-init - Module loading needs to be thought about more carefully, eg we do not need to
> load IB only modules on !IB, this script is also not really hot plug safe.
> This is the old sysv init script. It predates hotplug, and really predates !IB. But, if you use sysv initscripts, it's what you need. I agree it could use some updating though.
> 
Ok, we can start with it and improve it in the near feature.

> rdma.mlx4-setup.sh - We certainly need something like. I'm hoping @rleon's rdma-tool will provide a
> common interface to setting the port wire protocol so this should be refactored into something device
> generic not mlx4
> Right. This is entirely hardware specific and only should be kept as a stop gap measure at best if not dropped entirely. Right now it only deals with port types. For mlx5 based hardware, the preferred way to set the port types is using the mstconfig tool to set the port type in the card's firmware so that no runtime setup is needed. As such, this might be a relic we could get rid of if the mstconfig tool can be used on mlx4 devices too.

Right, mstconfig (mlxconfig) is the modern tool that used for changing the port link type, it can also work on mlx4, as mention in the mstconfig man page "Supported  devices:  4th  Generation  devices: ConnectX3, ConnectX3-Pro (FW 2.31.5000 and above).  5th Generation devices: ConnectIB, ConnectX4, ConnectX4-LX, ConnectX5."
so we can definitely remove it.

> 
> rdma.mlx4.conf
> Only needed if we are doing runtime setting of port types instead of firmware setting of port types.
> 
We can remove it once we remove the  rdma.mlx4-setup.sh

> rdma.mlx4.sys.modprobe - Same comment as for iwarp module loading
> rdma.mlx4.user.modprobe - Seems reasonable to include as is
> rdma.modules-setup.sh - Debian does not use dracut
> Yep...
This commit keeps rdma.modules-setup.sh in redhat directory and there is another commit in this pull request that adds initramfs for debian.
"debian: Adding initramfs hook for debian"

> 
> rdma.service
> Systemd service file, I assume debian doesn't need this?
> 
AFAIK, debian are using systemd files, therefore I think it is needed, we can consider to improve it by adding more functionality like stop and restart.

> rdma.sriov-init - Not really sure what these do..
> rdma.sriov-vfs
> These two are a holdover from when the upstream virt stuff could not manage mlx4 SRIOV VFs. I can't say for certain that the virt stuff handles it properly today, but it's getting there if not there already. You needed the init script to manage the setting of gids and pkeys on each VF, and the other file to configure what those gids and pkeys on each vf should be. Of note is the fact that if you need these, then your guests are tied to specific VFs if you want them to always have the right gids/pkeys. And migration is not supported with this method of setup. This was really only to get SRIOV working when it wasn't really supported by qemu/kvm/libvirt on mlx4 devices. Once these fully support managing the gids/pkeys out of a pool of VFs on their own, these files become totally obsolete.
> 
> rdma.udev-ipoib-naming.rules - Seems fine to include as is
> rdma.udev-rules

larrystevenwise:
I still haven't added cxgb* kernel change to load the iwarp upper level driver (ULD).  I'm kind of hesitant to add this, because there are more than one ULD and why should the driver opportunistically just load the one?  Further,  I don't see the benefit of adding logic in the kernel to load ULDs.  IMO this should be done by the system based on the role of the system.

jgunthorpe:
@larrystevenwise I would like to see consistency across all the RDMA drivers. Right now some effectively auto load and some do not, and if you install the user space package then everything force loads at boot (used or not), except the stuff that hasn't been updated yet :( Basically it is an inconsistent mess.

Globally in the kernel we expect everything to demand autoload, that is the gold standard for modules, so I think you need a really good reason to have functionality that must be user loaded.

Ideally we would come up with some way to trigger demand loading of iwarp/rocee drivers from users space...

larrystevenwise:
@jgunthorpe Not just from userspace.  Kernel RDMA usage could also trigger loading.  IE we _could_ come up with some kernel design that triggers the loading based on when applications (kernel and user) need to utilize some device.  Something that comes to mind is the rdma-cm.  When it determines that a cm_id needs device X, it could somehow load the rdma provider for X if it is registered.  For applications listening on the wildcard address, all registered provider drivers would be loaded.  This would require low level pci drivers that support rdma to register with the rdma core the module that should be loaded.   Or maybe pci drivers that support rdma export some function pointer that gets called by the rdma core when an application wants to utilize a cm_id that routes to the pci device...  Just thinking out loud.

jgunthorpe:
@larrystevenwise Yes, I was thinking along the same too..

PR #146, opened 2017-06-12
A couple of bug fixes

ramranix:
Ram Amrani (2):
  qedr: Configure byte_len in read/send WC
  qedr: Convert immediate data to network order

### DIFF
--- a/providers/qedr/qelr_verbs.c
+++ b/providers/qedr/qelr_verbs.c
@@ -1583,6 +1583,7 @@ static int process_req(struct qelr_qp *qp, struct qelr_cq *cq, int num_entries,
 		case IBV_WC_RDMA_READ:
 		case IBV_WC_SEND:
 		case IBV_WC_BIND_MW:
+			wc->byte_len = qp->wqe_wr_id[qp->sq.cons].bytes_len;
 			DP_VERBOSE(cxt->dbg_fp, QELR_MSG_CQ,
 				   "POLL REQ CQ: IBV_WC_RDMA_READ / IBV_WC_SEND\n");
 			break;


providers/qedr/qelr_verbs.c
```diff
@@ -922,14 +922,17 @@ static inline void qelr_edpm_set_inv_imm(struct qelr_qp *qp,
...
 {
+	uint32_t data;
+
 	if (!edpm->is_edpm)
 		return;
 
-	memcpy(&edpm->dpm_payload[edpm->dpm_payload_offset],
-	       &inv_key_or_imm_data, sizeof(inv_key_or_imm_data));
+	data = htobe32(inv_key_or_imm_data);
+
```
jgunthorpe: The byte swapping here is nutzo..

```
                rwqe->imm_data = htobe32(htole32(wr->imm_data));
                qelr_edpm_set_inv_imm(qp, &edpm, rwqe->imm_data);
+	data = htobe32(inv_key_or_imm_data);
￼```

Can you fix this so it makes sense - and would pass sparse?

wr->imm_data is defined to be be32, so you probably just need to memcpy that and remove all the swaps completely.

**Comment by ramranix** [comment]
This certainly seems complex.
Where and how does sparse run? I'd like to check this myself too before I
send a patch.
(It seems Travis didn't mind).

Thanks,
Ram

**Comment by jgunthorpe** [comment]
sparse is turned off for your driver right now because of stuff like I quoted. To turn on sparse edit CMakeLists.txt and remove the no sparse annotation for qedr, travis will then run sparse on your driver.

```
add_subdirectory(providers/qedr) # NO SPARSE
```

To run sparse locally, just do

```
mkdir build-sparse
cd build-sparse
CC=cgcc cmake -GNinja ..
ninja

# see only qedr related sparse warnings
touch ../providers/qedr/*
ninja
```

providers/qedr/qelr_verbs.c
jgunthorpe: BTW, regarding 'inv_key_or_imm_data' , the verbs spec is that wc->imm_data is BE, and a NIC must achieve a certain order on the wire for compat.

However, when imm_data is being used as the invalidate_rkey (eg in SEND_INVALIDATE) then it is in host endian. 

I have a patch that adds a transparent union to make this sparse compatible.

rleon:
Let's close for now and reopen once you are ready.

PR #147, opened 2017-06-18
qedr: Configure byte_len in read/send WC

ramranix:


jgunthorpe:
This seems like something you'd want backported.. adding a fixes and a long description might help that process.

ramranix:
OK. I'll resend.

PR #150, opened 2017-06-21
[v2] qedr: Set immediate data sent via EDPM as is.

ramranix:
v2 adds a Signed-off-by.

Reminding of my comment in v1, that I am going to handle the Sparse errors after a re-code that I'll introduce soon.

***

Enhanced Direct Packet Mode doesn't require flipping of bytes as it
is sent directly to the transmitting HW and isn't processed by the
FW that requires specific order.

Fixes: c0965e4fe6fe ("libqedr (qelr) verbs")


rleon:
ok, let's see

ramranix:
I was just about to remove the '.'

Thanks.

Ram

rleon:
I removed it

PR #153, opened 2017-06-28
debian: Let rmda-core conflict with infiniband-diags < 2.0.0

bdrung:
rdma-core ships /usr/sbin/rdma-ndd which was part of infiniband-diags until versions 2.0.0. Thus rdma-core and infiniband-diags < 2.0.0 cannot be installed in parallel and needs to declare a conflict.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -23,6 +23,8 @@ Homepage: https://github.com/linux-rdma/rdma-core
 Package: rdma-core
 Architecture: linux-any
 Depends: ${misc:Depends}, ${shlibs:Depends}
+Breaks: infiniband-diags (<< 2.0.0)
+Replaces: infiniband-diags (<< 2.0.0)
 Description: RDMA core userspace infrastructure and documentation
  rdma-core provides the basic boot time support (eg systemd integration)
  for systems that use the Linux kernel's RDMA subystem.


hnrose:
This is not strictly true. It is true for default configuration of infiniband-diags < 2.0 but those can be built without --enable-rdma-ndd. Do we need to be this strict ? @weiny2

Also, typo: rdma-core -> rdma-core in title

jgunthorpe:
@hnrose, for Bebian it only matters how the infiniband-diags packages were built in Debian, not what options are possible upstream.

weiny2:
I don't understand Debian install configuration but does this really "replace" infiniband-diags < 2.0.0

Or is this just the nature of the mess I created?

dledford:
It doesn't replace infiniband-diags < 2.0.0, it conflicts with it.  So, as a matter of practice, an attempt to upgrade to the latest rdma-core without also upgrading infiniband-diags will result in a conflict and fail.  However, replacing an older rdma with the latest rdma-core while simultaneously upgrading infiniband-diags to anything => 2.0.0 will succeed.  It merely forces the packages to be updated in unison so things work properly.

bdrung:
@dledford summed it up nicely. @hnrose rdma-core could be installed in parallel with infiniband-diags < 2.0 if built without --enable-rdma-ndd (which is not the case for Debian).

PR #157, opened 2017-06-29
Device hot plug support

yishaih:
This series adds a device hot plug support based on the RFC and its follows discussion.

### DIFF
--- a/libibverbs/device.c
+++ b/libibverbs/device.c
@@ -59,41 +59,52 @@ int __ibv_get_async_event(struct ibv_context *context,
 			  struct ibv_async_event *event);
 void __ibv_ack_async_event(struct ibv_async_event *event);
 
-static pthread_once_t device_list_once = PTHREAD_ONCE_INIT;
-static int num_devices;
-static struct ibv_device **device_list;
+static pthread_mutex_t dev_list_lock = PTHREAD_MUTEX_INITIALIZER;
+static int initialized;
+static struct list_head device_list = LIST_HEAD_INIT(device_list);
 
-static void count_devices(void)
-{
-	num_devices = ibverbs_init(&device_list);
-}
 
 struct ibv_device **__ibv_get_device_list(int *num)
 {
-	struct ibv_device **l;
-	int i;
+	struct ibv_device **l = NULL;
+	struct verbs_device *device;
+	int num_devices;
+	int i = 0;
 
 	if (num)
 		*num = 0;
 
-	pthread_once(&device_list_once, count_devices);
+	pthread_mutex_lock(&dev_list_lock);
+	if (!initialized) {
+		int ret = ibverbs_init();
+		initialized = (ret < 0) ? ret : 1;
+	}
 
+	if (initialized < 0) {
+		errno = -initialized;
+		goto out;
+	}
+
+	num_devices = ibverbs_get_device_list(&device_list);
 	if (num_devices < 0) {
 		errno = -num_devices;
-		return NULL;
+		goto out;
 	}
 
 	l = calloc(num_devices + 1, sizeof (struct ibv_device *));
 	if (!l) {
 		errno = ENOMEM;
-		return NULL;
+		goto out;
 	}
 
-	for (i = 0; i < num_devices; ++i)
-		l[i] = device_list[i];
+	list_for_each(&device_list, device, entry) {
+		l[i] = &device->device;
+		i++;
+	}
 	if (num)
 		*num = num_devices;
-
+out:
+	pthread_mutex_unlock(&dev_list_lock);
 	return l;
 }
 default_symver(__ibv_get_device_list, ibv_get_device_list);
--- a/libibverbs/device.c
+++ b/libibverbs/device.c
@@ -99,6 +99,7 @@ struct ibv_device **__ibv_get_device_list(int *num)
 
 	list_for_each(&device_list, device, entry) {
 		l[i] = &device->device;
+		ibverbs_device_hold(l[i]);
 		i++;
 	}
 	if (num)
@@ -111,6 +112,10 @@ default_symver(__ibv_get_device_list, ibv_get_device_list);
 
 void __ibv_free_device_list(struct ibv_device **list)
 {
+	int i;
+
+	for (i = 0; list[i]; i++)
+		ibverbs_device_put(list[i]);
 	free(list);
 }
 default_symver(__ibv_free_device_list, ibv_free_device_list);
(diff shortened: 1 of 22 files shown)


libibverbs/device.c
```diff
@@ -59,41 +59,54 @@ int __ibv_get_async_event(struct ibv_context *context,
...
 
-	pthread_once(&device_list_once, count_devices);
+	pthread_once(&init_once, init_resources);
+	if (initialized < 0) {
+		errno = -initialized;
+		return NULL;
+	}
 
+	pthread_mutex_lock(&dev_list_lock);
+	num_devices = ibverbs_get_device_list(&device_list);
```
jgunthorpe: Since we grab a mutex we don't need the pthread_once, just do the init within this mutex.

libibverbs/device.c
```diff
@@ -262,6 +267,8 @@ struct ibv_context *__ibv_open_device(struct ibv_device *device)
 	context->cmd_fd = cmd_fd;
 	pthread_mutex_init(&context->mutex, NULL);
 
+	ibverbs_device_hold(device);
+
```
jgunthorpe: I feel like this should fail early here if the device is already known to be deleted and not grab another ref on it. eg consider the case where the dev node has been reclaimed by another device. There is a race here of some kind with the kernel. We need to do design this so that the kernel cannot re-use the cdev numbers or sysfs file names, and do the timestamp test appropriately to ensure that we are not opening the wrong instance of a ibv_device.

yishaih: This ref count is needed to keep the ibv_device memory regardless whether its kernel device was deleted. Cleaning the memory upon device deletion is wrong and may crash the application. 
Application can get the 'IBV_EVENT_DEVICE_FATAL' event to know about its device deletion and upon ibv_close_device the memory will be freed.

The rare race that the kernel device was deleted and was re-claimed by another device just before that ibv_open_device is called  is orthogonal to this series and might happen without this series as well.

This can be handled in some future support in kernel side or in the user area by using the timestamp creation just after calling cmd_fd = open(devpath, O_RDWR | O_CLOEXEC); as part of __ibv_open_device().

libibverbs/device.c
yishaih: This can be done, will upload soon a series with this small change.

libibverbs/device.c
jgunthorpe: I'm not talking about clearing memory, I am saying ibv_open_device should not return a device we already know to be gone. 

I disagree about the race, it is very relevant to this series and must be addressed here, like you say, by checking the timestamp in proper places. Previously we expressly did not support hotplug while verbs apps are running, now we do, so we can no longer ignore these fringe cases.

You need to think of something to make sysfs safe as well, maybe openat()...

yishaih: If we want to address that in this series, we can go with a simple solution in the user area as follows:
As part of __ibv_open_device just *after* that the cmd_fd was opened successfully we can make sure that this is still the sysfs device that the application asked to open by re-reading the timestamp and comparing to what was already set on the ibv_device, In case that TS was changed it points on some "other" device and the API will return an error. Otherwise we can safely continue to allocate the context by calling the kernel device. In case that by that time the device was somehow dropped the API will fail and an error will be returned as expected. 

Does it make sense ?

jgunthorpe: That is what I was asking for originally.

Plus you have to do some solution for everything that touches sysfs. Again, maybe openat() is the answer there.

> Otherwise we can safely continue to allocate the context by calling the kernel device. In case that by that time the device was somehow dropped the API will fail and an error will be returned as expected.

I think you also need to check that the device hasn't been discovered to be removed before even attempting to open.

yishaih: Why do we need a solution for everything that touches sysfs ? same TS on the same sysfs device name entry should guaranteed same device .
 In addition, extra checking before attempting to open looks redundant, as just after checking it might be dropped and there is no atomicy with kernel here. However, doing the check after opening the cmf_fd will make sure that if later the device will be dropped the alloc context command will fail and the application will get an error as we already agreed.

jgunthorpe: > Why do we need a solution for everything that touches sysfs ? same TS on the same sysfs device name entry should guaranteed same device .

It is the same basic problem, hotplug could re-use the sysfs files between charddev open and sysfs read, so that needs to be protected by a TS check also. This is why openat might be nice, instead of passing uverbs_sys_path as a char *, pass in a fd for  the base sysfs dir and use openat.  The common code can confirm that the dir fd was opened without races, and the kernel will not allow a new driver to 'take over' an open dir fd, AFAIK. So if there is a race then openat will just fail not return bogus data.

> In addition, extra checking before attempting to open looks redundant, as just after checking 

It seems necessary to get a consistent and proper errno code in all cases.

yishaih: We saw that despite that openat was called the kernel driver could destroy and re-create its uverbs file (e.g. uverbs0) and as such we didn't find any added value by using this system call.

We'll go with the TS check solution and upload a version once be ready.
Re errno: EIO will be returned in case TS was changed after opening the cmd_fd or in case the device was dropped as part of the alloc context command that follows the open.

As such, application can know that this was the case of device removal when this errno was set.

jgunthorpe: When the kernel recreats the /sys/.../uverbs0 it does so with a new directory inode (or it should). Since openat is inode based, the dirfd held open in libibverbs will continue to refer to the old, now empty,  directory from the original device, even though a path path based view will continue to see the new, replaced sysfs directory. The goal is not to block the kernel from replacing the directory, but to hold on to the original directory inode across the hotplug event. This allows a more transactional view of sysfs and is safe against hotplug - openat's on the dirfd will fail with ENOENT after a reuse event.

yishaih: We made some extra checking re the option to use the openat system call and found that it can be done. However, it requires holding  an open 'FD' per ibv_device.  In some clouds environments which uses SRIOV and many NIC(s) it might be large numbers of 'FD's which is a limited resource per process.  The alternative solution with TS doesn't require to hold any additional 'FD' and as such is preferred,

jgunthorpe: I think you should do as I suggested, and hold the fd during the open_device and close it once the device is successfully opened - particularly for the downcall into the driver. 

sysfs access outside the opendevice flow needs to be identified and handled via TS checking, I'm not sure how much of that there is..

yishaih: Open the FD as part of ibv_open_device is too late as it might open the new device and as such openat *won't* fail.  If TS is used by that step, there is no extra need to use open/openat/close.

jgunthorpe: You still need to protect all sysfs access, hold the dirfd across sysfs accesses. I have some rework in a branch here, let me finish it and you can see how it would work.

yishaih: We already agreed that TS should be used to detect dropped devices as part of get device list flow.

This discussion is for further cases which are not directly related to this series and can happen also today, we prefer going ahead and take the series, any further handling can be done on top of this series.

PR #158, opened 2017-06-29
mlx4: Cleanup resources upon device fatal

yishaih:
Cleanup driver resources upon device fatal on closing commands. (e.g.
destroy qp/srq/cq etc.)

### DIFF
--- a/providers/mlx4/mlx4.c
+++ b/providers/mlx4/mlx4.c
@@ -43,6 +43,8 @@
 #include "mlx4.h"
 #include "mlx4-abi.h"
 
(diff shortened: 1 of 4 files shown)


jgunthorpe:
This seems horrible, why was it merged?

yishaih:
We found this as a simple solution which matches our customers needs who uses mlx4,  it has no affect on other vendors.

jgunthorpe:
It is an ugly work around to some kind of bug, and the new environment variable didn't even get a man page entry. Bad Bad Bad.

PR #160, opened 2017-07-04
libqedr: fix sparse warnings

ramranix:
Patch 1:
  Properly use __be and __le types and their conversions.
  Do not use values stored in firmware structures.

Patch 2:
Enable sparse for libqedr.


rleon:
Thanks

PR #161, opened 2017-07-10
libqedr: fix bugs introduced in "sparse fixes"

ramranix:
Fix byte swapping function (pointer must advance).
Fix WQE size to be multiples of 16 bytes, not 8 bytes.
Pass a pointer to wqe_size to allow it to be updated.
Update the immediate value before the wqe is copied.

Fixes: 259ddded0e16 ("libqedr: fix sparse warnings")


rleon:
Thanks

PR #162, opened 2017-07-10
Debian packaging changes

bdrung:
This chunk of patches for the Debian package removes support for Debian 8 (jessie). Only Debian 9 (stretch) and newer will be supported. Are you fine with this change?

PS: I am currently testing the rdma-core package before the finally uploading it to Debian.


jgunthorpe:
Dropping  jessie means we cannot do build tests for any packaging in travis CI. Do you think this is worth giving up?

buildlib/cbuild also needs to be updated to remove jessie and add stretch.

bdrung:
No. It's better to keep the Travis CI builds. I removed these two commit. I will let the Debian package for Debian unstable divert by these two commits instead.

rleon:
@bdrung, it is hard to see in this UI what exactly did you drop.

bdrung:
I dropped
* debian: Switch to debhelper 10
* debian: Switch to automatically created -dbgsym packages

These two commits are not shown any more.

rleon:
Thanks

PR #163, opened 2017-07-11
Update check_type.h

byronyi:
The path of header included is wrong.

### DIFF
--- a/ccan/check_type.h
+++ b/ccan/check_type.h
@@ -52,7 +52,7 @@
 #define check_types_match(expr1, expr2)		\
 	((typeof(expr1) *)0 != (typeof(expr2) *)0)
 #else
-#include <ccan/build_assert/build_assert.h>
+#include <ccan/build_assert.h>
 /* Without typeof, we can only test the sizes. */
 #define check_type(expr, type)					\
 	BUILD_ASSERT_OR_ZERO(sizeof(expr) == sizeof(type))


byronyi:
I am porting libibverbs to another build system (bazel from Google, for TensorFlow) and discovered this include path issue.

Fixing it doesn't solve my problem though...porting the old standalone version of these libraries to bazel is much easier for my case, but mysteriously none of the standalone versions could work with my IB driver. I am looking for a vendor-neutral RDMA user libraries that I could build using some simple rules, specifically libibverbs and librdmacm. My own OFED vendor (Mellanox) distributes the private version of these libraries, so they couldn't be merged into an open source project...

jgunthorpe:
Those bazel scripts will not even build the old libraries entirely correctly.. At least it doesn't link them with the map file, which will result in a subtly broken shared library.

In principle you can make this work, but you need to have bazel duplicate all the pre-setup that cmake does under build/ before attempting to compile. This includes symlinking the include files, generating the config.h and setting up any shims to fix old distros. Then you just need bazel to largely duplicate the compiler options that cmake selects. (eg run ninja -nv to see all the steps). 

eg since you hit this error I can tell you don't have a correct config.h. Review what cmake produces.

The compiler and linker options are also important, you need to make sure bazel duplicates the important ones (such as the map file, and explicit linking)

byronyi:
Thanks for your comments! Seems to be a considerable effort that I cannot afford though...plus we need to build binaries for different distro and hardware platform...I think I will just leave it to the OFED distributed installed on the target environment.

By the way, is there any planned schedule to make rdma-core into the repository of the downstream distros?

rleon:
AFAIK,
SUSE is already shipping it, Fedora has it in rawhide and according to this https://github.com/linux-rdma/rdma-core/pull/162 pull request, the debian will be very soon.

Thanks

PR #166, opened 2017-07-13
Add rdma service for kernel boot support

bdrung:
Currently upstream does not provide a rdma or openibd service. Only the RedHat package ships a rdma service and rdma modules configuration files. The ibacm and srp_daemon services depend on the openibd service.

Make RedHat's rdma service available to all distros by cherry-picking the basic files for the rdma service to run and trim them down to the minimum. Do not pick workarounds or other quirk that might not needed any more. Then replace the openibd service dependency by the rdma service.


bvanassche:
Shouldn't the rdma / openibd service be split into multiple services - one for IPoIB, one for the SRP initiator driver, one for the SRP target driver, one for the iSER initiator driver, one for the iSER target driver, one for RDS and one for NFSoRDMA? I think that would allow us to get rid of the rdma.conf configuration file and also that this would allow systemd to load these services concurrently.

rleon:
@bvanassche wouldn't it be prone for initialization races?

bdrung:
How would you then configure what services you want to have?

bvanassche:
Since this patch has also been posted on the linux-rdma mailing list, let's continue the discussion there.

rleon:
Please open new PR.

PR #165, opened 2017-07-12
More sparse work for more providers

jgunthorpe:
This does the mellanox providers and vmw_pvrdma

PR #168, opened 2017-07-15
travis: Enforce kernel coding style on rdma-core

rleon:
The rdma-core library is tightly coupled to the linux-rdma kernel
and developed mostly by the same kernel developers. Enforcement of
the same coding style as used for kernel development has potential
to reduce errors and improve readability.

Despite the fact that checkpatch is written as general tool, the
differences between kernel and user space requires from us to
skip kernel types, file movement and execute permissions checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,7 @@ before_script:
   - rm $LATEST_GCC_LINARO_TAR
 script:
   - buildlib/travis-build
+  - buildlib/travis-checkpatch
   - buildlib/github-release
 deploy:
   # Deploy assets to Github releases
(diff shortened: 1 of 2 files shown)


rleon:
@dledford thanks

@jgunthorpe it is actually checks for SOB in the patches without need to install anything.

jgunthorpe:
Lets see how it goes.. checkpatch is often strangely opinionated..

jgunthorpe:
This aborts travis for any sparse warnings, not just errors, so I don't think it is really going to work well enough as is..

eg I re-based the #165 to trigger this new check and see how it goes off the rails with warnings.

rleon:
@jgunthorpe , I see.

We can ignore warnings and fail for errors only

rleon:
But it is bad, I don't know if we catch something with such limited check

rleon:
@jgunthorpe https://github.com/linux-rdma/rdma-core/pull/170

PR #170, opened 2017-07-19
travis: relax checkpatch checks by skipping warnings

rleon:
The commit ecc5ff0c5090 ("travis: Enforce kernel coding style on rdma-core")
was proven as too optimistic, but the quality of rdma-core doesn't allow
us to treat warnings as errors.

The following patch relaxes the checks for the warnings.

### DIFF
--- a/buildlib/travis-checkpatch
+++ b/buildlib/travis-checkpatch
@@ -2,7 +2,9 @@
 # Copyright 2017 Mellanox Technologies Ltd.
 # Licensed under BSD (MIT variant) or GPLv2. See COPYING.
 
-set -e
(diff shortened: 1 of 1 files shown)


jgunthorpe:
checkpatch is something that needs human review to decide if the warnings matter or not. To make that easier it would be nice if checkpatch could run as a dedicated github 'check' pass, but I don't know how to do that...

It isn't so much that rdma-core is not mature, it is that checkpatch is setup for the kernel and has lots of hardwired things. eg the const detector is totally broken for some reason.

rleon:
We can relax it even more by adding IGNORE_TYPES or in slightly different approach we can check specific things.

Right now, it will fail for errors only.

PR #172, opened 2017-07-24
Fixes for packaging

jgunthorpe:
These are various problems I found while working on the boot scripts.

@jarodwilson @bdrung @bvanassche


srp_daemon/srp_daemon.c
```diff
@@ -1619,7 +1644,8 @@ static int get_config(struct config_t *conf, int argc, char *argv[])
 	while (1) {
 		int c;
 
-		c = getopt(argc, argv, "caveod:i:j:p:t:r:R:T:l:Vhnf:");
+		optind = 1;
```
bvanassche: Shouldn't this assignment occur outside the while-loop?

jgunthorpe: Yep, I'm going to move these two patches to another series and post them to the list. This has gone beyond packaging changes

srp_daemon/srp_daemon.c
```diff
@@ -1950,6 +1950,60 @@ static void ts_sub(const struct timespec *a, const struct timespec *b,
...
+	sigaction(SIGINT, &sa, NULL);
+	sigaction(SIGTERM, &sa, NULL);
+	sigaction(SRP_CATAS_ERR, &sa, NULL);
+	return 0;
+
+close_pipe:
+	cleanup_wakeup_fd();
+	return -1;
+}
+
```
bvanassche: Nit: a single blank line here should be sufficient.

srp_daemon/srp_daemon.c
```diff
@@ -1991,12 +2045,16 @@ static int ibsrpdm(int argc, char *argv[])
 		}
 	}
 
+	ret = setup_wakeup_fd();
```
bvanassche: Please don't add a call to setup_wakeup_fd() here. ibsrpdm doesn't call get_received_signal() and hence doesn't need wakeup_pipe[] to be initialized.

srp_daemon/srp_daemon.c
```diff
@@ -2017,23 +2075,25 @@ static int ibsrpdm(int argc, char *argv[])
 	pthread_mutex_unlock(&res->sync_res->retry_mutex);
 
 	free_res(res);
+
 umad_done:
 	umad_done();
-
+cleanup_wakeup:
```
bvanassche: Since ibsrpdm doesn't use wakeup_pipe[], please also remove this cleanup call.

srp_daemon/srp_daemon.c
jgunthorpe: Got it

srp_daemon/srp_daemon.c
jgunthorpe: I did this only because the original code setup wakeup_fd before calling ibsrpdm, and I saw stuff in ibsrpdm that looks like it was using threads. If you are certain then I will drop it.

bvanassche: Yes, I'm sure that ibsrpdm doesn't need wakeup_pipe[] to be initialized and also that it would be an improvement if the signal actions are left unchanged for ibsrpdm. There are four calls to get_received_signal() in main() but these are all in code that cannot be reached when executed as ibsrpdm.

jgunthorpe:
Branch updated:
 - The srp_daemon LOG_PERROR and followup patches are dropped (will be in a new PR)
 - 'Call systemctl properly from udev' also fixes start_on_all_ports
 - New trivial patch 'srp_daemon: start_on_all_ports uses a bashism'

This is up to date with all requested changes on github and the mailing list for this series.

srp_daemon/start_on_all_ports
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
bvanassche: In the description, please consider changing "tart_on_all_ports" into "start_on_all_ports" and "compatability" into "compatibility".

srp_daemon/srp_daemon.rules
```diff
@@ -1 +1 @@
-ACTION=="add", SUBSYSTEM=="infiniband_mad", KERNEL=="*umad*", PROGRAM:="/usr/bin/systemctl show srp_daemon -p ActiveState", RESULT=="ActiveState=active", ENV{SYSTEMD_WANTS}+="srp_daemon_port@$attr{ibdev}:$attr{port}.service"
```
bvanassche: Please change "Reviwed-by" into "Reviewed-by".

srp_daemon/start_on_all_ports
jgunthorpe: got it

srp_daemon/srp_daemon.rules
jgunthorpe: got it thanks

debian/rules
```diff
@@ -34,6 +34,9 @@ override_dh_auto_test:
 
 override_dh_auto_install:
 	DESTDIR=$(CURDIR)/debian/tmp ninja -C build-deb install
+	# Debian kmod has a conf file called mlx4.conf that unnecessarily
+	# tries to autoload mlx4_en, call our conf file rdma-mlx4.conf instead
+	mv $(CURDIR)/debian/tmp/etc/modprobe.d/mlx4.conf $(CURDIR)/debian/tmp/etc/modprobe.d/rdma-mlx4.conf
```
bdrung: mlx4.conf is only part of Ubuntu's kmod package. See https://launchpad.net/ubuntu/+source/kmod/24-1ubuntu1 and https://bugs.launchpad.net/maas/+bug/1115710

I am against renaming the file. kmod should stop providing that file and rely on rdma-core.

bdrung:
Besides my two smaller comments on the mailing list and my comment about the rename of mlx4.conf, I am fine with the changes (I skipped the redhat-prefixed ones)

debian/rules
talatb: We already open bug [1] on this issue in kmod package,  and i'll ask them to fix it.

https://bugs.launchpad.net/ubuntu/+source/kmod/+bug/1693503

jgunthorpe: Okay then.

@talatb To be clear, the mlx4.conf in kmod is *not* replaced by the one in rdma-core, they do totally different things.

The mlx4.conf in kmod is attempting to autoload ethernet support. I believe that the kernel now does that automatically via request_module.

Also, I haven't moved redhat/rdma.mlx4.sys.modprobe (which does the mlx4-setup thing) into common code because Mellanox said they prefer to do this setup another way. If so I recommend removing it from redhat as well.

jgunthorpe:
All comments have been addressed and the conflicts with the just merged branch resolved.

bdrung:
You missed one, see pull request #175

jgunthorpe:
@brung, I got that one, I just did it wrong :(

PR #173, opened 2017-07-26
Various related fixes for daemons

jgunthorpe:
These were found during my kernel-boot work
- Broadly make iwpmd work sanely as a daemon with systemd, do not fork, use the right daemon name
  for syslog, and do not send stderr to /dev/null
- Fix duplicated log messages in srp_daemon by removing LOG_PERROR when in systemd mode

@bvanassche @tatyana-en @larrystevenwise


larrystevenwise:
iwpmd changes look fine

PR #180, opened 2017-08-03
debian: Merge iwpmd into rdma-core

jgunthorpe:
Now that iwpmd automatically runs only if there is iWarp hardware
installed, like rdma-ndd, there is no much point in keeping the
extra package. iwpmd is new feature to Debian packaging so there
is no issue with compat.

What do you think @bdrung ?


debian/rdma-core.lintian-overrides
```diff
@@ -1,2 +1,3 @@
 # The rdma-ndd service is started by udev.
 rdma-core: systemd-service-file-missing-install-key lib/systemd/system/rdma-ndd.service
+rdma-core: systemd-service-file-missing-install-key lib/systemd/system/iwpmd.service
```
bdrung: Sort these please

bdrung:
Besides my one comment, I approve this merge proposal. I tested that the iwpmd services is not loaded on a mlx4 system.

jgunthorpe:
Updated with @bdrung's comment and @dledford's merge resolution

PR #182, opened 2017-08-09
verbs: Change the provider library file name suffix

jgunthorpe:
This should have been changed when the private provider API was
introduced..

The intent of the 'rdmav2' suffix was to indicate the provider is ABI
compatible with the libibverbs loading it, with the new scheme the only
providers which are compatible have the same ABI private version.

Change the suffix to be rdmav15 to match the current private version,
and set things up to be able to track the map file/etc.


rleon:
thanks

PR #184, opened 2017-08-17
mlx5: Return pointer to CQ doorbell

yishaih:
Small change in mlx5 driver to return the CQ doorbell directly.


yishaih:
There is no ABI/API change but just a bug fix in mlx5 driver.
The API talked about void* but by mistake the mlx5 code returned void**, this patch fixes that.
We don't care about potential break for direct mlx5 applications in this arm_cq flow as code was buggy and couldn't be used properly.

Taking this patch, see mailing list as well.

PR #188, opened 2017-08-23
mlx5: Fix ABI break from revising the UAR pointer

jgunthorpe:
Provide two implementations of mlx5dv_init_obj, one that has the
historical behaviour that has existed until now of returning the
void **uar and a new version that returns the 'void *' version
renamed to arb_db.

Apps that use this feature must refer to it as arb_db, they will not
compile on pre-rdma-core 15 releases, and they will not dynamically
link to old versions either. This provides a sane level of safety for
the end users of this library.

Fixes: c6e3439aaa93 ("mlx5: Return pointer to CQ doorbell")

### DIFF
--- a/providers/mlx5/mlx5.c
+++ b/providers/mlx5/mlx5.c
@@ -42,6 +42,8 @@
 #include <sched.h>
 #include <sys/param.h>
 
+#include <util/symver.h>
+
 #include "mlx5.h"
 #include "mlx5-abi.h"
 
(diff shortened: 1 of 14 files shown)


providers/mlx5/mlx5.c
```diff
@@ -716,7 +716,8 @@ static int mlx5dv_get_srq(struct ibv_srq *srq_in,
 	return 0;
 }
 
-int mlx5dv_init_obj(struct mlx5dv_obj *obj, uint64_t obj_type)
```
rleon: This removal is supposed to break static compilation.

jgunthorpe:
Updated to fix static compilation.

PR #186, opened 2017-08-21
cbuild updates to better support opensuse

jgunthorpe:
There were tested by @nmorey 

This can probably be merged after the release.

PR #189, opened 2017-08-24
Add SUSE Spec file and us it for SUSE targets

nmorey:
This series includes PR https://github.com/linux-rdma/rdma-core/pull/186 to avoid conflicts

 @jgunthorpe: Notifying you for awareness


rleon:
@nmorey please rebase, I merged PR #186, to have conversations preserve in the context of discussion.

thanks

nmorey:
@rleon done

suse/rdma-core.spec
```diff
@@ -0,0 +1,698 @@
...
+Device-specific plug-in ibverbs userspace drivers are included:
+
+- libcxgb3: Chelsio T3 iWARP HCA
+- libcxgb4: Chelsio T4 iWARP HCA
+- libhfi1: Intel Omni-Path HFI
+- libhns: HiSilicon Hip06 SoC
+- libi40iw: Intel Ethernet Connection X722 RDMA
+- libipathverbs: QLogic InfiniPath HCA
+- libmlx4: Mellanox ConnectX-3 InfiniBand HCA
+- libmlx5: Mellanox Connect-IB/X-4+ InfiniBand HCA
```
rleon: Connect-IB/X-4+ -> Connect-IB/X-4/5/6+

suse/rdma-core.spec
```diff
@@ -0,0 +1,698 @@
...
+Obsoletes:      ibacm-devel < %{version}-%{release}
+
+%description devel
+RDMA core development libraries and headers.
+
+%package -n     libibverbs
+Summary:        Library & drivers for direct userspace use of InfiniBand/iWARP/RoCE hardware
+Group:          System/Libraries
+Requires:       %{name}%{?_isa} = %{version}-%{release}
+Provides:       libcxgb3-rdmav2 = %{version}-%{release}
```
nmorey: I dont think it matters.
The goals is this is to handle updates from the old rdma packages where we had a -rdmav2 for each provider.
We don't expected people (and we don't either) who switched to the new rdma-core repo to have RPM dependencies to <provider>-rdmav<rdma_version>

suse/rdma-core.spec
nmorey: The descriptions are a direct copy of the redhat spec file. It might be worth fixing  both specs in a later patch.

nmorey:
Do I need to modify another script for Travis ?
I'll post the patches on the ML

nmorey:
Should I add another section in the MAINTAINERS for the suse directory or add my name to the "OVERALL PACKAGE" one ?

rleon:
There is buildlib/travis-build for Travis runs, there is need to take a look there. You can test your changes to that file with "cbuild travis" script.

Regarding supported cards - no problem - I'll fix it all other the place.

Please add another section for SuSE and I'll update Debian and RH after that.

MAINTAINERS
```diff
@@ -40,6 +40,11 @@ F:	*/CMakeLists.txt
 F:	*/lib*.map
 F:	buildlib/
 
+SUSE PACKAGING
```
rleon: It is supposed to be sorted in alphabetical order.

suse/rdma-core.spec
rleon: @nmorey I don't know much about packaging, this line "Provides:       libcxgb3-rdmav2" and the word "provides" worried me.

nmorey: It's RPM plumbing.
Basically the Obsolete will force the installation of libibverbs if the user had one of the -rdmav2 package installed when updating.
The Provides is simply so packages who had dependencies to those can still be installed.
I'll ask our QA team if these can be dropped

MAINTAINERS
nmorey: OK. Will fix.

nmorey:
Unless I'm missing something, none of the spec files are tested in travis.
What would you want me to test ?
The obvious choice would be to run a `cbuild pkg leap` but would it work in Travis ?

rleon:
We are testing debian, but I see this Jason's comment in the code:
"the RPM stuff will have to be audited by hand."

So i'm not sure now that spec can be tested in the travis.

jgunthorpe:
Testing the spec files in travis is hard.. We'd need to have travis build docker containers on the fly which is very very expensive.

If we want to do better testing I think we'd need to stop using travis and using something like Google Cloud Container Builder which allows us to supply our own baseline docker containers instead of using being forced to use the crappy one travis provides.

But that is a much larger and complex project for some other day....

nmorey:
@jgunthorpe My last company used to have integration jobs that lasted 6 hours so building docker images did not seem so bad :)

It wouldn't work for PR but for the master branch, it should be possible to hook something in OBS.
I know there are several SUSE projects on github that automatically triggers package updates in SUSE open build system.
It means there will be another build report to check but it could be a short term workaround.

@rleon  So does this mean you're happy with this PR ?
If yes, I'll post it to the ML

rleon:
@jgunthorpe we need to try https://packagecloud.io/ to build packages, it is in my back log :)

@nmorey I think so.

nmorey:
Patches submitted to linux-rdma ML

jgunthorpe:
@rleon, from what I can understand about package cloud it just handles the download side, you still need a CI system to run the build and upload the actual package files to it..  

eg cbuild can produce all the necessary package files, with a bit of extra plumbing it could upload them to packagecloud, and someone could run that manually after ever release.

The main failing of travis is that it does not let use specify which baseline distro the build steps run on, so we are trapped with the ancient trusty based image they use. Something like CircleCI or GCCB lets you specify a docker container as the base distro for each build step, so we could run builds under centos, suse, xenial, etc just like cbuild does. Those builds steps can produce package files and we could upload them to package cloud. 

But we can't do anything like that until we get a more suitable CI system.. 

Which is a project for another day ;)

jgunthorpe:
No, don't close it, we will merge it from github.. The ML posting is just to solicit a broader audiance. If there are comments then force push an update here and leave a note.

nmorey:
Ok. Please don't merge anything for the moment. i saw some weird systemd warnings on uninstall.
i'll check that out ASAP

nmorey:
Ok there are some issues with the services. I'll have to update the series.

I'm having another issue with the new rdma-load-modules@.service
@jgunthorpe The path to systemd-modules-load is hardcoded to /lib/systemd.
This works in debian and RHEL/Fedora (because of their symlink from /lib to /usr/lib) but does not work for us as it is in /usr/lib/systemd (without the RHEL symlink)

I'm not sure what is the best way to fix this ?
I think a CMAKE_INSTALL_SYSTEMD_BINDIR could be added (defaults to /lib/systemd).

jgunthorpe:
All looks fine to me.. We should merge this before 15 due to the udev.md at least..

rleon:
@jgunthorpe , agree it is a lot of work.
Regarding packagecloud.io, it is used together with https://github.com/packpack/packpack as an engine to create packages.

A couple of weeks before I researched the possibility to use https://aws.amazon.com/codebuild/ but stopped there.

jgunthorpe:
@rleon packpack is basically a slightly different version of buildlib/cbuild...

AWS codebuild looks a lot like circle ci and GCCB: you give it a docker container and it runs commands for you in it on a compute node. All the same stuff these days

It is all basically the same stuff now.. We know what docker containers we need because cbuild uses them.. The next step is to pick a docker-based replacement for travis and move over..

It also looks like all of these services need a container registry too. We really should get OFA to fund a AWS/GCE account for these sorts of things.

rleon:
@jgunthorpe I prefer to avoid OFA involvement in this upstream project.

I didn't calculate exactly the projected costs, but I don't expect more than 10$ per month.
https://cloud.google.com/container-registry/pricing

jgunthorpe:
@rleon the OFA is the best source of monies.. Having one vendor pay looks bad, even if it is small. We should talk about this at plumbers.

rleon:
@jgunthorpe, There is no one vendor pay right now, and I don't have plans to involve my employee in payments.

I think that Linux Foundation is the best choice to support such project.

PR #190, opened 2017-08-24
cbuild: fix return code handling

nmorey:
Properly return error codes from commands/subcommands (make, pkg, build-images)
so cbuild exit code makes sense.
Note that docker_cmd can only returns 0 as it raises an exception on failure.

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -480,7 +480,7 @@ os.symlink({tarfn!r},os.path.join("SOURCES",tarfn));
         else:
             opts.extend(["python","go.py"]);
 
-        docker_cmd(args,*opts)
+        res = docker_cmd(args,*opts)
 
         print
         for path,jnk,files in os.walk(os.path.join(tmpdir,"RPMS")):
@@ -488,6 +488,7 @@ os.symlink({tarfn!r},os.path.join("SOURCES",tarfn));
                 print "Final RPM: ",os.path.join("..",I);
                 shutil.move(os.path.join(path,I),
                             os.path.join("..",I));
+        return res
 
 def run_deb_build(args,env):
     image_id = get_image_id(args,env.image_name());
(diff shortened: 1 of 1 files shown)


nmorey:
@jgunthorpe  I'm no python expert so your feedback is welcome.
With this fix, it should be possible to use cbuild pkg in travis build to check RPM packaging (and debian too)

jgunthorpe:
I don't understand this patch.. Since docker_cmd always returns 0 why are we adding more paths to propagate the zero up?

This is python, if things fail then we should throw exceptions not do return codes.

What is the goal here? The top level cbuild  already fails if the docker commands fail:

```
  File "buildlib/cbuild", line 785, in <module>
    if not args.func(args):
  File "buildlib/cbuild", line 614, in cmd_pkg
    run_travis_build(args,env);
  File "buildlib/cbuild", line 604, in run_travis_build
    docker_cmd(args,*opts);
  File "buildlib/cbuild", line 361, in docker_cmd
    return subprocess.check_call(["sudo","docker"] + cmd);
  File "/usr/lib/python2.7/subprocess.py", line 541, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['sudo', 'docker', 'run', '--read-only', '--rm=true', '-v', '/tmp/tmpIHv60e:/home/travis/build', '-v', '/home/jgg/work/plumbing/.git:/home/travis/build/src/.git:ro', '-w', '/home/travis/build/src', '-u', '2009', '-e', 'TRAVIS_COMMIT_RANGE=ff13a6f3ab1f6e220ea637d389186736ad859e5b..HEAD', '-e', 'HOME=/home/travis', '-e', 'TMPDIR=/home/travis/build/tmp', 'build-rdma-core/travis', '/bin/bash', '/home/travis/build/go.sh']' returned non-zero exit status 1
```

nmorey:
The thing is cbuild also fails if docker worked...
If I run cbuild pkg leap, everything is OK but the cmd_pkg function returns None which causes the script to exit(100)
This propagates proper return codes.
It is not necessary for docker (although it doesn'nt hurt) but it seemed better than a hardcoded return 0.

jgunthorpe:
Oh I understand, that final if is just garbage..

Does this do what you want?

https://github.com/jgunthorpe/rdma-plumbing/commit/649ed18524b7f0f37e40ada9d3bdfb97d8497838

jgunthorpe:
Replaced by #197

PR #191, opened 2017-08-24
Improve symbol version macros

jgunthorpe:
This allows libibverbs to be built statically and actually work reliably. When building statically the compat symbols will be totally discarded and normal un-versioned symbols are provided for the latest symbol.

This also fixes the problems with uclibc by providing an operating mode that produces shlibs that do not include any compat symbols, allowing them to work properly with dynamic linkers that  do not support symbol versions. @abrodkin @vennila-21


jgunthorpe:
The checkpatch error is bogus, ignore the travis failure.

PR #192, opened 2017-08-24
verbs: Encode the private ABI version number in verbs_register_driver

jgunthorpe:
This causes the symbol table to look like:

   208: 000000000000c000   142 FUNC    GLOBAL DEFAULT   13 verbs_register_driver_15@@IBVERBS_PRIVATE_15

Which encodes the expected private ABI version inside the symbol name
as well as with the symbol version.

This ensures that the ABI version is checked even if the linking
environment does not include symbol versions, for instance if an
end user is linking a provider static library to the system dynamic
libibverbs.


jgunthorpe:
The check patch errors are bogus, ignore the travis failure.

PR #193, opened 2017-08-28
Fix mthca/qib support in udev

nmorey:


nmorey:
This should probably go to v15.
I need to run a few more tests but it seems that with those, I should be able to drop the redhat scripts fro; the suse RPM

PR #194, opened 2017-08-28
travis: Limit checkpatch checks on pull requests only

rleon:
Checkpatch is not always reliable and it forces us to merge pull
requests despite the fact that they are marked as failed.

The travis triggers for every such merge and propagates such wrong
status to master branch.

Let's avoid this check on any type of git update except pull request.

### DIFF
--- a/buildlib/travis-checkpatch
+++ b/buildlib/travis-checkpatch
@@ -17,7 +17,7 @@ if [ "x$TRAVIS_COMMIT_RANGE" != "x" ]; then
 	wget -q https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/scripts/checkpatch.pl \
 	        https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/scripts/spelling.txt
 	DIR_FOR_PATCHES_TO_CHECK=$(mktemp -d)
-	git format-patch --no-cover-letter $TRAVIS_COMMIT_RANGE -o $DIR_FOR_PATCHES_TO_CHECK/
+	git format-patch --no-cover-letter $TRAVIS_COMMIT_RANGE ^$TRAVIS_BRANCH -o $DIR_FOR_PATCHES_TO_CHECK/
 	CHECKPATCH_OPT="--no-tree --ignore PREFER_KERNEL_TYPES,FILE_PATH_CHANGES,EXECUTE_PERMISSIONS,USE_NEGATIVE_ERRNO,CONST_STRUCT $DIR_FOR_PATCHES_TO_CHECK/*"
 	perl checkpatch.pl $CHECKPATCH_OPT
 	if [ $? -ne 0 ]; then
--- a/buildlib/travis-checkpatch
+++ b/buildlib/travis-checkpatch
@@ -2,6 +2,12 @@
 # Copyright 2017 Mellanox Technologies Ltd.
 # Licensed under BSD (MIT variant) or GPLv2. See COPYING.
 
+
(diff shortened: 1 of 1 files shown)


nmorey:
There is probably another fix needed. The job for this build in Travis checked 12 patches becuase the TRAVIS_COMMIT_RANGE include the commits already in master but not in your branch (some patches were applied before you forked)

Maybe something like this:
```
diff --git a/buildlib/travis-checkpatch b/buildlib/travis-checkpatch
index ef7df4e6..75040c1e 100755
--- a/buildlib/travis-checkpatch
+++ b/buildlib/travis-checkpatch
@@ -11,7 +11,7 @@ if [ "x$TRAVIS_COMMIT_RANGE" != "x" ]; then
    wget -q https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/scripts/checkpatch.pl \
            https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/scripts/spelling.txt
    DIR_FOR_PATCHES_TO_CHECK=$(mktemp -d)
-   git format-patch --no-cover-letter $TRAVIS_COMMIT_RANGE -o $DIR_FOR_PATCHES_TO_CHECK/
+   git format-patch --no-cover-letter $TRAVIS_COMMIT_RANGE ^$TRAVIS_BRANCH -o $DIR_FOR_PATCHES_TO_CHECK/
    CHECKPATCH_OPT="--no-tree --ignore PREFER_KERNEL_TYPES,FILE_PATH_CHANGES,EXECUTE_PERMISSIONS,USE_NEGATIVE_ERRNO,CONST_STRUCT $DIR_FOR_PATCHES_TO_CHECK/*"
    perl checkpatch.pl $CHECKPATCH_OPT
    if [ $? -ne 0 ]; then
```

rleon:
Thanks, I pushed your fix

nmorey:
It seems to work better this way ;)

jgunthorpe:
cbuild needs a matching change to define TRAVIS_BRANCH as well...

nmorey:
But when running cbuild, I don't think TRAVIS_EVENT_TYPE is set to pull_request so we shouldn't end up in this code

jgunthorpe:
cbuild needs to be able to run checkpatch or there is no good way to do it locally, I had fixed all that up after Leon first added checkpatch. So it needs to set TRAVIS_EVENT_TYPE and TRAVIS_BRANCH now?

nmorey:
I guess so.
TRAVIS_EVENT_TYPE should be pull_request and TRAVIS_BRANCH should be base (the variable used for the range)
It emulates a pull request but as locally there is no pseudo merge (the one github does with master and the PR) we shouldn't hit the case where already upstreamed patches are in the devel branch but not in base.

rleon:
Something like that (I didn't test it yet):

diff --git a/buildlib/cbuild b/buildlib/cbuild
index 4c3ef36b..b08180dd 100755
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -580,6 +580,8 @@ def run_travis_build(args,env):
             "-w",os.path.join(home_build,"src"),
             "-u",str(os.getuid()),
             "-e","TRAVIS_COMMIT_RANGE=%s..HEAD"%(base),
+            "-e","TRAVIS_BRANCH=%s"%(base),
+            "-e","TRAVIS_EVENT_TYPE=pull_request",
             "-e","HOME=%s"%(home),
             "-e","TMPDIR=%s"%(os.path.join(home_build,"tmp")),
         ];

rleon:
https://github.com/linux-rdma/rdma-core/pull/196

PR #195, opened 2017-08-28
suse: switch fully to the new udev mechanism

nmorey:
Do not use redhat services and scripts and use the new udev system


nmorey:
Need to review a lot more stuff first in redhat scripts

PR #198, opened 2017-08-29
Use systemd ListenNetlink to open the RDMA_NL_GROUP_LS ibacm socket

jgunthorpe:
This makes it less likely that messages from the kernel that should be handled by acm are lost during bootup due to various boot time races.

For after v15

### DIFF
--- a/ibacm/ibacm.socket
+++ b/ibacm/ibacm.socket
@@ -10,6 +10,10 @@ Before=rdma-hw.target
 ListenStream=6125
 BindToDevice=lo
 
+# Bind to PF_NETLINK, NETLINK_RDMA, RDMA_NL_GROUP_LS
+# Supported in systemd > 234
+ListenNetlink=rdma 4
+
 [Install]
 # Standard for all sockets
 WantedBy=sockets.target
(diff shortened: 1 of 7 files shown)


weiny2:
In general this looks like a great addition but I have a question.

It's been a while since I have looked at any code but does this work most of the time with the default timeout the kernel is waiting on for an ibacm response?  I'm curious if you think this removes a need to delay a port from coming up until after ibacm is started like you and I discussed back at Open Fabrics?

jgunthorpe:
@weiny2, on the timeout, not really sure. I expect bootup to be pretty fast in most cases, so I would expect the timeout to be OK.

It probably does not eliminate the need to have the port be down at bootup, but it makes ordering of the 'port up' in the bootup process more robust, as it no longer needs to directly sync with ibacm. Just ensure the port is set up after rdma-hw.target, and if ibacm is installed then it will capture 100% of the messages from the kernel.

If I recall OPA's issue was things like opa_vnic automatically start doing LS queries as soon as they are loaded. We don't enforce any ordering between module loads and listening socket setup, so the issue is still racy.

PR #201, opened 2017-08-30
dev/suse spec

nmorey:
Minors cleanups in SUSE spec file and remove packaging of obsolete RH scripts.

### DIFF
--- a/suse/rdma-core.spec
+++ b/suse/rdma-core.spec
@@ -57,6 +57,7 @@ BuildRequires:  pkgconfig
 BuildRequires:  pkgconfig(libsystemd)
 BuildRequires:  pkgconfig(libudev)
 BuildRequires:  pkgconfig(systemd)
+BuildRequires:  pkgconfig(udev)
 %ifnarch s390 s390x
 BuildRequires:  valgrind-devel
 %endif
@@ -67,6 +68,7 @@ BuildRequires:  pkgconfig(systemd)
 Requires:       dracut
 Requires:       kmod
 Requires:       systemd
+Requires:       udev
 
 # SUSE previously shipped rdma as a stand-alone
 # package which we're supplanting here.
@@ -408,6 +410,9 @@ rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
 
 %post
 %service_add_post rdma-sriov.service
+# we ship udev rules, so trigger an update.
+/sbin/udevadm trigger --subsystem-match=infiniband --action=change || true
+/sbin/udevadm trigger --subsystem-match=infiniband_mad --action=change || true
 
 %preun
 %service_del_preun -n rdma-sriov.service
@@ -438,6 +443,8 @@ rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
 
 %post -n srp_daemon
 %service_add_post srp_daemon.service srp_daemon_port@.service
+# we ship udev rules, so trigger an update.
+/sbin/udevadm trigger --subsystem-match=infiniband_mad --action=change
 
 %preun -n srp_daemon
 %service_del_preun srp_daemon.service srp_daemon_port@.service
--- a/suse/rdma-core.spec
+++ b/suse/rdma-core.spec
@@ -447,10 +447,12 @@ rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
 /sbin/udevadm trigger --subsystem-match=infiniband_mad --action=change
 
 %preun -n srp_daemon
-%service_del_preun srp_daemon.service srp_daemon_port@.service
+%service_del_preun srp_daemon.service
+%service_del_postun -n  srp_daemon_port@.service
 
 %postun -n srp_daemon
-%service_del_postun srp_daemon.service srp_daemon_port@.service
+%service_del_postun srp_daemon.service
+%service_del_postun -n  srp_daemon_port@.service
 
 #
 # iwpmd
--- a/suse/rdma-core.spec
+++ b/suse/rdma-core.spec
@@ -678,7 +678,6 @@ rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
 %{_sbindir}/rdma-ndd
 %{_sbindir}/rcrdma-ndd
 %{_unitdir}/rdma-ndd.service
-%{_mandir}/man8/rdma-ndd.*
 %{_mandir}/man8/rdma-ndd.8*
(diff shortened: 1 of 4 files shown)


nmorey:
@dledford It might make sense for RH to pick some of the changes i did in patch 2.
It pretty much drops all the legacy fixes and adds a service for the sriov stuff.
The only thing not in this is the support for the tech previews.

If RH switch to these sriov scripts, they should probably be moved to kernel-boot.
With a quick fix as RH uses /usr/libexec and SUSE uses /usr/lib

suse/rdma.sriov-rules
```diff
@@ -0,0 +1,7 @@
+ACTION=="remove", GOTO="rdma_sriov_end"
+SUBSYSTEM!="infiniband", GOTO="rdma_sriov_end"
+
+# Automatically load general RDMA ULP modules when RDMA hardware is installed
+TAG+="systemd", ENV{ID_RDMA_INFINIBAND}=="1", ENV{SYSTEMD_WANTS}+="rdma-sriov.service"
+
```
jgunthorpe: redhat/rdma.sriov-vfs talks about ethernet ports, so this probably needs to be a wider..

Arguably we need ID_RDMA_SRIOV as well to do this.

suse/rdma.sriov-service
```diff
@@ -0,0 +1,14 @@
...
+Conflicts=emergency.target emergency.service
+Before=network.target remote-fs-pre.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/lib/rdma-set-sriov-vf
+
+[Install]
+WantedBy=sysinit.target
```
jgunthorpe: Rules triggered by udev should not be installed, otherwise they run unconditionally

suse/rdma.sriov-service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Initialize SRIOV for RDMA devices
+RefuseManualStop=true
+DefaultDependencies=false
+Conflicts=emergency.target emergency.service
+Before=network.target remote-fs-pre.target
+
```
jgunthorpe: The before's should be:

```
# Partially support distro network setup scripts that try to run after hardware is configured.
# Run them after any SRIOV devices have been created
Wants=network-pre.target
Before=network-pre.target
# RDMA is not ready until all SRIOV devices are created
Before=rdma-hw.target
```

suse/rdma-core.spec
```diff
@@ -513,9 +518,8 @@ rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
 %{_udevrulesdir}/90-rdma-hw-modules.rules
 %{_udevrulesdir}/90-rdma-ulp-modules.rules
 %{_udevrulesdir}/90-rdma-umad.rules
-%{_udevrulesdir}/98-rdma.rules
+%{_udevrulesdir}/98-rdma-sriov.rules
 %config %{sysmodprobedir}/50-libmlx4.conf
-%{_libexecdir}/rdma-init-kernel
 %{_libexecdir}/rdma-set-sriov-vf
```
jgunthorpe: redhat/rdma.mlx4.sys.modprobe is not a config file, at least not when installed in /usr/

suse/rdma-core.spec
```diff
@@ -67,6 +68,7 @@ BuildRequires:  pkgconfig(systemd)
 Requires:       dracut
 Requires:       kmod
 Requires:       systemd
+Requires:       udev
```
jgunthorpe: Isn't this implied by requires: systemd?

suse/rdma-core.spec
```diff
@@ -57,6 +57,7 @@ BuildRequires:  pkgconfig
 BuildRequires:  pkgconfig(libsystemd)
 BuildRequires:  pkgconfig(libudev)
 BuildRequires:  pkgconfig(systemd)
+BuildRequires:  pkgconfig(udev)
 %ifnarch s390 s390x
```
jgunthorpe: Not sure why this is needed.. We don' t need udev installed to build rdma-core, just libudev

suse/rdma.sriov-rules
nmorey: I think the ethernet part is for mlx4 boards in ethernet mode which would be marked as ID_RDMA_INFINIBAND anyway (use mlx4_core driver)

suse/rdma-core.spec
nmorey: For SUSE we do. Our buildsystem checks commands in %post.
Without this it fails saying udevadm does not exists

suse/rdma-core.spec
nmorey: It probably is, but at least it's explicit (and avoid some warning in the build system)

suse/rdma-core.spec
nmorey: Agreed. Early mistakes in the first packaging. the .conf suffix is misleading

suse/rdma.sriov-service
nmorey: If this is triggered by udev, does it have an impact ?

But OK

suse/rdma.sriov-service
nmorey: Noted. Will fix

suse/rdma.sriov-rules
jgunthorpe: Right, but fixing that is on my hitlist too :) So it is OK for now

suse/rdma.sriov-service
jgunthorpe: Yes, ordering is important! The sriov script may also need a 'udevadm settle' at the end to make the ordering predictable.

nmorey:
@jgunthorpe I fixed the two issue in a branch here:
https://github.com/nmorey/rdma-core/tree/dev/suse-spec

It's late and I'm off work so I won't do a PR right now (specially if Leon is in a hurry to accept it before I can do a full review)
Could you check the udev rule fix to make sure I got what you expected ?

jgunthorpe:
@nmorey Seems OK

rleon:
@nmorey , I'm not in hurry :)
You are the SuSE maintainer and I trust you that you are knowing what you are doing in regards of SuSE packaging.

nmorey:
@rleon I know. I'm confident on the packaging side of things but for all the systemd/udev stuff I'd gladly have some feedback first ;)

rleon:
So post to the ML before :)

PR #207, opened 2017-08-31
Small verbs fixes

jgunthorpe:
This is a precursor to a larger series. The big item switches verbs/init.c to use ccan lists and slightly reworks how the earlier patch to update the device list works.


jgunthorpe:
Updated:
 - A revised version of check_snprintf and the commit message
 - Remove list_init as ccan list_add does it.

jgunthorpe:
Updated:
- Replaced // comment with /*

rleon:
@jgunthorpe , Yishai planned to review it today.

Thanks

jgunthorpe:
Updated with the reviewed-bys

jgunthorpe:
The trusty packages for llvm have gone missing... If they don't come back we will have to drop clang until we can move off of travisci..

PR #210, opened 2017-09-05
[Resubmit] libqedr: Add iWARP support for user-space lib

ramranix:
Resubmitting iWARP libqedr support, now that iWARP was accepted upstream on August 22nd.
https://www.spinics.net/lists/linux-rdma/msg53677.html
There's, of course, no issues if a more advanced libqedr is used with an older non-iWARP driver.

Thanks,
Ram

Introduce IS_IWARP / IS_ROCE macros for libqedr as there are no global
macros for this in rdma-core.
There is a second doorbell required for RQ.
QP states in user lib are not updated, as they are updated implicitly by
the driver and not driven by modify_qp as they are in ib / roce.
For this reason, decisions based on states are encapsulated in IS_ROCE.

Signed-off-by: Michal Kalderon  <Michal.Kalderon@cavium.com>

### DIFF
--- a/providers/mlx5/buf.c
+++ b/providers/mlx5/buf.c
@@ -320,6 +320,36 @@ static void free_huge_buf(struct mlx5_context *ctx, struct mlx5_buf *buf)
 		mlx5_spin_unlock(&ctx->hugetlb_lock);
 }
 
+void mlx5_free_buf_extern(struct mlx5_context *ctx, struct mlx5_buf *buf)
+{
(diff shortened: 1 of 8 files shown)


jgunthorpe:
There are some sparse problems and checkpatch problems here..

jgunthorpe:
Ram, you probably also need to add your own Signed-off-by in addition to Michal's as you are handling the patch too..

PR #214, opened 2017-09-07
Rework the verb init scheme

jgunthorpe:
This changes how verbs providers register with the core and are bound to drivers. 

The end goal is to have textual a list of 'modalias' that each provider supports, similar to the kernel. This list can then ultimately be used to demand load the modules instead of loading every module like we do today.

This series does the work to bring all providers to use common code to match their supported devices, and provides the core code the supported device list. This is done by extending the verbs_driver_ops to include the matching information that the driver needs.

Quite a lot of duplicated driver code is removed in the process, and verbs startup reads fewer sysfs files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,7 @@ set(PACKAGE_VERSION "16")
 # When this is changed the values in these files need changing too:
 #   debian/libibverbs1.symbols
 #   libibverbs/libibverbs.map
-set(IBVERBS_PABI_VERSION "15")
+set(IBVERBS_PABI_VERSION "16")
 set(IBVERBS_PROVIDER_SUFFIX "-rdmav${IBVERBS_PABI_VERSION}.so")
 
 #-------------------------
--- a/debian/libibverbs1.symbols
+++ b/debian/libibverbs1.symbols
@@ -1,7 +1,7 @@
 libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.0@IBVERBS_1.0 1.1.6
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
- (symver)IBVERBS_PRIVATE_15 15
+ (symver)IBVERBS_PRIVATE_16 16
  ibv_ack_async_event@IBVERBS_1.0 1.1.6
  ibv_ack_async_event@IBVERBS_1.1 1.1.6
  ibv_ack_cq_events@IBVERBS_1.0 1.1.6
--- a/libibverbs/driver.h
+++ b/libibverbs/driver.h
@@ -96,14 +96,56 @@ struct verbs_qp {
 	struct verbs_xrcd       *xrcd;
 };
 
+enum {
+	VERBS_MATCH_SENTINEL = 0,
+	VERBS_MATCH_PCI = 1,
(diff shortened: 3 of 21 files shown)


rleon:
@jgunthorpe don't forget to post this series in the mailing list.

Thanks

jgunthorpe:
@relon, Oh! that is really strange, I thought I already sent it.. Well that explains the silence.

jgunthorpe:
Force pushed:
- Rebase and resolve conflict with bnxt patch

jgunthorpe:
Updated with comments and acks from Steve Wise

jgunthorpe:
Updated:
- Add Lijun Ou's Ack
- Up the verbs private ABI to 16 in the first patch when verbs_register_driver signature is changed

rleon:
@jgunthorpe , can you please post updated version to the mailing list? I want to review it there and will take it if no objections raise.

Thanks

jgunthorpe:
I think only four lines have changed since the mailing list, what is the point of posting it again? Let's not make the patch process more tedious than it needs to be :( You can format-patch it locally and review the diffs or use github's interface, or look at the original posting.. That is what I have been doing.

rleon:
> what is the point of posting it again? 
I would like to have similar rules for everyone without exception.

In this case, you are changing code in many places and I aware at least of one driver author who doesn't read github at all.

Thanks

jgunthorpe:
@rleon everyone should follow the same rules, post it once to the mailing list and collect minor comments, force push with a note about changes. No need to repost again and again for very small changes.

Every driver author was CC'd on the original posting according to the maintainers file. We do not need to follow the kernel version of re-posting again and again for every tiny change because we are not applying patches by PR.

jgunthorpe:
Updated with Ram's ack and comments:

```
diff --git a/providers/qedr/qelr_main.c b/providers/qedr/qelr_main.c
index 1297e8e16acee5..fa2be455edc93e 100644
--- a/providers/qedr/qelr_main.c
+++ b/providers/qedr/qelr_main.c
@@ -223,7 +223,7 @@ static void qelr_free_context(struct ibv_context *ibctx)
        free(ctx);
 }
 
-static struct verbs_device *qedr_device_alloc(struct verbs_sysfs_dev *sysfs_dev)
+static struct verbs_device *qelr_device_alloc(struct verbs_sysfs_dev *sysfs_dev)
 {
        struct qelr_device *dev;
 
@@ -235,11 +235,11 @@ static struct verbs_device *qedr_device_alloc(struct verbs_sysfs_dev *sysfs_dev)
 }
 
 static const struct verbs_device_ops qelr_dev_ops = {
-       .name = "qelr",
+       .name = "qedr",
        .match_min_abi_version = QELR_ABI_VERSION,
        .match_max_abi_version = QELR_ABI_VERSION,
        .match_table = hca_table,
-       .alloc_device = qedr_device_alloc,
+       .alloc_device = qelr_device_alloc,
        .uninit_device = qelr_uninit_device,
        .alloc_context = qelr_alloc_context,
        .free_context = qelr_free_context,
```

rleon:
@jgunthorpe let's collect latest Acked-by and merge it.

Thank you for the patience

rleon:
Merged manually

PR #216, opened 2017-09-10
mlx4: Add inline-receive support

yishaih:
Add mlx4 support for inline-receive, the matching kernel part was already merged into 4.14.

### DIFF
--- a/debian/ibverbs-providers.symbols
+++ b/debian/ibverbs-providers.symbols
@@ -1,6 +1,8 @@
 libmlx4.so.1 ibverbs-providers #MINVER#
  MLX4_1.0@MLX4_1.0 15
  mlx4dv_init_obj@MLX4_1.0 15
+ mlx4dv_query_device@MLX4_1.0 15
+ mlx4dv_create_qp@MLX4_1.0 15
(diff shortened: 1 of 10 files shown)


debian/ibverbs-providers.symbols
```diff
@@ -1,6 +1,8 @@
 libmlx4.so.1 ibverbs-providers #MINVER#
  MLX4_1.0@MLX4_1.0 15
  mlx4dv_init_obj@MLX4_1.0 15
+ mlx4dv_query_device@MLX4_1.0 15
+ mlx4dv_create_qp@MLX4_1.0 15
 libmlx5.so.1 ibverbs-providers #MINVER#
```
jgunthorpe: ben is going to want these sorted

providers/mlx4/libmlx4.map
```diff
@@ -3,5 +3,7 @@
 MLX4_1.0 {
 	global:
 		mlx4dv_init_obj;
+		mlx4dv_query_device;
+		mlx4dv_create_qp;
 	local: *;
```
jgunthorpe: Sort

PR #219, opened 2017-09-13
suse: relax requirement on dracut

nmorey:


nmorey:
Forget that it neds to be dropped altogether

PR #221, opened 2017-09-18
Bump version to be v16

rleon:
Update library version to be v16.


rleon:
Right, I wanted to have one commit with all relevant changes, so I'll be able to compare with it next time, I'm updating version.

PR #222, opened 2017-09-20
ibacm: Close MAD port when umad_register fails

yuvalshaia:
Cleanup of the open port is needed when program fails to register MAD
port.


rleon:
Thanks

PR #233, opened 2017-10-25
ibacm: In acm_util.c:acm_if_iter_sys, try IPv4 if IPv6 doesn't find any appropriate interfaces

hnrose:
Also, improve some other related acm log messages.


hnrose:
@shefty , @weiny2:
Just want to make sure you saw this. These patches came out of recent issue on OFA users list where they had machine running ACM but IPv6 was disabled in kernel.

hnrose:
@weiny2 This error occurs when IPv6 is disabled in kernel.

PR #236, opened 2017-10-29
vmw_pvrdma: Add SRQ support

bryantan-vmware:
Enable userlevel applications to use SRQs.

Reviewed-by: Adit Ranadive <aditr@vmware.com>
Reviewed-by: Aditya Sarwade <asarwade@vmware.com>
Reviewed-by: Jorgen Hansen <jhansen@vmware.com>
Reviewed-by: Nitish Bhat <bnitish@vmware.com>

### DIFF
--- a/buildlib/fixup-include/rdma-vmw_pvrdma-abi.h
+++ b/buildlib/fixup-include/rdma-vmw_pvrdma-abi.h
@@ -57,6 +57,8 @@
 #define PVRDMA_UAR_CQ_ARM_SOL		BIT(29)		/* Arm solicited bit. */
 #define PVRDMA_UAR_CQ_ARM		BIT(30)		/* Arm bit. */
 #define PVRDMA_UAR_CQ_POLL		BIT(31)		/* Poll bit. */
+#define PVRDMA_UAR_SRQ_OFFSET		8		/* SRQ doorbell. */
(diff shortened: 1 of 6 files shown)


buildlib/fixup-include/rdma-vmw_pvrdma-abi.h
```diff
@@ -157,6 +159,8 @@ struct pvrdma_resize_cq {
 
 struct pvrdma_create_srq {
 	__u64 buf_addr;
+	__u32 buf_size;
+	__u32 reserved;
 };
```
jgunthorpe: What is up with this kernel uAPI change? If srq was not supported until now why did a create_srq  struct exist?

bryantan-vmware: That is a mistake on our part--we should not have had a `pvrdma_create_srq` struct before this.

Is it preferable in this situation to create a `pvrdma_create_srq_v2`? We also noticed that we didn't update the ABI version either, which we'd need to do if we went with the above approach.

jgunthorpe: It looks like it was never used in the kernel at all, so it should be OK to change it however you like.

You should not need to update the ABI because the kernel create_srq call will currently return failure and this is good enough.

dledford:
Kernel patch was committed today, so this is now clear to merge.

PR #238, opened 2017-10-30
CQ moderation

yishaih:
This series is the supplementary part of the kernel series that was sent for 4.15 to support CQ moderation.


dledford:
CQ Moderation patches were merged today, so this is now clear to be merged.

PR #239, opened 2017-10-31
libibumad/umad.c: In get_port, ignore sysfs rate file errors

hnrose:
This can cause ibpanic in ibstat when width is not set properly
as can occur when QSPF is not plugged into port.

ibpanic: [7851] main: stat of IB device 'mlx5_1' failed: Invalid argument

It's caused by kernel sysfs.c:show_rate returning -EINVAL
and that error being treated as failure in umad.c:get_port.
With this change, Rate is displayed as 0 with ibstat for this scenario.


hnrose:
Just added Tested-by SOB to original commit message

PR #240, opened 2017-11-02
librdmacm: Remove unneeded condition

yuvalshaia:
The check makes no difference - remove it.

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -880,8 +880,8 @@ static int rdma_resolve_addr2(struct rdma_cm_id *id, struct sockaddr *src_addr,
 	CMA_INIT_CMD(&cmd, sizeof cmd, RESOLVE_ADDR);
 	id_priv = container_of(id, struct cma_id_private, id);
 	cmd.id = id_priv->handle;
-	if ((cmd.src_size = src_len))
-		memcpy(&cmd.src_addr, src_addr, src_len);
+	cmd.src_size = src_len;
+	memcpy(&cmd.src_addr, src_addr, src_len);
 	memcpy(&cmd.dst_addr, dst_addr, dst_len);
 	cmd.dst_size = dst_len;
 	cmd.timeout_ms = timeout_ms;


rleon:
@jgunthorpe Can it be that "=" was written instead of "=="?

Thanks

jgunthorpe:
@rleon Looks like the assignment is necessary to me..

rleon:
@jgunthorpe Thanks, I looked into the code now and cmd.src_size is not initialized, so yes assignment is needed.

librdmacm/cma.c
```diff
@@ -880,8 +880,8 @@ static int rdma_resolve_addr2(struct rdma_cm_id *id, struct sockaddr *src_addr,
 	CMA_INIT_CMD(&cmd, sizeof cmd, RESOLVE_ADDR);
 	id_priv = container_of(id, struct cma_id_private, id);
 	cmd.id = id_priv->handle;
-	if ((cmd.src_size = src_len))
-		memcpy(&cmd.src_addr, src_addr, src_len);
+	cmd.src_size = src_len;
```
gladkovdmitry17: @yuvalshaia @rleon @jgunthorpe Just out of curiosity - why this change is needed?
It assigns to the `ucma_abi_resolve_addr` command a source size of an address and if this size is 0, it won't do the an excessive call of the `memset` function.
The `cmd->src_addr` is 0, because the structure is zeromemd in the `CMA_INIT_CMD`.

rleon: @gladkovdmitry17 IMHO, it makes code cleaner.

yuvalshaia: Yeah, the existing code is unusual, just see how len is not checked in dest_addr

yuvalshaia:
Closing this pull request and submitting new one with a revised commit message

librdmacm/cma.c
gladkovdmitry17: @rleon @yuvalshaia Ok, it was just out of curiosity.
Yes, it will be similar to `dest_addr`

PR #241, opened 2017-11-02
Package travis test

rleon:
Add ability to test package builds on latest main distributions: CentOS, Ubuntu and SuSE.


rleon:
First two patches were merged, third patch needs to be resubmitted

PR #242, opened 2017-11-02
verbs: Do not block QP attr_masks used by older kernels

jgunthorpe:
At least the RDMA CM in 4.3 will give user space structs with these bits
set. Those kernels require user space to pass those set bits back to the
kernel. Blocking them causes RDMA CM to be unusable on older kernels.

Fixes: 3ca7a1031486 ("ibverbs: Add support for packet pacing")
Reported-by: Ram Amrani <Ram.Amrani@cavium.com>
Signed-off-by: Jason Gunthorpe <jgunthorpe@ziepe.ca>

### DIFF
--- a/libibverbs/cmd.c
+++ b/libibverbs/cmd.c
@@ -1182,10 +1182,10 @@ int ibv_cmd_query_qp(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 	struct ibv_query_qp_resp resp;
 
 	/*
-	 * Masks over IBV_QP_DEST_QPN are not supported by
-	 * that not extended command.
+	 * Starting with IBV_QP_RATE_LIMIT the attribute must go through the
+	 * _ex path.
 	 */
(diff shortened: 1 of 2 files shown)


jgunthorpe:
revised with Leon's comment to remove the #if 0

libibverbs/cmd.c
```diff
@@ -1182,10 +1182,10 @@ int ibv_cmd_query_qp(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 	struct ibv_query_qp_resp resp;
 
 	/*
-	 * Masks over IBV_QP_DEST_QPN are not supported by
-	 * that not extended command.
+	 * Starting with IB_QP_RATE_LIMIT the attribute must go through the
```
yishaih: IB_QP_XXX -> IBV_QP_XXX, same for below in line 1355.

jgunthorpe:
revised with Yishaih's comment on spelling

PR #245, opened 2017-11-08
librdmacm: Make return value from ucma_init consist

yuvalshaia:
When check_abi_version fails we must be consist with rest of exit flows
of the function and return -1 and set errno with the actual error code.


rleon:
It was already applied.

Thanks

PR #246, opened 2017-11-09
libbnxt_re bug fix series

dsharma283:
Patch 1: address the memory barrier changes Jason asked while reviewing a similar change in bnxt_re driver.

Patch2: Fixes psn calculations when sending out a zero length packet. This was found during internal testing.


dsharma283:
Will resend corrected pull request.

PR #250, opened 2017-11-14
PCI write end padding

yishaih:
This series is the supplementary part of the kernel series that was merged  into 4.15 'for-next'  branch by Doug few days ago. 

The series exposes the "PCI write end padding" device capability and enables QP and WQ creation with that option.


dledford:
Explanation in man page was clear and describes the situation well.  The actual code to implement this is almost trivial as it just involves a few flags.

PR #251, opened 2017-11-14
ABI consistency check

nmorey:
Add scripts to generate an ABI dump and compare against a ref.
Both tools (abi-dumper and abi-consistency-check) were imported as the
system version do not work on Travis.
Travis will automatically do that if a ref is present.
Ref should be added at stable-* branch creation (see doc update)

It seems the 1st patch is too big for the ML. So a PR will have to do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,8 @@ addons:
       - valgrind
       - sparse
       - wget
+      - abi-compliance-checker
+      - abi-dumper
 
       # 32 bit support packages
       - gcc-multilib
--- a/Documentation/stable.md
+++ b/Documentation/stable.md
@@ -63,3 +63,27 @@ Note that for Option 3, if the patch deviates from the original upstream patch (
 ## Versioning
 
 See versioning.md for setting package version on a stable branch.
+
+
+## Creating a stable branch
+
+Stable branch should be created from a release tag of the master branch.
+The first thing to do on a master branch is to commit the mainstream release ABI infos
+so that latters patches/fixes can be checked against this reference.
+
+To do that, the creator of the branch should run
+```
+./buildlib/cbuild build-images travis
+mkdir ABI
+touch ABI/.gitignore
+git add ABI/.gitignore
+git commit -m "ABI Files"
+./buildlib/cbuild pkg travis
+git add ABI/*
+git commit --amend
+```
+
+'cbuild pkg travis' will fail as the ABI verification step files, but it will
+produce the ABI reference files.
+
+Note that the ABI directory must NOT be committed at any point in the master branch.
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -38,6 +38,7 @@ useful for debugging certain kinds of build problems."""
 
 import argparse
 import collections
+import filecmp
 import grp
 import imp
 import inspect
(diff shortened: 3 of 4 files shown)


jgunthorpe:
This is a little bigger than I thought it would be..

See my list comments on the need for additional COPYING.* files

We might be better off just adding a git pull for the stuff into the travis setup pass?

nmorey:
@jgunthorpe It would work too.
We can probably do that from the gen-abi script.
This way it'll work manually and in travis.

nmorey:
@jgunthorpe Would you be OK with git submodules?
Better than some git pull somewhere with a hardcoded SHA1

jgunthorpe:
@nmorey I have never tested submodoules.

Philosophically, I guess it depends on your perspective..

Should we 'fix' the distro environment, eg in travis to have the right tools

Or should we internalize the tools inside rdma-core ?

Long term I would like to see us move to CircleCI and use docker containers we control for all of our builds. In this case we'd just provision those containers with the necessary versions of the tools we need.

So, I'd say if the issue here is that trusty is too old but something like FC26 would include the right tools, then let us fix the issue in the travis scripts like we do for the ARM compiler.

rleon:
why don't we put it in container from the beginning?

nmorey:
@rleon Tried adding them in the .travis.yml file but the versions of abi-dumper and abi-compliance-checker are incompatible... Hence trying to import them.

rleon:
@nmorey, I mean, something like
./buildlib/cbuild pkg check-abi
and execute it like we are doing with distros

nmorey:
@rleon Does that solve the issue ?
We have to find a distro with compatible dumper/checker tools and be sure they will always be !

@jgunthorpe @rleon 
Update the PR with submodules. It's much smaller now ;)

jgunthorpe:
@rleon, yah, we could probably use the existing FC26 container for this.

@nmorey, I think when dealing with external dependencies we should try hard to deal with it through the docker side and not try to embed the dependencies into the rdma-core git. This should be more maintainable long term, assuming the things we depend on are halfway sane and don't keep breaking compat.

rleon:
@nmorey , we really-really would like to see this check in container.

nmorey:
@rleon @jgunthorpe This should be better.
The version in travis are compatible but not with outside (more recent) versions of abi-dumper which was what I saw initially.
It means ABI dump file MUST be generated from the travis container to ensure compatibility.
Thus the ABI specific built is always done (I reverted the fix @rleon asked that it should run only if the ABI is present).
The impact on build time should be minimal. A simple call to ninja is much faster than spawning a container for RHEL/SUSE validations.

I created a branch with ABI refs to make sure it works. Yuu can see the results there:
https://travis-ci.org/nmorey/rdma-core/builds/304727022

rleon:
@nmorey once we will convert all our checks to containers, we will start to explore different options to run them (CircleCI, SemaphoreCI or my private server).

Thanks

nmorey:
@rleon Meaning you'd rather a dedicated abi_check container ?

rleon:
@nmorey I think it is better, but don't change anything yet, @jgunthorpe in favor of reusing existing container.

nmorey:
@rl

nmorey:
@relon Ideally I'd like this series to get in as is so I can backport it to stable-v15 and add some level of ABI checking.
And we can work on this to tweak it the way you want before the stable-v16 branch.

jgunthorpe:
I think this latest approach is fine. Using the two packages in trusty is fine, we  can move it into a dedicated and more up to date contiainer trivially down the road.

Let me look at these patches, confused why we are changing cbuild??

nmorey:
@jgunthorpe I added some code to cbuild to extract the ABI dump file from the container (same way RPM/deb are exctrated from ther containers).

This is needed to be able to generate the reference ABI on stable branches.
If I just run the ./buildlib/gen-abi on my SUSE Leap, the generated file are not compatible with the trusty version of abi-compliance-checker.

jgunthorpe:
@nmorey I made some revisions for you so it fits better into the buildlib scheme:

https://github.com/jgunthorpe/rdma-plumbing/tree/abi_check

- Put all the checking into the existing 'check-build' python script, eliminates one build pass
- Have cbuild copy the ABI files directly into the source tree home 'ABI/'
- Don't run ABI check at all if there is no ABI/ directory, so -stable has to create an empty ABI/
  directory in git to turn on the feature

nmorey:
@jgunthorpe It looks good.
abi-dumper suggest the build to be done with -g -Og, but it seems to work OK like this.

nmorey:
@rleon  I repushed my branch with Jason's patches.
If you're OK with this too, can you get this in ? :)

rleon:
Thanks to everyone.

jgunthorpe:
@nmorey, ah is that why it got so many bogus warnings.. clang might be okay with just -g? Lets keep an eye on it.

nmorey:
-g should probably be enough. As we mainly concern ourself with the external ABI, there are few chances that -O[23] would screw up with those...

We don't really care if internally gcc decide to invert parameters for an inline function

PR #254, opened 2017-11-17
Support for the MC join flag in RDMA CM

jgunthorpe:
This adds support for the 'send only, full member' multicast join type added to a recent InfiniBand spec.

This join type allows the joiner to avoid recieving traffic from the group, without some of the downsides of the existing send only join.

This is from the mailing list, but I fixed several more problems while applying it:
 - Issues with debian packagin
 - Missing man page updates
 - English
 - check patch failures

@hnrose @shefty


jgunthorpe:
@hrose: Done, thanks

PR #257, opened 2017-11-19
mlx5: Add striding RQ support via the DV API

yishaih:
This series exposes to mlx5 DV users the ability to create an RQ with the striding functionality.
The kernel part was already merged into 4.15.


jgunthorpe:
The ABI changes look OK to me

PR #259, opened 2017-11-20
libibumad: Add support for SA MCMemberRecord

hnrose:
Add umad_sa_mcm.h to support SA MCMemberRecord
and add umad_sa_mcm_rereg_test.c (rewrite of infiniband-diags umad_rereg_test) using this new header.
Update umad_sa.h with some common enums, defines, and helpers shared with SA PathRecord and MultiPathRecord.

### DIFF
--- a/libibumad/tests/umad_sa_mcm_rereg_test.c
+++ b/libibumad/tests/umad_sa_mcm_rereg_test.c
@@ -1,0 +1,582 @@
+/*
+ * Copyright (c) 2017 Mellanox Technologies Ltd. All rights reserved.
+ * Copyright (c) 2006-2009 Voltaire, Inc. All rights reserved.
+ *
+ * This software is available to you under a choice of one of two
+ * licenses.  You may choose to be licensed under the terms of the GNU
+ * General Public License (GPL) Version 2, available from the file
+ * COPYING in the main directory of this source tree, or the
+ * OpenIB.org BSD license below:
+ *
+ *     Redistribution and use in source and binary forms, with or
+ *     without modification, are permitted provided that the following
+ *     conditions are met:
+ *
+ *      - Redistributions of source code must retain the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer.
+ *
+ *      - Redistributions in binary form must reproduce the above
+ *        copyright notice, this list of conditions and the following
+ *        disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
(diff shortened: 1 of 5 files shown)


libibumad/tests/umad_sa_mcm_rereg_test.c
```diff
@@ -0,0 +1,559 @@
...
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+ * BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+#if HAVE_CONFIG_H
+#  include <config.h>
```
jgunthorpe: Drop HAVE_CONFIG_H it is old

libibumad/tests/umad_sa_mcm_rereg_test.c
```diff
@@ -0,0 +1,559 @@
...
+#define MAX_PORT_GUIDS 64
+
+void build_user_mad_addr(uint8_t *umad);
+void build_mcm_rec(struct umad_sa_packet *sa, uint8_t method,
+		   const uint8_t mgid[], const uint8_t port_gid[],
+		   uint64_t tid, int create);
+
+/* Use null MGID to request SA assigned MGID */
+static uint8_t null_mgid[16] = {
+	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
```
jgunthorpe: static const

hnrose:
@jgunthorpe 

Finally got it down to 1 error which I don't understand:

2119-----------------------------------------------------------------------------------

2120/tmp/tmp.3iQWBBx3cc/0001-libibumad-Add-umad_sa_mcm.h-for-umad-SA-MCMemberReco.patch

2121-----------------------------------------------------------------------------------

2122ERROR: Macros with complex values should be enclosed in parentheses

2123#72: FILE: libibumad/umad_sa_mcm.h:41:

2124+#  define END_C_DECLS   }

2125

2126total: 1 errors, 0 warnings, 204 lines checked

2127

2128NOTE: For some of the reported defects, checkpatch may be able to

2129      mechanically convert to the typical style using --fix or --fix-inplace.

2130

2131/tmp/tmp.3iQWBBx3cc/0001-libibumad-Add-umad_sa_mcm.h-for-umad-SA-MCMemberReco.patch has style problems, please review.

2132------------------------------------------------------------------------------

Line 41 in umad_sa_mcm.h is same as in many other files (as is line 40) for BEGIN_C_DECLS and END_C_DECLS.

Any idea what it doesn't like ? Why does it think this is "complex macro" ?

jgunthorpe:
That one is bogus, we can ignore it.

However, IMHO we should never use the BEGIN_C_DECLS approach in public header files, they should all directly open code the '#ifdef __cplusplus' to avoid symbol pollution.

hnrose:
@jgunthorpe 
You wrote:
IMHO we should never use the BEGIN_C_DECLS approach in public header files, they should all directly open code the '#ifdef __cplusplus' to avoid symbol pollution.

It's in libibverbs/verbs.h and others in rdma-core.

How does open coding this eliminate symbol pollution ? Wouldn't it effectively be the same thing if those macros are replaced with the open coding ? Does the difference when !cplusplus somehow matter ? If so, I'll submit another patch to open code all such occurrences in rdma-core.

jgunthorpe:
BEGIN_C_DECLS is the symbol that creates the pollution. We shouldn't define it or use it, as the user may be using it for something else.

Changing all files would be great.

hnrose:
I just fixed up umad_sa_mcm.h to open code the extern "C" when __cplusplus is defined. I'll do the others in a separate patch.

jgunthorpe:
Looks fine now

rleon:
@hnrose please rebase.

Thanks

hnrose:
@rleon done as requested; baking now...

PR #265, opened 2017-11-23
Add tunnel support

yishaih:
This series is the supplementary part of the kernel series that was merged into 4.15.

It introduces the following support for tunneling packets:
1) RSS on the tunneled packet.
2) Tunneling offloads support for mlx5 devices via the DV API.

### DIFF
--- a/debian/ibverbs-providers.symbols
+++ b/debian/ibverbs-providers.symbols
@@ -14,4 +14,5 @@ libmlx5.so.1 ibverbs-providers #MINVER#
  mlx5dv_query_device@MLX5_1.0 13
  mlx5dv_create_cq@MLX5_1.1 14
  mlx5dv_set_context_attr@MLX5_1.2 15
+ mlx5dv_create_qp@MLX5_1.3 16
  mlx5dv_create_wq@MLX5_1.3 16
--- a/libibverbs/man/ibv_create_qp_ex.3
+++ b/libibverbs/man/ibv_create_qp_ex.3
@@ -72,6 +72,26 @@ uint64_t               rx_hash_fields_mask;    /* RX fields that should particip
 .in -8
 };
 .fi
+
+.nf
+enum ibv_rx_hash_fields {
+.in +8
+IBV_RX_HASH_SRC_IPV4            = 1 << 0,
+IBV_RX_HASH_DST_IPV4            = 1 << 1,
+IBV_RX_HASH_SRC_IPV6            = 1 << 2,
+IBV_RX_HASH_DST_IPV6            = 1 << 3,
+IBV_RX_HASH_SRC_PORT_TCP        = 1 << 4,
+IBV_RX_HASH_DST_PORT_TCP        = 1 << 5,
+IBV_RX_HASH_SRC_PORT_UDP        = 1 << 6,
+IBV_RX_HASH_DST_PORT_UDP        = 1 << 7,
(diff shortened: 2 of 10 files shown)


debian/ibverbs-providers.symbols
```diff
@@ -15,3 +15,4 @@ libmlx5.so.1 ibverbs-providers #MINVER#
  mlx5dv_create_cq@MLX5_1.1 14
  mlx5dv_set_context_attr@MLX5_1.2 15
  mlx5dv_create_wq@MLX5_1.3 16
+ mlx5dv_create_qp@MLX5_1.3 16
```
jgunthorpe: Sorted please, or Ben will send patch to sort it..

libibverbs/verbs.h
```diff
@@ -241,9 +241,14 @@ enum ibv_rx_hash_fields {
...
 	IBV_RX_HASH_SRC_PORT_UDP	= 1 << 6,
-	IBV_RX_HASH_DST_PORT_UDP	= 1 << 7
+	IBV_RX_HASH_DST_PORT_UDP	= 1 << 7,
 };
 
+/* In some systems/compilers enum range might be limited
+ * to signed 4 bytes.
+ */
+#define IBV_RX_HASH_INNER (1ULL << 31)
+
```
jgunthorpe: Why not put this in the enum? It just needs to be (1UL << 31)

providers/mlx5/libmlx5.map
```diff
@@ -21,4 +21,5 @@ MLX5_1.2 {
 MLX5_1.3 {
 	global:
 		mlx5dv_create_wq;
+		mlx5dv_create_qp;
 } MLX5_1.2;
```
jgunthorpe: sorted

libibverbs/man/ibv_create_qp_ex.3
```diff
@@ -72,6 +72,27 @@ uint64_t               rx_hash_fields_mask;    /* RX fields that should particip
...
+IBV_RX_HASH_DST_PORT_TCP        = 1 << 5,
+IBV_RX_HASH_SRC_PORT_UDP        = 1 << 6,
+IBV_RX_HASH_DST_PORT_UDP        = 1 << 7,
+.in -8
+};
+
+/* When set with one of the L3/L4 fields,
+ * then the RSS will be applied on the inner packet. */
+#define IBV_RX_HASH_INNER       = (1ULL << 31)
+
```
jgunthorpe: This is a terrible man page. At least explain what an outer and inner packet are.

libibverbs/verbs.h
yishaih: The concern was that enum is signed 4 bytes, we suspect that using 1UL << 31 as part of the enum might bring some warning in some compiler/setup. If you think that this is safe we can give that a try and add it.

jgunthorpe: We went over this in the kernel once. 

https://www.spinics.net/lists/linux-rdma/msg36828.html

The issue with this construction has always been that (1<<31) can be *negative* since '1' is technically signed, and 1 << 31 can be a signed 32 bit value. This causes the same problems as last weeks thing with comp_mask..

The only thing needed to make this safe is to make the shift unsigned:
 
  IBV_RX_HASH_INNER = 1U << 31,

Shifts over 31 will need 1ULL to avoid 32 overflow.

PR #267, opened 2017-11-27
Bump version to 15.1

nmorey:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+rdma-core (15.1-1) unstable; urgency=low
+
+  * New upstream stable release.
+
+ -- Nicolas Morey-Chaisemartion <nmoreychaisemartin@suse.com>  Mon, 27 Nov 2017 10:37:05 +0100
+
 rdma-core (15-1) unstable; urgency=low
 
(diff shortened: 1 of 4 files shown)


nmorey:
@rleon Please don't merge this one yet.
I just wanted @jgunthorpe review to make sure I did the right thing for the debian changelog.

I still have a couple of tests to run to make sure the release process will work :)

rleon:
no problem, if i remember well, there is no need to change debian/changelog

debian/changelog
```diff
@@ -1,3 +1,9 @@
+rdma-core (15.1-1) unstable; urgency=low
+
+  * New upstream stable release.
+
+ -- Nicolas Morey-Chaisemartion <nmoreychaisemartin@suse.com>  Mon, 27 Nov 2017 10:37:05 +0100
+
 rdma-core (15-1) unstable; urgency=low
```
jgunthorpe: We are not actually maintaing any of the packaging change logs in rdma-core upstream, so you can just revise the last entry to have the right version number.

It is necessary in debian for the changelog version number to match the version number in the control file.

PR #268, opened 2017-11-30
Debian 15.1-1 changes

bdrung:
This patch set contains the packaging changes from the Debian unstable 15.1-1 upload.


nmorey:
Isn't that (or at least part of it) meant for the stable branch ?

PR #269, opened 2017-12-04
Update library version to be 17.0

rleon:


rleon:
I updated debian/* just for the consistency.

jgunthorpe:
You do have to update debian/changelog or the debian packages dont' build properly.

PR #271, opened 2017-12-08
register_to_trap(): Specify number of iterations once 

bvanassche:
Master SM changes can take more than three seconds. Hence increase
the time during which to wait for a master SM to appear from 3 to 5
seconds.

Signed-off-by: Bart Van Assche <bart.vanassche@wdc.com>
Cc: Hal Rosenstock <hal@dev.mellanox.co.il>
Cc: Nicolas Morey-Chaisemartin <NMoreyChaisemartin@suse.de>


nmorey:
OK for me

hnrose:
:+1:

PR #275, opened 2017-12-14
vmw_pvrdma: Remove usage of BIT() from vmw_pvrdma

bryantan-vmware:
BIT() should not be used in the UAPI headers. It has been removed
in the kernel vmw_pvrdma-abi.h, so update the fixup-include header
accordingly, and remove the definition for BIT() in vmw_pvrdma code.


bryantan-vmware:
There's a stray whitespace change in there, as this is the first time I've actually copied the header file from the kernel into this repo. It's now completely sync'd up, aside from the extra copyright license line found in the kernel headers.

I can remove the whitespace change if that is desired.

jgunthorpe:
It should be byte-for-byte identical with the kernel, so go ahead and include the copyright license too

PR #276, opened 2017-12-14
Various small fixes

jgunthorpe:
There are several places that open code container_of and do it more poorly than the ccan version. Replace them all with ccan container_of then fix all the compile errors from the additional checking ccan container_of does.

Remove container_of from verbs.h to avoid namespace pollution and to ensure we use the ccan version.

Tidy some uses of nested container_of

Remove some pointless assignments and unnecessary function prototypes.

### DIFF
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -67,18 +67,6 @@ union ibv_gid {
 	} global;
 };
 
-#ifndef container_of
-/**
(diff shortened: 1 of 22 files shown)


rleon:
> Remove container_of from verbs.h to avoid namespace pollution and to ensure we use the ccan version.

My gut tells me that we will see complains about it in near future.

PR #277, opened 2017-12-14
Have cbuild map the git alternates

jgunthorpe:
When running commands in the travis configuration it needs access to
the git history to run checkpatch, but if the local repo is using
alternates that fails.

The simple solution is to just map the alternate directories into the
container as well.


rleon:
It is unclear to me the benefits of alternates on such small repo.

Thanks

PR #278, opened 2017-12-18
srp_daemon fixes

nmorey:
This patch series fixes two issues of srp_daemon which occur when the master SM switches.
Both causes a failure to register to SM traps.


nmorey:
@rleon  Can we get this one in ?

rleon:
What did you decide about  "stable versions"?
Why isn't enough to write #v14 so all versions after it will get it?

nmorey:
@rleon agreed. Do you want me to resubmit for this ? Or let's agree to just put the oldest version after this PR ;)

rleon:
better to repush, people will copy/paste :)

nmorey:
Updated :)

rleon:
Thanks, LIFO :)

PR #279, opened 2017-12-25
Remove the obsolete libibcm library

rleon:
The libibcm library was marked as obsolete long time ago,
the following patch is first step before removing ucm from
the kernel.


dledford:
I did a quick check of Red Hat's products.  Using a rhel7 host with the EPEL repo enabled, I did a yum deplist \*, which resulted in checking 484,894 lines worth of output covering 24,269 total binary packages and the result was that not a single package in all of rhel7 + epel linked against libibcm.  So, although we can't remove it from rhel7, removing it from rhel8 should be fairly straightforward.  As such, I don't think Red Hat has any problems with removing libibcm from the upstream rdma-core.  They'll add it back in for rhel7 builds, and drop it in rhel8.

rleon:
Thanks Doug,

Now, I'm more puzzled than before :)
How can we ensure that this library is dropped in rhel8?

Thanks

dledford:
All we have to do upstream is go ahead and remove it from rdma-core.  For rhel8, Red Hat will mark it as gone.  For rhel7, since they aren't allowed to drop a library that shipped with the initial .0 release, they'll add it back in and build it going forward on their own.  Basically, they'll just revert this commit in rhel7 and fix up build failures as they arise, and in rhel8 they'll know that the library is gone and nothing further needs done on our part.

rleon:
Jason explained to me that RH takes latest rdma-core to their stable releases (7.X) and such change (removal) can theoretically cause to RH to stop to do it and stick with some old library.

Can you please sync with Don to ensure that he is ok with removal and internal reverting till RHEL 7.X is updated with latest rdma-core?

Thanks

dledford:
I've email the appropriate people inside Red Hat.

dledford:
The response was basically what I expected: "Given how few bugs there have been about this library (maybe 1 in the last 2 years), I think it can safely be removed and I can carry it forward internally" (slightly paraphrased from Jarod Wilson).  So, unless someone from SuSE or Debian objects, I think we can merge this.

bdrung:
libibcm has no reverse dependencies in Debian. The [code search for libibcm](https://codesearch.debian.net/search?q=libibcm&perpkg=1) does not reveal any users. So from the Debian/Ubuntu perspective, this library can go away.

rleon:
Thanks for prompt response.
Let's wait for response from @nmorey

PR #280, opened 2017-12-25
Add thread domain support

yishaih:
This series enables grouping verbs objects together for the purpose of optimizing the behavior of the driver in a multi-threaded environment.
The first use of this will be for mlx5 to group UAR registers in a way that avoids needing locking on the post send path.
The series follows an RFC and a detailed discussion over it in the ML, the matching kernel side was sent to rdma-next for 4.16.

### DIFF
--- a/providers/mlx5/verbs.c
+++ b/providers/mlx5/verbs.c
@@ -153,6 +153,56 @@ struct ibv_pd *mlx5_alloc_pd(struct ibv_context *context)
 	return &pd->ibv_pd;
 }
 
+static struct mlx5_bf *mlx5_attach_dedicated_bf(struct ibv_context *context)
+{
+	return NULL;
+}
+
+static void mlx5_detach_dedicated_bf(struct ibv_context *context, struct mlx5_bf *bf)
+{
+}
+
+struct ibv_td *mlx5_alloc_td(struct ibv_context *context, struct ibv_td_init_attr *init_attr)
+{
+	struct mlx5_td	*td;
+
+	if (init_attr->comp_mask) {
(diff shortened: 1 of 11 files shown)


providers/mlx5/verbs.c
```diff
@@ -154,13 +154,102 @@ struct ibv_pd *mlx5_alloc_pd(struct ibv_context *context)
...
+
+	bfreg_dyn_index = mlx5_get_bfreg_index(ctx);
+	if (bfreg_dyn_index < 0) {
+		errno = ENOENT;
+		return NULL;
+	}
+
+	bfreg_total_index = ctx->start_dyn_bfregs_index + bfreg_dyn_index;
+	/* Check whether this bfreg index was already mapped and ready to be used */
+	if (ctx->bfs[bfreg_total_index].reg)
```
rzambre: @yishaih is this just a sanity check or is there a scenario where this would evaluate to true? `mlx5_get_bfreg_index()` returns an unused `bfreg` (whose count is 0). Locks on `dyn_bfregs_mutex` ensures that no two threads will get the same `bfreg` so for an unused `bfreg`, `reg` wouldn't be set only by the thread that gets it. Each dynamic `bfreg` is used only once and so the check would never evaluate to true. Just want to make sure I am not missing anything.

yishaih: It may be evaluate to true post mlx5_detach_dedicated_bf(), then it will be available back for further usage.

rzambre: Yes, makes sense. Thanks!

PR #281, opened 2018-01-08
Revise provider initialization

jgunthorpe:
This make all the context creation uniform by having every driver create a struct verbs_context using the same startup flow.

Long ago we added a new and flow, which was intended to support a stable provider ABI. Now that the ABI is private we do not need that.

By having every single provider create a verbs_context we can now rely on things like the priv existing and a sane init ordering where the priv can exist even before the first command runs.

Having priv always available is a valuable component to introduce support for the new ioctl kabi.

Finally, the last three patches rework the ops setup to use a struct of function pointers scheme and provides defaults for every function. The defaults return ENOSYS which fixes a bunch of cases where the user could have called a function without driver support and got a crash.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,7 @@ set(PACKAGE_VERSION "17.0")
 # When this is changed the values in these files need changing too:
 #   debian/libibverbs1.symbols
 #   libibverbs/libibverbs.map
-set(IBVERBS_PABI_VERSION "16")
+set(IBVERBS_PABI_VERSION "17")
 set(IBVERBS_PROVIDER_SUFFIX "-rdmav${IBVERBS_PABI_VERSION}.so")
 
 #-------------------------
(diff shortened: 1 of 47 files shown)


jgunthorpe:
Rebase onto master and incorporate the thread domain/parent domain changes.

jgunthorpe:
Updated with a minor checkpatch fix and the change to use void * arith discussed on the mailing list:

```diff
--- a/libibverbs/device.c
+++ b/libibverbs/device.c
@@ -248,8 +248,7 @@ void *_verbs_init_and_alloc_context(struct ibv_device *device, int cmd_fd,
                return NULL;
        }
 
-       context = (struct verbs_context *)((uint8_t *)drv_context +
-                                          (uintptr_t)context_offset);
+       context = drv_context + (uintptr_t)context_offset;
 
        if (verbs_init_context(context, device, cmd_fd))
                goto err_free;
@@ -288,7 +287,7 @@ LATEST_SYMVER_FUNC(ibv_open_device, 1_1, "IBVERBS_1.1",
         * then it closes cmd_fd and returns NULL
         */
        context_ex = verbs_device->ops->alloc_context(device, cmd_fd);
-       if (context_ex == NULL)
+       if (!context_ex)
                return NULL;
 
        if (context_ex->create_cq_ex) {
```

dledford:
Jason, can you make a change and repush this (or in some other way trigger a recheck by travis)?  I'm wondering if the travis fixes I merged fix the travis failures here.

jgunthorpe:
@dledford yes I will sort it..

PR #283, opened 2018-01-15
Fixes for cbuild and travis

jgunthorpe:
Fix the unconditional travis failures with the latest suse tumbleweed docker images
Fix perplexing travis failures with no output from sparse.
Use clang 5.0

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -73,6 +73,7 @@ class DockerFile(object):
         self.lines = ["FROM %s"%(src)];
 
 class Environment(object):
+    python_cmd = "python";
     aliases = set();
     use_make = False;
     proxy = True;
@@ -324,6 +325,9 @@ class tumbleweed(ZypperEnvironment):
     name = "tumbleweed";
     specfile = "suse/rdma-core.spec";
     rpmbuild_options = [ "--without=curlmini" ];
+    # tumbleweed no longer includes python2 in the base docker images, only python3,
+    # use that instead.
+    python_cmd = "python3";
 
 # -------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


buildlib/cbuild
```diff
@@ -474,17 +478,17 @@ def run_rpm_build(args,spec_file,env):
...
 os.setuid({uid:d});
 
 # Get RPM to tell us the expected tar filename.
 for ln in subprocess.check_output(["rpmspec","-P",{tspec_file!r}]).splitlines():
-   if ln.startswith("Source:"):
-      tarfn = ln.strip().partition(' ')[2].strip();
-os.symlink({tarfn!r},os.path.join("SOURCES",tarfn));
+   if ln.startswith(b"Source:"):
+      tarfn = ln.strip().partition(b' ')[2].strip();
+os.symlink({tarfn!r},os.path.join(b"SOURCES",tarfn));
```
nmorey: Where is this b coming from and why ?

jgunthorpe: subprocess.check_output returns a bytes, so 'ln' is a bytes, 'tarfn' is a bytes, etc.

nmorey: oh ok. Sorry I'm a total python noobie.
This cleared it for me:
https://stackoverflow.com/questions/6269765/what-does-the-b-character-do-in-front-of-a-string-literal.

I'm good with the first patch then :)

jgunthorpe: Oh, I thought you were asking something else.. Yes, the need for the 'b'  is related to very invasive unicode handling in python3. The two alternates here are to use the 'b'  or to call 'decode()' early on. The latter suffers from a couple of subtle faults if the spec file has non-ASCII characters, while 'b' works properly in all cases, if not a little ugly.

dledford: I hate python....

bdrung: It's better to use `decode()` to get proper strings.

jgunthorpe: @bdrung, yes that is what I thought was being asked.. There are subtle faults with using decode - rpmbuild doesn't do this so we shouldn't either.

bdrung: @jgunthorpe Isn't the file to be in a certain format? e.g. most files in `debian/` needs to be UTF-8. Then `decode()` is the right choice.

PR #284, opened 2018-01-17
Clock info

yishaih:
This series adds the ability to query mlx5 core clock info and translates timestamps from HCA core clock units to nanoseconds using the clock info.

The supplementary kernel part was sent upstream for 4.16.


jgunthorpe:
The travis failure is checkpatch and looks ignorable to me

PR #286, opened 2018-01-23
Small bunch of cppcheck fixes 

rleon:
Various fixes spotted by cppcheck followed Ubuntu security team review:
https://bugs.launchpad.net/ubuntu/+source/rdma-core/+bug/1732892

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -2934,7 +2934,7 @@ static void acm_set_options(void)
 		if (s[0] == '#')
 			continue;
 
-		if (sscanf(s, "%32s%256s", opt, value) != 2)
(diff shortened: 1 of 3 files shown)


jgunthorpe:
Everything looks fine except the uninitialized variables stuff, it looks like that tool is somehow choking on list_for_each, and we certainly are not going to add bogus initializers to evey list_for_each user. I'd probably drop that patch as we already have gcc and clang doing that analysis correctly.

rleon:
@jgunthorpe, I didn't like that commit either, so wrote in commit message that it is false alarm

rleon:
removed, rebased and repushed

PR #289, opened 2018-02-07
mlx5: Add missing include file in mlx5dv.h

yishaih:
The 'off_t' type requires the #include <sys/types.h> in some environments.

This patch adds this header file explicitly in mlx5dv.h and prevents the
need from applications to explicitly add it.

Reported-by: Shahaf Shuler <shahafs@mellanox.com>
Signed-off-by: Yishai Hadas <yishaih@mellanox.com>

### DIFF
--- a/providers/mlx5/mlx5dv.h
+++ b/providers/mlx5/mlx5dv.h
@@ -35,6 +35,7 @@
 
 #include <stdio.h>
 #include <linux/types.h> /* For the __be64 type */
+#include <sys/types.h>
 #include <endian.h>
 #if defined(__SSE3__)
 #include <limits.h>


jgunthorpe:
SUSv2 requires sys/types.h to get off_t, so this is absolutely right.

The check-build script actually tests for stuff like this. I have been wondering lately if we should run it from the distro packaging so we don't ship broken headers..

rleon:
Yes and no.
Yes arguments are obvious.
No arguments - we are running latest versions of distros, so most probably the headers will be fine there. The problems usually are related to backward compatibility - attempts to run modern rdma-core on old distro and this is not tested in our CI.

nmorey:
@yishaih For next time, please add the oldest branch it should be backported to 

In this case:
```
Cc: stable@linux-rdma.org #v15
```

I will push this on v15 and v16 stable branches

rleon:
@nmorey we followed Documentation/stable.md, at the beginning of "Option 1", you wrote that CC without tags is enough.

nmorey:
@rleon I'm not complaining :) Thanks for having the cc to stable added.

I will never NOT take a patch that is missing the version.
But add the versions, it's less work for me and fewer chances I make a mistake or forget something !

Take it as a good practice hint to make sure everything works as smooth as possible and not as me being a grumpy maintainer requiring this ;)

BTW, pushed to both v15 and v16 stable branches

rleon:
Thanks, I'll be more strict next time :)

nmorey:
I'm also OK with people tagging me and writing it out in the PRs :)
If it's the only thing missing, no need for a patch respin

PR #297, opened 2018-02-26
ibacm: Support FDR and EDR rates

mattaezell:
Signed-off-by: Matt Ezell <ezellma@ornl.gov>

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -1818,32 +1818,44 @@ static void acm_server(bool systemd)
 enum ibv_rate acm_get_rate(uint8_t width, uint8_t speed)
 {
 	switch (width) {
-	case 1:
+	case 1: /* 1x */
 		switch (speed) {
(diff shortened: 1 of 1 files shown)


nmorey:
Might be worth pushing to stable branches

nmorey:
@jgunthorpe @rleon How do you feel about backporting this kind of "fixes" to the stable branch (and this one in particular)?

IMHO, from a distro PoV, it makes sense to add these patches to older rdma-core releases so that newer HW works well to.
And it's small enough to not cause any problems

rleon:
@nmorey, I have no preference. 

Sorry

jgunthorpe:
I think it is worth backporting this simply because it is pretty much a stealth ABI change. If software starts using these new values it will silently work differently (but not really wrongly) without this patch. Better to have a consistent ABI everywhere??

nmorey:
@jgunthorpe Agreed. Backporting to stable-v15+ branches

PR #299, opened 2018-03-05
Ml 05 04 2018

rleon:


rleon:
Merged with checkpatch error in ccan/bitmap.h file

PR #300, opened 2018-03-05
mlx5: Fix referencing uninitialized var in create_qp error flow

yishaih:
Fix an error flow in mlx5 create qp.


nmorey:
Shouldn't this be backported to all stable branches ?

nmorey:
Backported to stable-v1[567]

PR #308, opened 2018-03-20
Remove uapi structure copies from drivers

jgunthorpe:
This converts most of the drivers to have largely empty -abi.h headers that contain only simple mechanical declarations, such as:

```c
DECLARE_DRV_CMD(mlx5_alloc_ucontext, IB_USER_VERBS_CMD_GET_CONTEXT,
		mlx5_ib_alloc_ucontext_req_v2, mlx5_ib_alloc_ucontext_resp);
```

Similar to the new DECLARE_CMD macro, this automatically constructs a driver request and response structure using the kernel uapi headers:

```c
struct mlx5_alloc_ucontext {
     struct ibv_get_context ibv_cmd;
     union {
          _STRUCT_mlx5_ib_alloc_ucontext_req_v2; // Inlined version of below
          struct mlx5_ib_alloc_ucontext_req_v2 drv_payload;
     };
};

struct mlx5_alloc_ucontext_resp {
    struct ib_uverbs_get_context ibv_resp;
    union {
        _STRUCT_mlx5_ib_alloc_ucontext_resp; // Inlined version of below
        struct mlx5_ib_alloc_ucontext_resp drv_payload;
    };
};
```

As with kern-abi.h, a codegen script is used to produce the _STRUCT_* macros from the real kernel ABI headers.

Everything is tied together with the enum tag to minimize typing and avoid errors, the macros match the correct core structs via the tag to the driver structs. This is preparation for a future bulk change to push the new ioctl API down into the drivers.

The macro also contains static_assertions that ensure padding isn't being added when the structs are concatenated, which detected several bugs.

Patches for the remaining drivers are still pending on the kernel changes.

This work supports the 32/64 compat by forcing the userspace to use the same memory layouts as the kernel after the kernel layouts have been validated to be good.

### DIFF
--- a/buildlib/make_abi_structs.py
+++ b/buildlib/make_abi_structs.py
@@ -24,8 +24,8 @@ def in_struct(ln,FO,nesting=0):
     """Copy a top level structure over to the #define output, keeping track of
     nested structures."""
     if nesting == 0:
-        if ln == "};":
-            FO.write("}\n\n");
+        if re.match(r"(}.*);",ln):
+            FO.write(ln[:-1] + "\n\n");
             return find_struct;
 
     FO.write(ln + " \\\n");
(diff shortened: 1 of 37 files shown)


jgunthorpe:
Updated to drop the hns patch, that one needs newer kernel headers

PR #313, opened 2018-03-24
Fixes so that code will compile using the musl c library and GCC 6.3.0

raphaelcohn:
Changes are mostly to headers to either use the standard location or
to include a missing header that is implicit with glibc.

A small problem with a private constant has also been corrected.

These changes also ensure that all current warnings are fixed. This
is important, as there is not other easy way to fail the build for
genuine missing function definitions.

### DIFF
--- a/libibumad/umad.c
+++ b/libibumad/umad.c
@@ -34,7 +34,7 @@
 
 #include <config.h>
 
-#include <sys/poll.h>
+#include <poll.h>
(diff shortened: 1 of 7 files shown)


jgunthorpe:
The header files changes looked OK, but the fxstat thing breaks the build on glibc, so you'll need to come up with something else there..

raphaelcohn:
Yep...

PR #320, opened 2018-04-09
Limit server to 127.0.0.1 & don't serve requests from librdmacm

gerd-rausch:
Please consider the following commits that add configuration options (with compile-time defaults) to:
1) Limit incoming ibacm connection to loopback (IPv4 127.0.0.1) for security reasons
2) Add an option to ibacm to not serve librdmacm requests, but only respond to 'ib_acme'
   and kernel (via NETLINK socket) requests:
   This is achieved by creating two ".port" files instead of a single one.


gerd-rausch:
Hi Nicolas,

I would gladly update the man-page, but was unable to find one for "ibacm_addr.cfg".
I assumed the only documentation available was within the template file generated via
"ib_acme -O" (that I updated).

What am I missng?

Thanks,
  Gerd

gerd-rausch:
Hi Jason,

I presume we want to preserve TCP-loopback "127.0.0.1" and any/open,
just add unix?

How about selecting this via a new option called "server_mode", with:

#      Selects which clients can connect to this server:
#      unix) Uses unix-domain sockets, hence limits service to the same machine.
#      loop) Limits incoming connections for server_port to 127.0.0.1.
#      open) Allows incoming connections from any client.

What should the default be? "open" or "unix?

A slightly different issue (but somewhat related, not completely off-topic):

I talked to some of my colleagues that have anecdotal reports of
"ibacm" not performing well at large scale:
They're talking about 50+k librdmacm client connections.

With TCP I would expect them to potentially run out client sockets
(out of ports on 127.0.0.1), and also "select" as used today in ibacm,
is known to waste a lot of cycles in scenarios where a lot of
mostly idle client connections are served.

Is there any test/data on "ibacm" performance based on scale?

Thanks,

  Gerd

gerd-rausch:
Hi,

A prototype implementation of this (i.e. unix-domain support) can be found at:

https://github.com/gerd-rausch/rdma-core.git, branch "ibacm-unix-domain":

### Caveats:
* The change to ibacm/ibacm.socket is untested.
  Since it's so trivial, it should just work (famous last words)

* The behavior of when to use unix-domain sockets is inconsitent:
  o "ibacm" server uses it if either called with "--systemd",
    or if "ibacm_opts.cfg" says so.
  o "ib_acme" uses it if it was a compile-time default or specified via "-S /..."
  o "librdmacm" uses it if "ibacm.port" file is non-existent, can't be read,
    or the port# in there ends up being zero.

* The change to the man-page ib_acme.1 does not depend on the compile-time default
  for server-mode:
  It always claims "localhost" even if unix-domain socket was chosen to be the default.

Thanks,

  Gerd

PS: I need to switch context and work on something else... won't be able to spend
    cycles on this until Monday, but wanted to share what I got so far.

gerd-rausch:
+Aron

gerd-rausch:
Hi Jason,

I'm sure you are digging out after the workshop last week but I wanted
to checkin on a couple of things.

1. Do we all agree on what the new acme option names are?  Agreeing on
   the configuratin stuff, even if other changes are made to the
   implementation, reduces the potential mess if I have to do a build for
   Exadata before the patches are accepted upstream. They want me to do
   a build tomorrow... :-)

   # server_mode:
   # Selects how clients can connect to this server:
   # unix - Use unix-domain sockets, hence limits service to the same machine.
   # loop - Limit incoming connections for server_port to 127.0.0.1.
   # open - Allow incoming connections from any TCP client (internal or external).

   server_mode sounds good to me. socket_something would also work but
   then one might want to tweak the options values from what is listed
   above. I think I'm fine with it as-is.

   Question: The server_mode is global so the same mechanism is used for
   both acme and librdmacm? Below I propose that acme have its own
   channel and that it is always a Unix Domain Socket (although one
   could have "server_mode_client" and "server_mode_acme" if it is
   really necessary to have all three options for both clients and
   acme.)

2. We just discussed one minor change to the implementation which is to
   leave the port file name for librdmacm as-is and make the new one
   a dedicated communication channel for acme. This has the benefit of
   not having to touch librdmacm and it seems reasonable to let acme
   have its own dedicated channel for talking to the server.

   # acme_plus_kernel_only:
   # If set to 'true', 'yes' or a non-zero number
   # ibacm will only serve requests originating
   # from the kernel or the ib_acme utility.

   Question: Does it make sense to have a mode where kernel requests are
   not served but userspace is?
   
   kernel  userspace  ib_acme
    no      yes        yes
    yes     no         yes
    yes     yes        yes

   Because acme is always allowed to communicate with the server,
   perhaps this option name should reference just kernel vs. userspace
   and a note can be added that ib_acme is always able to communicate
   using its own channel which is always a Unix Domain Socket (unless
   server_mode_acme is implemented.)

Aron

gerd-rausch:
On Tue, Apr 17, 2018 at 03:38:48PM -0500, Aron Silverton wrote:
> Hi Jason,
> 
> I'm sure you are digging out after the workshop last week but I wanted
> to checkin on a couple of things.
> 
> 1. Do we all agree on what the new acme option names are?  Agreeing on
>    the configuratin stuff, even if other changes are made to the
>    implementation, reduces the potential mess if I have to do a build for
>    Exadata before the patches are accepted upstream. They want me to do
>    a build tomorrow... :-)
> 
>    # server_mode:
>    # Selects how clients can connect to this server:
>    # unix - Use unix-domain sockets, hence limits service to the same machine.
>    # loop - Limit incoming connections for server_port to 127.0.0.1.
>    # open - Allow incoming connections from any TCP client (internal or external).
> 
>    server_mode sounds good to me. socket_something would also work but
>    then one might want to tweak the options values from what is listed
>    above. I think I'm fine with it as-is.

This seems unnceessary in a systemd world, for systemd I think I
already had it support multiple listen sockets of any type
automatically, right?

So don't care too much what you do for non-systemd
systems. server_mode sounds find, just document clearly this is not
used when using systemd...

>    Question: The server_mode is global so the same mechanism is used for
>    both acme and librdmacm? Below I propose that acme have its own
>    channel and that it is always a Unix Domain Socket (although one
>    could have "server_mode_client" and "server_mode_acme" if it is
>    really necessary to have all three options for both clients and
>    acme.)

Always using unix domain by default seems fine to me, but I wouldn't
say it is private..

> 2. We just discussed one minor change to the implementation which is to
>    leave the port file name for librdmacm as-is and make the new one
>    a dedicated communication channel for acme. This has the benefit of
>    not having to touch librdmacm and it seems reasonable to let acme
>    have its own dedicated channel for talking to the server.
> 
>    # acme_plus_kernel_only:
>    # If set to 'true', 'yes' or a non-zero number
>    # ibacm will only serve requests originating
>    # from the kernel or the ib_acme utility.
> 
>    Question: Does it make sense to have a mode where kernel requests are
>    not served but userspace is?
>    
>    kernel  userspace  ib_acme
>     no      yes        yes
>     yes     no         yes
>     yes     yes        yes

Again, doesn't systemd mode already handle these combinations?

I'd probably just go for the simple solution of always and only using
unix domain sockets when not using systemd.

Realistically if someone is using a non-systemd OS still they are
unlikely to upgrade to this new rdma-core version. Oracle may be the
only exception.

Jason

gerd-rausch:
Hi Jason,

On 04/17/2018 07:42 PM, Jason Gunthorpe wrote:
> On Tue, Apr 17, 2018 at 03:38:48PM -0500, Aron Silverton wrote:
>>
>>    server_mode sounds good to me. socket_something would also work but
>>    then one might want to tweak the options values from what is listed
>>    above. I think I'm fine with it as-is.
> 
> This seems unnceessary in a systemd world, for systemd I think I
> already had it support multiple listen sockets of any type
> automatically, right?
> 

The objective was to retain compatibility and being able to
support non-systemd installations.

> So don't care too much what you do for non-systemd
> systems. server_mode sounds find, just document clearly this is not
> used when using systemd...

Will do.

>>    
>>    kernel  userspace  ib_acme
>>     no      yes        yes
>>     yes     no         yes
>>     yes     yes        yes
> 
> Again, doesn't systemd mode already handle these combinations?
> 

It doesn't handle the combination where ibacm serves ib_acme but not
librdmacm (the second patch, that adds 'acme_plus_kernel_only'.

Not the patch that adds 'server_mode' talked about above.

> I'd probably just go for the simple solution of always and only using
> unix domain sockets when not using systemd.
> 

That would render some old installations with an older librdmacm,
that do not use unix-domain sockets yet non-functional.


> Realistically if someone is using a non-systemd OS still they are
> unlikely to upgrade to this new rdma-core version. Oracle may be the
> only exception.
> 

Forward looking yes.
But for compatibility sake, IMHO it is best to retain what used to work.

Thanks,

  Gerd

jgunthorpe:
On Wed, Apr 18, 2018 at 03:30:57PM +0000, gerd-rausch wrote:
>    Hi Jason,
>    On 04/17/2018 07:42 PM, Jason Gunthorpe wrote:
>    > On Tue, Apr 17, 2018 at 03:38:48PM -0500, Aron Silverton wrote:
>    >>
>    >> server_mode sounds good to me. socket_something would also work but
>    >> then one might want to tweak the options values from what is listed
>    >> above. I think I'm fine with it as-is.
>    >
>    > This seems unnceessary in a systemd world, for systemd I think I
>    > already had it support multiple listen sockets of any type
>    > automatically, right?
>    >
>    The objective was to retain compatibility and being able to
>    support non-systemd installations.
>    > So don't care too much what you do for non-systemd
>    > systems. server_mode sounds find, just document clearly this is not
>    > used when using systemd...
>    Will do.
>    >>
>    >> kernel userspace ib_acme
>    >> no yes yes
>    >> yes no yes
>    >> yes yes yes
>    >
>    > Again, doesn't systemd mode already handle these combinations?
>    >
>    It doesn't handle the combination where ibacm serves ib_acme but not
>    librdmacm (the second patch, that adds 'acme_plus_kernel_only'.
>    Not the patch that adds 'server_mode' talked about above.

So you want an option to not write the path file for librdmacm to
locate the server?

>    > I'd probably just go for the simple solution of always and only using
>    > unix domain sockets when not using systemd.
>    >
>    That would render some old installations with an older librdmacm,
>    that do not use unix-domain sockets yet non-functional.

Don't do that? Why would you want to allow an upgrade of ibacm without
matching librdmacm? This is what package dependencies are for :)

Jason

PR #321, opened 2018-04-20
umad: Do not check for umad sysfs files in umad_init

jgunthorpe:
Now that we don't load the umad module if the HW doesn't use it (eg for roce only hardware) umad_init is failing to read the ABI version from the kernel.
    
Applications still want to use some libibumad services that are not related to the char device, so move the version check to umad_open_port instead.


jgunthorpe:
Here you go @nmorey

libibumad/man/umad_init.3.md
```diff
@@ -24,22 +25,22 @@ int umad_done(void);
...
-**umad_init()** and **umad_done()** return 0 on success, and -1 on error.
-Error is returned from **umad_init()** if infiniband umad can't be opened, or
-the abi version doesn't match. There are no errors currently returned by
-**umad_done().**
+Always 0.
+
+# COMPATIBILITY
 
-# NOTES
+Versions prior to releaes 18 of the library require **umad_init()** to be
```
hnrose: typo: releaes -> release

libibumad/man/umad_init.3.md
```diff
@@ -24,22 +25,22 @@ int umad_done(void);
...
+called prior to using any other library functions. Old versions could return a
+failure code of -1 from **umad_init()**.
 
-If an error occurs during the library initialization, no further use of the
-umad library should be attempted.
+For compatibility applications should continue to cal **umad_init()**, and
```
hnrose: typo: cal -> call

libibumad/umad.c
```diff
@@ -618,9 +627,13 @@ int umad_open_port(const char *ca_name, int portnum)
 {
 	char dev_file[UMAD_DEV_FILE_SZ];
 	int umad_id, fd;
+	unsigned int abi_version = get_abi_version();
 
 	TRACE("ca %s port %d", ca_name, portnum);
 
+	if (!abi_version)
+		return -EOPNOTSUPP;
```
hnrose: This patch looks right to me but I want to test this with the various tools to be sure and won't be able to get to that until either Monday or Tuesday.

jgunthorpe:
Updated with 

```c
diff --git a/libibumad/man/umad_init.3.md b/libibumad/man/umad_init.3.md
index 3110ccfb708825..12bb7a7db25bac 100644
--- a/libibumad/man/umad_init.3.md
+++ b/libibumad/man/umad_init.3.md
@@ -33,11 +33,11 @@ Always 0.
 
 # COMPATIBILITY
 
-Versions prior to releaes 18 of the library require **umad_init()** to be
+Versions prior to release 18 of the library require **umad_init()** to be
 called prior to using any other library functions. Old versions could return a
 failure code of -1 from **umad_init()**.
 
-For compatibility applications should continue to cal **umad_init()**, and
+For compatibility applications should continue to call **umad_init()**, and
 check the return code, prior to calling other **umad_** functions.  If
 **umad_init()** returns an error then, no further use of the umad library
 should be attempted.
```

libibumad/umad.c
hnrose: I ran some tests using OpenSM and infiniband-diags and this is working as expected.

Only thing it does is change when error occurs. It used to occur at umad_init time which is always the first thing and now won't occur until first umad_open_port. This might surprise some users but don't see a way around that.

:+1

PR #322, opened 2018-04-20
Limit server to 127.0.0.1 & don't serve requests from librdmacm (take #2)

gerd-rausch:
This is take#2 of the previous pull-request (#320), with the following changes:

* Supports unix-domain sockets (now the default, even for systemd's ibacm.socket)
* The .port file-name for librdmacm is retained, but the one for ib_acme is chaged
  (in order to be able to serve older (e.g. statically linked) versions of librdmacm)

Not included:
* A note that config optin "server_mode" is not considered when run as "--systemd":
  Reason: This behavior is not limited to "server_mode" but applies to others as well (e.g. server_port)
  A more general explanation of this should be included in the man-page(s) / config_file or it should be extended to each and every option affected.
* Documentation of the new options in a man-page (as I was unable to locate a man-page containing such documentation).
  Options are documented in the generated file "ibacm_opts.cfg" though, with the corresponding defaults.


rleon:
@gerd-rausch , please fix checkpatch errors.

Thanks

gerd-rausch:
@rleon done. Please let me know if you encounter any other issues.

PR #326, opened 2018-05-13
GRH required

yishaih:
This series is the supplementary part of the kernel series that was sent upstream into rdma-next for 4.18.

It exposes a new port capability flag: IBV_PORT_GRH_REQUIRED, when this port capability flag is set, the applications must create all AH with GRH configured.


jgunthorpe:
updated with revised kernel headers

providers/mthca/verbs.c
```diff
@@ -709,6 +709,13 @@ int mthca_destroy_qp(struct ibv_qp *qp)
 struct ibv_ah *mthca_create_ah(struct ibv_pd *pd, struct ibv_ah_attr *attr)
 {
 	struct mthca_ah *ah;
+	struct ibv_port_attr port_attr;
+
+	if (mthca_query_port(pd->context, attr->port_num, &port_attr))
+		return NULL;
+
+	if (port_attr.flags & IBV_QPF_GRH_REQUIRED && !attr->is_global)
+		return NULL;
```
jgunthorpe: What is the point of this? The mthca driver never sets RDMA_CORE_CAP_IB_GRH_REQUIRED and isn't roce..

providers/mlx4/verbs.c
```diff
@@ -1328,6 +1334,10 @@ struct ibv_ah *mlx4_create_ah(struct ibv_pd *pd, struct ibv_ah_attr *attr)
 	if (query_port_cache(pd->context, attr->port_num, &port_attr))
 		return NULL;
 
+	if (port_attr.flags & IBV_QPF_GRH_REQUIRED &&
+	    !attr->is_global)
```
jgunthorpe: Same question here.. Is there some expected future use for this in mlx4? This at least makes more sense since it will be set for roce ports..

providers/mthca/verbs.c
yishaih: The idea was to be consistent for all drivers and have it ready from day one for a case that some kernel support will be added.

providers/mlx4/verbs.c
yishaih: mlx4 turn on this bit today for the roce case, so it's applicable.

providers/mthca/verbs.c
jgunthorpe: mthca is never going to be changed, and doing it in this hacky way creates a big performance regression for it, prefer to see it dropped I think..

yishaih: OK, will drop the mthca part from this patch.

PR #329, opened 2018-05-16
MPLS and GRE flow specifications

yishaih:
This series adds new two flow specifications: GRE and MPLS. 
The matching kernel series was already sent into rdma-next for 4.18.


jgunthorpe:
Remember to re-do the first commit once the kernel patches are accepted so it has the right commit id.

jgunthorpe:
Updated with final kernel headers:

```sh
kernel-headers/update ../k.o e818e255a58d64e86c8c93e3aa52498b1a3d1760
```

PR #332, opened 2018-05-17
Update library version to be 19.0

rleon:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,7 @@ endif()
 set(PACKAGE_NAME "RDMA")
 
 # See Documentation/versioning.md
-set(PACKAGE_VERSION "18.0")
+set(PACKAGE_VERSION "19.0")
 # When this is changed the values in these files need changing too:
 #   debian/libibverbs1.symbols
 set(IBVERBS_PABI_VERSION "18")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ endif()
 set(PACKAGE_NAME "RDMA")
 
 # See Documentation/versioning.md
-set(PACKAGE_VERSION "18.0")
+set(PACKAGE_VERSION "19.0")
 # When this is changed the values in these files need changing too:
 #   debian/libibverbs1.symbols
 set(IBVERBS_PABI_VERSION "18")
```
jgunthorpe: Check this!! I don't think 18 is right, I see this hunk since v18:

```diff
 /*
  * Must change the PRIVATE IBVERBS_PRIVATE_ symbol if this is changed. This is
  * the union of every op the driver can support. If new elements are added to
@@ -211,9 +206,6 @@ struct verbs_context_ops {
        struct ibv_td *(*alloc_td)(struct ibv_context *context,
                                   struct ibv_td_init_attr *init_attr);
        void (*async_event)(struct ibv_async_event *event);
-       int (*attach_counters_point_flow)(struct ibv_counters *counters,
-                                         struct ibv_counter_attach_attr *attr,
-                                         struct ibv_flow *flow);
```

rleon: I was under compression that this symbol should be updated once PABI is changed by developers who break PABi and not supposed to be updated in release. I really-really don't want to dig into the code during release.

PR #333, opened 2018-05-17
gcc 8 fixes

rleon:


rleon:
I have 2-3 extra patches, once I'll finish them, i'll send bundle to the ML.

providers/mlx5/mlx5.c
```diff
@@ -305,14 +305,14 @@ man cpuset
 */
 static void mlx5_local_cpu_set(struct ibv_device *ibdev, cpu_set_t *cpu_set)
 {
-	char *p, buf[1024];
+	char *p, *buf = NULL;
```
yishaih: This is used in the 'else' flow below (317) as a buffer, the code here is wrong.

PR #338, opened 2018-05-29
ibacm: Option "acme_plus_kernel_only" is ignored if controlled by "--…

gerd-rausch:
…systemd"

Commit "95df9ed596" introduced an option "acme_plus_kernel_only",
that made a distinction between communcation with
* "ib_acme" via socket "/run/ibacm-unix.sock"
* "librdmacm" via symlink "/run/ibacm.sock".

Since systemd does not parse "ibacm_opts.cfg" (nor should it),
simply ignore this option if "ibacm" is controlled by "systemd":
Assume that option is set to "no".

Also make that fact known to the generated comments of "ibacm_opts.cfg"

Signed-off-by: Gerd Rausch <gerd.rausch@oracle.com>

### DIFF
--- a/ibacm/ibacm.socket
+++ b/ibacm/ibacm.socket
@@ -1,3 +1,21 @@
+# Please copy this file to /etc/systemd/system/ibacm.socket
+# before modification, if not done already.
+#
+# When using socket-based activation of the 'ibacm' service
+# ibacm's configuration option 'acme_plus_kernel_only' is ignored
+# (i.e. an implicit 'acme_plus_kernel_only no')
+#
+# In order to get the equivalent behavior of
+# configuration 'acme_plus_kernel_only yes'
+# Please add a comment (i.e. a '#' character) in front
+# of the line 'Symlinks' below, and ensure that file
+# file '/run/ibacm.sock' does not exist:
+# e.g. by using "rm -f /run/ibacm.sock" after modifying
+# the copy of this file that lives in /etc/systemd/system.
+#
+# Please also remember to reload the systemd configuration by running:
+# % systemctl --system daemon-reload
+
 [Unit]
 Description=Socket for InfiniBand Address Cache Manager Daemon
 Documentation=man:ibacm
@@ -7,7 +25,8 @@ Before=rdma-hw.target
 # ibacm.socket always starts
 
 [Socket]
-ListenStream=/run/ibacm.sock
+ListenStream=/run/ibacm-unix.sock
+Symlinks=/run/ibacm.sock
 
 # Bind to PF_NETLINK, NETLINK_RDMA, RDMA_NL_GROUP_LS
 # Supported in systemd > 234
--- a/ibacm/src/acme.c
+++ b/ibacm/src/acme.c
@@ -217,6 +217,10 @@ static void gen_opts_temp(FILE *f)
 	fprintf(f, "# If set to 'true', 'yes' or a non-zero number\n");
 	fprintf(f, "# ibacm will only serve requests originating\n");
 	fprintf(f, "# from the kernel or the ib_acme utility.\n");
+	fprintf(f, "# Please note that this option is ignored if the ibacm\n");
+	fprintf(f, "# service is started on demand by systemd,\n");
+	fprintf(f, "# in which case this option is treated\n");
+	fprintf(f, "# as if it were set to 'no'\n");
 	fprintf(f, "\n");
 #if IBACM_ACME_PLUS_KERNEL_ONLY_DEFAULT
 	fprintf(f, "acme_plus_kernel_only yes\n");


ibacm/ibacm.socket
```diff
@@ -7,7 +7,8 @@ Before=rdma-hw.target
 # ibacm.socket always starts
 
 [Socket]
-ListenStream=/run/ibacm.sock
```
jgunthorpe: I think a comment describing how to get the 'acme_plus_kernel_only' behavior here would be appropriate.

Guessing it is 
```
ListenStream=/run/ibacm-unix.sock
SocketMode=0600
```
 ?

gerd-rausch:
On 2018-06-04 15:38, Jason Gunthorpe wrote:
> *@jgunthorpe* requested changes on this pull request.
> 
> ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> 
> In ibacm/ibacm.socket <https://github.com/linux-rdma/rdma-core/pull/338#discussion_r192900728>:
> 
>> @@ -7,7 +7,8 @@ Before=rdma-hw.target
>  # ibacm.socket always starts
>  
>  [Socket]
> -ListenStream=/run/ibacm.sock
> 
> I think a comment describing how to get the 'acme_plus_kernel_only' behavior here would be appropriate.

True, I'll update the pull-request.

> 
> Guessing it is
> 
> |ListenStream=/run/ibacm-unix.sock SocketMode=0600 |
> 

SocketMode should remain at its default value of 0666:
Unprivileged processes using librdmacm
access /run/ibacm-unix.sock via symlink /run/ibacm.sock.

Thanks,

  Gerd

jgunthorpe:
What is the point of the option if sockets have the same permissions?

gerd-rausch:
On 2018-06-04 15:53, Jason Gunthorpe wrote:
> What is the point of the option if sockets have the same permissions?
> 

To make librdmacm not talk to the "ibacm" daemon, while both "ib_acme" and
the kernel (via netlink) DO talk to the "ibacm" daemon.

It's an Oracle thing:
Some existing application has trouble when "ibacm" is turned on.

So we need some way to distinguish "ib_acme" from "librdmacm" (regardless of permissions).
Hence the symlink.

Thanks,

  Gerd

jgunthorpe:
Weird. Ok

ibacm/ibacm.socket
```diff
@@ -1,3 +1,21 @@
+# Please copy this file to /etc/systemd/system/ibacm.socket
+# before modification, if not done already.
+#
```
jgunthorpe: This isn't quite the right instruction for overriding a system unit file, it should be done in a drop in that contains only the updated information eg

```
[Socket]
Symlinks=
````

Assuming that works for symlinks, otherwise the direction is Ok.

Something like this is general guidance
https://stackoverflow.com/questions/34802345/systemd-how-to-clear-an-entry-when-overriding-a-unit-file

gerd-rausch:
On 2018-06-12 20:30, Jason Gunthorpe wrote:
> *@jgunthorpe* commented on this pull request.
> 
> ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> 
> In ibacm/ibacm.socket <https://github.com/linux-rdma/rdma-core/pull/338#discussion_r194947754>:
> 
>> @@ -1,3 +1,21 @@
> +# Please copy this file to /etc/systemd/system/ibacm.socket
> +# before modification, if not done already.
> +#
> 
> This isn't quite the right instruction for overriding a system unit file, it should be done in a drop in that contains only the updated information eg
> 
> |[Socket] Symlinks= |
> 
> Assuming that works for symlinks, otherwise the direction is Ok.
> 
> Something like this is general guidance
> https://stackoverflow.com/questions/34802345/systemd-how-to-clear-an-entry-when-overriding-a-unit-file
> 


Hi Jason,

According to "man 5 systemd.unit", section "Example 2. Overriding vendor settings":

  There are two methods of overriding vendor settings in unit files: copying the unit file from /usr/lib/systemd/system to
  /etc/systemd/system and modifying the chosen settings. Alternatively, one can create a directory named unit.d/ within
  /etc/systemd/system and place a drop-in file name.conf there that only changes the specific settings one is interested in. Note
  that multiple such drop-in files are read if present.

  The advantage of the first method is that one easily overrides the complete unit, the vendor unit is not parsed at all anymore.
  It has the disadvantage that improvements to the unit file by the vendor are not automatically incorporated on updates.

  The advantage of the second method is that one only overrides the settings one specifically wants, where updates to the unit by
  the vendor automatically apply. This has the disadvantage that some future updates by the vendor might be incompatible with the
  local changes.

  Note that for drop-in files, if one wants to remove entries from a setting that is parsed as a list (and is not a dependency),
  such as ConditionPathExists= (or e.g.  ExecStart= in service units), one needs to first clear the list before re-adding all
  entries except the one that is to be removed. See below for an example.


I was describing the first method in the comments, i.e. copying & overriding the complete unit.
Or at least I thought so.

If I read the documentation correctly, lists (e.g. "Symlinks") need not be reset first.

And the testing was consistend with this documentation: i.e. it worked for me.


Thanks,

  Gerd

jgunthorpe:
My point was to describe the second method in the comment, as I think that is the preferred method by distros/etc

gerd-rausch:
Hi Jason,

On 2018-06-19 11:36, Jason Gunthorpe wrote:
> My point was to describe the second method in the comment, as I think that is the preferred method by distros/etc
> 

Resetting "Symlinks" via, e.g.:

echo -e '[Socket]\nSymlinks=' >/etc/systemd/system/ibacm.socket.d/no_symlink.conf

does _not_ work.

Furthermore, if said file "no_symlink.conf" contains:

[Socket]
Symlinks=
Symlinks=/run/foo.sock

instead of the expected "/run/foo.sock" symlink only, we end up
with both "/run/foo.sock" _and_ "/run/ibacm.sock" symlinks.

The documentation is not very explicit about the expected behavior with regards
to resetting lists in unit override files.

Any suggestions?
Go with the previously pushed approach or chase down systemd bugs (/features)?

Thanks,

  Gerd

jgunthorpe:
Okay, existing comment is fine then. This is one of the weird areas in systemd where they have lists in the config language but not all lists act like the same :(

jgunthorpe:
I revised this to be only one commit, and merged it..

PR #340, opened 2018-06-04
Fix SSE support for older GCC     

nmorey:
On older GCC versions (at least on v4.8), SSE is not enabled by default for x86.
This causes rdma-core compilation to fail on util/mmio.c as it uses SSE functions.
    
This series adds a CMake helper that enables -msse if available and
adds a HAVE_SSE_EXTENSIONS flags.
mmio.c was also tweaked to support compiler without SSE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,6 +318,8 @@ if (NOT HAVE_SPARSE)
   RDMA_DoFixup("${HAVE_STDATOMIC}" "stdatomic.h")
 endif()
 
+RDMA_Check_SSE(HAVE_TARGET_SSE)
+
 # Enable development support features
 # Prune unneeded shared libraries during linking
 RDMA_AddOptLDFlag(CMAKE_EXE_LINKER_FLAGS SUPPORTS_AS_NEEDED "-Wl,--as-needed")
@@ -640,4 +642,7 @@ if (NOT HAVE_C_WREDUNDANT_DECLS)
 endif()
 if (NOT HAVE_GLIBC_UAPI_COMPAT)
   message(STATUS " libc netinet/in.h and linux/in.h do NOT coexist")
+endif()
+if (NOT HAVE_TARGET_SSE)
+  message(STATUS " attribute(target(\"sse\")) does NOT work")
 endif()
--- a/buildlib/RDMA_EnableCStd.cmake
+++ b/buildlib/RDMA_EnableCStd.cmake
@@ -85,3 +85,46 @@ int main(int argc, char *argv[])
 
   set(${TO_VAR} "${HAVE_WORKING_STRICT_ALIASING}" PARENT_SCOPE)
 endfunction()
+
+function(RDMA_Check_SSE TO_VAR)
+  set(SSE_CHECK_PROGRAM "
+    #include <string.h>
+    #if defined(__i386__)
+      #include <xmmintrin.h>
+    #endif
+    int __attribute__((target(\"sse\"))) main(int argc, char *argv[])
+    {
+        int ret = 0;
+
+        #if defined(__i386__)
+	__m128 tmp = {};
+	tmp = _mm_loadl_pi(tmp, (__m64 *)&main);
+	_mm_storel_pi((__m64 *)&main, tmp);
+        ret = memchr(&tmp, 0, sizeof(tmp)) == &tmp;
+        #endif
+        return ret;
+    }")
+
+  CHECK_C_SOURCE_COMPILES(
+    "${SSE_CHECK_PROGRAM}"
+    HAVE_TARGET_SSE
+    FAIL_REGEX "warning")
+
+  if(NOT HAVE_TARGET_SSE)
+    # Older compiler, we can work around this by adding -msse instead of
+    # relying on the function attribute.
+    set(CMAKE_REQUIRED_FLAGS "-msse")
+    CHECK_C_SOURCE_COMPILES(
+      "${SSE_CHECK_PROGRAM}"
+      NEED_MSSE_FLAG
+      FAIL_REGEX "warning")
+    set(CMAKE_REQUIRED_FLAGS)
+
+    if(NEED_MSSE_FLAG)
+      set(SSE_FLAGS "-msse" PARENT_SCOPE)
+    else()
+      message(FATAL_ERROR "Can not figure out how to turn on sse instructions for i386")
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ include(RDMA_DoFixup)
 include(publish_headers)
 include(rdma_functions)
 include(rdma_man)
+include(SSEFlags)
```
jgunthorpe: Often we've just been inlining these a little lower in cmakefile, after the compiler is basically setup.

But there is a big grouping of c compiler related stuff now in buildlib/RDMA_EnableCStd.cmake, so you could add a new function there instead of this file..

buildlib/SSEFlags.cmake
```diff
@@ -0,0 +1,26 @@
...
+  }"
+  HAS_SSE_EXTENSIONS)
+
+set(CMAKE_REQUIRED_FLAGS)
+
+if(HAS_SSE_EXTENSIONS)
+  set(SSE_FLAGS "-msse -DHAVE_SSE_EXTENSIONS")
+endif()
+
+add_definitions(${SSE_FLAGS})
```
jgunthorpe: add_definitions is not for compiler options, -msse should be added to CMAKE_C_FLAGS

Review how buildlib/RDMA_EnableCStd.cmake works..

buildlib/SSEFlags.cmake
```diff
@@ -0,0 +1,26 @@
+# check for SSE extensions
+include(CheckCSourceRuns)
+set(SSE_FLAGS)
+
```
jgunthorpe: These lines are probably not necessary..

buildlib/SSEFlags.cmake
```diff
@@ -0,0 +1,26 @@
...
+set(CMAKE_REQUIRED_FLAGS "-msse")
```
jgunthorpe: The standard test protocol is to run the test in the current environment and look for failure.

If there is failure, then try and fix it with some compiler change.

So the above shouldn't be done, and the test should be changed to only check xmm it on i386, otherwise always pass.

buildlib/SSEFlags.cmake
```diff
@@ -0,0 +1,26 @@
...
+      b = _mm_add_ps(a,b);
+      _mm_storeu_ps(vals,b);
+      return 0;
+  }"
+  HAS_SSE_EXTENSIONS)
+
+set(CMAKE_REQUIRED_FLAGS)
+
+if(HAS_SSE_EXTENSIONS)
+  set(SSE_FLAGS "-msse -DHAVE_SSE_EXTENSIONS")
```
jgunthorpe: We don't use -D arguments, update buildlib/config.h.in instead with

```c
#cmakedefine HAVE_SSE_EXTENSIONS 1
```

nmorey:
@jgunthorpe I think this update answers to all your concerns.
Regarding moving the SSE stub to another file, it makes sense. But as this is also needed into stable, I want to limit the impact.

(The real reason is rdma-core >= 15 do not build on SLES12-SP* for i586 due to this bug and I need to solve this ASAP )

But I changed the use of SSE_FLAGS and only apply it to the mmio.c file which should limit the impact anyway.
The HAVE_SSE_EXTENSIONS define was dropped altogether to use the Gcc __SSE__ flag which is check by the xmm include anyway

buildlib/RDMA_EnableCStd.cmake
```diff
@@ -85,3 +85,43 @@ int main(int argc, char *argv[])
 
   set(${TO_VAR} "${HAVE_WORKING_STRICT_ALIASING}" PARENT_SCOPE)
 endfunction()
+
+function(RDMA_Check_SSE TO_VAR)
+  CHECK_C_SOURCE_COMPILES("
+    #include <xmmintrin.h>
```
jgunthorpe: This all looks fine now,

But I still think this block needs to be enclosed in #if defined(__i386__) so it doesn't fail on other architectures. The 'SEE extensions do NOT work' should only be printed if it is a problem, and it isn't a problem on arm, etc.

buildlib/RDMA_EnableCStd.cmake
```diff
@@ -85,3 +85,43 @@ int main(int argc, char *argv[])
...
+      {
+          __m128 a, b;
+          float vals[4] = {0};
+          a = _mm_loadu_ps(vals);
+          b = a;
+          b = _mm_add_ps(a,b);
+          _mm_storeu_ps(vals,b);
+          return 0;
+      }"
+      NEED_SSE_FLAG)
```
jgunthorpe: you could put this in a variable with set() then use it twice..

jgunthorpe:
regarding needed a separate file, just check the asm produced for mmio.c to make sure gcc is not using see instructions in weird places. I doubt it is.

nmorey:
@jgunthorpe  Hopefully all good this time.

jgunthorpe:
Hurm. Something is wrong, travis reports:

```
-- Performing Test HAVE_SSE_EXTENSIONS
-- Performing Test HAVE_SSE_EXTENSIONS - Failed
-- Performing Test NEED_SSE_FLAG
-- Performing Test NEED_SSE_FLAG - Success
-- Performing Test SUPPORTS_AS_NEEDED
```

And travis always had working SSE. Hmm

It is because the cmake test doesn't use __attribute__((target("sse")))

And I guess that is what is actually wrong with your old compiler, not that it needs -msse, but that the target attribute doesn't work.

I'll send you an email

nmorey:
Yes and no.
Yes the test needs to have the attribute target.

But no it has nothing to do with our build issue:
```
abuild@portia:~> cat /tmp/test.c
#include <xmmintrin.h>
abuild@portia:~> gcc -o /tmp/test.o -c /tmp/test.c 
In file included from /tmp/test.c:1:0:
/usr/lib/gcc/i586-suse-linux/4.8/include/xmmintrin.h:31:3: error: #error "SSE instruction set not enabled"
 # error "SSE instruction set not enabled"
   ^
```

jgunthorpe:
Ah, okay, the include files didn't get fixed until gcc 5.

jgunthorpe:
Updated with the patch sent to the mailing list

nmorey:
Your commit broke the build fo all non i386/x86_64 archs :s

PR #341, opened 2018-06-04
Mlx5 static rate limit

yishaih:
This series fixes mlx5 to work properly with IB static rate limit by converting from the IB static rate value to
to mlx5 internal static rate value.


jgunthorpe:
Did these go on the mailing list?

nmorey:
Backported to stable-15+

PR #342, opened 2018-06-04
verbs: Allow creation of inner MPLS flow spec

yishaih:
This change allows user to provide an inner MPLS flow specification among the filters list.
The matching kernel code was already accepted into rdma-next for 4.18 as part of the MPLS series.


jgunthorpe:
This seems like it needs a man page update..

MPLS was described as an ordered spec, so what does INNER mean  with respect to the ordering? Sounds like it means skip the first? Anyhow, should be documented in a man page.

PR #344, opened 2018-06-11
buildilb: Fix -msse breakage on ARM builds

jgunthorpe:
Travis was broken when this patch was accepted, so this was missed.
Non-x86 arches do not like the attribute(target) on main, so conditionalize
the entire thing.

Cc: stable@linux-rdma.org # v15


jgunthorpe:
@nmorey

PR #348, opened 2018-06-15
debian: Add ibacm-dev package

rfehren:
Hi Benjamin,

could you please apply this patch, build and upload the new deb to unstable? The new package opa-ff (currently in the NEW queue) needs ibacm-dev a as a build dependency. It would also be nice if you could put the package under the debian-hpc salsa group umbrella.

Thanks,

Roland


rfehren:
Resubmitting with signed-off

PR #350, opened 2018-06-15
debian: Add ibacm-dev package

rfehren:
Signed-off-by: Roland Fehrenbacher <rf@q-leap.de>

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -45,6 +45,24 @@ Description: RDMA core userspace infrastructure and documentation
     service in userspace for iWARP drivers to claim TCP ports through the
     standard socket interface.
 
+Package: ibacm-dev
+Section: libdevel
+Architecture: linux-any
+Multi-Arch: same
+Depends: ${misc:Depends}
+Breaks: ibacm (<< 18.1-1)
+Replaces: ibacm (<< 18.1-1)
+Description: Development files for the ibacm library
+ The IB ACM implements and provides a framework for name, address, and
+ route (path) resolution services over InfiniBand.
+ It is intended to address connection setup scalability issues running
+ MPI applications on large clusters.  The IB ACM provides information
+ needed to establish a connection, but does not implement the CM protocol.
+ A primary user of the ibacm service is the librdmacm library.
+ .
+ This package is needed to compile ibacm providers. It contains just the ibacm
+ header files. The corresponding library is part of the ibacm package.
+
 Package: ibacm
 Architecture: linux-any
 Depends: lsb-base (>= 3.2-14~),
--- a/debian/ibacm-dev.install
+++ b/debian/ibacm-dev.install
@@ -1,0 +1,4 @@
+usr/include/infiniband/acm.h
+usr/include/infiniband/acm_prov.h
+usr/share/man/man7/ibacm.7
+usr/share/man/man7/ibacm_prov.7
--- a/debian/ibacm.install
+++ b/debian/ibacm.install
@@ -2,12 +2,8 @@ etc/init.d/ibacm
 lib/systemd/system/ibacm.service
 lib/systemd/system/ibacm.socket
 usr/bin/ib_acme
-usr/include/infiniband/acm.h
-usr/include/infiniband/acm_prov.h
 usr/lib/*/ibacm/libibacmp.so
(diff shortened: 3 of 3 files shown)


jgunthorpe:
I'm pretty sure we don't provide an appropriately controlled ABI for the plugins, nor do we have the development processes to avoid breaking it at this time..  So I don't really want to advertise that 3rd parties could make a plugin outside the rdma-core tree and then go on to support that..

Particularly since the plugin mechanism used by acm relies on importing symbols from the ibacm executable rather than more robust dynamic linking..

Could you consider contributing your plugin?

jgunthorpe:
And please just force push your branches not open/close PRs.. it looses the discussion.

jgunthorpe:
Also, we don't install the devel stuff for RPM - so even if we want to do this, it has to be done uniformly for all distros.

jgunthorpe:
I guess this is related to opa-ff? @weiny2  ?

rfehren:
>>>>> "JG" == Jason Gunthorpe <notifications@github.com> writes:

    JG> I'm pretty sure we don't provide an appropriately controlled ABI
    JG> for the plugins, nor do we have the development processes to
    JG> avoid breaking it at this time.. So I don't really want to
    JG> advertise that 3rd parties could make a plugin outside the
    JG> rdma-core tree and then go on to support that..

Well, the reason this came up is, that opa-ff already has such a plugin.
Compiling opa-ff and packaging for Debian needs the headers that are
proposed to go to ibacm-dev.

    JG> Particularly since the plugin mechanism used by acm relies on
    JG> importing symbols from the ibacm executable rather than more
    JG> robust dynamic linking..

    JG> Could you consider contributing your plugin?

You'll have to ask Intel :)

rfehren:
>>>>> "JG" == Jason Gunthorpe <notifications@github.com> writes:

    JG> Also, we don't install the devel stuff for RPM - so even if we
    JG> want to do this, it has to be done uniformly for all distros.

In both spec files shipped with rdma-core, there is a virtual package
ibacm-devel, so in some sense, the ibacm-dev Debian package makes things
more uniform :)

jgunthorpe:
It is a historical virtual pack, pre rdma-core that stuff was packaged, now it isn't...

rfehren:
Hmm, kind of inconsistent: Current Fedora ships the man pages, but not the headers:
https://src.fedoraproject.org/rpms/rdma-core/blob/master/f/rdma-core.spec
...
%{_mandir}/man7/ibacm.*
%{_mandir}/man7/ibacm_prov.*
...
Wouldn't be too hard to fix that and put a real ibacm-devel into the spec files, rather than a 'provides' by ibacm.
Anyway, opa-ff needs the stuff ... Personally, I have no interest in it, other than that I want to get opa-ff packaged properly for Debian.

jgunthorpe:
Yes, we might want to drop those man pages too and put them under documentation or something.

At the end of the day, this is Intel's problem, we don't support external providers for verbs or acm in rdma-core anymore, they have to sort it out, either by stepping up to provide acceptable support for external acm providers or include their acm provider in rdma-core.

rfehren:
>>>>> "JG" == Jason Gunthorpe <notifications@github.com> writes:

    JG> Yes, we might want to drop those man pages too and put them
    JG> under documentation or something.

    JG> At the end of the day, this is Intel's problem, we don't support
    JG> external providers for verbs or acm in rdma-core anymore, they
    JG> have to sort it out, either by stepping up to provide acceptable
    JG> support for external acm providers or include their acm provider
    JG> in rdma-core.

Ok, never mind.

weiny2:
It's been a while since I have worked on this stuff.  Adding John @JFleck

JFleck:
Hey Jason, I've read the thread and am working through the options.

bdrung:
One month passed. Any updates? We should either provide a ibacm-dev package or the library support should be dropped (remove headers from ibacm).

jgunthorpe:
I'm not willing to support a stable ABI for ibacm, so if nobody stands up to do that (and has a really good reason to do so) then we should remove the headers and related to make it very clear this is not supported.

jgunthorpe:
Lets remove the headers please.

PR #351, opened 2018-06-20
NULL MR

yishaih:
This series introduces the NULL MR type and its related verbs.

It is used to force local HCA operations to skip the PCI bus access, while keeping track of the processed length in the ibv_sge handling.

The matching kernel part was sent into rdma-next few days ago.


libibverbs/driver.h
```diff
@@ -204,6 +219,7 @@ struct verbs_context_ops {
 	struct ibv_dm *(*alloc_dm)(struct ibv_context *context,
 				   struct ibv_alloc_dm_attr *attr);
 	struct ibv_mw *(*alloc_mw)(struct ibv_pd *pd, enum ibv_mw_type type);
+	struct ibv_mr *(*alloc_null_mr)(struct ibv_pd *pd);
 	struct ibv_pd *(*alloc_parent_domain)(
```
jgunthorpe: Why is this still wrong?

jgunthorpe: Never mind, wrong file...

jgunthorpe:
Updated with the right 'Update kernel headers'

PR #355, opened 2018-07-08
libqedr: add support for SRQ's

yuvalbason:
Enable SRQ's for rdma-core.

SRQ rdma-core implementation is pretty straight forward, except perhaps the
Way that the driver udpates FW producers: they are updated using a mapped buffer
that FW reads, and not doorbells Like RQ / SQ.

Signed-off-by: Yuval Bason <yuval.bason@cavium.com>
Signed-off-by: Michal Kalderon <michal.kalderon@cavium.com>


jgunthorpe:
You need to fix the build.

providers/qedr/qelr_verbs.c
```diff
@@ -333,6 +508,11 @@ static void qelr_chain_free_rq(struct qelr_qp *qp)
 	qelr_chain_free(&qp->rq.chain);
 }
 
+static inline bool qelr_qp_has_srq(struct qelr_qp *qp)
+{
+	return !!qp->srq;
```
jgunthorpe: !! obfuscation isn't needed if casting to bool

yuvalbason: will fix.

jgunthorpe:
Updated with the final kernel header commit

PR #356, opened 2018-07-09
mlx5 flow steering

yishaih:
This series adds support for mlx5 flow steering to match the kernel series that was sent to rdma-next.

It's done in a way that enables the driver to get its specific device attributes in a raw data to match its underlying specification while still using the generic ibv_flow object for cleanup and code sharing.


jgunthorpe:
Updated with right kernel headers commit

PR #368, opened 2018-08-14
librdmacm: Update message in example tools when no RDMA devices

patrickmacarthur:
Most of the librdmacm demos call rdma_create_event_channel() as their
first call.  Currently, when there are no RDMA devices detected, the
user gets a message like "rdma_create_event_channel: No such device".
This can be confusing as it implies that there is an issue with the
event channel mechanism, rather than an issue with the RDMA device
configuration.

Since we know why this function sets errno to ENODEV, print a specific
error message indicating that no RDMA devices were detected, so that the
user knows where to start troubleshooting.

This fixes every librdmacm example program, except for udpong's server
side, which due to the control flow does not have an obvious call that
will check for the presence of an RDMA device before entering its main
loop.

Signed-off-by: Patrick MacArthur <patrick@patrickmacarthur.net>

### DIFF
--- a/librdmacm/examples/CMakeLists.txt
+++ b/librdmacm/examples/CMakeLists.txt
@@ -7,10 +7,10 @@ rdma_executable(cmtime cmtime.c)
 target_link_libraries(cmtime LINK_PRIVATE rdmacm ${CMAKE_THREAD_LIBS_INIT} rdmacm_tools)
 
 rdma_executable(mckey mckey.c)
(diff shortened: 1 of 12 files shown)


patrickmacarthur:
I think the Travis check timed out; can someone re-run it?

jgunthorpe:
This looks OK to me.. @rleon?

dledford:
It passes my review too.

PR #370, opened 2018-08-15
Add Microsoft Azure RDMA device to MLX4 HCA table

longlimsft:
Microsoft Azure supports Mellanox MLX4 RDMA devices in HPC VM SKUs. The device
is virtualized and exposed over VMBUS, and registered as a MLX4 Infininband
hardware in RDMA layer.

This devcie is on VMBUS, not on PCI tree. Add it to the HCA table so it can
be properly discovered by user-mode driver.


jgunthorpe:
'3daf2e8ca732094bab99bd1f1c86b501' is a stable ID (like MS UUID or something?)

Looks fine if yes..

longlimsft:
Yes, it is a stable ID for device type exposed from Hyper-v host.

PR #373, opened 2018-08-28
Generate and install pkgconfig files for the RDMA libraries

bluca:
It's very useful for applications and other projects to be able to use pkgconfig to build with external libraries. Generate and install pc files for the three RDMA libraries.

Here's how the files look after a Debian build:

```
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libibverbs
Description: RDMA Core Userspace Library
Version: 20.0
Libs: -L${libdir} -libverbs
Libs.private: -lnl-route-3 -lnl-3 -lpthread
Cflags: -I${includedir} -I/usr/include/libnl3
Requires: 
```

```
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: librdmacm
Description: RDMA Core Userspace Library
Version: 20.0
Libs: -L${libdir} -lrdmacm
Libs.private:
Cflags: -I${includedir} 
Requires: 
```

```
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: libibumad
Description: RDMA Core Userspace Library
Version: 20.0
Libs: -L${libdir} -libumad
Libs.private: 
Cflags: -I${includedir} 
Requires: libibverbs
```


bdrung:
These changes look good to me (haven't tested it).

librdmacm/CMakeLists.txt
```diff
@@ -41,3 +41,7 @@ install(TARGETS rspreload DESTINATION "${CMAKE_INSTALL_LIBDIR}/rsocket/")
 # FIXME: Maybe we can get rid of them?
 rdma_install_symlink("librspreload.so" "${CMAKE_INSTALL_LIBDIR}/rsocket/librspreload.so.1")
 rdma_install_symlink("librspreload.so" "${CMAKE_INSTALL_LIBDIR}/rsocket/librspreload.so.1.0.0")
+
+set(PC_LIB_NAME "rdmacm")
+configure_file(${CMAKE_SOURCE_DIR}/buildlib/template.pc.in ${CMAKE_CURRENT_BINARY_DIR}/lib${PC_LIB_NAME}.pc @ONLY)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/lib${PC_LIB_NAME}.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
```
rleon: 1. Three lines below worth to put into separate function in buildlib/rdma_functions.cmake
2. I don't see "set(PC_REQUIRES "libibverbs")" line here, is it ok?

Thanks

debian/control
```diff
@@ -127,6 +127,8 @@ Architecture: linux-any
 Multi-Arch: same
 Depends: ibverbs-providers (= ${binary:Version}),
          libibverbs1 (= ${binary:Version}),
+         libnl-3-dev,
+         libnl-route-3-dev,
```
rleon: rdma-core section has already this dependency,
Isn't this enough? sorry for my dumb question, I'm not debian guru.

Source: rdma-core
Maintainer: Linux RDMA Mailing List <linux-rdma@vger.kernel.org>
Uploaders: Benjamin Drung <benjamin.drung@profitbricks.com>,
           Talat Batheesh <talatb@mellanox.com>
Section: net
Priority: optional
Build-Depends: cmake (>= 2.8.11),
               debhelper (>= 9),
               debhelper (>= 9.20160709) | dh-systemd,
               dpkg-dev (>= 1.17),
               libnl-3-dev,
               libnl-route-3-dev,
               libudev-dev,

librdmacm/CMakeLists.txt
bluca: 1) sure will do that
2) does it require it? I don't see rdmacm linking to it here?

debian/control
bluca: Yes - in that section that's a build dependency for these packages. Here it's a runtime dependency of the dev library, and as the commit msg says it's needed if a third party wants to do static linking with ibverbs, as they'll need to install and static link to everything that ibverbs depends on, which in this case is nl3

bluca:
@rleon all done, thanks

rleon:
One latest change, please fix title from "Genearate" to be "Generate" in this PR and in first commit too.

bluca:
I always make that typo! @rleon fixed

rleon:
Thanks

debian/control
jgunthorpe: No, this is wrong, we don't support static linking in Debian packaging today so don't need the -dev packages here. If we add static libraries someday it would become necessary.

debian/libibverbs-dev.install
```diff
@@ -13,6 +13,7 @@ usr/include/infiniband/verbs_api.h
 usr/lib/*/libibverbs*.so
 usr/lib/*/libmlx4.so
 usr/lib/*/libmlx5.so
+usr/lib/*/pkgconfig/libibverbs.pc
```
jgunthorpe: Where are the pkgconfigs for libmlx5 and 4?

buildlib/rdma_functions.cmake
```diff
@@ -265,3 +265,13 @@ function(rdma_finalize_libs)
...
   endforeach()
 endfunction()
+
+# Generate a pkg-config file
+function(rdma_pkg_config PC_LIB_NAME PC_LIB_PRIVATE PC_INCLUDE_PRIVATE PC_REQUIRES)
+  set(PC_LIB_NAME "${PC_LIB_NAME}")
+  set(PC_LIB_PRIVATE "${PC_LIB_PRIVATE}")
+  set(PC_INCLUDE_PRIVATE "${PC_INCLUDE_PRIVATE}")
+  set(PC_REQUIRES "${PC_REQUIRES}")
+  configure_file(${CMAKE_SOURCE_DIR}/buildlib/template.pc.in ${CMAKE_CURRENT_BINARY_DIR}/lib${PC_LIB_NAME}.pc @ONLY)
```
jgunthorpe: Don't we need a build/lib/pkgconfig/  as well specifying options to 'use in place' ?

CMakeLists.txt
```diff
@@ -368,6 +368,10 @@ if (ENABLE_RESOLVE_NEIGH)
     set(NL_KIND 3)
     set(NL_INCLUDE_DIRS ${NL3_INCLUDE_DIRS})
     set(NL_LIBRARIES ${NL3_LIBRARIES})
+    set(NL_CFLAGS ${NL3_CFLAGS})
+    set(NL_LDFLAGS ${NL3_LDFLAGS})
+    # CMake separates ldflags with a semicolon, cannot be used in a command
+    string(REPLACE ";" " " NL_LDFLAGS "${NL_LDFLAGS}")
```
jgunthorpe: Why is this needed here but not below for NL1 ? Isn't this just list to string conversion?

debian/control
bluca: The static library is built and shipped as you can see here:
https://packages.debian.org/buster/amd64/libibverbs-dev/filelist

CMakeLists.txt
bluca: Because nl1 only adds one flag, so cmake doesn't add a semicolon.

debian/control
jgunthorpe: packages.debian.org is not working, it is showing you a file list from a pre rdma-core package, our packaging does not enable static libraries. download the .deb from buster and check yourself with dpkg-deb -c.

debian/libibverbs-dev.install
jgunthorpe: providers/mlx5 and providers/mlx4, they are intended as application linkable libraries and as provider plugins, eg dpdk links to them.

buildlib/rdma_functions.cmake
jgunthorpe: rdma-core is designed to be built an run 'in-place' when the -DIN_PLACE=1 is used. It sets up rpaths and other things to enable it.

If you want to compile another application against it using pkgconfig I would expect you'd configure that application telling pkgconfig to search rdma-core/build/lib/pkgconfig for the .pc files before searching system directories, and then those .pc files should mirror the IN_PLACE setup and include -I and -L directives pointing to the build directory, not to system paths.

CMakeLists.txt
jgunthorpe: That is relying a bit too much on cmake internal behavior, if the flags are cmake lists then they should always be converted.. Probably the helper function should be doing this conversion to keep with cmake convention of using lists for arguments.

rleon:
Should we revert the PR?

CMakeLists.txt
bluca: Sure, I'll change it that way

debian/control
bluca: Uh, hadn't noticed it was broken, interesting (and annoying). I'll revert this commit then, thanks for spotting it.

buildlib/rdma_functions.cmake
bluca: Ok, will do.

bluca:
https://github.com/linux-rdma/rdma-core/pull/374

PR #376, opened 2018-09-12
mlx5 flow actions

yishaih:
This series exposes the ability to create flow actions which can mutate packet headers via the mlx5 DV API.
Once created the flow action must be attached to a steering rule for it to take effect.

The kernel part was already accepted to 'for-next'.


providers/mlx5/mlx5dv.h
```diff
@@ -244,6 +244,26 @@ struct ibv_flow_action *mlx5dv_create_flow_action_esp(struct ibv_context *ctx,
...
+ * @actions_sz: The size of *actions* buffer in bytes.
+ * @actions: A buffer which contains modify actions provided in device spec
+ *	     format.
+ * @ft_type: Defines the flow table type to which the modify
+ *	     header action will be attached.
+ *
+ * Return a valid ibv_flow_action if successful, NULL otherwise.
+ */
+struct ibv_flow_action *
+mlx5dv_create_flow_action_modify_header(struct ibv_context *ctx,
```
jgunthorpe: We are not getting a man page for this??

yishaih: Just uploaded a man page as part of this patch, see latest.

PR #378, opened 2018-09-17
bnxt_re: Add shared receive queue support to broadcom's roce driver

dsharma283:
Adding support to enable shared receive queue creation,destruction,
query, post-wqes and poll for cqes. The patch is tested for
functionality.

Signed-off-by: Devesh Sharma <devesh.sharma@broadcom.com>

### DIFF
--- a/providers/bnxt_re/verbs.c
+++ b/providers/bnxt_re/verbs.c
@@ -342,39 +342,56 @@ static uint8_t bnxt_re_poll_scqe(struct bnxt_re_qp *qp, struct ibv_wc *ibvwc,
 	return pcqe;
 }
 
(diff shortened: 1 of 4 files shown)


providers/bnxt_re/verbs.c
```diff
@@ -1373,32 +1399,222 @@ int bnxt_re_post_recv(struct ibv_qp *ibvqp, struct ibv_recv_wr *wr,
...
+		free(srq);
+		return NULL;
+	}
+
+	return srq;
+}
+
+static void bnxt_re_srq_free_queue(struct bnxt_re_srq *srq)
+{
+	if (srq->srwrid)
```
jgunthorpe: This if is redundant, free null is OK.

PR #379, opened 2018-09-17
libhns: Add mw support for hip08 in user space

eason-liuyx:
This patchset adds memory window (mw) support for hip08,
including mw alloc, dealloc and bind.

Signed-off-by: Yixian Liu <liuyixian@huawei.com>

### DIFF
--- a/providers/hns/hns_roce_u_hw_v2.c
+++ b/providers/hns/hns_roce_u_hw_v2.c
@@ -45,6 +45,18 @@ static void set_data_seg_v2(struct hns_roce_v2_wqe_data_seg *dseg,
 	dseg->len = htole32(sg->length);
 }
 
+static void set_atomic_seg(struct hns_roce_wqe_atomic_seg *aseg,
+			   struct ibv_send_wr *wr)
+{
+	if (wr->opcode == IBV_WR_ATOMIC_CMP_AND_SWP) {
+		aseg->fetchadd_swap_data = htole64(wr->wr.atomic.swap);
+		aseg->cmp_data  = htole64(wr->wr.atomic.compare_add);
+	} else {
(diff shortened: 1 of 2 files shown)


providers/hns/hns_roce_u_verbs.c
```diff
@@ -175,6 +175,77 @@ int hns_roce_u_dereg_mr(struct verbs_mr *vmr)
...
+	if ((mw->pd != qp->pd) || (mw->pd != bind_info->mr->pd))
+		return EINVAL;
+
+	if (mw->type == IBV_MW_TYPE_2)
+		return EINVAL;
+
+	if (!bind_info->mr && bind_info->length)
+		return EINVAL;
+
+	if (bind_info->mw_access_flags & IBV_ACCESS_ZERO_BASED)
```
jgunthorpe: This should list the flags this code does support, not the flags it doesn't support.

providers/hns/hns_roce_u_verbs.c
```diff
@@ -175,6 +175,77 @@ int hns_roce_u_dereg_mr(struct verbs_mr *vmr)
...
+{
+	int ret;
+
+	ret = ibv_cmd_dealloc_mw(mw);
+	if (ret)
+		return ret;
+
+	free(mw);
+
+	return ret;
```
jgunthorpe: return 0

eason-liuyx: You are right, I will fix it.

providers/hns/hns_roce_u_verbs.c
eason-liuyx: This code is used to do correctness checking, hip08 only support VA, not ZBVA. Should I keep this check or just delete it?

PR #389, opened 2018-09-27
misc fix and add local invalidate support for libhns

oulijun:
This patch series add a small feature and fix a bug for atomic.


jgunthorpe:
This will need to be rebased and forced pushed

jgunthorpe:
Reminder to rebase and force push to clear the travis error

PR #392, opened 2018-10-08
libhns: Add mw support for hip08 in user space

eason-liuyx:
This patchset adds memory window (mw) support for hip08,
including mw alloc, dealloc and bind.

Signed-off-by: Yixian Liu liuyixian@huawei.com
---
v2->v3:
1. according to Jason'comments, give the list of supported, not unsupported.
v1->v2:
1. revise according to Jason'comments, remove unsupported checking, reture 0.

### DIFF
--- a/providers/hns/hns_roce_u.c
+++ b/providers/hns/hns_roce_u.c
@@ -62,10 +62,13 @@ static const struct verbs_match_ent hca_table[] = {
 };
 
 static const struct verbs_context_ops hns_common_ops = {
+	.alloc_mw = hns_roce_u_alloc_mw,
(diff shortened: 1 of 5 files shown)


jgunthorpe:
This should be a single patch, we shouldn't be adding APIs piece wise like this

eason-liuyx:
> This should be a single patch, we shouldn't be adding APIs piece wise like this

Okay, my consideration is it only refers to code movement so I do it.
I will keep the structure of this patch set unchanged and add the new API in another patch.

Thanks
Eason

eason-liuyx:
Hi Jason,

Do you plan to merge this patch set recently?

PR #395, opened 2018-10-09
ibacm: remove endpoint IP address from provider when address deleted

markhaywood:
When addresses are deleted, ibacm needs to call the provider to
let it know that the address has been deleted and allow it to
remove the address from its addr_info[].


markhaywood:
I'd appreciate any clues about the travis_ci failure. I'm not sure what I'm supposed to do about that
since I think it is unrelated to the proposed code change.

markhaywood:
Thanks. The second commit was a futile attempt to resolve the travis failure. I've removed it.

PR #399, opened 2018-10-15
librdmacm: Allow RDMA CM use with connected QP when it is not managing the QP

jswaro:
The RDMA CM documentation indicates it supports being used for communication
management only (I.E. application manages QP). This change enables using the
RDMA CM for connected QP communication management without it managing the QP.
Since the Connection Response event is consumed in the library on the active
side, the application must ensure it closes the race of transitioning the QP
to RTR.

This change only impacts the current error path when QP are not managed
and is consistent with the other unmanaged QP checks done throughout the code.

Signed-off-by: Steve Welch <swelch@systemfabricworks.com>
Signed-off-by: James Swaro <jswaro@cray.com>

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -1023,7 +1023,7 @@ static int ucma_modify_qp_rtr(struct rdma_cm_id *id, uint8_t resp_res)
 	uint8_t link_layer;
 
 	if (!id->qp)
-		return ERR(EINVAL);
+		return 0;
 
 	/* Need to update QP attributes from default values. */
 	qp_attr.qp_state = IBV_QPS_INIT;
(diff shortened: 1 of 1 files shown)


jswaro:
@jgunthorpe 

Hi Jason, 

Steve identified an implementation bug with librdmacm that is inconsistent with the documentation. I'm hoping you are the right person to ask for a review of this change. If not, could you help me with finding the appropriate reviewer for this change?

jgunthorpe:
If you are making this PR you need to add your own signed-off-by as well as Steve's.

I'm not sure who will review rdmacm changes these days, but seemed reasonable to me.

jswaro:
Thanks for the feedback. I'll wait and see who reviews.

PR #400, opened 2018-10-16
ibacm: Fix improper refcnt and address compare issues

Hakon-Bugge:
In the default provider, acmp, the port's sa_dest.refcnt is
initialized to one in acmp_init_dest(). When a port associated with
the device is added, the refcnt is set once again to one in
acmp_port_up().

For a system with VFs enabled, the ports for both the PF and the VFs
are added. Here an extract from the log file on a system with a single
VF enabled:

    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 is up
    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 is up

Now, when/if this ports goes down, acmp_port_down() will be called. It
decrements the refcnt and waits endlessly until it becomes zero. For
the first call to acmp_port_down(), it becomes zero, but the
subsequent call will loop forever in said busy-waiting loop, due to
the refcnt becoming minus one.

Fixed by changing the atomic_set() to atomic_inc() in acmp_port_up()
and test for one instead of the while loop in acmp_port_down().

Also added printing of instance number to ease debugging.

With the fix, the extract from the log file yields:

    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 1 is up
    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 2 is up
    acmp_port_down: mlx4_0 1
    acmp_port_down: mlx4_0 1 2 is down
    acmp_port_down: mlx4_0 1
    acmp_port_down: mlx4_0 1 1 is down

### DIFF
--- a/ibacm/prov/acmp/src/acmp.c
+++ b/ibacm/prov/acmp/src/acmp.c
@@ -188,7 +188,8 @@ struct acmp_ep {
 	struct list_head      active_queue;
 	struct list_head      wait_queue;
 	enum acmp_state       state;
-	struct acmp_addr      addr_info[MAX_EP_ADDR];
+	int		      nmbr_eps;
+	struct acmp_addr      *addr_info;
 	atomic_t              counters[ACM_MAX_COUNTER];
 };
 
@@ -1044,7 +1045,7 @@ acmp_addr_lookup(struct acmp_ep *ep, uint8_t *addr, uint16_t type)
 {
 	int i;
 
-	for (i = 0; i < MAX_EP_ADDR; i++) {
+	for (i = 0; i < ep->nmbr_eps; i++) {
 		if (ep->addr_info[i].type != type)
 			continue;
 
@@ -2366,13 +2367,20 @@ static int acmp_add_addr(const struct acm_address *addr, void *ep_context,
 
 	acm_log(2, "\n");
 
-	for (i = 0; (i < MAX_EP_ADDR) &&
+	for (i = 0; (i < ep->nmbr_eps) &&
 	     (ep->addr_info[i].type != ACM_ADDRESS_INVALID); i++)
 		;
 
-	if (i == MAX_EP_ADDR) {
-		acm_log(0, "ERROR - no more space for local address\n");
-		return -1;
+	if (i == ep->nmbr_eps) {
+		++ep->nmbr_eps;
+		ep->addr_info = realloc(ep->addr_info, ep->nmbr_eps * sizeof(*ep->addr_info));
+		if (!ep->addr_info) {
+			--ep->nmbr_eps;
+			acm_log(0, "ERROR - no more space for local address\n");
+			return -1;
+		}
+		/* Added memory is not initialized */
+		memset(ep->addr_info + i, 0, sizeof(*ep->addr_info));
 	}
 	ep->addr_info[i].type = addr->type;
 	memcpy(&ep->addr_info[i].info, &addr->info, sizeof(addr->info));
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -72,7 +72,6 @@
 #define src_index   data[1]
 #define dst_index   data[2]
 
-#define MAX_EP_ADDR 4
 #define NL_MSG_BUF_SIZE 4096
 #define ACM_PROV_NAME_SIZE 64
 #define NL_CLIENT_INDEX 0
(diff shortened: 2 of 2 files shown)


ibacm/src/acm.c
```diff
@@ -2016,7 +2016,7 @@ acm_ep_insert_addr(struct acmc_ep *ep, const char *name, uint8_t *addr,
 		return EINVAL;
 
 	memset(tmp, 0, sizeof tmp);
-	memcpy(tmp, addr, addr_len);
+	memcpy(tmp, addr, acm_addr_len(addr_type));
```
jgunthorpe: It is really strange to pass in addr_len and not use it. Are you sure the bug isn't that the caller is using the wrong addr_len for the address type?

Otherwise I suggest deleting addr_len as a parameter as well

ibacm/prov/acmp/src/acmp.c
```diff
@@ -2643,7 +2644,7 @@ static void acmp_port_up(struct acmp_port *port)
 	acmp_set_dest_addr(&port->sa_dest, ACM_ADDRESS_LID,
 			   (uint8_t *) &sm_lid, sizeof(sm_lid));
 
-	atomic_set(&port->sa_dest.refcnt, 1);
+	instance = atomic_inc(&port->sa_dest.refcnt) - 1;
```
jgunthorpe: Why are systems with a VF adding the PF multiple times, that seems strange, maybe explain that in the commit message.

I don't know this code at all, so I'm not sure who will review this patch

@weiny2 do you care?

Hakon-Bugge: From the commit message: "For a system with VFs enabled, the ports for both the PF and the VFs
are added."

ibacm/src/acm.c
Hakon-Bugge: Yes, I agree. An oversight from me. Just pushed a v2 where the addr_len is re-factored out.

Hakon-Bugge: I do see a s/form/from/ fix. You want me to fix it?

ibacm/prov/acmp/src/acmp.c
jgunthorpe: That doesn't explain why the VF and PF ports are intermingled in this algorithm..

Hakon-Bugge: Well, the design of this algorithm is somewhat outside the scope of this bug fix. We discussed it on the mailing list, see https://www.spinics.net/lists/linux-rdma/msg69934.html

Bottom line, the endpoints are designated by the **Node** GUID, hence ports from both VFs and PFs are associated with the same struct _acmp_port_.

Hakon-Bugge: Just sent a new version with more elaborate commit message on this point.

Hakon-Bugge:
I am a novice when it comes to git and a virgin (excuse the phrase) when it comes to github. I have done the changes Jason requested (if I understood him correctly), but you have to click on the right SHA to see it.
Please let me know if I am on the critical line here.

jgunthorpe:
I think we are just waiting for a review...

jgunthorpe:
Oh, why did you close it? Confused.

Hakon-Bugge:
Oh, sorry, pushed the wrong button.

markhaywood:
It looks like there are travis_ci issues (lines greater than 75 characters) that need to be resolved.
Otherwise, your changes look good to me.

Hakon-Bugge:
I see the travis error is:

> [48/48] Header C++ check for /tmp/tmpb...2l/usr/local/include/rdma/rdma_verbs.h[48/48] Header C++ check for /tmp/tmpbkO62l/usr/local/include/rdma/rsocket.h
  File "../buildlib/check-build", line 506, in <module>
    v(args);
  File "../buildlib/check-build", line 146, in test_verbs_private
    check_verbs_abi(args,fn);
  File "../buildlib/check-build", line 131, in check_verbs_abi
    raise ValueError("Provider library has unknown file name format %r"%(fn));
ValueError: Provider library has unknown file name format 'libhns-rdmav21.a'
The command "buildlib/travis-build" exited with 1.

Not related to the commit AFAICT.

jgunthorpe:
Yes, that travis thing is my mistake

markhaywood:
Does the travis problem to be resolved so that this pull request might proceed?

Hakon-Bugge:
@jgunthorpe Still a little confused. You wrote above, that only a review was missing. Got one from Mark. Same company, yes, but leaning towards what @rleon said on the rdma list.

PR #405, opened 2018-10-23
Build static libraries for rdma-core

jgunthorpe:
Restore static library support. Before rdma-core we did produce some simplistic static libraries, and this was lost when things became more complicated.

This version is a complete solution and provides a simple way for applications to use the static libraries, including static providers, without having to resort to strange linker options.


jgunthorpe:
Updated
 - Don't build static libraries on suse, per policy [Nicholas]]

jgunthorpe:
updated with nicholas's version of the rpm patch to make it use a define

PR #412, opened 2018-11-01
ibacm: Add comments explaining mc join bits

Hakon-Bugge:

### DIFF
--- a/ibacm/prov/acmp/src/acmp.c
+++ b/ibacm/prov/acmp/src/acmp.c
@@ -1380,15 +1380,15 @@ static void acmp_init_join(struct ib_sa_mad *mad, union ibv_gid *port_gid,
 		IB_COMP_MASK_MC_SCOPE | IB_COMP_MASK_MC_JOIN_STATE;
 
 	mc_rec = (struct ib_mc_member_rec *) mad->data;
-	acmp_format_mgid(&mc_rec->mgid, pkey | 0x8000, tos, rate, mtu);
+	acmp_format_mgid(&mc_rec->mgid, pkey | 0x8000, tos, rate, mtu); /* Full partition mmbr (0x8000) */
 	mc_rec->port_gid = *port_gid;
 	mc_rec->qkey = htobe32(ACM_QKEY);
-	mc_rec->mtu = 0x80 | mtu;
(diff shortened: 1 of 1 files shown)


ibacm/prov/acmp/src/acmp.c
```diff
@@ -1380,15 +1380,15 @@ static void acmp_init_join(struct ib_sa_mad *mad, union ibv_gid *port_gid,
 		IB_COMP_MASK_MC_SCOPE | IB_COMP_MASK_MC_JOIN_STATE;
 
 	mc_rec = (struct ib_mc_member_rec *) mad->data;
-	acmp_format_mgid(&mc_rec->mgid, pkey | 0x8000, tos, rate, mtu);
+	acmp_format_mgid(&mc_rec->mgid, pkey | 0x8000, tos, rate, mtu); /* Full partition mmbr (0x8000) */
```
weiny2: Would it be worth adding #defines for values like this?  I'm ok with the patch as I'm familiar with the meanings of these bits.  But if it was troublesome enough for you to want to add a comment here maybe we should have something like:

#define RDMA_PKEY_FULL_MEMBER 0x8000
#define RDMA_MTU_EXACT 0x80
#define RDMA_RATE_EXACT 0x80

It's been a while so I apologize if they have been defined somewhere like in the umad headers.

Hakon-Bugge: The most important, agnostic to the use of defines or comments, is to use terms that can be found in the relevant collaterals, i.e. IBTA spec in this setting.
I used some hours this morning to try to understand the code, and said to myself, if have have done this, just let me document it for others as well.

jgunthorpe: I think Ira is right, this should be done by adding new #defines, not by sprinkling more magic constant values and comments  around

Hakon-Bugge: @jgunthorpe , @weiny2 I'll work on this after I have gotten the functional commits in.

Hakon-Bugge: I am deleting this PR, replaced by PR #427

Hakon-Bugge:
I am deleting this PR, replaced by PR #427

PR #413, opened 2018-11-06
ibacm: Unable to assign EP name for limited pkey

Hakon-Bugge:
Using limited IB pkeys, ibacm compares pkeys retrieved using the OF-UV
API, ibv_query_pkey(), with the sysfs interface of ipoib,
/sys/class/net/<some_ipoib_if>/pkey.

The problem is that the latter always has the full member bit set
(0x8000). But the partition could be limited.

Using a limited partition, we see the following error in the ibacm.log
file (using the limited partition 0x0123):

acm_find_ep: pkey 0x123
acm_ep_up: creating endpoint for pkey 0x123
acm_ep_up: ERROR - unable to assign EP name for pkey 0x123

This is fixed by ORing in the full membership bit when comparing to a
pkey retrieved from the ipoib sysfs. With the fix, the above log lines
will read:

cm_find_ep: pkey 0x123
acm_ep_up: creating endpoint for pkey 0x123
acm_assign_ep_names: device mlx4_0, port 2, pkey 0x123
acmp_get_ep: dev 0x21280001a17c96 port 2 pkey 0x123
acmp_open_endpoint: creating endpoint for pkey 0x123
acmp_get_port: dev 0x21280001a17c96 port 2 pkey 0x123
acmp_ep_join: mlx4_0-2-0x123
acmp_join_group: mlx4_0 2 pkey 0x123, sl 0x0, rate 0x3, mtu 0x4
acmp_ep_join: join for mlx4_0-2-0x123 complete
acm_ep_ip_iter_cb: Added 192.168.213.213 mlx4_0 2 0x8123 from test_if

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -2071,7 +2071,9 @@ static void acm_ep_ip_iter_cb(char *ifname, union ibv_gid *gid, uint16_t pkey,
 
 	dev = acm_get_device_from_gid(gid, &port_num);
 	if (dev && ep->port->dev == dev
-	    && ep->port->port.port_num == port_num && ep->endpoint.pkey == pkey) {
+	    && ep->port->port.port_num == port_num &&
+		/* pkey retrieved from ipoib has always full mmbr bit set */
+		(ep->endpoint.pkey | IB_PKEY_FULL_MEMBER) == pkey) {
 		if (!acm_ep_insert_addr(ep, ip_str, addr, addr_len, addr_type)) {
 			acm_log(0, "Added %s %s %d 0x%x from %s\n", ip_str,
 				dev->device.verbs->device->name, port_num, pkey,


Hakon-Bugge:
Travis error seems unrelated to the commit:

[356/356] Linking C executable bin/srp_daemon[356/356] Linking C executable bin/srp_daemon
+../buildlib/check-build --src .. --cc clang-5.0
[86/86] Header check for /home/travis/...ild-travis/include/util/udma_barrier.h[86/86] Header check for /home/travis/...-core/build-travis/include/util/util.h
ABI check skipped, no ABI/ directory.
[48/48] Header C++ check for /tmp/tmpF...hD/usr/local/include/rdma/rdma_verbs.h[48/48] Header C++ check for /tmp/tmpFN49hD/usr/local/include/rdma/rsocket.h
  File "../buildlib/check-build", line 506, in <module>
    v(args);
  File "../buildlib/check-build", line 146, in test_verbs_private
    check_verbs_abi(args,fn);
  File "../buildlib/check-build", line 131, in check_verbs_abi
    raise ValueError("Provider library has unknown file name format %r"%(fn));
ValueError: Provider library has unknown file name format 'libhns-rdmav21.a'

ibacm/src/acm.c
```diff
@@ -2070,7 +2070,9 @@ static void acm_ep_ip_iter_cb(char *ifname, union ibv_gid *gid, uint16_t pkey,
 
 	dev = acm_get_device_from_gid(gid, &port_num);
 	if (dev && ep->port->dev == dev
-	    && ep->port->port.port_num == port_num && ep->endpoint.pkey == pkey) {
+	    && ep->port->port.port_num == port_num &&
+		/* pkey retrieved from ipoib has always full mmbr bit set */
+		(ep->endpoint.pkey | 0x8000) == pkey) {
```
weiny2: I'm a bit worried about this.  Unless the user is using AF_IB I don't believe the Pkey will be set correctly to use the AF_INET/INET6 address assigned to the IPoIB interface.  I think the error should stand.  OR a Pkey table search should be used in the case of an EP being addressed with AF_IB.

Hakon-Bugge: If you echo a limited pkey into /sys/class/net/ipoib/create_child, the new device's /sys/class/net/ib0.0123/pkey will be the full partition. Isn't that correct?

weiny2: Yes but IPoIB will not be able to use that partition if the membership of the port is limited for that partition.

Hakon-Bugge: @weiny2 Not quite right. If the peer is a full member, communication flows smoothly. This is actually how orcl multitenant virtualized rdbms works, which is the use-case requiring this bug-fix.

weiny2: How can you have more than 1 limited member port in a subnet?  Mcast traffic would fail between any 2 limited members.  I guess if there is only 1 limited member you are ok but that seems to be a pretty narrow use case.

Hakon-Bugge: @weiny2 You are right that limited<->limited cannot communicate, and that is why we use limited partitions :-) VMs from different tenants shall not be able to communicate, and they do not. But they need I/O, so the array of I/O servers are full members of this partition. Trust me, orcl has actively been using partition in the IB fabric for a while.

Hakon-Bugge: No, the IPoIB subnet shall **not** be able to ping between node A and B (since they are different tenants in my case). That is the virtue of limited partitions. But A and B can communicate with X. And that is my use case.

There is no Spoofing. The fact is that if IPoIB runs on a limited partition, its sysfs returns the full member, which at least in this context, is wrong. One could argue to fix IPoIB in this context, I can agree to that.

Try for yourself, make a limited partition and make ibacm work with it. It doesn't!

Are you saying that a large number of Exadata employments using VMs and partitioning is deemed not to work if you toss ibacm into the equation? Well, they do not right now, but this commit fixes that, and I do not see what harm it causes.

As stated, the IPoIB sysfs returns always the full membership partition. If the actual membership is full, ORing in 0x8000 is a NOP. If the actual membership is limited, the ORing makes a match for the endpoint, and makes ibacm useful.

Hakon-Bugge: @weiny2 Did what I wrote above make any sense to you?

weiny2: My apologies.  I see that the comment I meant to submit a few days ago was marked "Pending".  I believe it is submitted now.

Hakon-Bugge:
Pushed a variant using the define from PR #427 

A yea or nay from @jgunthorpe @dledford @rleon appreciated.

PR #415, opened 2018-11-07
Python binding

rleon:
In this patch set we introduce python bindings (pyverbs) for
rdma-core in its initial form. It will cover libibverbs and will be
expanded to support librdmacm in the future.
    
Prior to developing pyverbs, the existing python-rdma package [1] was
evaluated, but a new package was built using newer Cython and Python
versions, in order to provide a more comfortable user experience.
    
The main goal of pyverbs is to provide the user a simple API that
makes working with verbs easier. Among other things, it allows
setting up RDMA resources quickly.
    
In the future, it will also suggest default configuration (e.g.
default values for QP attributes, which the user can override) and
further bring up shortcuts will be provided (e.g. QPs can change state
internally, without user intervention).
The above means a significant cut in tests development time, which
will allow developers to send a feature with a test included,
increasing rdma-core's robustness.

A short example based on the preliminary abilities of this PR:
    
    >>>import pyverbs.device as d
    >>>ctx = d.Context(name='mlx5_0')
    >>>attr = ctx.query_device()
    print(attr)
    FW version            : 16.24.0185
    Node guid             : 9803:9b03:0000:e4c6
    Sys image GUID        : 9803:9b03:0000:e4c6
    Max MR size           : 0xffffffffffffffff
    [...]
    
Known limitations:
This PR contains an updated rdma-core.spec files, but it is not final.
   
Future plan:
Extend pyverbs' coverage over rdma-core (further RDMA objects such as PD, MR etc.).
Enrich examles section.
Provide tests for existing and new features.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,8 @@ Build-Depends: cmake (>= 2.8.11),
                ninja-build,
                pandoc,
                pkg-config,
-               python,
+               python3-dev,
+               cython3,
                valgrind [amd64 arm64 armhf i386 mips mips64el mipsel powerpc ppc64 ppc64el s390x]
 Standards-Version: 4.2.1
 Vcs-Git: https://github.com/linux-rdma/rdma-core.git
@@ -324,3 +325,10 @@ Description: Tools for Infiniband attached storage (SRP)
  In conjunction with the kernel ib_srp driver, srptools allows you to
  discover and use Infiniband attached storage devices which use the
  SCSI RDMA Protocol (SRP).
+
+Package: python3-pyverbs
+Architecture: linux-any
+Depends: rdma-core (>= 21),
+         ${python3:Depends}
+Description: Pyverbs is a python bindings for rdma-core.
+ It allows an easy development in python language of RDMA applications.
--- a/debian/rules
+++ b/debian/rules
@@ -7,7 +7,7 @@ export DEB_BUILD_MAINT_OPTIONS=hardening=+all
 COHERENT_DMA_ARCHS = amd64 arm64 i386 ia64 powerpc powerpcspe ppc64 ppc64el s390x sparc64 x32
 
 %:
-	dh $@ --with systemd --builddirectory=build-deb
+	dh $@ --with python3,systemd --builddirectory=build-deb
 
 override_dh_auto_clean:
 	dh_auto_clean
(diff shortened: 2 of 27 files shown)


suse/rdma-core.spec
```diff
@@ -325,6 +330,14 @@ rdma-ndd is a system daemon which watches for rdma device changes and/or
 hostname changes and updates the Node Description of the rdma devices based
 on those changes.
 
+%package -n python3-pyverbs
+Summary: Python3 API over IB verbs
```
nmorey: Minor incoherency in indent here. Nothing blocking though

rleon: fixed

debian/control
```diff
@@ -324,3 +325,11 @@ Description: Tools for Infiniband attached storage (SRP)
  In conjunction with the kernel ib_srp driver, srptools allows you to
  discover and use Infiniband attached storage devices which use the
  SCSI RDMA Protocol (SRP).
+
+Package: python3-pyverbs
+Architecture: linux-any
+Depends: rdma-core (>= 21),
+         python3
```
bdrung: Use `${python3:Depends}` instead.

debian/control
```diff
@@ -324,3 +325,11 @@ Description: Tools for Infiniband attached storage (SRP)
...
+Description: Pyverbs is a Python API over IB verbs. It allows an easy
```
bdrung: The description consists of two parts:
1. The short description one the first line
2. The long description on the second line and following

It looks like you put the beginning of the long description in the first line.

debian/rules
```diff
@@ -34,6 +34,8 @@ override_dh_auto_configure:
 			-DCMAKE_INSTALL_RUNDIR:PATH=/run \
 			-DCMAKE_INSTALL_UDEV_RULESDIR:PATH=/lib/udev/rules.d \
 			-DENABLE_STATIC=1 \
+			-DPYTHON_EXECUTABLE:PATH=/usr/bin/python3 \
+			-DCMAKE_INSTALL_PYTHON_ARCH_LIB:PATH=/usr/lib/python3/dist-packages \
```
bdrung: Please also change the `dh` line to:
```
dh $@ --with python3,systemd --builddirectory=build-deb
```

debian/control
```diff
@@ -15,7 +15,8 @@ Build-Depends: cmake (>= 2.8.11),
                ninja-build,
                pandoc,
                pkg-config,
-               python,
+               python3-dev,
+               cython3,
```
bdrung: Please also depend on dh-python and sort these entries.

rleon:
Thanks, I'll fix and repush.

debian/control
rleon: done

debian/control
rleon: done

debian/control
rleon: done

debian/rules
rleon: done

rleon:
* Fixed enum redefinition spotted by gcc in FC29
* Added packaging tests for ubuntu OS

rleon:
Now with working travis and cbuild

rleon:
Rewrote debian patch to correctly work on debian-8 (without pyverbs) and other deb distros.
There is one leftover, centos6 builds but doesn't pass packaging.

rleon:
Status:
tumbleweed - pass 
ubuntu-14.04 - pass                                                                                    3 opensuse-15.0 - pass                                                                                   ubuntu-18.04 - pass                                                                                    5 debian-experimental - pass                                                                             centos7 - pass                                                                                       
centos6 - FAIL, looking on the wrong spec file.                                                      
debian-8 - pass                                                                                      
debian-9 - pass                                                                                      
ubuntu-16.04 - FAIL with error relocation R_X86_64_PC32                                                                             
   against undefined symbol `_Py_NoneStruct' can not be used when making a shared object;              13   recompile with -fPIC                                                                               
travis - pass                                                                                        
fc29 - pass

rleon:
Talked with Jason, need to close centos6 for packaging

PR #418, opened 2018-11-08
Use NAME_MAX and PATH_MAX for file name buffers

wscullin:
Replace 128 / 256 char file and path names with NAME_MAX and PATH_MAX

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -67,6 +67,7 @@
 #include <util/util.h>
 #include "acm_mad.h"
 #include "acm_util.h"
+#include <linux/limits.h>
 
 #define src_out     data[0]
 #define src_index   data[1]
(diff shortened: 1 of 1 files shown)


ibacm/src/acm.c
```diff
@@ -216,14 +217,14 @@ static struct sa_data {
 static const char *acme = IBACM_BIN_PATH "/ib_acme -A";
 static const char *opts_file = ACM_CONF_DIR "/" ACM_OPTS_FILE;
 static const char *addr_file = ACM_CONF_DIR "/" ACM_ADDR_FILE;
-static char log_file[128] = IBACM_LOG_FILE;
+static char log_file[NAME_MAX] = IBACM_LOG_FILE;
```
Hakon-Bugge: The log_file is actually a path, you must use PATH_MAX here (e.g. see ibacm_opts.cfg file).

ibacm/src/acm.c
```diff
@@ -216,14 +217,14 @@ static struct sa_data {
...
 static int log_level = 0;
-static char lock_file[128] = IBACM_PID_FILE;
+static char lock_file[NAME_MAX] = IBACM_PID_FILE;
```
Hakon-Bugge: Ditto

Hakon-Bugge: .. and in acm_set_options() please use strncpy() instead of strcpy when copying into these three variables.

PR #429, opened 2018-11-21
libibumad: Restore umad_sa_set_rate_mtu_or_life to original implementation ( using defines)

hnrose:
umad_sa_set_rate_mtu_or_life was changed to mask selector
with 0x3f (UMAD_SA_RATE_MTU_PKT_LIFE_MASK) rather than 0x3.
While this should work due to using left shift by 6 on 8 bit variable,
a more literal translation is better.
    
Fixes: 0ff0e21b89e2 ("ibacm: Use helper functions and existing defines to avoid literal use")

### DIFF
--- a/libibumad/umad_sa.h
+++ b/libibumad/umad_sa.h
@@ -138,6 +138,7 @@ enum {
 
 #define UMAD_SA_SELECTOR_SHIFT		6
 #define UMAD_SA_RATE_MTU_PKT_LIFE_MASK	0x3f
+#define UMAD_SA_SELECTOR_MASK		0x3
 
 /*
  *  sm_key is not aligned on an 8-byte boundary, so is defined as a byte array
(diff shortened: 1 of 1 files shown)


hnrose:
On Wed, Nov 21, 2018 at 1:19 PM Leon Romanovsky <notifications@github.com>
wrote:

> *@rleon* requested changes on this pull request.
>
> Hal,
> Can we do a little bit more formal commit message? Without "IMO", proper
> title and less hardcore description as it is now :)
>
I updated commit message. What more would you propose ? Not sure what you
mean by less hardcore description. Are you saying to remove the verbiage as
to how umad_sa_set_rate_mtu_or_life was changed by commit 0ff0e21b89e2 ?

rleon:
@hnrose, yes, let's remove the code from the commit message.

hnrose:
Done

jgunthorpe:
Fixes line should be at the end, directly before signed-off-by, and the title can omit the .h file reference

hnrose:
Done

rleon:
Thanks, i pushed it manually

PR #431, opened 2018-11-22
ibacm: Fix improper refcnt

Hakon-Bugge:
In the default provider, acmp, the port's sa_dest.refcnt is
initialized to one in acmp_init_dest(). When a port associated with
the device is added, the refcnt is set once again to one in
acmp_port_up().

For a system with VFs enabled, the ports for both the PF and the VFs
are added. This because ibacm uses the Node GUID, Port, and pkey
triple to designate an endpoint. Since the Node GUID is the very same
for the PF and VFs, we will get several port up events.

    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 is up
    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 is up

Now, when/if this ports goes down, acmp_port_down() will be called. It
decrements the refcnt and waits endlessly until it becomes zero. For
the first call to acmp_port_down(), it becomes zero, but the
subsequent call will loop forever in said busy-waiting loop, due to
the refcnt becoming minus one.

Fixed by changing the atomic_set() to atomic_inc() in acmp_port_up()
and test for one instead of the while loop in acmp_port_down().

Also added printing of instance number to ease debugging.

With the fix, the extract from the log file yields:

    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 1 is up
    acmp_port_up: mlx4_0 1
    acmp_port_up: mlx4_0 1 2 is up
    acmp_port_down: mlx4_0 1
    acmp_port_down: mlx4_0 1 2 is down
    acmp_port_down: mlx4_0 1
    acmp_port_down: mlx4_0 1 1 is down

Reviewed-by: Mark Haywood <mark.haywood@oracle.com>

### DIFF
--- a/ibacm/prov/acmp/src/acmp.c
+++ b/ibacm/prov/acmp/src/acmp.c
@@ -2658,6 +2658,7 @@ static void acmp_port_up(struct acmp_port *port)
 	__be16 pkey_be;
 	__be16 sm_lid;
 	int i, ret;
+	int instance;
 
 	acm_log(1, "%s %d\n", port->dev->verbs->device->name, port->port_num);
 	ret = ibv_query_port(port->dev->verbs, port->port_num, &attr);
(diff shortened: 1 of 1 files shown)


Hakon-Bugge:
Can this be merged please?

PR #436, opened 2018-12-02
mlx5: Enrich DEVX support

yishaih:
This series enriches DEVX support in few aspects: 
- it enables the DEVX counter object to be used by mlx5dv_create_flow API.
- It enables interoperability between DEVX and verbs by introducing the required DV APIs.


dledford:
Then the easy way to do it would be to run the script against the commit I referenced, commit that, then run it again against the commit that you originally targeted.  My complaint is that doing it the way it was done in this PR picks up extraneous, unrelated changes with no explanation of why they are there.  For something as important as the kabi, I think that's a bad idea.

PR #437, opened 2018-12-04
Mlx5 devx

dledford:
This is how I would prefer the kABI updates be done.  When I reviewed the original patchset, it was clear that the reserved[] changes were not part of this patchset or relevant.  Having a separate commit clears things up.

@yishaih @jgunthorpe If you guys are OK with this, I'll commit it instead.


jgunthorpe:
I think this is going to be very hard to explain to people how to construct this, I really prefer the simpler single patch sync approach. It is already such a pain to the github work flow to manage these updates.

All I'm checking for the 'Update kernel headers' is that they match the kernel at the noted commit ID, I don't want to review the patch. I already looked at it for the kernel side.

People are already doing this wrong. They are sending patches that were not made by the script, that don't actually match the kernel commit ID. ie they manually took just the diff hunks they though were relevant, or they grabbed diffs from an old version of the patchset and just revised the commit message..

Lets compromise, if you want an additional patch for aesthetic reasons then push one as its own push (don't need a PR) and leave the feature PRs as a single update at the start. The feature PR will get cleaned up when it is rebased, force pushed and linked to the final kernel commit. Maintainer discretion if we add the extra commit or not, and we don't bother the submitters with this complexity.

dledford:
I'm thinking maybe a git pre-hook to reject changes to the kernel-headers directory that aren't initiated by the script, with a reject message giving a simple example of how to use the script?

Personally, I would like to start doing a stand alone update to rdma-core each time we update the kernel headers.

In any case, there is no need to rebase this series.  In order to split the kABI update into two commits, I did a rebase -i on HEAD, dropped the first commit, made two new ones, and the final two were simply rebased.  They already reference the final commits IDs because the needed commit was in for-next already having been committed by you previously.  So there's no need to rebase again, it's all final and can be merged right now.

rleon:
"Regular updates" - yes and no :)
If we continue our policy that every change to kernel API should be correlated to change in rdma-core, those "updates" will be done naturally.

yishaih:
Doug, personally I prefer to continue with one commit for kernel headers per series as both Jason and Leon pointed to keep things simple.

As here you already did the split and as you pointed it's binary equal to my original series I'm fine with merging your code, I'll merge it and close my PR.

PR #459, opened 2019-01-27
mlx5:  Async DEVX commands

yishaih:
This series exposes DEVX APIs to run command asynchronously without blocking and once the response is ready read it.

Specifically, the mlx5dv_devx_obj_query_async() API was introduced to let applications running many query commands which might take time in parallel.

The matching kernel series was already sent into rdma-next.


yishaih:
The kernel headers patch was updated to match the kernel commit ID that was accepted.

PR #465, opened 2019-02-03
ibacm: Compare logical partitions instead of pkeys

Hakon-Bugge:
Using ibacm in an environment with limited pkeys, it is unable to
obtain a Path Record (PR) in some situations.

The use of limited pkey implies communication over the logical
partition where one peer is a full member and the other is limited.

When comparing pkey values, the above difference will yield a
mismatch, even if the logical partitions of the two pkeys compared are
the very same.

Fixed by introducing a helper function which compares the logical
partitions.

Querying a PR on the peer which uses the full partition, we see
without this commit:

ib_acme -s 192.168.200.203 -d  192.168.200.200 -v
Service: /usr/local/var/run/ibacm-unix.sock
Destination: 192.168.200.200
Source: 192.168.200.203
Path information
  dgid: fe80::21:2800:1a1:7c97
  sgid: fe80::21:2800:1ce:ac18
  dlid: 243
  slid: 246
  flow label: 0x0
  hop limit: 0
  tclass: 0
  reversible: 1
  pkey: 0x2b4
  sl: 0
  mtu: 4
  rate: 7
  packet lifetime: 16
SA verification: failed Cannot assign requested address

The verbose ibacm log contains:

acm_get_ep_address: SLID(246) DLID(243)
acm_get_ep_address: notice - could not find SLID(246) DLID(243)
acm_svr_query_path: notice - could not find local end point address

With this commit, the query with verification has success and we see
in the log:

acm_get_ep_address: SLID(246) DLID(243)
acm_same_partition: pkey_a: 0x02b4 pkey_b: 0x82b4
[]
acm_query_response: status 0x0

Another example is a kernel client that requests PRs over
netlink. Without this commit, the statistics counters show:

ib_acme -P col
/usr/local/var/run/ibacm-unix.sock
Error Count : 234
Resolve Count : 234
[]

With this commit, we see:

ib_acme -P col
/usr/local/var/run/ibacm-unix.sock
Error Count : 0
Resolve Count : 228
No Data : 0
Addr Query Count : 0
Addr Cache Count : 0
Route Query Count : 1
Route Cache Count : 226

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -820,6 +820,14 @@ acm_is_path_from_port(struct acmc_port *port, struct ibv_path_record *path)
 	return 0;
 }
 
+static bool acm_same_partition(uint16_t pkey_a, uint16_t pkey_b)
+{
+
+	acm_log(2, "pkey_a: 0x%04x pkey_b: 0x%04x\n", pkey_a, pkey_b);
+
+	return ((pkey_a | IB_PKEY_FULL_MEMBER) == (pkey_b | IB_PKEY_FULL_MEMBER));
+}
+
 static struct acmc_addr *
 acm_get_port_ep_address(struct acmc_port *port, struct acm_ep_addr_data *data)
 {
@@ -837,7 +845,7 @@ acm_get_port_ep_address(struct acmc_port *port, struct acm_ep_addr_data *data)
 	list_for_each(&port->ep_list, ep, entry) {
 		if ((data->type == ACM_EP_INFO_PATH) &&
 		    (!data->info.path.pkey ||
-		     (be16toh(data->info.path.pkey) == ep->endpoint.pkey))) {
+		     acm_same_partition(be16toh(data->info.path.pkey), ep->endpoint.pkey))) {
 			for (i = 0; i < MAX_EP_ADDR; i++) {
 				if (ep->addr_info[i].addr.type)
 					return &ep->addr_info[i];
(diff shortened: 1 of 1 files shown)


Hakon-Bugge:
v1 -> v2:
       * Fixed Travis error, incorrect placed opening bracked

Hakon-Bugge:
@hnrose wrote on the ML:

_I'm concerned about how open the door is in ACM for limited <-> limited
communication. Maybe that's just up to the path resolution or cache
preload doing the right thing (meaning if both endpoints are limited
only, no PR should be available)._

@Hakon-Bugge responded:

_I am concerned also, but this commit doesn't alter anything. Without having tested this, I guess that if you set the loopback_prot to "none" and do keep the default route_prot value ("sa"), then ACM will always ask the SA first for a PR, and what you describe above will not happen._

and @Hakon-Bugge responded again:

> I am not concerned anymore. First, the behaviour of ACM is not changed with this patch. Secondly, with route_prot set to "sa", ACM will query the SA for the PR, and for two limited ports, this PR request will fail. Thirdly, if route_prot set to "acm", ACM will post an MC send, but the MC packet will have the limited pkey value in its BTH, and as such, not be received by the ACM peer. And since it is not received, the active side will not receive any response. And then no communication will be attempted between the two limited pkey ports.

This commit is needed in order to support ACM where a limited peer needs to communicate with the corresponding full peer.

jgunthorpe:
@Hakon-Bugge There has been so much round and round.. Please confirm this is the patch you want merged. Thanks

Hakon-Bugge:
@jgunthorpe Affirmative. Please merge.

rleon:
Thanks

PR #466, opened 2019-02-03
Ibacm fix acme ep port display v2

Hakon-Bugge:


Hakon-Bugge:
v2 -> v3:
       * Cleaned up use of accessor defines for hdr.data
       * Removed explicit define of port_num in struct acm_hdr as data
         became on too short

PR #467, opened 2019-02-04
Support for no port mapping requirements

larrystevenwise:
A soft iwarp driver that uses the host TCP stack via a kernel mode
socket does not need port mapping.  In fact, if the port map daemon,
iwpmd, is running, then iwpmd must not try and create/bind a socket to
the actual port for a soft iwarp connection, since the driver already
has that socket bound.

Yet if the soft iwarp driver wants to interoperate with hard iwarp
devices that -are- using port mapping, then the soft iwarp driver's
mappings still need to be maintained and advertised by the iwpm protocol.

This series updates iwpmd V4 of the IWPM ABI to support soft iwarp.

The ABI version define has been created in rdma_netlink.h so both kernel
and user code can share it.  The iwcm and iwpmd negotiate the ABI version
to use with a new HELLO netlink message.

The corresponding kernel changes are at:
https://github.com/larrystevenwise/linux/tree/topic/no-port-map-v2

### DIFF
--- a/iwpmd/iwarp_pm.h
+++ b/iwpmd/iwarp_pm.h
@@ -108,7 +108,8 @@ enum {
         IWPM_UNKNOWN_MAPPING_ERR,
         IWPM_CLIENT_DEV_INFO_ERR,
         IWPM_USER_LIB_INFO_ERR,
-        IWPM_REMOTE_QUERY_REJECT
+        IWPM_REMOTE_QUERY_REJECT,
+        IWPM_VERSION_MISMATCH_ERR,
 };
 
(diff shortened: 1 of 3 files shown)


larrystevenwise:
The rdma_netlink.h needs to be done using the rdma-core tools.  I'm "remembering" how to do that. :)

larrystevenwise:
Is this ready to merge?  If I fix the warnings, then new warnings will be added for the additional lines in the patch.  The iwpmd code really needs to be reformatting, but I hesitate to do that as part of this patch.

PR #468, opened 2019-02-04
Ibacm log wrong client

Hakon-Bugge:
This is a small fix enabling logging of an unknown netlink client ID, and also introduces a log once functionality.


jgunthorpe:
It is a new way for travis to fail, just ignore it I guess.

PR #471, opened 2019-02-06
librxe: Resolve dmac in create AH

Kamalheib:
Add support for resolving the destination MAC address when creating the AH.
The dmac will be used by the driver to mark loopback packets if the dmac and the smac are equal.


Kamalheib:
@jgunthorpe @rleon @yishaih Is something blocking this PR from being merged?

PR #473, opened 2019-02-18
mlx5: Fix incorrect error handling when SQ wqe count is 0

yishaih:
Driver allocates memory based on wqe count and checks whether the memory
allocation succeeded. However, this memory check should not be performed
when wqe count is 0.


nmorey:
Picked for stables releases too

PR #474, opened 2019-02-20
Verbs: Enhanced ODP functionality 

yishaih:
This series enhanced the ODP functionality in the user space area as follows:
- It reports ODP related capabilities on XRC and SRQ.
- It handles some page fault flow in the mlx5 driver .

The matching kernel part was already merged into for-next.

### DIFF
--- a/libibverbs/cmd.c
+++ b/libibverbs/cmd.c
@@ -265,6 +265,15 @@ int ibv_cmd_query_device_ex(struct ibv_context *context,
 		}
 	}
 
+	if (attr_size >= offsetof(struct ibv_device_attr_ex, xrc_odp_caps) +
+			sizeof(attr->xrc_odp_caps)) {
+		if (resp->response_length >=
+		    offsetof(struct ib_uverbs_ex_query_device_resp, xrc_odp_caps) +
+		    sizeof(resp->xrc_odp_caps)) {
+			attr->xrc_odp_caps = resp->xrc_odp_caps;
+		}
+	}
+
 	return 0;
 }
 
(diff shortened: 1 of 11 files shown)


libibverbs/cmd.c
```diff
@@ -265,6 +265,15 @@ int ibv_cmd_query_device_ex(struct ibv_context *context,
 		}
 	}
 
+	if (attr_size >= offsetof(struct ibv_device_attr_ex, xrc_odp_caps) +
```
jgunthorpe: All these attr_size checks are obsolete and should be completely deleted. Now that verbs and the providers share the same code it is not possible for attr to not have xrc_odp_caps.

providers/mlx5/srq.c
```diff
@@ -179,28 +293,38 @@ int mlx5_alloc_srq_buf(struct ibv_context *context, struct mlx5_srq *srq)
...
+		srq->waitq_head = -1;
+		srq->waitq_tail = -1;
 	}
 
+	srq->wrid = malloc(srq->max * sizeof(*srq->wrid));
+	if (!srq->wrid) {
+		mlx5_free_buf(&srq->buf);
+		return -1;
+	}
 	memset(srq->buf.buf, 0, buf_size);
```
jgunthorpe: This memset is now in a goofy place

libibverbs/cmd.c
yishaih: 'attr_size' is the application's input size which comes from the static inline function
ibv_query_device_ex(),  applications that were complied with legacy libibverbs may not have this field in place, that's why this check is still needed here as done also above.

providers/mlx5/srq.c
yishaih: Correct, will move to the original place.

PR #476, opened 2019-02-24
Pyverbs additions

noaos:
This pull requeset adds pyverbs support for PD and MR related classes,
divided as follows:
- General fixes prior to addition of new classes:
  - Rename Gid class to GID and use as will be done from now on for all
    acronyms.
  - Unify close() functions of pyverbs objects to a single cpdef
    function.
  - Switch to Python-style properties, replacing the deprecated legacy
    syntax.
- Add PD and MR related classes: PD, MR, MW, DM and DMMR. This part also
  adds support for ibv_query_device_ex(), as it is needed for new
  capabilities checks. Those patches also include unittests (for path
  only).
- Add support for ibv_query_port() in Context class. This is a very
  basic capability that doesn't depend on any missing infrastructure.
- Some technical/build-related commits: update MAINTAINERS file and fix
  build issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ addons:
 
       # pyverbs
       - python3-dev
-      - cython3
+      - python3-pip
 
 service:
     - docker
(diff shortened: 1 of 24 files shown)


noaos:
A typo was fixed in one of the commits ('Documentation: update pyverbs'), PR was updated.

rleon:
DCO check is known failure.

Thanks

PR #487, opened 2019-03-06
Predictable RDMA Interface Names

rleon:
Generalize the naming scheme for RDMA devices, so users will always
see names based on topology/GUID information. Such naming scheme has
big advantage that the names are fully automatic, fully predictable
and they stay fixed even if hardware is added or removed (i.e. no
reenumeration takes place) and that broken hardware can be replaced
seamlessly.

### DIFF
--- a/Documentation/udev.md
+++ b/Documentation/udev.md
@@ -147,3 +147,52 @@ In their unit files.
 
 `rdma-hw.target` is also a synchronization point that orders after the low level,
 pre `sysinit.target` RDMA related units have been started.
+
+# Stable names
+
+The library provides general utility and udev rule to automatically perform
+stable IB device name assignments, so users will always see names based on
+topology/GUID information. Such naming scheme has big advantage that the
+names are fully automatic, fully predictable and they stay fixed even if
+hardware is added or removed (i.e. no reenumeration takes place) and that
+broken hardware can be replaced seamlessly.
+
+The name is combination of link type (Infiniband, RoCE, iWARP, OPA or USNIC)
+and the chosen naming policy, like NAME_KERNEL, NAME_PCI, NAME_GUID, NAME_ONBOARD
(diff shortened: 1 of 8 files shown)


Documentation/udev.md
```diff
@@ -147,3 +147,34 @@ In their unit files.
...
+# Stable names
+
+The library provides general utility and udev rule to automatically perform
+stable IB device name assignments, so users will always see names based on
+topology/GUID information. Such naming scheme has big advantage that the
+names are fully automatic, fully predictable and they stay fixed even if
+hardware is added or removed (i.e. no reenumeration takes place) and that
+broken hardware can be replaced seamlessly.
+
+The name is combination of link type (Infiniband, RoCE, iWARP or OPA)
```
paravmellanox: usnic too.

rleon: Thanks, I'll fix

kernel-boot/rdma-persistent-naming.rules
```diff
@@ -0,0 +1,27 @@
...
+# * NAME_PCI
+#   pci = 0000:00:0c.4
+#   Device type = IB
+#   mlx5_0 -> ibp0s12f4
+# * NAME_GUID
+#   GUID = 5254:00c0:fe12:3455
+#   Device type = RoCE
+#   mlx5_0 -> rocex525400c0fe123455
+#
+ACTION=="add", SUBSYSTEM=="infiniband", PROGRAM="rdma_rename %k NAME_FALLBACK"
```
jgunthorpe: I'm a little unclear how this works, what does udevd see after the rename is done? Does it create the right systemd unit files and so on?

rleon: Yes,
http://www.reactivated.net/writing_udev_rules.html#external-naming

PR #489, opened 2019-03-06
ibacm: remove include of net/if.h due to conflict

rleon:
Certain versions of linux-headers nad glibc-headers had both an enum with
identifiers name IFF_* (e.g.: IFF_UP). See e.g.:
https://bugzilla.redhat.com/show_bug.cgi?id=1300256 .

Remove inclusion if <net/if.h> (<linux/if.h> is included through the
netlink3 headers).

Fixes: db109a15 ('ibacm: Replace ioctl with netlink')
Signed-off-by: Tzafrir Cohen <tzafrirc@mellanox.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -360,23 +360,23 @@ else()
   set(HAVE_FULL_SYMBOL_VERSIONS 1)
 endif()
 
-if (${NO_PYVERBS})
-  set(CYTHON_EXECUTABLE "")
-else ()
-  # Look for Python. We prefer some variant of python 3 if the system has it.
-  FIND_PACKAGE(PythonInterp 3 QUIET)
-  if (NOT ${PythonInterp_FOUND})
-    FIND_PACKAGE(PythonInterp REQUIRED)
-  endif()
-  FIND_PACKAGE(cython)
-endif()
-
 # Look for Python. We prefer some variant of python 3 if the system has it.
 FIND_PACKAGE(PythonInterp 3 QUIET)
-if (NOT ${PythonInterp_FOUND})
+if (PythonInterp_FOUND)
+  # pyverbs can only use python3:
+  if (NO_PYVERBS)
+    set(CYTHON_EXECUTABLE "")
+  else()
+    FIND_PACKAGE(cython)
+  endif()
+else()
+  # But we still must have python (be it 2) for the build process:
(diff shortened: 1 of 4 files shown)


markhaywood:
Is there a requirement on the netlink3 version? I know some versions of netlink did not include
<linux/if.h>.

rleon:
@markhaywood, Tzafrir found it by doing cross-compilation over ~50 different distros with default libnl3, and I hope that it will esnure that all distros libnl3 are working.

markhaywood:
But ibacm has been building without this change, right? I'm curious why or what has changed.

jgunthorpe:
Hakon's patch to add libnl support to ibacm broke it. See the fixes line

markhaywood:
Thanks Jason. I missed that. Did Hakon's patch not get built by Travis?
I looked at the commit, but it wasn't clear to me.

jgunthorpe:
It is an obscure distro that fails, I suppose. Travis doesn't run all the distros today.

rleon:
Yes, it was some old, but LTS version of Ubuntu,

markhaywood:
I think this was fixed some time ago in libnl3.
https://github.com/thom311/libnl/commit/50a76998ac36ace3716d3c979b352fac73cfc80a#diff-67dd36fc2860cf9eeb76dc889f5ce447

If this is going to be fixed here, then I think it will have be fixed to be included (or not) conditionally?

rleon:
not on already released systems

markhaywood:
I understand that it breaks on some already released systems. But I think your proposed fix will break it on others. For example, I am building on a system with version 3.2.28. I'm fairly certain that your proposed fix will break the ibacm built there. That is why I proposed a fix that conditionally includes net/if.h. But maybe I am missing something.

rleon:
Hi,

Can you please point me to your proposed fix? And can you please check my
current patch?

Thanks

jgunthorpe:
Lets find a situation where it breaks before speculating please

markhaywood:
Yep, you're right. It builds fine with and without your change. I don't understand why. But given that it does, I don't see any issue with your change.

markhaywood:
I guess it is possible that after the netlink changes, the header file (regardless of the source) is
no longer needed.

PR #490, opened 2019-03-07
travis: Drop SuSE package test due to Travis CI failures

rleon:
Remove SuSE from the list of tested packages due to random
sigfaults while running zypper command under Travis CI environment.


nmorey:
could we switch to leap instead of tumbleweed? it should be much more stable

rleon:
I'll try it in next hour or so.

Thanks

rleon:
https://github.com/linux-rdma/rdma-core/pull/491

PR #495, opened 2019-03-12
verbs: Fix attribute returning

yishaih:
Fix copying of returned attributes to chained buffers.


jgunthorpe:
My tag should be reviewed-by

PR #512, opened 2019-04-15
rsockets: Align rsocket behavior with socket behavior

shefty:
This series adjusts the behavior or rsockets to align with sockets.  Specifically, it adds a service thread to processing connection requests and accept them without using the application thread to drive the CM states.  This allows a single-threaded application to connect to itself, and fixes an issue in rsockets where newly accepted sockets were forced to inherit the nonblocking semantics of the listening socket.  This conflicted with normal socket behavior.

The last patch fixes the behavior of rpoll() in multi-threaded use cases.  It uses a signaling mechanism to wake up blocked threads whenever the state of an rsocket might have changed, so that the polling threads can re-check the rsocket states.

An email thread for the issues should be available here:

http://mail.openjdk.java.net/pipermail/nio-dev/2018-December/005632.html

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008-2014 Intel Corporation.  All rights reserved.
+ * Copyright (c) 2008-2019 Intel Corporation.  All rights reserved.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
@@ -46,6 +46,7 @@
 #include <string.h>
 #include <netinet/tcp.h>
 #include <sys/epoll.h>
+#include <sys/eventfd.h>
 #include <search.h>
 #include <time.h>
 #include <byteswap.h>
@@ -117,6 +118,10 @@ static struct rs_svc listen_svc = {
 	.context_size = sizeof(struct pollfd),
 	.run = cm_svc_run
 };
+
+static uint32_t pollcnt;
+static bool suspendpoll;
+static int pollsignal = -1;
 
 static uint16_t def_iomap_size = 0;
 static uint16_t def_inline = 64;
@@ -479,8 +484,8 @@ static int rs_notify_svc(struct rs_svc *svc, struct rsocket *rs, int cmd)
 	msg.cmd = cmd;
 	msg.status = EINVAL;
 	msg.rs = rs;
-	write_all(svc->sock[0], &msg, sizeof msg);
-	read_all(svc->sock[0], &msg, sizeof msg);
+	write_all(svc->sock[0], &msg, sizeof(msg));
+	read_all(svc->sock[0], &msg, sizeof(msg));
 	ret = rdma_seterrno(msg.status);
 	if (svc->cnt)
 		goto unlock;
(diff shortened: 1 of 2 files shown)


shefty:
I've asked the JDK team to validate the additional changes.

librdmacm/rsocket.c
```diff
@@ -118,6 +118,10 @@ static struct rs_svc listen_svc = {
 	.run = listen_svc_run
 };
 
+static uint32_t pollcnt;
+static bool suspendpoll;
+static int pollsignal[2];
```
mjruhl: 0 is a valid socket, shouldn't this be inited to -1?

librdmacm/rsocket.c
```diff
@@ -2983,19 +2987,123 @@ static uint64_t rs_time_us(void)
...
+	ret = socketpair(AF_UNIX, SOCK_STREAM, 0, pollsignal);
+	if (ret)
+		goto unlock;
+
+	/* Avoid hangs clearing the signal (reading fd) */
+	ret = set_fd_nonblock(pollsignal[0], true);
+	if (ret) {
+		close(pollsignal[0]);
+		close(pollsignal[1]);
+		pollsignal[0] = 0;
```
mjruhl: -1?

librdmacm/rsocket.c
shefty: I was relying on that both fd's can't be 0.  But I can update these to -1.

librdmacm/rsocket.c
```diff
@@ -1026,6 +1035,11 @@ static void rs_free(struct rsocket *rs)
 		rdma_destroy_id(rs->cm_id);
 	}
 
+	if (rs->accept_queue[0] > 0 || rs->accept_queue[1] > 0) {
+		close(rs->accept_queue[0]);
+		close(rs->accept_queue[1]);
+	}
```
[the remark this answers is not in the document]
shefty: See above comment.  Both can't be zero, and I don't think it's possible to open a socketpair() but only get one valid fd.

librdmacm/rsocket.c
```diff
@@ -2983,19 +2987,123 @@ static uint64_t rs_time_us(void)
...
+ * monitored rsocket, independent of whether that thread actually
+ * reads an event from an fd.  In other words, we must wake up all
+ * polling threads whenever poll() indicates that there is a new
+ * completion to process, and when rpoll() will return a successful
+ * value after having blocked.
+ */
+static void rs_poll_stop(void)
+{
+	ssize_t __attribute__((unused)) rc;
+	char c = 5;
```
weiny2: NIT: any special meaning to '5'?

librdmacm/rsocket.c
```diff
@@ -2983,19 +2987,123 @@ static uint64_t rs_time_us(void)
...
+}
+
+/* When an event occurs, we must wait until the state of all rsockets
+ * has settled.  Then we need to re-check the rsocket state prior to
+ * blocking on poll().
+ */
+static int rs_poll_enter(void)
+{
+	pthread_mutex_lock(&mut);
+	if (suspendpoll) {
```
weiny2: Why use suspend poll rather than a trylock yeild loop here?  Does rs_poll_arm need to be called each time?

librdmacm/rsocket.c
```diff
@@ -3166,6 +3172,12 @@ int rpoll(struct pollfd *fds, nfds_t nfds, int timeout)
 		if (ret)
 			break;
 
+		if (timeout >= 0) {
+			timeout -= (int) ((rs_time_us() - start_time) / 1000);
+			if (timeout <= 0)
+				return 0;
```
weiny2: Shouldn't we at least do a poll with a timeout of 0 here?

shefty: The functional equivalent of poll() with a timeout of 0 has already been done (rs_poll_arm).  Here we're simply calculating the new wait timeout, and if no additional waiting is needed, we're done.

librdmacm/rsocket.c
shefty: We're not waiting on a lock here, but yes, we need to call rs_poll_arm() immediately before calling poll() to ensure that we notice all state changes on the rsockets that the thread is monitoring before blocking.  And that any state change after we block will generate an event on the fd.  There are threads currently in the poll() call that we need to wake-up and exit before we can proceed.

In other terms, we need all threads to *enter* poll together.  So threads currently in poll must exit and come get in line with threads waiting here.

librdmacm/rsocket.c
shefty: Yes - I have 5 daughters.  Thanks for asking.  :)

librdmacm/rsocket.c
```diff
@@ -2973,6 +2973,14 @@ ssize_t rwritev(int socket, const struct iovec *iov, int iovcnt)
 	return rsendv(socket, iov, iovcnt, 0);
 }
 
+static uint64_t rs_time_us(void)
+{
+	struct timeval now;
+
+	gettimeofday(&now, NULL);
```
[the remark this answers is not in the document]
shefty: updated - added a new patch to change all instances of gettimeofday to clock_gettime

librdmacm/rsocket.c
```diff
@@ -1026,6 +1035,11 @@ static void rs_free(struct rsocket *rs)
 		rdma_destroy_id(rs->cm_id);
 	}
 
+	if (rs->accept_queue[0] > 0 || rs->accept_queue[1] > 0) {
```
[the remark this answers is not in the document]
shefty: These are a socketpair, so it's safe to close both.

shefty:
Updated based on comments.

librdmacm/rsocket.c
```diff
@@ -2936,19 +2991,123 @@ ssize_t rwritev(int socket, const struct iovec *iov, int iovcnt)
...
+ */
+static int rs_pollinit(void)
+{
+	int ret = 0;
+
+	pthread_mutex_lock(&mut);
+	if (pollsignal[0] >= 0 || pollsignal[1] >= 0)
+		goto unlock;
+
+	ret = socketpair(AF_UNIX, SOCK_STREAM, 0, pollsignal);
```
jgunthorpe: Don't use a sockpair for something that is an eventfd.

librdmacm/rsocket.c
```diff
@@ -4258,11 +4436,7 @@ static void *udp_svc_run(void *arg)
 
 static uint32_t rs_get_time(void)
 {
-	struct timeval now;
-
-	memset(&now, 0, sizeof now);
-	gettimeofday(&now, NULL);
-	return (uint32_t) now.tv_sec;
+	return (uint32_t) (rs_time_us() / 1000000);
```
jgunthorpe: Use of 32 bit seconds should be avoided.

librdmacm/rsocket.c
```diff
@@ -2936,19 +2991,123 @@ ssize_t rwritev(int socket, const struct iovec *iov, int iovcnt)
...
+ * signal state checks whenever a thread updates the state of a
+ * monitored rsocket, independent of whether that thread actually
+ * reads an event from an fd.  In other words, we must wake up all
+ * polling threads whenever poll() indicates that there is a new
+ * completion to process, and when rpoll() will return a successful
+ * value after having blocked.
+ */
+static void rs_poll_stop(void)
+{
+	ssize_t __attribute__((unused)) rc;
```
jgunthorpe: gross - read can always fail with various things, you need to handle it.

shefty: This is clearing one possible queued event.  It doesn't matter if we read an actual value.  I'll switch to eventfd EFD_SEMAPHORE

shefty:
Updated - switched socketpair to eventfd.  Timers converted to 64-bit values.  Maintain errno value from previous call (e.g. poll) when reading/writing the eventfd.

rleon:
So what is the bottom line?
merge or close?

shefty:
Here's the status as of yesterday: the changes fixed the previous issues that they were seen, but there are 3 failures against the JDK test suite.  I don't have more details on those yet.  I don't know if these are new failures, or if this is the first time they've tried running the full test suite.  I would continue sitting on this PR until the testing can be completed.

rleon:
ok, let's add some new fancy label.

shefty:
Update: I was able to get more information on the test failures.  The tests require that nonblocking connect() calls make progress independent from the application thread.  These patches add a thread to progress accepting connections on the listen side.  A similar change will be needed on the connect side.

The tests had work-arounds in them to run with the current upstream rsocket code.  Those were removed to run against this PR.

What's here are improvements over the existing code, but other changes are needed.  I will work on building those over what's here.  I would say wait to merge until the new updates are ready, so that they all hit the same release.

shefty:
Added a patch to use a service thread to drive the connection state on the active side.  This is similar to that used for passive/accept.  Because of the similarities, the listen service thread patch rsockets was slightly updated -- mostly just renames -- in order to re-use the same code on both the active and passive sides.

A fix was applied to the accept patch to use the accept_queue socketpair in rpoll(), in place of the rdma_cm_id event channel fd.  This was a missed change in the original submission.

rleon:
@shefty, I want to release tomorrow and without your explicit merge request won't include this PR.

Thanks

shefty:
That is fine.  I would hold off on including this in the next release and merge after that.  There is still one issue in the JDK testing that is unresolved, and I haven't been able to re-create yet with a reproducer test.

shefty:
IMO, this is ready for merging.  Any other fixes can be on top of what's here.  I have a report of JDK tests occasionally failing to connect, but I have been unsuccessful reproducing.  What's here does allow the majority of the tests to pass, whereas, they failed before.  And this aligns the behavior closer to sockets.

PR #514, opened 2019-04-17
mlx5: Support scatter to CQE over DCT QP

yishaih:
Scatter to CQE which is a performance feature has never been enabled on DCT QP, this series enables it in case the kernel supports it as well.

The matching kernel patch was already sent.


jgunthorpe:
Generally we can't use commit IDs from the for-rc branch in rdma-core as it causes problems for the next commit that updates the kernel headers. You'll have to wait until we merge for-rc to for-next I guess.

nmorey:
The commit says "Fixes" but this seems like a new feature. So I'm guessing this should NOT be backported to stable branches right ?

rleon:
On Thu, May 02, 2019 at 08:41:45AM -0700, Nicolas Morey-Chaisemartin wrote:
> The commit says "Fixes" but this seems like a new feature. So I'm guessing this should NOT be backported to stable branches right ?

It is a fix solved by introduction of new extra bit to ensure
backward/forward compatibility. You are right, there is no need
to backport it.

Thanks

PR #517, opened 2019-04-17
ibacm: acme does not work if server_mode != unix

markhaywood:
Running the ibacm server in mode loopback or open and then trying to run
ib_acme against it, fails:

$ ib_acme -S 127.0.0.1 -v -f i -s 10.196.100.60 -d 10.196.1.60
*** Error in `ib_acme': double free or corruption (fasttop): 0x000000000177c380 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7c619)[0x7f540121d619]
/lib64/libc.so.6(freeaddrinfo+0x28)[0x7f5401282fe8]
ib_acme[0x4049eb]
ib_acme[0x401841]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x7f54011c2c05]
ib_acme[0x40315b]

There are several errors in ib_acm_connect() that need to be fixed.

Afterwards:

$ ib_acme -S 127.0.0.1 -v -f i -s 10.196.100.60 -d 10.196.1.60
Service: 127.0.0.1
Destination: 10.196.1.60
Source: 10.196.100.60
Path information
  dgid: fe80::10:e000:128:d021
  sgid: fe80::10:e000:128:d021
  dlid: 19
  slid: 19
  flow label: 0x0
  hop limit: 0
  tclass: 0
  reversible: 1
  pkey: 0xffff
  sl: 0
  mtu: 4
  rate: 7
  packet lifetime: 0
SA verification: success

return status 0x0

Fixes: 579b6bf8 ("ibacm: Adding new configuration option 'server_mode'")

### DIFF
--- a/ibacm/src/libacm.c
+++ b/ibacm/src/libacm.c
@@ -76,22 +76,19 @@ int ib_acm_connect(char *dest)
 			      res->ai_protocol);
 		if (sock == -1) {
 			ret = errno;
-			goto err1;
+			goto freeaddr;
 		}
 
 		((struct sockaddr_in *) res->ai_addr)->sin_port =
 			htobe16(server_port);
 		ret = connect(sock, res->ai_addr, res->ai_addrlen);
-		if (ret)
-			goto err2;
-
+		if (ret) {
+			close(sock);
+			sock = -1;
+		}
+freeaddr:
 		freeaddrinfo(res);
-
-err2:
-		close(sock);
-		sock = -1;
-err1:
-		freeaddrinfo(res);
+		return ret;
 	} else {
 		struct sockaddr_un addr;
 


ibacm/src/libacm.c
```diff
@@ -59,38 +59,37 @@ static void acm_set_server_port(void)
...
+				ret = errno;
+				close(sock);
+				sock = -1;
+			}
 		}
-
-		((struct sockaddr_in *) res->ai_addr)->sin_port =
-			htobe16(server_port);
-		ret = connect(sock, res->ai_addr, res->ai_addrlen);
-		if (ret)
```
iweiny: I see a lot of clean up in this patch which I'm not sure is necessary.  Not wrong but...  For example AFAICT this hunk is only changing the return value to be errno.  The rest of the logic seems to be the same between the old and new code.  Furthermore, the new code does not follow the kernel convention of branch on failure.  So I'm not sure it is worth the change.  But since this is not the kernel I'm not going to argue against the change.  What I will argue for is a separate clean up patch which is separate from the logic change.

You also don't mention in the changelog that you are making the return meaning of ib_acm_connect consistent with this change and the one below.  That would help.

markhaywood: There are four things being fixed by this patch:

1 - Removing the double setting of hint.ai_family and instead setting hint.ai_protocol.
2 - Making the return from the function consistent (i.e., return errno or -1) since the calling function, query_svcs(), assumed errno which was sometimes not the case.
3 - Fixed double freeaddinfo() in the server_mode != unix case successful return path.
4 - fixed the closing of the socket in the server_mode != unix successful return path.

Maybe this needed to be added to the Changelog?

The clean up of the patch, was mostly a result of fixing 3 and 4. Is it really preferred convention to use goto instead of if-else construct? I have to admit that I find that both surprising and ugly. It is true that I didn't need to touch the server-mode == unix case. However, I thought it best if the two cases were consistent. If that means using goto instead of if-else and (in my opinion) being less elegant, I can do that. 

I'll see if I can provide two patches. One that fixes the bugs (using the goto convention). And then (assuming that it seems worthwhile), one that structures the code using if-else.

ibacm/src/libacm.c
```diff
@@ -80,18 +80,16 @@ int ib_acm_connect(char *dest)
...
-		}
-
-		((struct sockaddr_in *) res->ai_addr)->sin_port =
-			htobe16(server_port);
-		ret = connect(sock, res->ai_addr, res->ai_addrlen);
-		if (ret) {
-			ret = errno;
-			close(sock);
-			sock = -1;
+		} else {
```
weiny2: Again this does not follow the kernel coding style which is to bail on error and not have the "non-error flow" in an else statement.  Jason do you care about that?

ibacm/src/libacm.c
weiny2: Sorry for the late response.

I think the fact that you list 4 things being fixed proves my point that there are too many things going on to properly review if this is correct.

github is confusing me right now.  Did you split the patch into 2?  and repush?  Ah I see it shows the new patch with a new version of this one...  Let me look at those.

ibacm/src/acme.c
```diff
@@ -980,8 +980,12 @@ static int query_svcs(void)
 	for (i = 0; svc_list[i]; i++) {
 		ret = ib_acm_connect(svc_list[i]);
 		if (ret) {
-			printf("%s,unable to contact service: %s\n",
-				svc_list[i], strerror(errno));
+			if (ret == -1)
```
weiny2: What is the semantic difference between -1 and errno?  It seems like -1 is just an error without any "meaning"?  Why?

ibacm/src/libacm.c
```diff
@@ -59,18 +59,22 @@ static void acm_set_server_port(void)
...
+		hint.ai_protocol = IPPROTO_TCP;
 
 		ret = getaddrinfo(dest, NULL, &hint, &res);
-		if (ret)
-			return ret;
+		if (ret) {
+			if (ret == EAI_SYSTEM)
+				return errno;
+			else
+				return -1;
```
weiny2: Why can't we return errno here?

ibacm/src/libacm.c
rleon: I'm not Jason, but we care.

ibacm/src/libacm.c
markhaywood: Before I go through a third iteration. Let me make sure I understand the requirements.

You only want to see one patch for each thing I'm fixing above? Seems excessive, but I can do that.

You (or rather, kernel convention) requires a goto or redundant code to bail out on error? That is don't use the if/else construct. That is counter to years of experience and education. But if that's the Linux way, I'll conform.

And yes, to split the patches I have created a new version of the original patch and then added the subsequent patch and the force pushed. If that's not the preferred method, please let me know what is.

ibacm/src/libacm.c
markhaywood: I don't believe that getaddrinfo() sets errno unless it returns EAI_SYSTEM. I suppose we could set errno to a valuea and return that. But personally, I think manufacturing an errno would be worse than not setting it since obviously the author(s) of getaddrinfo() had that choice themselves and opted not to set it.

weiny2: We don’t have to do anything with errno.  You are trying to define ib_acm_connect() to return an errno value.  So we could EIO or ENODATA?  I’ve not searched extensively for possible “general” errors...

OTOH  What about all the other EAI_* values?
gai_strerror() returns a string for those.  Perhaps it would be better to put the print in this function and simply return -1 or 0 to the caller?  I’ve not got the code in front of me but from your patch ib_acm_connect() is only called from 1 place right?

ibacm/src/libacm.c
weiny2: Yes.  But we are _not_ looking for redundant code.  The point is to not have redundant code if possible.  But also don’t put the “good” non-error code in an if/else.

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#centralized-exiting-of-functions

We may be hitting an issue with this convention as well.

https://www.kernel.org/doc/html/v4.10/process/coding-style.html#functions

Is ib_acm_connect() trying to do too many things?

ibacm/src/libacm.c
markhaywood: You asked why we didn't return errno. I explained why. Whether we chose to set errno and return it or return an errno value seems immaterial.

As for why don't we return EIO or ENODATA - like I said, "I think manufacturing an errno would be worse than not setting one since obviously the author(s) of getaddrinfo() had that choice themselves and opted not to set it."  I don't see what that could possibly accomplish other than provide misleading information.

I had considered using gai_strerror() as you suggested, but thought I might receive negative feedback on printing errors from a function that seemed to be written expressly not to print error messages. I assume the original authors intended libacm functions to not print to stdout or stderr (though I did find an exception).

ibacm/src/acme.c
markhaywood: Because getaddrinfo() in ib_acm_connect() does not always set errno on failure.

ibacm/src/libacm.c
markhaywood: It handles two cases. Is that too many things?

ibacm/src/libacm.c
weiny2: Ah yea, you are correct WRT printing in ib_acm_connect()...

Ok the problem as I see it is that ib_acm_connect() is not following it's own "contract".  Since there is no man page for it, this becomes kind of difficult to determine.  But I think  we need to make ib_acm_connect() follow the same pattern as the library calls it calls.                    

getaddrinfo(), socket(), connect() all return some error and set errno.  Why not make all the error handling the same and follow that convention?                                                                 
             
08:14:34 > git diff
diff --git a/ibacm/src/libacm.c b/ibacm/src/libacm.c                                                                                 
index 66ed298a39ab..f79041973d54 100644                                                                                              
--- a/ibacm/src/libacm.c                                                                                                             
+++ b/ibacm/src/libacm.c
@@ -56,7 +56,7 @@ static void acm_set_server_port(void) 
 int ib_acm_connect(char *dest)
 {              
        struct addrinfo hint, *res;                                                                                                  
-       int ret;                                                                                                                     
+       int ret = 0;                                                                                                                 
 
        acm_set_server_port();                                                                                                       
        memset(&hint, 0, sizeof hint);                                                                                               
@@ -68,22 +68,18 @@ int ib_acm_connect(char *dest)                                                                                   
                                                                                                                                     
        sock = socket(res->ai_family, res->ai_socktype, res->ai_protocol);                                                           
        if (sock == -1) {
-               ret = errno;
-               goto err1; 
+               ret = -1;
+               goto freeaddr;                                                                                                       
        }       
                                                                                                                                     
        ((struct sockaddr_in *) res->ai_addr)->sin_port = htobe16(server_port);                                                      
        ret = connect(sock, res->ai_addr, res->ai_addrlen);                                                                          
-       if (ret)
-               goto err2;
-       
-       freeaddrinfo(res);                                                                                                           
-       return 0;
+       if (ret) {
+               close(sock);                                                                                                         
+               sock = -1;
+       }       
        
-err2:                                    
-       close(sock);                      
-       sock = -1;
-err1:  
+freeaddr:                                
        freeaddrinfo(res);                
        return ret;
}

weiny2: <sigh> that didnot format well.

`08:14:34 > git di
diff --git a/ibacm/src/libacm.c b/ibacm/src/libacm.c                                                                                 
index 66ed298a39ab..f79041973d54 100644                                                                                              
--- a/ibacm/src/libacm.c                                                                                                             
+++ b/ibacm/src/libacm.c
@@ -56,7 +56,7 @@ static void acm_set_server_port(void) 
 int ib_acm_connect(char *dest)
 {              
        struct addrinfo hint, *res;                                                                                                  
-       int ret;                                                                                                                     
+       int ret = 0;                                                                                                                 
 
        acm_set_server_port();                                                                                                       
        memset(&hint, 0, sizeof hint);                                                                                               
@@ -68,22 +68,18 @@ int ib_acm_connect(char *dest)                                                                                   
                                                                                                                                     
        sock = socket(res->ai_family, res->ai_socktype, res->ai_protocol);                                                           
        if (sock == -1) {
-               ret = errno;
-               goto err1; 
+               ret = -1;
+               goto freeaddr;                                                                                                       
        }       
                                                                                                                                     
        ((struct sockaddr_in *) res->ai_addr)->sin_port = htobe16(server_port);                                                      
        ret = connect(sock, res->ai_addr, res->ai_addrlen);                                                                          
-       if (ret)
-               goto err2;
-       
-       freeaddrinfo(res);                                                                                                           
-       return 0;
+       if (ret) {
+               close(sock);                                                                                                         
+               sock = -1;
+       }       
        
-err2:                                    
-       close(sock);                      
-       sock = -1;
-err1:  
+freeaddr:                                
        freeaddrinfo(res);                
        return ret;
}                                         
`

weiny2: I have no idea where that "return -1" removal line is coming from in the previous comment.  It is not part of my patch and I can't seem to edit it out of my comment.  :-(  github is not being friendly to me today...   :-(

markhaywood:
I'm not sure what happened, but some of this conversation has 
disappeared from github. And your suggestion for changes were not being 
rendered in a readable form.

I'm really not sure how to proceed here since you have an issue with the 
changes that I made with the error handling. But I'm not sure what you 
want me to do to change it.

weiny2:
This is what I was trying to add to my comment as a suggestion.

```
11:45:32 > git diff
diff --git a/ibacm/src/libacm.c b/ibacm/src/libacm.c
index 66ed298a39ab..f79041973d54 100644
--- a/ibacm/src/libacm.c
+++ b/ibacm/src/libacm.c
@@ -56,7 +56,7 @@ static void acm_set_server_port(void)
 int ib_acm_connect(char *dest)
 {
        struct addrinfo hint, *res;
-       int ret;
+       int ret = 0;
 
        acm_set_server_port();
        memset(&hint, 0, sizeof hint);
@@ -68,22 +68,18 @@ int ib_acm_connect(char *dest)
 
        sock = socket(res->ai_family, res->ai_socktype, res->ai_protocol);
        if (sock == -1) {
-               ret = errno;
-               goto err1;
+               ret = -1;
+               goto freeaddr;
        }
 
        ((struct sockaddr_in *) res->ai_addr)->sin_port = htobe16(server_port);
        ret = connect(sock, res->ai_addr, res->ai_addrlen);
-       if (ret)
-               goto err2;
-
-       freeaddrinfo(res);
-       return 0;
+       if (ret) {
+               close(sock);
+               sock = -1;
+       }
 
-err2:
-       close(sock);
-       sock = -1;
-err1:
+freeaddr:
        freeaddrinfo(res);
        return ret;
 }
```

weiny2:
Ok now that my code example seems to be formatted in a way that can be read...

What I tried to say as part of pasting that code in:

getaddrinfo(), socket(), and connect() all return some sort of error and set errno on error.  Why not fix the code to have ib_acm_connect() do the same thing.  That keeps the semantics of the call the same for old users and we don't have to change ibacme at all.

Although there is an issue with my code...  need to change ret to be -1 from getaddrinfo.  This still ignores all the EAI_* errors from getaddrinfo but it looks like you are ok with that?

weiny2:
As for github issues this morning:  Sorry, I tried editing my response because of the reder issues and had all kinds of problems.  I was tethered to my cell phone at the time and it could have been that something did not get through...  Sorry about the churn.

markhaywood:
I don't believe that this "getaddrinfo(), socket(), and connect() all return some sort of error and set errno on error" is correct.  getaddrinfo() does not always set errno. The consumer of ib_acm_connect(), query_svcs(), has no way of knowing whether or not is should use strerror() when processing an error return from ib_acm_connect(). What's more, as far as I know, the value of errno upon return from freeaddrinfo() is undetermined isn't it?

weiny2:
My biggest concern here is that you have changed the return value of ib_acm_connect() from it's intended meaning.  I see that my earlier comment got lost in the github issues I had this morning, sorry...

ib_acm_connect()'s "contract" is difficult to determine because there is no man page for it.   But it seems like it is trying to do what I am suggesting.  Returning a simple 0 (success) or -1 (error) and letting errno be the reason that a user can decode if they want.

FWIW I think that what ib_acm_connect() does is wrong.  However, if you need to convince the community that it will not break them.

But, if you do that, then what will the return value be when getaddrinfo returns something other than EAI_SYSTEM?  IMO it is easier and cleaner to turn those into EIO or something which just means we could not communicate for the connection.  Returning -1 is, IMO, confusing because now we have 3 states the user has to contend with (success '0', unknown error '-1', and known error 'errno value').  This does not convey any additional information IMO and puts more burden on the caller...  But I could be wrong.  Regardless we need to see if changing this is going to affect anyone adversely.

Probably not but are you sure?

Also, in the end is it critical to change the semantics of ib_acm_connect() to fix your bug?  I've lost track.

markhaywood:
Fixing this problem does not require me fixing the return codes. I'll leave that for someone else. I've changed this fix to do the bare minimum.

PR #523, opened 2019-05-06
Add traffic in pyverbs

noaos:
This PR adds the support for RDMA traffic in pyverbs. It's
consisted roughly of three parts - Address handle support, QP
support and example (RC pingpong).

### DIFF
--- a/Documentation/pyverbs.md
+++ b/Documentation/pyverbs.md
@@ -209,7 +209,7 @@ device's own memory rather than a user-allocated buffer.
 import random
 
 from pyverbs.device import DM, AllocDmAttr
-from pyverbs.mr import DmMr
+from pyverbs.mr import DMMR
 import pyverbs.device as d
 from pyverbs.pd import PD
 import pyverbs.enums as e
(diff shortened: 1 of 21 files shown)


pyverbs/examples/rc_pingpong.py
```diff
@@ -0,0 +1,208 @@
...
+        self.qp = None
+        self.cq = None
+        self.dm = None
+        self.cc = None
+        self.recv_sge = None
+        self.is_server = False
+        self.num_cq_events = 0
+
+    def parse(self, args):
+        self.parser.add_argument('-m', '--mtu', dest='path_mtu', default=1024,
```
jgunthorpe: Please no - I hate this rc_pingpoing model we have already, do not copy it.

This stuff needs to be done as proper tests under the run tests.

noaos: I agree that we need tests with traffic.
What's wrong with an example that shows how to use pyverbs from scratch?

rleon: Because you are not giving an example, but reimplement C bad example to be in Python.

A person who is searching for a specific example needs to see clear flow without endless irrelevant options.

Examples are supposed to be small and concrete.

noaos: I removed the examples-related patches.

PR #527, opened 2019-05-13
ibverbs: remove node from list before free its space

iotcg:
Signed-off-by: Changcheng Liu <changcheng.liu@intel.com>

### DIFF
--- a/libibverbs/init.c
+++ b/libibverbs/init.c
@@ -486,6 +486,7 @@ out:
 				"Warning: no userspace device-specific driver found for %s\n",
 				sysfs_dev->sysfs_path);
 		}
+		list_del(&sysfs_dev->entry);
 		free(sysfs_dev);
 	}
 
--- a/libibverbs/init.c
+++ b/libibverbs/init.c
@@ -537,6 +537,7 @@ void ibverbs_device_put(struct ibv_device *dev)
 	struct verbs_device *verbs_device = verbs_get_device(dev);
 
 	if (atomic_fetch_sub(&verbs_device->refcount, 1) == 1) {
+		list_del(&verbs_device->entry);
 		free(verbs_device->sysfs);
 		if (verbs_device->ops->uninit_device)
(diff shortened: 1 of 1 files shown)


libibverbs/init.c
```diff
@@ -486,6 +486,7 @@ int ibverbs_get_device_list(struct list_head *device_list)
 				"Warning: no userspace device-specific driver found for %s\n",
 				sysfs_dev->sysfs_path);
 		}
+		list_del(&sysfs_dev->entry);
```
jgunthorpe: This is useless, the sysfs_list is on stack, and sysfs_dev->entry points to that on-stack list. There is no actual reason to list_del it when the sysfs_list will evaporate on funciton exit.

libibverbs/init.c
```diff
@@ -537,6 +537,7 @@ void ibverbs_device_put(struct ibv_device *dev)
 	struct verbs_device *verbs_device = verbs_get_device(dev);
 
 	if (atomic_fetch_sub(&verbs_device->refcount, 1) == 1) {
+		list_del(&verbs_device->entry);
```
jgunthorpe: No.. the ownership of the list belongs to ibverbs_get_device_list, and it holds a reference so long as the device is in that list. When the device is dropped off the list the reference is dropped which permits the device to free. This would just be a double list_del in that path.

libibverbs/init.c
iotcg: @jgunthorpe : get it.

libibverbs/init.c
iotcg: @jgunthorpe Where do the "list_del" operation? When the refcount is 0, the entry should be freed.

jgunthorpe: The refcount does not become 0 while the entry is on the list.

iotcg: atomic_fetch_sub will decrease 1 from verbs_device->refcount  and return it's original value. Since the original value is 1, so it current value should be 0, right?

PR #529, opened 2019-05-14
Incorporate infiniband-diags into rdma-core

jgunthorpe:
@weiny2 would like to stop maintaining infiniband-diags, and also bring it up to the level of CI, packaging, static analysis, etc that rdma-core has. Since it is a fairly small and complementary code base, just roll it into rdma-core. 

The packaging is based on what Debian @bdrung, Fedora @jarodwilson and suse @nmorey have already.

I've already sent many commits to infinbiand-diags that clean up warnings/etc so this drops in and passes travis CI.

This should not be merged until after we make the next rdma-core release.


nmorey:
SUSE packaging looks OK at first glance. I'll give it a spin in Factory to make sure.

CMakeLists.txt
```diff
@@ -45,6 +45,11 @@
 #   -DNO_PYVERBS=1 (default, build pyverbs)
 #      Invoke cython to build pyverbs. Usually you will run with this option
 #      is set, but it will be disabled for travis runs.
+#   -DWITH_IBDIAGS=False (default True)
```
dledford: The context here makes it clear we need to do a consistency review of the options to the build command.  Having mis-matched with_blah versus no_foo is ugly.  We should use with_blah and with_foo and the options should be consistent: True/False or 1/0, but not mixed (although I'm ok if we accept either True or 1, but the comments and usage info should be consistent).  We'll need to watch this on future patches too so we don't get inconsistent again after we fix the consistency issues we have now.

dledford:
Series overall looks good to me.  Certainly a welcome cleanup over what the package was.  My only concern is about pulling in the bits from opensm.  They don't seem like they are likely to change much, but will this create a burden in terms of making sure that our stuff doesn't bit rot while the opensm stuff stays current?  What about making opensm use those bits from rdma-core so that there is no duplication?  After all, I don't think you can legitimately have a system with opensm without having the base rdma-core runtime, but the opposite is true, so making opensm depend on rdma-core shouldn't be any big deal, no?

aron-silverton:
> Series overall looks good to me. Certainly a welcome cleanup over what the package was. My only concern is about pulling in the bits from opensm. They don't seem like they are likely to change much, but will this create a burden in terms of making sure that our stuff doesn't bit rot while the opensm stuff stays current? What about making opensm use those bits from rdma-core so that there is no duplication? After all, I don't think you can legitimately have a system with opensm without having the base rdma-core runtime, but the opposite is true, so making opensm depend on rdma-core shouldn't be any big deal, no?

IIRC, opensm "Requires: libibumad-devel" so it already depends on rdma-core-devel. Without commenting on the contents of the opensm files in question, it does seem better to have them in rdma-core than to risk bit rot or have "Requires: opensm-devel" in rdma-core.spec.

jgunthorpe:
There are three things taken from opensm 
- The complib qmap, this is just a simple tree data structure. We should drop it and replace it with something sensible from ccan. This is a good followup task
- The ibtypes.h header. These are all IBA structure layouts and don't change. Arguably umad or ibmad should provide some clean canonical version of this, but the file is so vast this is hard to create.. and opensm ends up with duplicates anyhow has it gets new features from IBA and has to support older rdma-core
- The node name map reader - this should eventually be exported from libibmad and consumed by opensm as it really logically belongs in libibmad

This is stuff to handle after we go ahead here as it is already pretty complicated to get to this point.

CMakeLists.txt
jgunthorpe: WITH_IBDIAGS is just some throwaway define to sequence getting the cmake and packaging done right. It is deleted later on.

But arguably -DWITH_IBDIAGS_COMPAT should be -DENABLE_IBDIAGS_COMPAT to match most of the other users. NO_PYVERBS is the mistake here.

dledford:
Ok, I'm fine merging it and fixing things up.  What's the Travis failure?  The log looks a mess of errors...

jgunthorpe:
travis is mad because checkpatch on all the historical infiniband-diags commits fails. We shouldn't merge it until the current release is tagged..

iweiny:
> There are three things taken from opensm
> 
> * The complib qmap, this is just a simple tree data structure. We should drop it and replace it with something sensible from ccan. This is a good followup task
> * The ibtypes.h header. These are all IBA structure layouts and don't change. Arguably umad or ibmad should provide some clean canonical version of this, but the file is so vast this is hard to create.. and opensm ends up with duplicates anyhow has it gets new features from IBA and has to support older rdma-core
> * The node name map reader - this should eventually be exported from libibmad and consumed by opensm as it really logically belongs in libibmad
> 
> This is stuff to handle after we go ahead here as it is already pretty complicated to get to this point.

Agree with Jason on these items.

iweiny:
> > There are three things taken from opensm
> > 
> > * The complib qmap, this is just a simple tree data structure. We should drop it and replace it with something sensible from ccan. This is a good followup task
> > * The ibtypes.h header. These are all IBA structure layouts and don't change. Arguably umad or ibmad should provide some clean canonical version of this, but the file is so vast this is hard to create.. and opensm ends up with duplicates anyhow has it gets new features from IBA and has to support older rdma-core
> > * The node name map reader - this should eventually be exported from libibmad and consumed by opensm as it really logically belongs in libibmad
> > 
> > This is stuff to handle after we go ahead here as it is already pretty complicated to get to this point.
> 
> Agree with Jason on these items.

That said I will add to the list:

* Get rid of the compat scripts.  They have been around for a long time and most of those are pretty poor performing.

Do we have any idea who, if anyone is using the following tools:

%{_sbindir}/ibcheckerrs
%{_sbindir}/ibchecknet
%{_sbindir}/ibchecknode
%{_sbindir}/ibcheckport
%{_sbindir}/ibcheckportwidth
%{_sbindir}/ibcheckportstate
%{_sbindir}/ibcheckwidth
%{_sbindir}/ibcheckstate
%{_sbindir}/ibcheckerrors
%{_sbindir}/ibdatacounts
%{_sbindir}/ibdatacounters
%{_sbindir}/ibdiscover.pl
%{_sbindir}/ibswportwatch.pl
%{_sbindir}/ibqueryerrors.pl
%{_sbindir}/iblinkinfo.pl
%{_sbindir}/ibprintca.pl
%{_sbindir}/ibprintswitch.pl
%{_sbindir}/ibprintrt.pl
%{_sbindir}/set_nodedesc.sh

jgunthorpe:
I left the compat scripts because red hat continues to package them, maybe others. The other included distros already dropped them. If RH wishes to stop including them then we can drop them, a decision for another day.

iweiny:
> I left the compat scripts because red hat continues to package them, maybe others. The other included distros already dropped them. If RH wishes to stop including them then we can drop them, a decision for another day.

Agreed...  I was just adding to your list of things to do after this merge.  I'm looking at the changes now.  So far nothing looks wrong.

aron-silverton:
FWIW, some of the tools provided by infiniband-diags (ibhosts and ibnetdiscover, maybe others) don't behave well when running in a VM. I provide an infiniband-diags-guest package that strips out everything except for ibstat and ibstatus, their man pages, and related _libdir stuff for use with VMs. It's been on the long-term plan to figure out how to either make the other utilities work in a VM or have them exit gracefully if they cannot, should not, or currently do not work in VMs. Something to add to the post-merge list of to-do items?

iweiny:
I was having some trouble testing some of the scripts while built in place:

14:36:02 > sudo ./ibnodes
Usage: ibhosts [-h] [<topology-file> | -y mkey -C ca_name -P ca_port -t timeout_ms]
Usage: ibswitches [-h] [<topology-file> | -y mkey -C ca_name -P ca_port -t timeout_ms]
14:36:24 > cat ./ibnodes 
#!/bin/bash /home/iweiny/rdma-core/build/infiniband-diags/scripts/ibnodes

So I tried to build an rpm to install...  and got this error:

CMake Error at infiniband-diags/man/cmake_install.cmake:49 (file):
  file INSTALL cannot find
  "/home/iweiny/rpmbuild/BUILD/rdma-core-24.0/buildlib/pandoc-prebuilt/f48a8d31ddfa68fad6c3badbc768ac703976c43f".
  cmake_install.cmake:88 (include)


FAILED: CMakeFiles/install.util
cd /home/iweiny/rpmbuild/BUILD/rdma-core-24.0 && /usr/bin/cmake -P cmake_install.cmake
ninja: build stopped: subcommand failed.
error: Bad exit status from /var/tmp/rpm-tmp.HC46kW (%install)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.HC46kW (%install)

Not really sure what is wrong but I dug around to see if I could find the man pages built anywhere.  I could not.  Along the way I noticed that in FC29 python-docutils no longer seems to include rst2man...  :-(

So converting to MD may be required sooner rather than later?  Was rst2man moved somewhere?  Or is it no longer supported.  Not even sure if that is related to the rpm error I'm seeing but I'm not seeing any of the rst files being generated into man pages in my builds.  So not sure what cmake magic is supposed to be working here...

iweiny:
On FC29 after getting rst2man from python3-docutils and building an rpm everything seems to be working well.

I don't think it is critical that the scripts run out of an "in tree" build.

Would be nice to convert the docs but I don't think it is critical now that I have found where rst2man went...

iweiny:
> FWIW, some of the tools provided by infiniband-diags (ibhosts and ibnetdiscover, maybe others) don't behave well when running in a VM. I provide an infiniband-diags-guest package that strips out everything except for ibstat and ibstatus, their man pages, and related _libdir stuff for use with VMs. It's been on the long-term plan to figure out how to either make the other utilities work in a VM or have them exit gracefully if they cannot, should not, or currently do not work in VMs. Something to add to the post-merge list of to-do items?

Part of the problem here is that these tools require the ability to send Subnet Mangement packets which are not allowed from VMs.  I _think_ saquery should work in a VM as it should only use GSI packets to talk to the SA which is required for normal traffic.  So that would be a good tool to include.

I've never tried running these on a VM as they are intended for an admin of the network not for "normal users" (IMO it was a mistake for MPI's to be using ibnetdiscover).  So I would have to see what errors the kernel returns when it fails the QP0 communications.  My gut tells me it is not an error which is easily recognized as a "hey you don't have permission for this access" and probably looks like "something is broken"...  If you are willing to help I am sure patches would be welcome by Jason and others.

jgunthorpe:
Update with 

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c6b5b3ddd815b9..ef2251603727b6 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@
 #   -DNO_PYVERBS=1 (default, build pyverbs)
 #      Invoke cython to build pyverbs. Usually you will run with this option
 #      is set, but it will be disabled for travis runs.
-#   -DWITH_IBDIAGS_COMPAT=True (default False)
+#   -DENABLE_IBDIAGS_COMPAT=True (default False)
 #      Include obsolete scripts. These scripts are replaced by C programs with
 #      a different interface now.
 
diff --git a/infiniband-diags/man/CMakeLists.txt b/infiniband-diags/man/CMakeLists.txt
index 3cc2c643c2287c..1f1f91c9f23abf 100644
--- a/infiniband-diags/man/CMakeLists.txt
+++ b/infiniband-diags/man/CMakeLists.txt
@@ -82,7 +82,7 @@ rdma_alias_man_pages(
   dump_fts.8 dump_mfts.8
   )
 
-if (WITH_IBDIAGS_COMPAT)
+if (ENABLE_IBDIAGS_COMPAT)
   rdma_man_pages(
     ibcheckerrors.8
     ibcheckerrs.8
diff --git a/infiniband-diags/scripts/CMakeLists.txt b/infiniband-diags/scripts/CMakeLists.txt
index ae4dcc5b7d9cbe..377a38845d67be 100644
--- a/infiniband-diags/scripts/CMakeLists.txt
+++ b/infiniband-diags/scripts/CMakeLists.txt
@@ -1,6 +1,6 @@
 function(_rdma_sbin_interp INTERP IFN OFN)
   configure_file("${IFN}" "${CMAKE_CURRENT_BINARY_DIR}/${OFN}" @ONLY)
-  file(WRITE "${BUILD_BIN}/${OFN}" "#!${INTERP} ${CMAKE_CURRENT_BINARY_DIR}/${OFN}")
+  file(WRITE "${BUILD_BIN}/${OFN}" "#!${INTERP}\nexec ${INTERP} ${CMAKE_CURRENT_BINARY_DIR}/${OFN} \"$@\"\n")
   execute_process(COMMAND "chmod" "a+x" "${BUILD_BIN}/${OFN}")
 
   install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${OFN}"
@@ -9,7 +9,7 @@ function(_rdma_sbin_interp INTERP IFN OFN)
 endfunction()
 
 function(_rdma_sbin_interp_link INTERP IFN OFN)
-  file(WRITE "${BUILD_BIN}/${OFN}" "#!${INTERP} ${CMAKE_CURRENT_SOURCE_DIR}/${IFN}")
+  file(WRITE "${BUILD_BIN}/${OFN}" "#!${INTERP}\nexec ${INTERP} ${CMAKE_CURRENT_SOURCE_DIR}/${IFN} \"$@\"\n")
   execute_process(COMMAND "chmod" "a+x" "${BUILD_BIN}/${OFN}")
 
   install(FILES "${IFN}"
@@ -84,7 +84,7 @@ rdma_sbin_perl_program(
 install(FILES "IBswcountlimits.pm"
   DESTINATION "${CMAKE_INSTALL_PERLDIR}")
 
-if (WITH_IBDIAGS_COMPAT)
+if (ENABLE_IBDIAGS_COMPAT)
   rdma_sbin_shell_program(
     ibcheckerrors.in
     ibcheckerrs.in
diff --git a/redhat/rdma-core.spec b/redhat/rdma-core.spec
index 72a05e58eda279..fee5f65fb20c6f 100644
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -20,7 +20,7 @@ BuildRequires: libudev-devel
 BuildRequires: pkgconfig
 BuildRequires: pkgconfig(libnl-3.0)
 BuildRequires: pkgconfig(libnl-route-3.0)
-BuildRequires: python-docutils
+BuildRequires: /usr/bin/rst2man
 BuildRequires: valgrind-devel
 BuildRequires: systemd
 BuildRequires: systemd-devel
@@ -286,7 +286,7 @@ easy, object-oriented access to IB verbs.
          -DCMAKE_INSTALL_DOCDIR:PATH=%{_docdir}/%{name}-%{version} \
          -DCMAKE_INSTALL_UDEV_RULESDIR:PATH=%{_udevrulesdir} \
          -DCMAKE_INSTALL_PERLDIR:PATH=%{perl_vendorlib} \
-         -DWITH_IBDIAGS_COMPAT:BOOL=True \
+         -DENABLE_IBDIAGS_COMPAT:BOOL=True \
 %if %{with_static}
          -DENABLE_STATIC=1 \
 %endif
diff --git a/suse/rdma-core.spec b/suse/rdma-core.spec
index a4614646fb2de9..0e5545a1aac714 100644
--- a/suse/rdma-core.spec
+++ b/suse/rdma-core.spec
@@ -64,7 +64,7 @@ BuildRequires:  pkgconfig(libsystemd)
 BuildRequires:  pkgconfig(libudev)
 BuildRequires:  pkgconfig(systemd)
 BuildRequires:  pkgconfig(udev)
-BuildRequires:  python-docutils
+BuildRequires:  /usr/bin/rst2man
 %if %{with_pyverbs}
 BuildRequires:  python3-Cython
 BuildRequires:  python3-devel
```

iweiny:
Much better with those changes.

aron-silverton:
> > FWIW, some of the tools provided by infiniband-diags (ibhosts and ibnetdiscover, maybe others) don't behave well when running in a VM. 
>>
>> snip
> 
> Part of the problem here is that these tools require the ability to send Subnet Mangement packets which are not allowed from VMs. I _think_ saquery should work in a VM as it should only use GSI packets to talk to the SA which is required for normal traffic. So that would be a good tool to include.
> 
Thanks, I will take a look at saquery in the VM. Our use case is debugging communications to/from VMs.
> I've never tried running these on a VM as they are intended for an admin of the network not for "normal users" (IMO it was a mistake for MPI's to be using ibnetdiscover). So I would have to see what errors the kernel returns when it fails the QP0 communications. My gut tells me it is not an error which is easily recognized as a "hey you don't have permission for this access" and probably looks like "something is broken"... If you are willing to help I am sure patches would be welcome by Jason and others.

This all predates me but it's been my goal to get rid of our -guest package by making the utilities say, more or less, "hey you don't have permission for this access" and exit gracefully. I will submit those patches for review if/when I ever get to it. :-)

debian/control
```diff
@@ -336,3 +337,104 @@ Depends: rdma-core (>= 21),
...
+Depends: libibnetdisc5 (= ${binary:Version}), ${misc:Depends}
+Description: InfiniBand diagnostics library
+ InfiniBand is a switched fabric communications link used in
+ high-performance computing and enterprise data centers. Its features
+ include high throughput, low latency, quality of service and
+ failover, and it is designed to be scalable.
+ .
+ This package contains the debug symbols associated with
+ libibnetdisc. They will automatically be used by gdb for debugging
+ libibnetdisc-related issues.
```
bdrung: `libibnetdisc` looks like copy-paste issue.

debian/control
```diff
@@ -336,3 +337,104 @@ Depends: rdma-core (>= 21),
...
+ failover, and it is designed to be scalable.
+ .
+ This package provides libraries required by the InfiniBand
+ diagnostic programs.
+
+Package: libibnetdisc5-dbg
+Section: libs
+Architecture: linux-any
+Depends: libibnetdisc5 (= ${binary:Version}), ${misc:Depends}
+Description: InfiniBand diagnostics library
```
bdrung: Short description should not be identical for `libibnetdisc5` and `libibnetdisc5-dbg`.

debian/control
```diff
@@ -336,3 +337,104 @@ Depends: rdma-core (>= 21),
...
+ InfiniBand is a switched fabric communications link used in
+ high-performance computing and enterprise data centers. Its features
+ include high throughput, low latency, quality of service and
+ failover, and it is designed to be scalable.
+ .
+ This package provides libraries required by the InfiniBand
+ diagnostic programs.
+
+Package: libibnetdisc5-dbg
```
bdrung: In Debian buster, we use automatic dbgsym packages. So adding `libibnetdisc5-dbg` would just be for older releases.

jgunthorpe: What do you want to do here? We have the -dbg packages for everything else currently and if we start using buster stuff then it won't build on lots of things..

PR #530, opened 2019-05-16
rdma-ndd: Execute rdma-ndd after device rename

rleon:
Attempts to run rdma-ndd before device renames produces the following
warnings in /var/log/messages, which are harmless, but annoying.

May 15 20:56:38 server rdma-ndd[23474]: Failed to open /sys/class/infiniband/mlx4_0/node_desc
May 15 20:56:38 server rdma-ndd[23474]: set Node Description failed on mlx4_0

Fixes: 6b4099d47be3 ("kernel-boot: Perform device rename to make stable names")

### DIFF
--- a/rdma-ndd/rdma-ndd.c
+++ b/rdma-ndd/rdma-ndd.c
@@ -131,10 +131,8 @@ static int update_node_desc(const char *device, const char *hostname, int force)
 	nd_file[sizeof(nd_file)-1] = '\0';
 
 	f = fopen(nd_file, "r+");
-	if (!f) {
-		syslog(LOG_ERR, "Failed to open %s\n", nd_file);
+	if (!f)
 		return -EIO;
-	}
 
 	if (!fgets(nd, sizeof(nd), f)) {
 		syslog(LOG_ERR, "Failed to read %s\n", nd_file);
@@ -175,7 +173,7 @@ static void set_rdma_node_desc(const char *hostname, int force)
 			continue;
 
 		if (update_node_desc(dent->d_name, hostname, force))
-			syslog(LOG_ERR, "set Node Description failed on %s\n",
+			syslog(LOG_DEBUG, "set Node Description failed on %s\n",
 			       dent->d_name);
 	}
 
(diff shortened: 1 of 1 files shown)


dledford:
Since device rename, once enabled, never ends up being just at bootup, but ends up being done at any time by admins, I think the better solution is to modify rdma-ndd to detect renames and not emit warnings about missing device files.

jgunthorpe:
We really need to generally fix everything to use the device index and not the device name.

rdma-ndd should immediately query the index and then each time it wants to touch sysfs it can re-query the device name from the index and use that, maybe with a loop.

The renamer helper should also cause a udev variable that holds the device index to be set so we can pass that through systemd to rdma-ndd and other things. 

Though exactly how to do that is a bit involved I fear.

rleon:
@dledford , I got the same comment internally from Yishai

rleon:
Simply removed warnings, anyway rdma-ndd continues to work if such "error" occurred.

jgunthorpe:
I liked the first version better, at least that works properly if nobody is changing the names. This just leaves rdma-ndd totally broken.

rleon:
We can do both variants

rleon:
First variant doesn't prevent from this scenario:

boot
start mlx5
start rdma-ndd
rmmod mlx5 
insmod mlx5 <--- we will get warning here

rleon:
> I liked the first version better, at least that works properly if nobody is changing the names. This just leaves rdma-ndd totally broken.

rdma-ndd doesn't exist in case of missing directory, so "totally broken" is a little bit overrated.

jgunthorpe:
Why would we get a warning on the insmod? rdma-ndd is suppose to be halted after rmmod by systemd and restarted again on the insmod. Perhaps this is broken too because of the issue with the systemd names.

rleon:
rdma-ndd won't be removed because it listens to /sys/class/infiniband/ and not to device insmod.

I didn't check it.

jgunthorpe:
It will be removed because systemd shuts it down when there are no rdma devices left running.

But you could get the same problem if you had multiple devices..

rdma-ndd needs to monitor for the renam e event as well, otherwise failure on add will result in a missed node description update - ie broken.

The only failure that should be suppressed is the fopen - if  that fails then we raced with rename and we will get a rename event to fix it up. The other errors still indicate a bug and should not be removed.

Also the ordering of get_udev_fd and set_rdma_node_desc is wrong

rleon:
I didn't remove other prints except for fopen.

jgunthorpe:
what about the 

```
syslog(LOG_ERR, "set Node Description failed on %s\n",
```
??

rleon:
It is part of fopen failure because it will print such warning for update_node_desc() failure too.

update_node_desc() has enough debug prints inside.

rleon:
Added "rename" event to the list of monitored events.

Anything else?

jgunthorpe:
I think you should keep the top level print and reduce the ones in update_node_desc to LOG_DBG or get rid of them all..

rleon:
> I think you should keep the top level print and reduce the ones in update_node_desc to LOG_DBG or get rid of them all..

done

PR #531, opened 2019-05-16
vmw_pvrdma: Use resource ids from physical device if available

aditr:
This change allows user-space to use physical resource numbers if
they are passed up from the driver. Doing so allows communication with
physical non-ESX endpoints (such as a bare-metal Linux machine or a
SR-IOV-enabled VM).

This is accomplished by separating the concept of the QP number from
the QP handle. Previously, the two were the same, as the QP number
was exposed to the guest and also used to reference  a virtual QP in
the device backend. With physical resource numbers exposed, the QP
number given to the guest is the QP number assigned to the physical
HCA's QP, while the QP handle is still the internal handle used to
reference a virtual QP. Regardless of whether the device is exposing
physical ids, the driver will still try to pick up the QP handle from
the backend if possible.

The MR keys exposed to the guest when the physical resource ids feature
is turned on are likewise now the MR keys created by the physical HCA,
instead of virtual MR keys.

The ABI has been updated to allow the return of the QP handle to the
guest library. The ABI version has been bumped up because of this
non-compatible change.

Reviewed-by: Jorgen Hansen <jhansen@vmware.com>
Signed-off-by: Adit Ranadive <aditr@vmware.com>

### DIFF
--- a/kernel-headers/rdma/vmw_pvrdma-abi.h
+++ b/kernel-headers/rdma/vmw_pvrdma-abi.h
@@ -49,7 +49,11 @@
 
 #include <linux/types.h>
 
-#define PVRDMA_UVERBS_ABI_VERSION	3		/* ABI Version. */
+#define PVRDMA_UVERBS_MIN_ABI_VERSION		3
+#define PVRDMA_UVERBS_MAX_ABI_VERSION		4
+
+#define PVRDMA_UVERBS_NO_QP_HANDLE_ABI_VERSION	3
+
 #define PVRDMA_UAR_HANDLE_MASK		0x00FFFFFF	/* Bottom 24 bits. */
 #define PVRDMA_UAR_QP_OFFSET		0		/* QP doorbell. */
 #define PVRDMA_UAR_QP_SEND		(1 << 30)	/* Send bit. */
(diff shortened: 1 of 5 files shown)


kernel-headers/rdma/vmw_pvrdma-abi.h
```diff
@@ -49,7 +49,11 @@
 
 #include <linux/types.h>
 
-#define PVRDMA_UVERBS_ABI_VERSION	3		/* ABI Version. */
```
jgunthorpe: When you resubmit this, kernel headers have to be in their own commit

jgunthorpe:
Whatever became of this?

aditr:
> Whatever became of this?
Yeah sorry I got pulled into something else and had to leave this on the backburner. I'll come back to this next week.

rleon:
Enough is enough

aditr:
> Enough is enough

What do you mean by that? Is there a deadline to close PRs? I wasn't aware. I would have closed it myself before if that had been the case. We have been having internal discussions about this and sorry about delaying this. I'm hoping we'll resolve those this week and will send another PR.

Thanks,
Adit

rleon:
> What do you mean by that? Is there a deadline to close PRs?
This PR was opened approximately 5 months ago. We prefer to keep our queue clean and see "real" PRs which are expected from us to be reviewed and merged now.

Thanks

kernel-headers/rdma/vmw_pvrdma-abi.h
aditr: @jgunthorpe Does this apply to the kernel changes too?

PR #535, opened 2019-05-29
suse: fix dracut support

nmorey:
SUSE dracut does not support /etc/dracut.conf.d but /usr/lib/dracut/modules.d
Replace redhat script by a simpler one that reuses the standard
udev/systemd scripts

Cc: stable@linux-rdma.org #v15

### DIFF
--- a/suse/module-setup.sh
+++ b/suse/module-setup.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+check() {
+	[ -n "$hostonly" -a -c /sys/class/infiniband_verbs/uverbs0 ] && return 0
+	[ -n "$hostonly" ] && return 255
+	return 0
+}
+
+depends() {
+	return 0
+}
+
+install() {
+	inst /etc/rdma/mlx4.conf
+	inst /etc/rdma/modules/infiniband.conf
+	inst /etc/rdma/modules/iwarp.conf
+	inst /etc/rdma/modules/opa.conf
+	inst /etc/rdma/modules/rdma.conf
+	inst /etc/rdma/modules/roce.conf
+	inst /usr/lib/mlx4-setup.sh
+	inst_multiple lspci setpci awk sleep
+	inst_rules 60-rdma-persistent-naming.rules 70-persistent-ipoib.rules 75-rdma-description.rules 90-rdma-hw-modules.rules 90-rdma-ulp-modules.rules
+	inst_multiple -o \
+                  $systemdsystemunitdir/rdma-hw.target \
+                  $systemdsystemunitdir/rdma-load-modules@.service
+}
+
+installkernel() {
+	hostonly='' instmods =drivers/infiniband =drivers/net/ethernet/mellanox =drivers/net/ethernet/chelsio =drivers/net/ethernet/cisco =drivers/net/ethernet/emulex =drivers/target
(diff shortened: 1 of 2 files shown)


suse/module-setup.sh
```diff
@@ -0,0 +1,31 @@
...
+install() {
+	inst /etc/rdma/mlx4.conf
+	inst /etc/rdma/modules/infiniband.conf
+	inst /etc/rdma/modules/iwarp.conf
+	inst /etc/rdma/modules/opa.conf
+	inst /etc/rdma/modules/rdma.conf
+	inst /etc/rdma/modules/roce.conf
+	inst /usr/lib/mlx4-setup.sh
+	inst_multiple lspci setpci awk sleep
+	inst_rules 70-persistent-ipoib.rules 75-rdma-description.rules 90-rdma-hw-modules.rules 90-rdma-ulp-modules.rules
```
rleon: What about persistent naming rule?

nmorey: Yup. Missed that one. Was working on a previous release and it didn't exist there.
Fixed in the new commit.
Seems like dracut is smart enough to package rdma_rename by itself (thus no inst command)

suse/module-setup.sh
```diff
@@ -0,0 +1,31 @@
...
+	inst /etc/rdma/modules/infiniband.conf
+	inst /etc/rdma/modules/iwarp.conf
+	inst /etc/rdma/modules/opa.conf
+	inst /etc/rdma/modules/rdma.conf
+	inst /etc/rdma/modules/roce.conf
+	inst /usr/lib/mlx4-setup.sh
+	inst_multiple lspci setpci awk sleep
+	inst_rules 60-rdma-persistent-naming.rules 70-persistent-ipoib.rules 75-rdma-description.rules 90-rdma-hw-modules.rules 90-rdma-ulp-modules.rules
+	inst_multiple -o \
+                  $systemdsystemunitdir/rdma-hw.target \
```
jgunthorpe: Does suse use systemd in the initrd that it needs this stuff?

nmorey: We do have systemd in initrd.
We initially used the redhat rdma init script which was not working (at least because some files were not packaed the same way or even not installed).

As we have systemd available, I figured it would be cleaner to use the udev/systemd scripts rather than copy/patch the redhat ones.

Someone at SUSE tested this patch (before the persistent name rules) and it worked well

nmorey: Note: It was tested on top of v16.5.
I simply checked that dracut managed to generate an image on master

rleon: Can you copy "*.rules" with the asterisk? It will ensure that any change in kernel-boot folder won't break SuSE dracut.

nmorey: Not sure about that.
Plus this is run on a full system so if it worked, it'd probably pull all the .service files, not on;y the rdma ones

jgunthorpe: Hum, interesting. So I'd like to get to a world where we don't need explicit module loading..

Do you have a list of tasks that require explicit module loading that suse does from the initrd?

About the only one I know of today is network booting over ipoib..

nmorey: The use case that showed the bug here is kdump over IPoIB which should require roughly the same things

PR #536, opened 2019-06-04
Adding siw user library to rdma-core

BernardMetzler:


redhat/rdma-core.spec
```diff
@@ -148,6 +148,8 @@ Provides: libocrdma = %{version}-%{release}
 Obsoletes: libocrdma < %{version}-%{release}
 Provides: librxe = %{version}-%{release}
 Obsoletes: librxe < %{version}-%{release}
+Provides: libsiw = %{version}-%{release}
```
jgunthorpe: Not needed

redhat/rdma-core.spec
```diff
@@ -148,6 +148,8 @@ Provides: libocrdma = %{version}-%{release}
...
+Obsoletes: libsiw < %{version}-%{release}
```
jgunthorpe: not needed

suse/rdma-core.spec
```diff
@@ -188,6 +188,7 @@ Obsoletes:      libmthca-rdmav2 < %{version}-%{release}
 Obsoletes:      libnes-rdmav2 < %{version}-%{release}
 Obsoletes:      libocrdma-rdmav2 < %{version}-%{release}
 Obsoletes:      librxe-rdmav2 < %{version}-%{release}
+Obsoletes:      libsiw-rdmav2 < %{version}-%{release}
```
jgunthorpe: not needed

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+			    struct ibv_device_attr *attr)
+{
+	struct ibv_query_device cmd;
+	uint64_t raw_fw_ver;
+	unsigned int major, minor, sub_minor;
+	int rv;
+
+	memset(&cmd, 0, sizeof(cmd));
+
+	rv = ibv_cmd_query_device(ctx, attr, &raw_fw_ver, &cmd, sizeof(cmd));
```
jgunthorpe: No, all udatas have to be in structs and declared in the various abi files. Do not pass loose values like this through udata, it makes a big mess to figure out.

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+{
+	return -EOPNOTSUPP;
+}
+
+static int siw_destroy_cq(struct ibv_cq *base_cq)
+{
+	struct siw_cq *cq = cq_base2siw(base_cq);
+	int rv;
+
+	pthread_spin_lock(&cq->lock);
```
jgunthorpe: Bit weird to hold the lock on something that is being freed? Does this do anything?

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+		return NULL;
+
+	if (ibv_cmd_create_srq(pd, &srq->base_srq, attr, &cmd.base, sizeof(cmd),
+			       &resp.base, sizeof(resp))) {
+		free(srq);
+		return NULL;
+	}
+	pthread_spin_init(&srq->lock, PTHREAD_PROCESS_PRIVATE);
+
+	if (resp.siw.srq_key != SIW_INVAL_UOBJ_KEY) {
```
jgunthorpe: success oriented flow, here and in other places like this

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+	case IBV_EVENT_COMM_EST:
+	case IBV_EVENT_QP_LAST_WQE_REACHED:
+		break;
+
+	default:
+		break;
+	}
+}
+
+#define _load_mmapped(a)	\
```
jgunthorpe: why do we need these wrappers?

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+		while (i < base_wr->num_sge) {
+			bytes += base_wr->sg_list[i].length;
+			if (bytes > (int)SIW_MAX_INLINE) {
+				if (siw_debug)
+					printf("libsiw: inline data: %d:%d\n",
+					       bytes, (int)SIW_MAX_INLINE);
+				return -EINVAL;
+			}
+			memcpy(data,
+			       (void *)(unsigned long)base_wr->sg_list[i].addr,
```
jgunthorpe: unintptr_t

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+	case IBV_EVENT_QP_LAST_WQE_REACHED:
+		break;
+
+	default:
+		break;
+	}
+}
+
+#define _load_mmapped(a)	\
+	atomic_load_explicit((a), memory_order_relaxed)
```
jgunthorpe: Are you sure it is relaxed? If this is for producer/consumer pointers it cannot be relaxed

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+	_store_mmapped(fp, flags);
+
+	return 0;
+}
+
+static int siw_db(struct siw_qp *qp)
+{
+	int rv;
+
+	rv = write(qp->base_qp.context->cmd_fd, &qp->db_req,
```
jgunthorpe: Huh? Drivers should not be calling write on the cmd_fd - what is this trying to do?

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+		wr = wr->next;
+	}
+	srq->rq_put = srq_put;
+
+	pthread_spin_unlock(&srq->lock);
+
+	return rv;
+}
+
+static struct {
```
jgunthorpe: const?

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+	{ SIW_OP_READ_LOCAL_INV, IBV_WC_RDMA_READ },
+	{ SIW_OP_SEND, IBV_WC_SEND },
+	{ SIW_OP_SEND_WITH_IMM, IBV_WC_SEND },
+	{ SIW_OP_SEND_REMOTE_INV, IBV_WC_SEND },
+	{ SIW_OP_FETCH_AND_ADD, IBV_WC_FETCH_ADD },
+	{ SIW_OP_COMP_AND_SWAP, IBV_WC_COMP_SWAP },
+	{ SIW_OP_RECEIVE, IBV_WC_RECV }
+};
+
+static struct {
```
jgunthorpe: const?

providers/siw/siw.c
```diff
@@ -0,0 +1,941 @@
...
+
+static const struct verbs_match_ent rnic_table[] = {
+	VERBS_NAME_MATCH("siw", NULL),
+	{},
+};
+
+static struct verbs_device_ops siw_dev_ops = {
+	.name = "siw",
+	.match_min_abi_version = 0,
+	.match_max_abi_version = INT_MAX,
```
jgunthorpe: No, use the right ABI #, the drivers that use 0, INT_MAX are broken.

providers/siw/siw.h
```diff
@@ -0,0 +1,100 @@
...
+	struct siw_cqe *queue;
+	pthread_spinlock_t lock;
+};
+
+struct siw_context {
+	struct verbs_context base_ctx;
+	uint32_t dev_id;
+};
+
+#undef offsetof
```
jgunthorpe: What? no way.

providers/siw/siw.h
```diff
@@ -0,0 +1,100 @@
...
+#ifdef __compiler_offsetof
+#define offsetof(TYPE, MEMBER) __compiler_offsetof(TYPE, MEMBER)
+#else
+#define offsetof(TYPE, MEMBER) ((size_t)&((TYPE *)0)->MEMBER)
+#endif
+
+#ifndef container_of
+#define container_of(ptr, type, member)                                        \
```
jgunthorpe: Use the builtin container_of, it is better than this

providers/siw/siw.h
```diff
@@ -0,0 +1,100 @@
...
+	({                                                                     \
+		const typeof(((type *)0)->member)(*__mptr) = (ptr);            \
+		(type *)((char *)__mptr - offsetof(type, member));             \
+	})
+#endif
+
+#define ctx_base2siw(base_ctx)                                                 \
```
jgunthorpe: Static inline for all of these

redhat/rdma-core.spec
BernardMetzler: OK, will remove

redhat/rdma-core.spec
BernardMetzler: OK, will remove

suse/rdma-core.spec
BernardMetzler: OK, will remove

providers/siw/siw.c
BernardMetzler: Can you please help me here - I do not see what is wrong with it
(it is basically rather the same as all the other provides query_device()
function)?

providers/siw/siw.c
BernardMetzler: OK will change that into pthread_spin_destroy(). wanted to make sure
nobody is still working on that CQ...

providers/siw/siw.c
BernardMetzler: OK will do that for all similar places as well...

providers/siw/siw.c
BernardMetzler: Probably not anymore. will change it.
I did this a while back when atomics were not available on some systems.

providers/siw/siw.c
BernardMetzler: OK

providers/siw/siw.c
BernardMetzler: All mapped SQ/RQ/CQ queues intentionally do not share producer/consumer
pointers. So the producer (libsiw for SQ/RQ, siw for CQ) advances it's
private queue index; the consumer (siw for SQ/RQ, libsiw for CQ) advances
its private queue index as well. The only shared synchronization point is
the current wqe/cqe itself. Here, a VALID flag is set by the producer and
cleared by the consumer. A delayed visibility of a changed state shall
only affect the usable queue depth.

providers/siw/siw.c
BernardMetzler: It rings the SQ DB in case a new SQE got written to the mmapped queue
and the siw kernel thread is likely idling. Basically, it pushes an empty
SQE as efficient as possible. It is what also librxe does as a HW DB surrogate.

providers/siw/siw.c
BernardMetzler: OK will make it 'static const'

providers/siw/siw.c
BernardMetzler: OK will make it 'static const' as well

providers/siw/siw.c
BernardMetzler: OK will fix it to one version of 0

providers/siw/siw.h
BernardMetzler: yes, I'll get rid of that stuff!

providers/siw/siw.h
BernardMetzler: right. this is legacy stuff where container_of was not there in all compilers I used...

will throw that all away

providers/siw/siw.h
BernardMetzler: yes, thanks

providers/siw/siw.c
jgunthorpe: Then use assert(pthread_try_lock) ?

providers/siw/siw.c
jgunthorpe: Grr.. the other providers are wrong too, I thought I got rid of all this stuff :(

All buffers passed to these ibv_cmd functions by drivers must be formed by the DECLARE_DRV_CMD() macros.

providers/siw/siw.c
jgunthorpe: If the kernel reads the producer value that was written by  libsiw then it must also observe all the prior stores of queue entries that libsiw did - and this is not a relaxed ordering model.

I'm confused why a valid bit would exist in a producer/consumer model, that seems unnecessary. At the valid bit set also needs to be a non-relaxed atomic.

providers/siw/siw.c
jgunthorpe: Probably use 1

providers/siw/siw.c
jgunthorpe: Oh that is gross.

If you want to signal the kernel cheaply (or vice versa) then please use an eventfd. Have userspace create one and then pass the FD # into the kernel during get_context. write() from userspace on the event fd can wake up the kernel.

See vfio_virqfd_enable for how this works.

providers/siw/siw.c
jgunthorpe: also the kernel needs to return 1 in the uverbs_abi

providers/siw/siw.c
BernardMetzler: OK I'll change it accordingly.

providers/siw/siw.c
BernardMetzler: sounds good. i'll assert if not getting the lock, and destroy it after the ibv_cmd. If the ibv_cmd fails, I still have to unlock without destroy and return.

providers/siw/siw.c
BernardMetzler: so let me take version 1 then. I'll have to update the kernel driver then as well, yes.

providers/siw/siw.c
BernardMetzler: OK let me make them all strict. on x86 there is no big difference anyway.

providers/siw/siw.c
BernardMetzler: I think the sole reason to allow for IB_USER_VERBS_CMD_POST_SEND
is to support sw rdma userlibs like rxe and siw. Since both have
mmapped SQ's, nothing is to be passed down during post_send,
except the fact the SQ got updated. It would be probably better
to drop that IB_USER_VERBS_CMD_POST_SEND support, and replace it
with a IB_USER_VERBS_CMD_DB_SEND, which just transfers the QP
handle, and is forwarded by ib_core largely uninterpreted.

Besides the effort of wrapping my brain around the whole
vfio_virqfd thing - do we really want to waste another fd just
for passing down that one piece of information? 

So I would be very happy if we could leave it as is for
now (same for rxe), and I'd volunteer to introduce that
IB_USER_VERBS_CMD_DB_SEND for a next release.

jgunthorpe: post_send was not intended to be used like this, it is some rxe abuse to do so. hfi/qib use it more sensibly.

eventfd will be much faster than a IB_USER_VERBS_CMD_DB_SEND - there is lots of overheard on the path. The design of the new api was always to have companion fds for the high speed path.

The whole vfio_virqqfd thing is pretty simple and gives you exactly what siw needs..

We want to look at the rdma-core side before merging kernel patches specifically because it is easier to see ABI abuses  like this from userspace :)

BernardMetzler: So I tried a little this stuff. To my understanding, it would have to be per QP?? What I get in the kernel is just being called on a re-registered handler with an opaque argument, which I can define during vfio_virqfd_enable().
That single argument doesnt help me much if it is not the QP itself. Writing to the eventfd is limited to an 8 bytes buffer which adds the contained value to some bogus kernel counter I can read afterwards from userland. I do not get that number (which would be the QP number in my case) down there in the driver.

If I do not did  it wrong -- I think one fd per QP is rather heavy stuff. one fd per context would be ok.

jgunthorpe: Yeah, you have that right. Sigh. OK let us leave it be then.

BernardMetzler: that's unfortunate. why do these guys do not pass the written value along? I'd be down there at my context in a little more than 200ns (measured on a simple i5 box). that would be pretty fast....the post_send command seem to take ages compared against that ;)

providers/siw/siw.c
jgunthorpe: Oh, my remarks here are not right, raw_fw_ver is just some hacky thing that got added, not a udata like in every other command.. Never mind then.

dledford:
As far as I can tell, all issues Jason raised have been addressed and the kernel patch has been taken.  I'm merging this in.  Thanks.

PR #537, opened 2019-06-04
kernel-boot: Reset buffer before refill

rleon:
In our flows, we fill buffer with larger number of literals than
in second run. It causes to situation where "junk" from previous
call is left in buffer and breaks second parsing.

Fixes: 6b4099d47be3 ("kernel-boot: Perform device rename to make stable names")

### DIFF
--- a/kernel-boot/rdma_rename.c
+++ b/kernel-boot/rdma_rename.c
@@ -240,10 +240,11 @@ static int by_pci(struct data *d)
 		return -ENOMEM;
 
 	ret = readlink(subsystem, buf, sizeof(buf)-1);
-	if (ret == -1) {
+	if (ret == -1 || ret == sizeof(buf)) {
 		ret = -EINVAL;
 		goto out;
 	}
+	buf[ret] = 0;
 
 	subs = basename(buf);
 	if (strcmp(subs, "pci")) {
@@ -251,7 +252,6 @@ static int by_pci(struct data *d)
 		ret = -EINVAL;
 		goto out;
 	}
-
 	/* Real devices */
 	ret = asprintf(&devpath, "/sys/class/infiniband/%s/device", d->curr);
 	if (ret < 0) {
@@ -261,10 +261,12 @@ static int by_pci(struct data *d)
 	}
 
 	ret = readlink(devpath, buf, sizeof(buf)-1);
-	if (ret == -1) {
+	if (ret == -1 || ret == sizeof(buf)) {
 		ret = -EINVAL;
 		goto out;
 	}
+	buf[ret] = 0;
+
 	pci = basename(buf);
 	/*
 	 * pci = 0000:00:0c.0


jgunthorpe:
Why do we need memsets when working with strings? Which call is not forcibly null terminating?

rleon:
> Why do we need memsets when working with strings? Which call is not forcibly null terminating?

readlink() does not append a null byte to buf.
http://man7.org/linux/man-pages/man2/readlink.2.html

jgunthorpe:
If readlink does not append the null byte then the buffer is garbage and the function should abort. You are supposed to check for rc == -1 || rc == sizeof(buf) to detect failure.

rleon:
There is no need to check  "rc == sizeof(buf)", this will be checked later because it is not a failure.

Anyway, I'm not talking about readlink which returns garbage but for situation where consecutive calls to readlink on the same output buffer.

In out case, we filled buf with 25 bytes for first run and for send overwrote with 21 bytes.
Such situation caused to be seen in buf 21 bytes of new data + 4 bytes of old data.

Thanks

jgunthorpe:
readlink always null terminates unless it fails because of insufficient buffer. For this application insufficient buffer is a failure of readlink and it should not continue past it.

rleon:
Please read man link, it doesn't put null buffer and I saw differences between Fedora and Ubuntu on that matter. 

Thanks

jgunthorpe:
The caller is supposed to add the zero based on the return value of readlink

rleon:
we don't care in "\0" in this string, so we don't add

jgunthorpe:
The code immediately  passes buf to basename, so it must be null terminated

Like this is how readlink should be used, not with memset.

```diff
diff --git a/kernel-boot/rdma_rename.c b/kernel-boot/rdma_rename.c
index 6f05a21315584c..550e6bd5141de8 100644
--- a/kernel-boot/rdma_rename.c
+++ b/kernel-boot/rdma_rename.c
@@ -222,7 +222,7 @@ static int by_pci(struct data *d)
        long domain, bus, slot, func, sun;
        char *devpath = NULL;
        char *subsystem;
-       char buf[256] = {};
+       char buf[256];
        char *pci, *subs;
        int ret;
 
@@ -231,11 +231,12 @@ static int by_pci(struct data *d)
        if (ret < 0)
                return -ENOMEM;
 
-       ret = readlink(subsystem, buf, sizeof(buf)-1);
-       if (ret == -1) {
+       ret = readlink(subsystem, buf, sizeof(buf));
+       if (ret == -1 || ret == sizeof(buf)) {
                ret = -EINVAL;
                goto out;
        }
+       buf[ret] = 0;
 
        subs = basename(buf);
        if (strcmp(subs, "pci")) {
@@ -252,11 +253,13 @@ static int by_pci(struct data *d)
                goto out;
        }
 
-       ret = readlink(devpath, buf, sizeof(buf)-1);
-       if (ret == -1) {
+       ret = readlink(devpath, buf, sizeof(buf));
+       if (ret == -1 || ret == sizeof(buf)) {
                ret = -EINVAL;
                goto out;
        }
+       buf[ret] = 0;
+
        pci = basename(buf);
        /*
         * pci = 0000:00:0c.0
```

rleon:
memset in advance ensures that buf[ret] is null, but I'll change for your variant.

PR #540, opened 2019-06-06
verbs: Introduce a new reg_mr API for virtual address space

yuvalshaia:
The virtual address that is registered is used as a base for any address
passed later in post_recv and post_send operations.

On some virtualized environment this is not correct.

A guest cannot register its memory so hypervisor maps the guest physical
address to a host virtual address and register it with the HW. Later on,
at datapath phase, the guest fills the SGEs with addresses from its
address space.
Since HW cannot access guest virtual address space an extra translation
is needed to map those addresses to be based on the host virtual address
that was registered with the HW.
This datapath interference affects performances.

To avoid this, a logical separation between the address that is
registered and the address that is used as a offset at datapath phase is
needed.
This separation is already implemented in the lower layer part
(ibv_cmd_reg_mr) but blocked at the API level.

Fix it by introducing a new API function which accepts an address from
guest virtual address space as well, to be used as offset for later
datapath operations.

### DIFF
--- a/debian/libibverbs1.symbols
+++ b/debian/libibverbs1.symbols
@@ -4,7 +4,8 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
  IBVERBS_1.5@IBVERBS_1.5 20
  IBVERBS_1.6@IBVERBS_1.6 24
- (symver)IBVERBS_PRIVATE_25 25
+ IBVERBS_1.7@IBVERBS_1.7 25.0-1
+ (symver)IBVERBS_PRIVATE_25 25.0-1
  ibv_ack_async_event@IBVERBS_1.0 1.1.6
  ibv_ack_async_event@IBVERBS_1.1 1.1.6
  ibv_ack_cq_events@IBVERBS_1.0 1.1.6
@@ -89,6 +90,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  ibv_read_sysfs_file@IBVERBS_1.0 1.1.6
  ibv_reg_mr@IBVERBS_1.0 1.1.6
  ibv_reg_mr@IBVERBS_1.1 1.1.6
+ ibv_reg_mr_iova@IBVERBS_1.7 25.0-1
  ibv_register_driver@IBVERBS_1.1 1.1.6
  ibv_rereg_mr@IBVERBS_1.1 1.2.1
  ibv_resize_cq@IBVERBS_1.0 1.1.6
--- a/libibverbs/man/ibv_reg_mr.3
+++ b/libibverbs/man/ibv_reg_mr.3
@@ -3,13 +3,17 @@
 .\"
 .TH IBV_REG_MR 3 2006-10-31 libibverbs "Libibverbs Programmer's Manual"
 .SH "NAME"
-ibv_reg_mr, ibv_dereg_mr \- register or deregister a memory region (MR)
+ibv_reg_mr, ibv_reg_mr_iova, ibv_dereg_mr \- register or deregister a memory region (MR)
 .SH "SYNOPSIS"
 .nf
 .B #include <infiniband/verbs.h>
 .sp
 .BI "struct ibv_mr *ibv_reg_mr(struct ibv_pd " "*pd" ", void " "*addr" ,
 .BI "                          size_t " "length" ", int " "access" );
+.sp
+.BI "struct ibv_mr *ibv_reg_mr_iova(struct ibv_pd " "*pd" ", void " "*addr" ,
+.BI "                               size_t " "length" ", uint64_t " "hca_va" ,
+.BI "                               int " "access" );
 .sp
 .BI "int ibv_dereg_mr(struct ibv_mr " "*mr" );
 .fi
@@ -52,11 +56,18 @@ Local read access is always enabled for the MR.
 .PP
 To create an implicit ODP MR, IBV_ACCESS_ON_DEMAND should be set, addr should be 0 and length should be SIZE_MAX.
 .PP
+.B ibv_reg_mr_iova()
+ibv_reg_mr_iova is the same as the normal reg_mr, except that the user is
+allowed to specify the virtual base address of the MR when accessed through
+a lkey or rkey. The offset in the memory region is computed as 'addr +
+(iova - hca_va)'. Specifying 0 for hca_va has the same effect as
+IBV_ACCESS_ZERO_BASED.
+.PP
 .B ibv_dereg_mr()
 deregisters the MR
 .I mr\fR.
 .SH "RETURN VALUE"
-.B ibv_reg_mr()
+.B ibv_reg_mr() / ibv_reg_mr_iova()
 returns a pointer to the registered MR, or NULL if the request fails.
 The local key (\fBL_Key\fR) field
 .B lkey
--- a/libibverbs/CMakeLists.txt
+++ b/libibverbs/CMakeLists.txt
@@ -21,7 +21,7 @@ configure_file("libibverbs.map.in"
 
 rdma_library(ibverbs "${CMAKE_CURRENT_BINARY_DIR}/libibverbs.map"
   # See Documentation/versioning.md
-  1 1.6.${PACKAGE_VERSION}
+  1 1.7.${PACKAGE_VERSION}
   all_providers.c
   cmd.c
   cmd_ah.c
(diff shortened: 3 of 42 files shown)


jgunthorpe:
Correct, please rebase it when you respin it, and don't forget to update the cmakefile and other debian/ files when you make this change, thanks.

yuvalshaia:
So i added a new section to libibverbs/libibverbs.map.in:
IBVERBS_1.7 {
        global: 
                ibv_reg_mr_iova;
} IBVERBS_1.6;

Is that the right way to do it?

CMakefile and debian files were already updated in v5

markhaywood:
Looks right to me. But I believe you need to also update debian/libibverbs1.symbols and libibverbs/CMakefileLists.txt. You might want to take a look at commit 2241d8ceed353c4809d521d98e66ac2fd11f4b88 to see what changes were made to introduce ibv_get_pkey_index().

jgunthorpe:
This needs to be rebased now that we have siw merged

debian/libibverbs1.symbols
```diff
@@ -4,7 +4,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
  IBVERBS_1.5@IBVERBS_1.5 20
  IBVERBS_1.6@IBVERBS_1.6 24
- (symver)IBVERBS_PRIVATE_22 22
+ (symver)IBVERBS_PRIVATE_25 25
```
jgunthorpe: This is missing lines, be sure to test it with 'buildlib/cbuild pkg bionic':

```
dpkg-gensymbols: warning: some new symbols appeared in the symbols file: see diff output below
dpkg-gensymbols: warning: debian/libibverbs1/DEBIAN/symbols doesn't match completely debian/libibverbs1.symbols
--- debian/libibverbs1.symbols (libibverbs1_25.0-1_amd64)
+++ dpkg-gensymbolsEhYGYT	2019-07-12 20:00:15.906417218 +0000
@@ -4,6 +4,7 @@
  IBVERBS_1.1@IBVERBS_1.1 1.1.6
  IBVERBS_1.5@IBVERBS_1.5 20
  IBVERBS_1.6@IBVERBS_1.6 24
+ IBVERBS_1.7@IBVERBS_1.7 25.0-1
  (symver)IBVERBS_PRIVATE_25 25
  ibv_ack_async_event@IBVERBS_1.0 1.1.6
  ibv_ack_async_event@IBVERBS_1.1 1.1.6
@@ -89,6 +90,7 @@
  ibv_read_sysfs_file@IBVERBS_1.0 1.1.6
  ibv_reg_mr@IBVERBS_1.0 1.1.6
  ibv_reg_mr@IBVERBS_1.1 1.1.6
+ ibv_reg_mr_iova@IBVERBS_1.7 25.0-1
  ibv_register_driver@IBVERBS_1.1 1.1.6
  ibv_rereg_mr@IBVERBS_1.1 1.2.1
  ibv_resize_cq@IBVERBS_1.0 1.1.6
```

libibverbs/man/ibv_reg_mr.3
```diff
@@ -52,11 +56,18 @@ Local read access is always enabled for the MR.
 .PP
 To create an implicit ODP MR, IBV_ACCESS_ON_DEMAND should be set, addr should be 0 and length should be SIZE_MAX.
 .PP
+.B ibv_reg_mr_iova()
+Special variant of memory registration used when addresses passed to
```
jgunthorpe: Instead:

ibv_reg_mr_iova is the same as the normal reg_mr, except that the user is allowed to specify the virtual base address of the MR when accessed through a lkey or rkey. The offset in the memory region is computed as 'addr + (iova - hca_va)'. Specifying 0 for hca_va has the same effect as IBV_ACCESS_ZERO_BASED.

shamoya:
Hi Yuval,
I don't understand how this can help in the use case of Virtualization  

A guest can (and does) register memory within its device (VA -> GPA) using this API.
In SRIOV (without vIOMMU) the Hypervisor will map the entire guest memory into the IOMMU (different HW, not the device), so a different entity controls the (GPA -> PA) translation.

So I don't see how this helps you unless your Hypervisor is implementing some kind of vIOMMU and you know the offset used in advanced.
Is this the case?

debian/libibverbs1.symbols
yuvalshaia: So looks like 25 is already taken, just rebased and saw it.

To be honest, i'm totally lost here, any help is appreciated.

1. Do i still need to bump up BAPI?
2. Do i still need this section:
IBVERBS_1.7 {
global:
ibv_reg_mr_iova;
} IBVERBS_1.6;
3. Do i need to change this in libibverbs.map.in
1 1.7.${PACKAGE_VERSION}
4. Do i need to add this to debian/libibverbs1.symbols
ibv_reg_mr_iova@IBVERBS_1.7 25

Yuval

yuvalshaia:
On Sun, Jul 14, 2019 at 06:03:15AM -0700, Ido Shamay wrote:
>    Hi Yuval,
>    I don't understand how this can help in the use case of Virtualization
> 
>    A guest can (and does) register memory within its device (VA -> GPA)
>    using this API.
>    In SRIOV (without vIOMMU) the Hypervisor will map the entire guest
>    memory into the IOMMU (different HW, not the device), so a different
>    entity controls the (GPA -> PA) translation.
> 
>    So I don't see how this helps you unless your Hypervisor is
>    implementing some kind of vIOMMU and you know the offset used in
>    advanced.
>    Is this the case?

I'm talking about device emulation, not passthrough.

rosenbaumalex:
Yuval,

I'm trying to see if I understand the use case:
> I'm talking about device emulation, not passthrough.

so guest driver adds a send/recv descriptors to the emulated device ring, with its Guest PA? 
and HV user space app reads these descriptors, does lookup for the correct lkey, and posts this to the local IB device, based on the hca_va which is already reg_mr'ed in this new API?
or is this for some other use case?

if app does a lookup for the lkey anyway, for each post_send, it can easily do the offset math from the the GPA(hca_va) to the host VA at the same time, no?

yuvalshaia:
On Mon, Jul 15, 2019 at 04:44:25AM -0700, Alex Rosenbaum wrote:
>    Yuval,
> 
>    I'm trying to see if I understand the use case:
> 
>      I'm talking about device emulation, not passthrough.
> 
>    so guest driver adds a send/recv descriptors to the emulated device
>    ring, with its Guest PA?
>    and HV user space app reads these descriptors, does lookup for the
>    correct lkey, and posts this to the local IB device, based on the
>    hca_va which is already reg_mr'ed in this new API?
>    or is this for some other use case?
> 
>    if app does a lookup for the lkey anyway, for each post_send, it can
>    easily do the offset math from the the GPA(hca_va) to the host VA at
>    the same time, no?

No, with this new API there is no need to have the lookup and math in
data-path as the guest base address is already registered with the host HW
- the hca_va is the guest base address.

Idea here is to minimize processing in data-path so with this patch the
only processing is for management purposes (save context etc).

Actually this is a must if in the future we would like to have virtio
driver in guest that bypass the emulated device's ring and post the WQEs
directly to the HW ring. In this case we will not have the chance to do
*any* calculation in the emulated device.

jgunthorpe:
> if app does a lookup for the lkey anyway, for each post_send, it can easily do the offset math from the the GPA(hca_va) to the host VA at the same time, no?

Offset math won't work in the rkey case - if the MR is emulated then the real HW must have the correct hca_va.

rosenbaumalex:
> > if app does a lookup for the lkey anyway, for each post_send, it can easily do the offset math from the the GPA(hca_va) to the host VA at the same time, no?
> 
> Offset math won't work in the rkey case - if the MR is emulated then the real HW must have the correct hca_va.

agreed, 
but what's the application logic example that requires using rkey's with the hca_va?

yuvalshaia:
On Mon, Jul 15, 2019 at 07:49:44AM -0700, Alex Rosenbaum wrote:
>    @rosenbaumalex commented on this pull request.
> 
>      No, with this new API there is no need to have the lookup and math
>      in data-path as the guest base address is already registered with
>      the host HW - the hca_va is the guest base address.
> 
>    there is no mr/lkey lookup?
>    you assume single mr for the entire Guest PA? which also must be
>    continues?

No, but the host mr keys are known to the guest.

(as a side note, to be honest, this part is not ready yet so yes, the
device still do lookup but since it is supported already by the driver then
it is up to me to do it also in the device)

jgunthorpe:
> > > if app does a lookup for the lkey anyway, for each post_send, it can easily do the offset math from the the GPA(hca_va) to the host VA at the same time, no?
> > 
> > 
> > Offset math won't work in the rkey case - if the MR is emulated then the real HW must have the correct hca_va.
> 
> agreed,
> but what's the application logic example that requires using rkey's with the hca_va?

?? All uses of rkeys are combined with a virtual address and thus need to know the hca_va

PR #541, opened 2019-06-12
azp: Set up CI with Azure Pipelines

jgunthorpe:
Azure Pipelines is intended to replace travis as the CI for rdma-core.

Build times are much faster at around 5 mins (vs 23 mins for travis) and
because it is built on pre-built containers it should not suffer from the
same flakiness that travis historically has. We can now also scale up the
number of distros we are testing.

This commit replicates all the functionality of travis, except for the
automatic .tar.gz upload on release. That requires a more careful
security analysis first.

However AZP is setup to use a bionic based container, not the ancient
xenial stuff. This container uses the built in bionic arm64 and 32 bit x86
cross compiler, including all the support libraries. This means it can now
cross build everything except for pyverbs.

The GitHub experience is similar with two notable differences:

1) PRs from non-team outside contributors will not run automatically.
   Someone on the github team has to trigger them with a '/azp run'
   comment. See azure-pipelines.md for why

2) Checkpatch warnings/errors now result in a green checkmark. Clicking
   through the check will show a count of errors/warnings and Azure's GUI
   will show an orange checkmark.

### DIFF
--- a/buildlib/azp-checkpatch
+++ b/buildlib/azp-checkpatch
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+import subprocess
+import urllib.request
+import os
+import re
+import tempfile
+import collections
+import sys
+
+base = os.environ["SYSTEM_PULLREQUEST_TARGETBRANCH"]
+if not re.match("^[0-9a-fA-F]{40}$", base):
+    base = "refs/remotes/origin/" + base
+
+with tempfile.TemporaryDirectory() as dfn:
+    patches = subprocess.check_output(
+        [
+            "git", "format-patch",
+            "--output-directory", dfn,
+            os.environ["SYSTEM_PULLREQUEST_SOURCECOMMITID"], "^" + base
+        ],
+        universal_newlines=True).splitlines()
+    if len(patches) == 0:
+        sys.exit(0)
+
+    ckp = os.path.join(dfn, "checkpatch.pl")
+    urllib.request.urlretrieve(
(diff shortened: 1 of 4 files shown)


aron-silverton:
Jason, we have started using travis via our fork of rdma-core at github.com/oracle/rdma-core. Will it be possible to use AZP from forks? Will the "$ buildlib/cbuild build-images travis" stuff still be maintained for doing a local CI-like build?

jgunthorpe:
For cbuild I added a azp specific image and runner that approximates the actual CI for local running.

As for running your own AZP.. Conceptually yes you can, all the parts are open, however the actual container registry we use for these builds is private (to control BW costs). So you end up having to setup some parallel installation.

Considering Oracle cares about UEK, not things like Centos I think you would be best served to run your own AZP (or otherwise) with your own containers and your own yaml description that focuses on testing UEK relevant scenarios.

aron-silverton:
> For cbuild I added a azp specific image and runner that approximates the actual CI for local running.
> 
Thanks.
<snip>

> Considering Oracle cares about UEK, not things like Centos I think you would be best served to run your own AZP (or otherwise) with your own containers and your own yaml description that focuses on testing UEK relevant scenarios.

I was trying to set things up so that any patches we want to submit upstream have passed CI before we send the PR. We can continue to use cbuild for that but it was nifty to just push to our fork and have it trigger travis.

jgunthorpe:
I think if this is the use, you should use send the PR and check the result here. Running cbuild before is a good idea to save time. 

The azp builder is quite a bit faster than travis so this is much more reasonable.

PR #543, opened 2019-06-18
mlx5: Introduce DEVX APIs for device asynchronous events

yishaih:
This series introduces DEVX APIs for device asynchronous events.

It includes:
- An option to create/destroy an asynchronous event channel.
- An option to subscribe for asynchronous events.
- An option to read asynchronous event.

Detailed description for using the new APIs exists as part of the man pages inside the series.

The matching kernel series was sent upstream for 'rdma-next'.


yishaih:
The kernel headers patch was updated to the accepted code in the kernel.

jgunthorpe:
This merge isn't supposed to happen until the kernel code actually makes it to for-next

PR #545, opened 2019-06-20
Enhance build system to allow separate installation

BtbN:
This is achieved by semi-automatically assigning a component name based
on the current source dir.
For static libs, a seperate build is not sensibly possible due to how
they are generated. But it's at least made sure all the dependencies are
cleanly built, and ultimately only the requested lib is built.

With this, one can for example do "make install-libibverbs" to install
just libibverbs and its related files. Man pages and tests are
intentionally assigned a separate target, like install-libibverbs-man.

The motivation behind this comes from packaging this in a modular fashion.
Luckily, static libs are not overly common, so users who want them will just have to accept it how it is.

### DIFF
--- a/Documentation/CMakeLists.txt
+++ b/Documentation/CMakeLists.txt
@@ -1,4 +1,6 @@
-install(FILES
+rdma_component()
+
+rdma_install(FILES
   ibacm.md
   ibsrpdm.md
   libibverbs.md
--- a/buildlib/publish_headers.cmake
+++ b/buildlib/publish_headers.cmake
@@ -25,6 +25,6 @@ function(publish_headers DEST)
 
   foreach(SFIL ${ARGN})
     get_filename_component(FIL ${SFIL} NAME)
-    install(FILES "${SFIL}" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${DEST}/" RENAME "${FIL}")
+    rdma_install(FILES "${SFIL}" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${DEST}/" RENAME "${FIL}")
   endforeach()
 endfunction()
--- a/buildlib/pyverbs_functions.cmake
+++ b/buildlib/pyverbs_functions.cmake
@@ -21,7 +21,7 @@ function(rdma_cython_module PY_MODULE)
       LIBRARY_OUTPUT_DIRECTORY "${BUILD_PYTHON}/${PY_MODULE}"
       PREFIX "")
     target_link_libraries(${SONAME} LINK_PRIVATE ${PYTHON_LIBRARIES} ibverbs)
-    install(TARGETS ${SONAME}
+    rdma_install(TARGETS ${SONAME}
       DESTINATION ${CMAKE_INSTALL_PYTHON_ARCH_LIB}/${PY_MODULE})
   endforeach()
 endfunction()
@@ -30,7 +30,7 @@ function(rdma_python_module PY_MODULE)
   foreach(PY_FILE ${ARGN})
     get_filename_component(LINK "${CMAKE_CURRENT_SOURCE_DIR}/${PY_FILE}" ABSOLUTE)
     rdma_create_symlink("${LINK}" "${BUILD_PYTHON}/${PY_MODULE}/${PY_FILE}")
-    install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/${PY_FILE}
+    rdma_install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/${PY_FILE}
       DESTINATION ${CMAKE_INSTALL_PYTHON_ARCH_LIB}/${PY_MODULE})
   endforeach()
 endfunction()
--- a/buildlib/rdma_functions.cmake
+++ b/buildlib/rdma_functions.cmake
@@ -10,6 +10,9 @@ set(RDMA_STATIC_LIBS "" CACHE INTERNAL "Doc" FORCE)
 # Global list of tuples of (PROVIDER_NAME LIB_NAME)
 set(RDMA_PROVIDER_LIST "" CACHE INTERNAL "Doc" FORCE)
 
+# Current component
+set(RDMA_CUR_INSTALL_COMP "${CMAKE_INSTALL_DEFAULT_COMPONENT_NAME}")
+
 set(COMMON_LIBS_PIC ccan_pic rdma_util_pic)
 set(COMMON_LIBS ccan rdma_util)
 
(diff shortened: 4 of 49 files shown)


jgunthorpe:
What does 'packaging in a modular fashion' mean?

Do mean some thing like debian does where there are lots of packages? Why do we need cmake support for that?

IMHO the mapping of files to packages is entirely the responsibility of the packaging and cmake should not be involved.

BtbN:
Writing ebuilds/pkg-builds for Gentoo/Arch primarily. Each package, like for example libibverbs, builds and installs only that component.
When it still was individual srcrmps in a big OFED tarball, that was no problem. But in the monolithic fashion that is rdma-core now, so far what has been done is build the desired targets, and then manually copy the files to the destdir.
With this it's possible to do it way cleaner.

jgunthorpe:
It is not really cleaner, you've just encoded the gentoo preferences for file/package mapping into cmake, when cmake is not supposed to be concerned with packaging.

In the cmake model you have to compile everything and do the 'make install' step, then take what you want from  DESTDIR. This is what every distro does, why is it a problem for gentoo? (BTW you cannot skip make install or stuff will not work!!)

I know nothing about gentoo, but I was just browsing the systemd ebuild and it looks like it just eninja installs the entire mess, while debian splits systemd into lots and lots of packages. Are you sure it makes sense in the gentoo world to split up rdma-core? The pieces are really quite small.

Honestly, I don't want to support anything other than 'cmake && ninja && ninja install' (already tested all the distros we support!)- the cmake stuff is already so complicated and adding more permutations to that mix is really not a good thing.

BtbN:
All the current packages are split for historical reasons(old OFED tarballs). So it's virtually impossible to change that without breaking each and every package in the tree that depends on one of them.
Gentoo commonly goes through quite some lengths to split packages that are one source tree upstream. For example Qt and KDE/Plasma.

I'm aware that static libs need special handling and only work with the final post-processing step. They are even unusual enough that I'd consider dropping them from the packages. Though this specific thing is actually accounted for by this. Or is there something else that'll break?

Even if this would be one single package in portage, that would not solve this problem. There would be useflags for all the individual components that are split packages right now. So during building, the same problem of needing to only build and install parts of the whole thing would arise.

jgunthorpe:
Does QT have a long list of separate install targets?

'make install' does many things, fixes rpath, performs variable substitutions, builds out man pages, the static library stuff, and probably more I've forgot. I will absolutely not support any distro that does not take its output to install from DESTDIR after 'make install'.

If you want to have fine grained useflags then the packaging needs to filter DESTDIR, like everyone else does. But I'm not sure the point, the pieces are so small, and the build is quite fast..

dledford:
Gentoo may split things up, so did Red Hat when they put rdma-core into
rhel7.  But the split up happens in the package manager scripts and
setups.  What package manager does Gentoo use?  Can it not do this?

-- 
    GPG KeyID: B826A3330E572FDD
    Fingerprint = AE6B 1BDA 122B 23B4 265B  1274 B826 A333 0E57 2FDD

BtbN:
Gentoo uses portage. It's a source based distribution, there are no binary packages. When a user installs a library, it's built from source, installed to staging dir, and then merged into the system.
So the usual binary-distro approach of building everything and then splitting the result into packages is not applicable.
Forcing a huge bunch of libraries, binaries and daemons onto the user, which potentially only wants/needs as dependency one single lib, is avoided as much as possible.

Installing individual subdirs, including all the final fixups like rpath changing, are already working fine as it is now, since cmake generates per-subdir build and install targets. It's only about the static libs and the pkg-config files, since they get special handling and are accounted to the top level CMakeLists target.

Would it be more acceptable if I just generate custom targets to install those specifically?

jgunthorpe:
Must do 'make install' - do not copy random things out of the build tree, I will not support that.

I still think you have to download and build the whole thing then copy the bits you want out of DESTDIR directory, then repeat the whole thing for the other bits. Are you sure portage can't share the source build between different 'packages' or something? This must be so common, many packages are giant multi-package things these days. I can't believe Gentoo is going around patching build systems all over the place???

BtbN:
It indeed does not support that. Every package goes through the usual prepare+config+compile+install+merge cycle. There is (pretty sure intentionally) no support for sharing build artifacts between packages.
It works fine for all packages, even though not intended by upstream at times. It also works great for rdma-core, due to the automatic target generation by cmake.
One can already do "make libibverbs{,/man,/examples} libibverbs{,/man,/examples}/install" and end up with a complete installation, minus the .pc file.
For static libs, there is no way around calling "make make_static".

So the only pieces missing for support are install targets for the individual .pc and .a files.
Copying those out of the cmake binary dir is working at the moment, though a bit ugly.

dledford:
So it's like the FreeBSD source stuff.

When you run this, let's say I ran make install-libibverbs, does it install that in the live filesystem, or in DESTDIR?

jgunthorpe:
The result of 'make ibverbs/etc' is NOT a complete installation, again we will not support any install flow that does not go through 'make install'. You will have bugs.

It sounds like what you really want is to conditionally compile & install less of the package, maybe that is a more maintainable cmake change?

BtbN:
> When you run this, let's say I ran make install-libibverbs, does it install that in the live filesystem, or in DESTDIR?

It does the same the original install target does (cmake -P BINDIR/cmake_install.cmake), it just adds another variable to identify the component, so it will only install things that had that component passed to its install() invocation.
All other things behave the same, i.e. it does honor DESTDIR.

> The result of 'make ibverbs/etc' is NOT a complete installation, again we will not support any install flow that does not go through 'make install'. You will have bugs.

Well, yes, that's kind of the whole idea? What's missing (except for the .pc and a potential .a file) when running "make/ninja libibverbs/install libibverbs/man/install libibverbs/examples/install"?

> It sounds like what you really want is to conditionally compile & install less of the package, maybe that is a more maintainable cmake change?

In the end, yes. But it'll also have to be split entirely. Undoing the historical split is not gonna be accepted for getting the packages into the official tree.
Sure, one could add a whole bunch of if(ENABLE_LIBIBVERBS) add_subdir(...) endif() into the top level CMakeLists, but I don't think this is necessary, since using individual targets already seems to work just fine, except for a few minor missing pieces.

jgunthorpe:
I still think this approach is very broken. rdma-core is integrated, so if you build something like srp_daemon it always builds the internal libibverbs, libibumad, etc and then links the srp_daemon to those internal libraries. It never uses the system libraries for building.

So if you then install srp_daemon it might not work because it wasn't built with the libraries on the system.

Pretending otherwise by hacking up cmake doesn't change the reality - it is one package now, and a source based system like Gentoo needs to install it in one shot - because we just don't support building the fragments independently against the system libraries anymore.

Surely portage has some means to handle package aggregations, it happens often enough..

dledford:
> I still think this approach is very broken. rdma-core is integrated, so if you build something like srp_daemon it always builds the internal libibverbs, libibumad, etc and then links the srp_daemon to those internal libraries. It never uses the system libraries for building.
> 
> So if you then install srp_daemon it might not work because it wasn't built with the libraries on the system.
> 
> Pretending otherwise by hacking up cmake doesn't change the reality - it is one package now, and a source based system like Gentoo needs to install it in one shot - because we just don't support building the fragments independently against the system libraries anymore.
> 
> Surely portage has some means to handle package aggregations, it happens often enough..

That all makes sense and I don't disagree with any of it.  That said, the patch still seems OK to me.  It enables partial installs and I'm perfectly fine saying something like "If you install X, then you must also install Y and Z that it was built against".

I am not a CMake expert, so pardon my ignorance on this, but would it be possible to update the patch to force installation of dependencies?  So if you install srp_daemon it would also install libibumad and libibverbs?  We obviously already have the dependency information for building, can we just reuse that during install?

jgunthorpe:
It would be a nightmare to figure out the dependencies for the 'install' step, AFAIK. And because we have to build the dependencies I really question the point - I would think almost everything would be a big time % of a full build anyhow, so why even optimize this?

AFAIK the only really useful purpose is that it lets gentoo push the DESTDIR filtering into cmake - and I feel pretty strongly that how the files are split up is a distro specific concern, not a cmake one.

PR #549, opened 2019-07-09
cbuild: Upgrade PyYAML load call to v5.1

rleon:
Use of PyYAML's yaml.load function without specifying the Loader=...
parameter, has been deprecated. In PyYAML version 5.1.

Before PyYAML 5.1, the PyYAML.load function could be easily exploited
to call any Python function. That means it could call any system command
using os.system().

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -258,7 +258,7 @@ class travis(APTEnvironment):
 
         # Load the commands from the travis file
         with open(".travis.yml") as F:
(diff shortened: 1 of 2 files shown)


jgunthorpe:
Why is the kernel boot thing in this PR?

rleon:
I'm finding bugs and fixing them.

The old PR name isn't correct anymore, but if you want, I will separate it to two PRs.

PR #550, opened 2019-07-10
pyverbs fixes

noaos:
The following is a series of fixes to pyverbs:
The first two patches clean the compilation warning introduces in
Fedora 30.
The third patch fixes a wrong casting.
The last one fixes the handling of bad work requests in post send and
post recv operations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,16 @@ endif()
 
 #-------------------------
 # Setup the basic C compiler
+# Check if the current compiler supports -fvar-tracking-assignments.
+# For that, we need to try and compile some code with it. Save the current
+# CMAKE_REQUIRED_FLAGS so we can restore it later.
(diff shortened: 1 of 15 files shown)


CMakeLists.txt
```diff
@@ -182,6 +182,13 @@ endif()
 
 #-------------------------
 # Setup the basic C compiler
+# Some compilation flags are not supported in clang, lets allow users to know
+# whether gcc or clang is used.
+if ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
```
rleon: I know that I reviewed it, but unfortunately, this is not enough, you need to do a feature test.
It will allow us to run and check everything with new clang once they will support this -fno-var-tracking-assignments without changing rdma-core at all. Otherwise, we won't be able to open this check, because it will never work on "old" platforms.

PR #551, opened 2019-07-11
Debian 24.0-2 changes

bdrung:
This patch set contains the packaging changes from the Debian unstable 24.0-2 upload.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,7 @@ Build-Depends: cmake (>= 2.8.11),
                python-docutils,
                python3-dev,
                valgrind [amd64 arm64 armhf i386 mips mips64el mipsel powerpc ppc64 ppc64el s390x]
-Standards-Version: 4.3.0
+Standards-Version: 4.4.0
 Vcs-Git: https://github.com/linux-rdma/rdma-core.git
 Vcs-Browser: https://github.com/linux-rdma/rdma-core
 Homepage: https://github.com/linux-rdma/rdma-core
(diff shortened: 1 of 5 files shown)


debian/control
```diff
@@ -66,7 +67,7 @@ Architecture: linux-any
 Multi-Arch: same
 Depends: ${misc:Depends}, ${shlibs:Depends}
 Provides: libcxgb3-1,
-          libefa-1,
```
jgunthorpe: There never was a libefa, so this shouldn't have been added, just delete it

bdrung: This package ships a `libefa.so.1` file. So providing libefa1 is nice to have.

dledford: Auto provides should pick up the current Provides automatically.  Usually, you only have to give a manual Provides: tag for when you want an application linked against an older libibverbs and using an older lib<provider> to still pass package system dependency resolution.  Since there never was an older libefa for things to link against, there should be no requirement for a manual provides of an older libefa version (the current libefa should not be libefa-1).  But, that's all based on my knowledge of rpm as the package manager.  I can't speak for Debian, but if apt works similarly, then Jason would be right.

jgunthorpe: The provides are only present to easy transition from the pre-rdma-core era. We don't need to add new provides for things that didn't exist prior to rdma-core.

bdrung: Debian does not have an "auto provides" functionality. You can remove the provided libefa1 if you like. It is only useful for users wanting to install that library and will not have an influence on the inter-package dependencies.

PR #556, opened 2019-07-25
cxgb4: fix chipversion initialization

bharatpotnuri:
With Commit `0c3cecfe04d5a`, device and vendor members of structure
verbs_match_ent are set to zero. This causes c4iw_alloc_context()
failure due to incorrect chipversion.
This patch reorders the code to fetch chipversion from ib device attributes.

Signed-off-by: Potnuri Bharat Teja <bharat@chelsio.com>

### DIFF
--- a/providers/cxgb4/dev.c
+++ b/providers/cxgb4/dev.c
@@ -144,22 +144,6 @@ static struct verbs_context *c4iw_alloc_context(struct ibv_device *ibdev,
 
 	verbs_set_ops(&context->ibv_ctx, &c4iw_ctx_common_ops);
 
-	switch (rhp->chip_version) {
-	case CHELSIO_T6:
-		PDBG("%s T6/T5/T4 device\n", __FUNCTION__);
-	case CHELSIO_T5:
-		PDBG("%s T5/T4 device\n", __FUNCTION__);
-	case CHELSIO_T4:
-		PDBG("%s T4 device\n", __FUNCTION__);
-		verbs_set_ops(&context->ibv_ctx, &c4iw_ctx_t4_ops);
-		break;
-	default:
-		PDBG("%s unknown hca type %d\n", __FUNCTION__,
-		     rhp->chip_version);
-		goto err_unmap;
-		break;
-	}
-
 	if (!rhp->mmid2ptr) {
 		int ret;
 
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Nit: it isn't uninitialized, they are deliberately set to 0.

bharatpotnuri:
> 
> 
> Nit: it isn't uninitialized, they are deliberately set to 0.

Right. Corrected it. Thanks!

providers/cxgb4/dev.c
```diff
@@ -196,6 +180,23 @@ static struct verbs_context *c4iw_alloc_context(struct ibv_device *ibdev,
 				context->status_page->write_cmpl_supported;
 	}
 
+	rhp->chip_version = CHELSIO_CHIP_VERSION(attr.vendor_part_id >> 8);
+	switch (rhp->chip_version) {
+	case CHELSIO_T6:
+		PDBG("%s T6/T5/T4 device\n", __FUNCTION__);
```
rleon: Macro PDBG worries me, it is a wrapper to syslog(LOG_DEBUG, ....), which is not visible without enabling full debug mode for systemd.

I afraid that the right thing will be to delete it.

bharatpotnuri: Yes, I have been thinking of removing it as part of my clean-up changes.
I am thinking of changing it to env variable based prints instead.
Do you have any better suggestions?
Thanks @rleon

bharatpotnuri:
Hey @rleon 
v25.0 is released with `verbs: Use CHARDEV info from netlink to bind drivers` and without this PR included. This will break user cxgb4. 
> https://github.com/linux-rdma/rdma-core/releases

Please add this to the immedieate dot release if there are no issues.
Thanks.

rleon:
> Hey @rleon
> v25.0 is released with `verbs: Use CHARDEV info from netlink to bind drivers` and without this PR included. This will break user cxgb4.
> 
> > https://github.com/linux-rdma/rdma-core/releases
> 
> I am not sure how to tag it. Please add this to the immediate dot release if there are no issues.
> Thanks.

@nmorey and this PR https://github.com/linux-rdma/rdma-core/pull/554/ should go to stable@ too.

providers/cxgb4/dev.c
rleon: No prints are supposed to be in library, I think that combination of general environment variable with prints to syslog is the right thing to do.

Se this commit https://github.com/linux-rdma/rdma-core/pull/554/commits/ef8aca40b140d234aadbdaedbb6a2199dd2dce6e

mlx5_dbg should be removed too.

jgunthorpe: LOG_DEBUG works fine, it depends on the journald configuration in /etc/ if it gets recorded or not.

jgunthorpe:
There are lot of checkpatch warnings, please fix them..

providers/cxgb4/dev.c
rleon: How will you enable log_debug for a specific application?

bharatpotnuri:
> 
> 
> There are lot of checkpatch warnings, please fix them..
My bad, missed checking them. Fixed now.

bharatpotnuri:
> > There are lot of checkpatch warnings, please fix them..
> 
> My bad, missed checking them. Fixed now.

The build checks were related to package generation.

jgunthorpe:
> > > There are lot of checkpatch warnings, please fix them..
> > 
> > 
> > My bad, missed checking them. Fixed now.
> 
> The build check failures were related to package generation.

Oops, please rebase

jgunthorpe:
Merged by hand. Please don't split dependent patches in a series across two pull requests.

PR #557, opened 2019-07-30
Fixes and code optimization for libhns

oulijun:
Here are some fixes and codes optimization for libhns.

### DIFF
--- a/providers/hns/hns_roce_u.c
+++ b/providers/hns/hns_roce_u.c
@@ -35,7 +35,6 @@
 #include <string.h>
 #include <pthread.h>
 #include <sys/mman.h>
-#include <fcntl.h>
 #include <unistd.h>
 
 #include "hns_roce_u.h"
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -183,7 +183,7 @@ struct hns_roce_srq {
 	unsigned int			srqn;
 	int				max;
 	unsigned int			max_gs;
-	int				wqe_shift;
+	unsigned int			wqe_shift;
 	int				head;
 	int				tail;
 	unsigned int			*db;
@@ -199,7 +199,7 @@ struct hns_roce_wq {
 	unsigned int			head;
 	unsigned int			tail;
 	unsigned int			max_gs;
-	int				wqe_shift;
+	unsigned int			wqe_shift;
 	int				offset;
 };
 
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -200,6 +200,7 @@ struct hns_roce_wq {
 	unsigned int			tail;
 	unsigned int			max_gs;
 	unsigned int			wqe_shift;
+	unsigned int			shift; /* wq size is 2^shift */
 	int				offset;
 };
 
@@ -247,11 +248,6 @@ struct hns_roce_u_hw {
 	uint32_t hw_version;
 	struct verbs_context_ops hw_ops;
 };
-
-static inline unsigned long align(unsigned long val, unsigned long align)
-{
-	return (val + align - 1) & ~(align - 1);
-}
 
 static inline struct hns_roce_device *to_hr_dev(struct ibv_device *ibv_dev)
 {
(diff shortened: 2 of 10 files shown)


providers/hns/hns_roce_u.h
```diff
@@ -183,7 +183,7 @@ struct hns_roce_srq {
 	unsigned int			srqn;
 	int				max;
 	unsigned int			max_gs;
-	int				wqe_shift;
+	unsigned int			wqe_shift;
```
rleon: The same "int wqe_shift" in struct hns_roce_wq, it should be updated too.

providers/hns/hns_roce_u.h
```diff
@@ -253,6 +254,32 @@ static inline unsigned long align(unsigned long val, unsigned long align)
 	return (val + align - 1) & ~(align - 1);
 }
 
+static inline int hr_ilog2(int n)
```
rleon: Use ccan/ilog.h and remove mlx5_ilog2() in favour of ccan too.

providers/hns/hns_roce_u.h
```diff
@@ -253,6 +254,32 @@ static inline unsigned long align(unsigned long val, unsigned long align)
...
+	if (n <= 0)
+		return -1;
+
+	for (shift = 0; (1 << shift) < n;)
+		shift++;
+
+	return shift;
+}
+
+static inline int hr_roundup_pow_of_two(int size)
```
rleon: utils/util.h - roundup_pow_of_two()

providers/hns/hns_roce_u.h
```diff
@@ -200,6 +200,7 @@ struct hns_roce_wq {
 	unsigned int			tail;
 	unsigned int			max_gs;
 	int				wqe_shift;
+	int				shift; /* wq size is 2^shift */
```
rleon: Why don't you use "unsigned"?

providers/hns/hns_roce_u_verbs.c
```diff
@@ -269,12 +269,7 @@ static int align_qp_size(int req)
 
 static int align_queue_size(int req)
```
rleon: Please remove this function completely

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -1098,6 +1098,9 @@ static int hns_roce_u_v2_modify_qp(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 
 	ret = ibv_cmd_modify_qp(qp, attr, attr_mask, &cmd, sizeof(cmd));
 
+	if (!ret)
```
rleon: I don't know if it is correct behavior.
Sometimes you need to change QP state to be ERR if ibv_cmd_modify_qp() fails

providers/hns/hns_roce_u.c
```diff
@@ -134,6 +134,14 @@ static struct verbs_context *hns_roce_alloc_context(struct ibv_device *ibdev,
 		}
 	}
 
+	if (hr_dev->hw_version == HNS_ROCE_HW_VER2) {
+		context->reset_state = mmap(NULL, to_hr_dev(ibdev)->page_size,
```
rleon: @jgunthorpe, doesn't it need some kernel patch?

rleon: ohh, i see kernel patch now.

Please don't mix random cleanups with patches which need kernel support

providers/hns/hns_roce_u.h
oulijun: Thanks, we will fix it

providers/hns/hns_roce_u.c
l00436852: Sorry for being late,  as the associated kernel-mode patch has received some suggestions, it is necessary to temporarily abandon the patch until the user-mode and kernel-mode new schemes are determined.

providers/hns/hns_roce_u_hw_v2.c
l00436852: Sorry for being late.
If ibv_cmd_modify_qp() fails，and QP state was changed to be ERR ,  modify_qp_to_err was called again in post_send or post_recv,  it may return fails too, this may be unnecessary because hr_qp->state was just used in post_send or post_recv. 

Thanks.

providers/hns/hns_roce_u.h
Li-Weihang: Hi Leon, we have pushed a v2 patch-set to fix your comments and remove the patch that needs-kernel-patch, could you please review again? Thank you :)

PR #560, opened 2019-08-05
mlx5: Report ODP capabilities for DC transport

yishaih:
Report ODP capabilities for DC through mlx5dv_query_device().

The matching kernel part was already sent to 'rdma-next'.


jgunthorpe:
Kernel series is now very different, resend this.

PR #563, opened 2019-08-12
libibumad: Support 128 devices

haimbo:
Extend the maximum of available InfiniBand devices
by changing the macro from 32 to 128.

### DIFF
--- a/libibumad/umad.h
+++ b/libibumad/umad.h
@@ -62,7 +62,7 @@ union umad_gid {
 	} global;
 } __attribute__((aligned(4))) __attribute__((packed));
 
-#define UMAD_MAX_DEVICES 32
+#define UMAD_MAX_DEVICES 128
 #define UMAD_ANY_PORT	0
 typedef struct ib_mad_addr {
 	__be32 qpn;


libibumad/umad.h
```diff
@@ -62,7 +62,7 @@ union umad_gid {
 	} global;
 } __attribute__((aligned(4))) __attribute__((packed));
 
-#define UMAD_MAX_DEVICES 32
```
jgunthorpe: While you are here please remove this internal constant from the public header.

It would be better to make a better interface than umad_get_cas_names() so we don't have such static limits in the first place.

haimbo: UMAD_MAX_DEVICES is used in other utilities/applications.
it would be problematic to move it from the public header.

jgunthorpe: Mark it an attribute deprecated enum then.

haimbo: Did you mean to deprecated macro, not enum, right ?

jgunthorpe: You can't use attribute deprecated on a #define

haimbo: Should it be done in a separate pull request?

jgunthorpe: A separate patch in this pull request is fine.

haimbo: Do you have an example code how to mark macros as deprecated?
Because there is no a standard/clean way to do it.

jgunthorpe:
```
enum { UMAD_MAX_DEVICES __attribute__((deprecated)) = 32};
```

haimbo: Thanks, your suggestion is good but it is supported from gcc-6 and above.
Attached link:
https://gcc.gnu.org/gcc-6/changes.html

jgunthorpe: Wrap it in a if gcc version test?

haimbo: Or I can do something like this:
enum __attribute__ ((deprecated)) {
        UMAD_MAX_DEVICES = 128
};

or this:
enum {
        UMAD_MAX_DEVICES = 128
} __attribute__ ((deprecated));

jgunthorpe:
This is obsoleted by the other PR, right?

rleon:
No answer for 16 days, I'm closing it.

PR #564, opened 2019-08-13
siw: Change user mmapped CQ notifications flags to 32bit.

BernardMetzler:
Signed-off-by: Bernard Metzler <bmt@zurich.ibm.com>

### DIFF
--- a/providers/siw/siw.c
+++ b/providers/siw/siw.c
@@ -173,7 +173,7 @@ static struct ibv_cq *siw_create_cq(struct ibv_context *ctx, int num_cqe,
 		goto fail;
 	}
 	cq->ctrl = (struct siw_cq_ctrl *)&cq->queue[cq->num_cqe];
-	cq->ctrl->notify = SIW_NOTIFY_NOT;
+	cq->ctrl->flags = SIW_NOTIFY_NOT;
 
 	return &cq->base_cq;
 fail:
@@ -482,13 +482,13 @@ static void siw_async_event(struct ibv_context *ctx,
 static int siw_notify_cq(struct ibv_cq *ibcq, int solicited)
 {
 	struct siw_cq *cq = cq_base2siw(ibcq);
-	atomic_ulong *notifyp = (atomic_ulong *)&cq->ctrl->notify;
 	int rv = 0;
 
 	if (solicited)
-		atomic_store(notifyp, SIW_NOTIFY_SOLICITED);
+		atomic_store((_Atomic(uint32_t) *)&cq->ctrl->flags,
+			SIW_NOTIFY_SOLICITED);
 	else
-		atomic_store(notifyp,
+		atomic_store((_Atomic(uint32_t) *)&cq->ctrl->flags,
 			SIW_NOTIFY_SOLICITED | SIW_NOTIFY_NEXT_COMPLETION);
 	return rv;
 }
--- a/kernel-headers/rdma/siw-abi.h
+++ b/kernel-headers/rdma/siw-abi.h
@@ -180,6 +180,7 @@ struct siw_cqe {
  * to control CQ arming.
  */
 struct siw_cq_ctrl {
-	__aligned_u64 notify;
+	__u32 flags;
+	__u32 pad;
 };
 #endif


jgunthorpe:
This needs a fixes line and cc stable

providers/siw/siw.c
```diff
@@ -482,7 +482,7 @@ static void siw_async_event(struct ibv_context *ctx,
 static int siw_notify_cq(struct ibv_cq *ibcq, int solicited)
 {
 	struct siw_cq *cq = cq_base2siw(ibcq);
-	atomic_ulong *notifyp = (atomic_ulong *)&cq->ctrl->notify;
+	atomic_uint *notifyp = (atomic_uint *)&cq->ctrl->flags;
```
jgunthorpe: Shouldn't this be atomic_uint32_t ?

dledford:
On Tue, 2019-08-13 at 07:51 -0700, Jason Gunthorpe wrote:
> This needs a fixes line and cc stable

I'm confused.  This is rdma-core, what stable do you mean?  And besides,
did we put siw into any prior rdma-core releases?  We *just* put siw
into the upstream kernel, so at worst siw *might* be in the last rdma-
core release, but that would be it.  Going to check...

Ok, siw was in version 25.  I think we put it there expecting siw to
merge last merge window and it didn't, or something like that.  So we do
need the fix to go back to be included in a 25.1 stable release.

Bernard, the proper Fixes: tag is:

Fixes: fb399ae6645c ("Adding Soft-iWarp user library")

As far as Cc: stable, we don't have that.  Instead, make a pull request
against master with that Fixes tag, and then make another pull request
against the version 25 branch (if Nicholas has opened one yet),
otherwise email Nicolas Morey-Chaisemartin <nmoreychaisemartin@suse.com>

-- 
    GPG KeyID: B826A3330E572FDD
    Fingerprint = AE6B 1BDA 122B 23B4 265B  1274 B826 A333 0E57 2FDD

providers/siw/siw.c
BernardMetzler: hmm, there is no such type for my gcc. unsigned int is 4 bytes on 32bit and 64bit arch's. so that would be OK as is? alternatively, we could call it a 'atomic_uint_fast32_t' -- which would be available....but it doesn't appear to be widely used to me.

jgunthorpe: It is spelled  _Atomic(uint32_t) and used widely in rdma-core

jgunthorpe:
Nicholas will take patches with the Cc: stable without extra effort on Bernard's part.

providers/siw/siw.c
BernardMetzler: great, thanks. will change

BernardMetzler:
Thanks Doug and Jason! I amended the commit message accordingly....

rleon:
1. Patch title should start from "siw: ..." to distinguish between different patches.
2. Please try to avoid empty commit messages.

nmorey:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

Just for clarification

On 8/13/19 5:18 PM, Doug Ledford wrote:
> On Tue, 2019-08-13 at 07:51 -0700, Jason Gunthorpe wrote:
> the proper Fixes: tag is:
> 
> Fixes: fb399ae6645c ("Adding Soft-iWarp user library")
> 

Yup this is good and most of the time enough.
I have script that check master for these Fixes commit and ask me to backport them (or not) so they do fall on my radar automatically
It never hurts to notify me though.

> As far as Cc: stable, we don't have that. 
We do have that.
stable@linux-rdma.org redirects to my mailbox and we/I expect the same format as upstream kernel (ie Cc: stable@linux-rdma.org # v15.1 v14)

Instead, make a pull request
> against master with that Fixes tag, and then make another pull request
> against the version 25 branch (if Nicholas has opened one yet),
> otherwise email Nicolas Morey-Chaisemartin <nmoreychaisemartin@suse.com>

Also works too ;) And yes the stable branch for 25 exists.

I'll backport these, as soon as these gets picked for master (or when I'm back from vacation)

Does anyone require a stable release right away ?

Nicolas
-----BEGIN PGP SIGNATURE-----

iQEzBAEBCAAdFiEEQtJThcGhwCuLGxxvgBvduCWYj2QFAl1S+IEACgkQgBvduCWY
j2RgIQgAnqbrI9QWBOw5nli7WPY+n6XMp/Tzr7i5KpLImr0Ml33ggTDEsfl14bUm
bIppvezyioDRmVnwrDMdGPNKbk+RrRRkji93hzGDPC3uz56/C+bghf3u9mbLVsUM
rDBlms/wFhezmnaH0ZW8FXfU68uFyTwmfTgUpg+LyCGgir/OiYPKCp5d6f2MXEUk
7hLaZ+pCuUV8+N3SVI6uN8Pg51acvRAu+/uCR74205utqesgB5q4sL1V47hGr22l
ctVFT4r7T5XR68yRbmqcS1awA19TaucYFDjLKXAfh3EOOi1oUxnGgaUHOazz98V1
FKjo1879kZ2uXJeRzhgrlnvkpBECTg==
=r9SB
-----END PGP SIGNATURE-----

providers/siw/siw.c
BernardMetzler: _Atomic(uint32_t) *notifyp = (_Atomic(uint32_t) *)&cq->ctrl->flags;
gives me a false positive from travis (seem to take the '*' as a multiplication):
>>>providers/siw/siw.c(485,): error : need consistent spacing around '*' (ctx:WxV)<<<

will cast within the atomic_store()

jgunthorpe: Ignore checkpatch warnings that are stupid.

BernardMetzler:
> 
> 
> Please read my request above and remove double SOB.

Sorry for the double-SOB! I also added to the commit message now.

rleon:
SOB should be last.

Thanks

PR #565, opened 2019-08-15
2019 08 15 fixes

galpress:
Hi,

Another round of fixes for the EFA provider.
First patch is a simple removal of unused fields from various structs.
Second patch is fixing an unprotected access to the wrid pool in poll_cq flow by acquiring the wqlock.

The third patch is a bit tricky as it turns out we forgot to fill the send operation type in the TX WQE (always passed 0 - invalid), which the device wrongly treated as send.
We fixed it in a backwards compatible way by changing the zero value to be send, and removed the invalid operation type.

Thanks,
Gal

### DIFF
--- a/providers/efa/efa.h
+++ b/providers/efa/efa.h
@@ -29,7 +29,6 @@ struct efa_context {
 
 struct efa_pd {
 	struct ibv_pd ibvpd;
-	struct efa_context *context;
 	uint16_t pdn;
 };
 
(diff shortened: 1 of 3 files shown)


providers/efa/efa.h
```diff
@@ -115,7 +113,6 @@ struct efa_ah {
 
 struct efa_dev {
 	struct verbs_device vdev;
-	uint8_t abi_version;
```
rleon: Are you sure about this change?

galpress: This field isn't used anywhere, do you see a reason to keep it?

rleon: I don't know, maybe for compatibility between providers and libibverbs.

PR #567, opened 2019-08-19
rdma-core tests infrastructure

noaos:
Currently, rdma-core provides no test suite for users/developers.
This means that for users, there is no way to sanity-check a fresh installation, and for developers, no comfortable API to write tests to be shipped along new features.
This series is changing the way we use pyverbs for tests and provides an easy-to-use classes and methods to make tests development fast. 
A testing.md file is available under the Documentation directory, explaining the design and usage in details.
This PR also unifies the output of the current API tests, so when executed in verbose mode, the output will be the same for all tests.


rleon:
Tested CX5 IB and RoCE, it worked.
RXE needs fixes in rdma-core.

Thanks

PR #569, opened 2019-08-23
Document how to setup rxe/siw.

marmistrz:


README.md
```diff
@@ -90,6 +90,25 @@ $ yum install cmake3 ninja-build pandoc
...
+# rdma link add <NAME> type <DRIVER> netdev <DEVICE>
+```
+
+Please note that you need version of `iproute2` recent enough is required for the
+command above to work.
+
+You can use either `ibv_devices` or `rdma link` to verify that the device was
+successfully added.
+
+Use of `rxe_cfg` is deprecated, please use the `rdma` command instead.
```
jgunthorpe: Could you update the rxe_cfg man page as well with this note?

marmistrz: Done.

dledford:
This is probably going to be a short lived set of instructions ;-)

My goal is to have a config file, a sysytemd unit file, and a set of instructions on creating a persistent configuration for rdma-core to enable both siw and rxe devices.  Maybe as soon as today.  This will be how it is done in the background, so still valuable for people doing things manually.

rleon:
Better to fix to autoload missing module :)

bvanassche:
Yes, I think that that message means that I should run the port mapper in user space. Some documentation about what to expect would be helpful. Should the udev rule iwpmd/iwpmd.rules start iwpmd automatically? If so, why did that not happen? If not, should it be documented that iwpmd should be started manually?

jgunthorpe:
The udev rule should start it automatically, it is probably a bug in that it was not updated for SIW, or SIW is incompatible..

I seem to remember we need to use something like Leon's renamer udev utility to reliably inject the is-iwarp-ness to udev?

PR #570, opened 2019-08-26
mlx5: Report ODP capabilities for DC transport

rleon:
Report ODP capabilities for DC through mlx5dv_query_device.

Signed-off-by: Michael Guralnik <michaelgur@mellanox.com>
Signed-off-by: Yishai Hadas <yishaih@mellanox.com>

### DIFF
--- a/providers/mlx5/man/mlx5dv_query_device.3
+++ b/providers/mlx5/man/mlx5dv_query_device.3
@@ -49,6 +49,9 @@ struct mlx5dv_cqe_comp_caps     cqe_comp_caps;
 struct mlx5dv_sw_parsing_caps sw_parsing_caps;
 uint32_t	tunnel_offloads_caps;
 uint32_t        max_dynamic_bfregs /* max blue-flame registers that can be dynamiclly allocated */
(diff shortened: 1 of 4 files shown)


rleon:
I want to repeat it again.
Please DON"T merge code which is not part of official for-next or for-rc.
https://git.kernel.org/pub/scm/linux/kernel/git/rdma/rdma.git/log/?h=for-next

ODP DC code is in wip/testing area yet.

Thanks

yishaih:
There is no direct dependency to any kernel series, the driver reads ODP DC caps directly from the device over DEVX.

In any case, the kernel series that turns on the device ODP caps was already accepted by Jason to his for-next branch.

rleon:
> In any case, the kernel series that turns on the device ODP caps was already accepted by Jason to his for-next branch.

There is no Jason's or Doug's for-next, we are working against official for-next and I posted link above.

Thanks

yishaih:
Agree, let's follow that.

Specifically,  this patch has no direct dependency on the kernel series, it uses the DEVX API to read device CAPs, if the kernel will not set them it will just report 0 as expected.

PR #575, opened 2019-09-10
Support UD on hip08

Li-Weihang:
This patchset enables hip08 to support Unreliable Datagram.

### DIFF
--- a/providers/hns/hns_roce_u.c
+++ b/providers/hns/hns_roce_u.c
@@ -81,6 +81,8 @@ static const struct verbs_context_ops hns_common_ops = {
 	.modify_srq = hns_roce_u_modify_srq,
 	.query_srq = hns_roce_u_query_srq,
 	.destroy_srq = hns_roce_u_destroy_srq,
+	.create_ah = hns_roce_u_create_ah,
+	.destroy_ah = hns_roce_u_destroy_ah,
 };
 
 static struct verbs_context *hns_roce_alloc_context(struct ibv_device *ibdev,
--- a/kernel-headers/rdma/hns-abi.h
+++ b/kernel-headers/rdma/hns-abi.h
@@ -80,4 +80,11 @@ struct hns_roce_ib_alloc_pd_resp {
 	__u32 pdn;
 };
 
+struct hns_roce_ib_create_ah_resp {
+	__u8    dmac[6];
+	__u16   vlan_id;
+	__u8    vlan_en;
+	__u8    reserved[7];
+};
+
 #endif /* HNS_ABI_USER_H */
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -68,6 +68,8 @@
 #define HNS_ROCE_TPTR_OFFSET		0x1000
 #define HNS_ROCE_STATIC_RATE		3 /* Gbps */
 
+#define	ETH_ALEN			6
+
 #define roce_get_field(origin, mask, shift) \
 	(((le32toh(origin)) & (mask)) >> (shift))
 
@@ -244,6 +246,25 @@ struct hns_roce_qp {
 	unsigned long			flags;
 };
 
+struct hns_roce_av {
+	uint8_t				port;
+	uint8_t				gid_index;
+	uint8_t				static_rate;
+	uint8_t				hop_limit;
+	uint32_t			flowlabel;
+	uint8_t				sl;
+	uint8_t				tclass;
+	uint8_t				dgid[HNS_ROCE_GID_SIZE];
+	uint8_t				mac[ETH_ALEN];
+	uint16_t			vlan_id;
+	uint8_t				vlan_en;
+};
+
+struct hns_roce_ah {
+	struct ibv_ah			ibv_ah;
+	struct hns_roce_av		av;
+};
+
 struct hns_roce_u_hw {
 	uint32_t hw_version;
 	struct verbs_context_ops hw_ops;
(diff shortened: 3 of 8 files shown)


providers/hns/hns_roce_u.c
```diff
@@ -85,6 +87,17 @@ static const struct verbs_context_ops hns_common_ops = {
 	.destroy_ah = hns_roce_u_destroy_ah,
 };
 
+static bool get_param_config(const char *var)
+{
+	char *env;
+
+	env = getenv(var);
```
rleon: Didn't we talk in ML that environmental variable is not the best way to configure library?

Li-Weihang: Hi leon,
Yes, you have point that.But could you please tell me

Li-Weihang: Hi leon,
Yes, you have pointed that, but sorry for failing to find another way to configure library. Could you please share more information? Thank you :)

rleon: In order to answer this question properly, I need to understand why and what do you mean by setting loopback to enable/disable.

First, we have IBV_QP_CREATE_BLOCK_SELF_MCAST_LB, second we have mlx5dv_create_qp() for more granular solution, third we have loopback in librdmacm,
https://github.com/linux-rdma/rdma-core/blob/master/Documentation/librdmacm.md

oulijun: Hi, leon Romanovsky
   This is the only version we currently think of that supports loopback.  In fact, I have always had a question for loopback.

if the rdma device want to supports loopback capability, it should be set in QPC(RC and other transport) or WQE(UD) for user.
I think that the indicator bit should expose to the user.  Also, I am learning the IB protocol chapter 11.
For example:
the chapter 11.2.2.1 CREATE ADDRESS HANDLE
 ...
 ...
 ...

• Address vector, for UD transports only, containing:
• Service level.
• Send Global Routing Header Flag.
• Destination LID. If destination is in same subnet, LID = final
destination; otherwise LID = router LID. DLID = 0xFFFF is illegal
except when the destination QP is QP0.
• Loopback Indicator (if the HCA supports Loopback Indicator
on Address Handles). DLID and Loopback Indicator are mutually
exclusive.

I think that the ULP should provider a paramter loopback for user.  Is this understood?
However, I didn't find it in anywhere.

oulijun: Because hipxx hardware set the loopback indicator bit in QPC and UD WQE. When the below condition established, the hardware will generate loopback packet in mac level.
1. smac == dmac
2. dgid == sgid
3. or the QPC.loopback indicator bit will be set for RC or UDWQE.loopback indicator bit will be set for UD

rleon: First, you need to provide loopback info through query HCA.

> o10-6.2-1.1: If the CI supports the Loopback Indicator, the CI shall report it
> through the Query HCA Verb. The CI shall accept the Loopback Indicator
> as an input modifier to the Create Address Handle, Modify Address 
> Handle, Modify QP and Modify EEC verbs. The CI shall support the Loopback
> Indicator as an output modifier of the Query QP, Query EEC and Poll
> for Completion verbs.

IMHO, providers don't support this indicator, because the value is little.

providers/hns/hns_roce_u.c
```diff
@@ -103,6 +118,8 @@ static struct verbs_context *hns_roce_alloc_context(struct ibv_device *ibdev,
 				&resp.ibv_resp, sizeof(resp)))
 		goto err_free;
 
+	loopback = get_param_config("HNS_ROCE_LOOPBACK");
```
jgunthorpe: A global for loopback? this is crazy, no, figure out something else.

providers/hns/hns_roce_u.c
oulijun: Hi, leon
   I am learned the IB protocol.  In protocol 11.2.1.2 QUERY HCA, Clearly mentioned support for query loopback. However, we can't find the output definition for ib_device_attr in ib_query_device verbs.  Besides, the wc have no the loopback member  when poll cq.
As a result, how to tell user the hw generate a loopback packet?

  I am also learned the  usage of IBV_QP_CREATE_BLOCK_SELF_MCAST_LB. I think that it is indicator the hw support multicast loopback. But, the hw is only support unicast and need to
support loopback,  should it be?
  At the same time I found, the FL of mlx wqe will be set when smac is equal to dmac.  I think it  is in violation of the IB agreement. Because mac is equal and loopback is the two equal conditions to determine whether to loop back？
Maybe I understand it wrong. I don't understand any other setup methods at the moment.

rleon: Why existing solutions don't work for you?

oulijun: From the current analysis, I still can't understand the tree methods you suggested can solve.
As use IBV_QP_CREATE_BLOCK_SELF_MCAST_LB, it should be support multicast. however, some hardwares only support unicast.
if use the third way, I think it is global. Is this method consistent with setting input parameters?

Li-Weihang: Hi Leon,
Thank you for your help. We have removed the patch about loopback from this patchset, and we will keep working on it. Do you have more suggestions on other patches?

rleon: Please close this PR and start a new one, it is hard to see the changes now.
The kernel-headers files should be updated with kernel-header/update script.

PR #576, opened 2019-09-10
EFA RDMA read support

galpress:
Support RDMA operations through the new verbs send API and report RDMA capabilities through the query device direct verb.


galpress:
Hi,
I've pushed the series so we can start the review process for these patches.
Please don't take it yet, I'll let you know when it can be merged.

PR #581, opened 2019-09-19
Use physical resource ids from device

aditr:
These patches allow the vmw_pvrdma driver to expose different values for qp handle and qp num to userspace.

### DIFF
--- a/providers/vmw_pvrdma/pvrdma-abi.h
+++ b/providers/vmw_pvrdma/pvrdma-abi.h
@@ -55,7 +55,7 @@ DECLARE_DRV_CMD(user_pvrdma_alloc_pd, IB_USER_VERBS_CMD_ALLOC_PD,
 DECLARE_DRV_CMD(user_pvrdma_create_cq, IB_USER_VERBS_CMD_CREATE_CQ,
 		pvrdma_create_cq, pvrdma_create_cq_resp);
 DECLARE_DRV_CMD(user_pvrdma_create_qp, IB_USER_VERBS_CMD_CREATE_QP,
-		pvrdma_create_qp, empty);
+		pvrdma_create_qp, pvrdma_create_qp_resp);
 DECLARE_DRV_CMD(user_pvrdma_create_srq, IB_USER_VERBS_CMD_CREATE_SRQ,
 		pvrdma_create_srq, pvrdma_create_srq_resp);
 DECLARE_DRV_CMD(user_pvrdma_alloc_ucontext, IB_USER_VERBS_CMD_GET_CONTEXT,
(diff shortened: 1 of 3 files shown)


rleon:
First patch should be created with kernel-headers/update script

aditr:
> First patch should be created with kernel-headers/update script

Sorry I wasn't aware of that. I just ran the script. Hmm, given that we haven't synced from the RDMA for-next branch recently, I'm getting a new file rvt-abi.h included in my commit. I'm guessing that's expected?

rleon:
Yes, you should take everything generated by script.

aditr:
@jgunthorpe Do you want me to resend this now that cxgb3 is removed?

rleon:
You need to clean your PR to have only two commits: one is update kernel headers to real SHA1 (in your current patch it is "To commit ?? ...") and second is your resource id patch.

Thanks

aditr:
> You need to clean your PR to have only two commits: one is update kernel headers to real SHA1 (in your current patch it is "To commit ?? ...") and second is your resource id patch.

Okay but I thought you should use the real SHA1 from when the kernel patch was committed? It hasn't been so I was asking Jason.

> 
> Thanks

rleon:
> > You need to clean your PR to have only two commits: one is update kernel headers to real SHA1 (in your current patch it is "To commit ?? ...") and second is your resource id patch.
> 
> Okay but I thought you should use the real SHA1 from when the kernel patch was committed? It hasn't been so I was asking Jason.
> 
> > Thanks

The idea is to create PR with wrong "update kernel headers" patch in parallel to kernel submission to easy review and once kernel part is accepted to update kernel header patch.

Take a look at how Yishai is doing that.

aditr:
> > > You need to clean your PR to have only two commits: one is update kernel headers to real SHA1 (in your current patch it is "To commit ?? ...") and second is your resource id patch.
> > 
> > 
> > Okay but I thought you should use the real SHA1 from when the kernel patch was committed? It hasn't been so I was asking Jason.
> > > Thanks
> 
> The idea is to create PR with wrong "update kernel headers" patch in parallel to kernel submission to easy review and once kernel part is accepted to update kernel header patch.
> 
> Take a look at how Yishai is doing that.

I looked at PR 543. I can push a force update with the right SHA1 once Jason says he is okay with the kernel patch. That seems to be the right approach to me.

jgunthorpe:
I  usually fix the kernel headers myself and force push, but it is always good if you beat me too it.

aditr:
> I usually fix the kernel headers myself and force push, but it is always good if you beat me too it.

Okay. I updated the kernel headers patch now that the kernel code is in the for-next branch.

jgunthorpe:
The first two patches should be squashed

aditr:
> The first two patches should be squashed
Okay, they are squashed now.

PR #582, opened 2019-09-22
README.md: Add Amazon Linux instructions to building section

galpress:
Building rdma-core on Amazon Linux 1/2 is similar to CentOS, state that
in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ NOTE: Fedora Core uses the name 'ninja-build' for the 'ninja' command.
 $ zypper install cmake gcc libnl3-devel libudev-devel ninja pkg-config valgrind-devel python3-devel python3-Cython
 ```
 
-## Building on CentOS 6/7
+## Building on CentOS 6/7, Amazon Linux 1/2
 
 Install required packages:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,22 +71,31 @@ NOTE: Fedora Core uses the name 'ninja-build' for the 'ninja' command.
...
 
 ```sh
 $ yum install epel-release
 $ yum install cmake3 ninja-build pandoc
 ```
 
+Amazon Linux 2:
+
+```sh
+$ amazon-linux-extras install epel
```
jgunthorpe: I'm not so excited about having instructions for a distro that isn't part of cbuild, can you add this thing to cbuild?

galpress: I might be wrong but cbuild works with containers, right?
You need to create an EC2 instance in order to run Amazon Linux.

rleon: From what I see Amazon provides the container to emulate EC2 instance, but documentation mentioned password, so unclear if we can use AWS containers.

https://docs.aws.amazon.com/AmazonECR/latest/userguide/amazon_linux_container_image.html

galpress: You're right, it seems Amazon Linux publishes official images https://hub.docker.com/_/amazonlinux.

The password in the link you sent is for customers who wish to pull the images from Amazon ECR, but we can just pull them from Docker Hub.

jgunthorpe: Should be easy enough to add it to cbuild then, all our other OS's come from docker hub already

galpress: Thanks, I will give it a shot.

rleon: Gal?

galpress: Didn't forget, I was OOO for a while 😃.

galpress:
I'm closing this PR for now, will push a new one when it's ready.

PR #586, opened 2019-09-26
General fixes

noaos:
This pull request fixes a few issues found in rdma-core:
1. In UD pingpong, change the default MTU value to be smaller than the default MTU for Ethernet devices, to avoid an error.
2. Fix a wrong assignment done in pyverbs.
3. Add missing errno reports in mlx4 (CQ creation, QP creation and AH creation).

### DIFF
--- a/providers/mlx4/verbs.c
+++ b/providers/mlx4/verbs.c
@@ -473,8 +473,10 @@ static struct ibv_cq_ex *create_cq(struct ibv_context *context,
 		return NULL;
 	}
 
-	if (cq_attr->wc_flags & ~CREATE_CQ_SUPPORTED_WC_FLAGS)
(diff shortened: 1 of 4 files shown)


providers/mlx4/verbs.c
```diff
@@ -473,8 +473,10 @@ static struct ibv_cq_ex *create_cq(struct ibv_context *context,
 		return NULL;
 	}
 
-	if (cq_attr->wc_flags & ~CREATE_CQ_SUPPORTED_WC_FLAGS)
+	if (cq_attr->wc_flags & ~CREATE_CQ_SUPPORTED_WC_FLAGS) {
+		errno = ENOTSUP;
```
rleon: don't like it, but it is how mlx4 is written

PR #587, opened 2019-09-29
Support UD on hip08

Li-Weihang:
This patchset enables hip08 to support Unreliable Datagram.

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -69,6 +69,9 @@
 #define HNS_ROCE_TPTR_OFFSET		0x1000
 #define HNS_ROCE_STATIC_RATE		3 /* Gbps */
 
+#define HNS_ROCE_ADDRESS_MASK		0xFFFFFFFF
+#define HNS_ROCE_ADDRESS_SHIFT		32
+
 #define roce_get_field(origin, mask, shift) \
 	(((le32toh(origin)) & (mask)) >> (shift))
 
@@ -203,6 +206,13 @@ struct hns_roce_wq {
 	unsigned int			wqe_shift;
 	unsigned int			shift; /* wq size is 2^shift */
 	int				offset;
+};
+
+/* record the result of sge process */
+struct hns_roce_sge_info {
+	unsigned int		valid_num; /* sge length is not 0 */
+	unsigned int		start_idx; /* start position of extend sge */
+	unsigned int		total_len; /* total length of valid sges */
 };
 
 struct hns_roce_sge_ex {
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -42,6 +42,7 @@
 #include <infiniband/verbs.h>
 #include <ccan/bitmap.h>
 #include <ccan/container_of.h>
+#include <linux/if_ether.h>
 
 #define HNS_ROCE_HW_VER1		('h' << 24 | 'i' << 16 | '0' << 8 | '6')
 
@@ -244,6 +245,25 @@ struct hns_roce_qp {
 	unsigned long			flags;
 };
 
+struct hns_roce_av {
+	uint8_t				port;
+	uint8_t				gid_index;
+	uint8_t				static_rate;
+	uint8_t				hop_limit;
+	uint32_t			flowlabel;
+	uint8_t				sl;
+	uint8_t				tclass;
+	uint8_t				dgid[HNS_ROCE_GID_SIZE];
+	uint8_t				mac[ETH_ALEN];
+	uint16_t			vlan_id;
+	uint8_t				vlan_en;
+};
+
+struct hns_roce_ah {
+	struct ibv_ah			ibv_ah;
+	struct hns_roce_av		av;
+};
+
 struct hns_roce_u_hw {
 	uint32_t hw_version;
 	struct verbs_context_ops hw_ops;
@@ -278,6 +298,11 @@ static inline struct hns_roce_srq *to_hr_srq(struct ibv_srq *ibv_srq)
 static inline struct  hns_roce_qp *to_hr_qp(struct ibv_qp *ibv_qp)
 {
 	return container_of(ibv_qp, struct hns_roce_qp, ibv_qp);
+}
+
+static inline struct hns_roce_ah *to_hr_ah(struct ibv_ah *ibv_ah)
+{
+	return container_of(ibv_ah, struct hns_roce_ah, ibv_ah);
 }
 
 int hns_roce_u_query_device(struct ibv_context *context,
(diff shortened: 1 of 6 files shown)


providers/hns/hns_roce_u_hw_v1.c
```diff
@@ -471,19 +470,20 @@ static int hns_roce_u_v1_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
...
 	for (nreq = 0; wr; ++nreq, wr = wr->next) {
 		if (hns_roce_wq_overflow(&qp->sq, nreq,
 					 to_hr_cq(qp->ibv_qp.send_cq))) {
 			ret = -1;
 			*bad_wr = wr;
 			goto out;
 		}
+
+		wqe_idx = (qp->sq.head + nreq) & (qp->sq.wqe_cnt - 1);
+
```
rleon: I didn't understand your problem and solution, but a couple of lines above you have hns_roce_wq_overflow() function which supposes to protect from overflow.

rleon: plus the "a %= b" construction is the better way to write it.

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -637,6 +637,109 @@ static void set_sge(struct hns_roce_v2_wqe_data_seg *dseg,
...
+		     (wr->send_flags & IBV_SEND_SIGNALED) ? 1 : 0);
+	roce_set_bit(ud_sq_wqe->rsv_opcode, UD_SQ_WQE_SE_S,
+		     (wr->send_flags & IBV_SEND_SOLICITED) ? 1 : 0);
+	roce_set_bit(ud_sq_wqe->rsv_opcode, UD_SQ_WQE_OWNER_S,
+		     ~(((qp->sq.head + nreq) >> qp->sq.shift) & 0x1));
+
+	switch (wr->opcode) {
+	case IBV_WR_SEND:
+		hr_op = HNS_ROCE_WQE_OP_SEND;
+		ud_sq_wqe->immtdata = 0;
```
rleon: You cleared it a second before.

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -609,28 +609,194 @@ static int hns_roce_u_v2_arm_cq(struct ibv_cq *ibvcq, int solicited)
...
+		    struct hns_roce_qp *qp, struct ibv_send_wr *wr,
+		    struct hns_roce_sge_info *sge_info)
+{
+	int i;
+
+	sge_info->valid_num = 0;
+	sge_info->total_len = 0;
+
+	for (i = 0; i < wr->num_sge; i++)
+		if (likely(wr->sg_list[i].length)) {
```
rleon: success-oriented flow, please

if (unlikely(,,,))
 continue;

..

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -609,28 +609,194 @@ static int hns_roce_u_v2_arm_cq(struct ibv_cq *ibvcq, int solicited)
...
+	if (wr->opcode == IBV_WR_ATOMIC_FETCH_AND_ADD ||
+	    wr->opcode == IBV_WR_ATOMIC_CMP_AND_SWP) {
+		dseg++;
+		set_atomic_seg((struct hns_roce_wqe_atomic_seg *)dseg, wr);
+		return 0;
+	}
+
+	if (wr->send_flags & IBV_SEND_INLINE) {
+		if (wr->opcode == IBV_WR_RDMA_READ) {
+			fprintf(stderr, "Not supported inline data!\n");
```
rleon: No fprints in libraries

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -609,28 +609,194 @@ static int hns_roce_u_v2_arm_cq(struct ibv_cq *ibvcq, int solicited)
...
+			return -EINVAL;
+		}
+
+		if (sge_info->total_len > qp->max_inline_data) {
+			fprintf(stderr,
+				"Failed to inline, data len=%d, max inline len=%d!\n",
```
rleon: ditto

providers/hns/hns_roce_u.h
```diff
@@ -68,6 +68,8 @@
 #define HNS_ROCE_TPTR_OFFSET		0x1000
 #define HNS_ROCE_STATIC_RATE		3 /* Gbps */
 
+#define	ETH_ALEN			6
```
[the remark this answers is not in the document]
Li-Weihang: OK, will modify, thank you.

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: OK, will remove it, thank you.

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: OK, thank you for your advice :)

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: Hi Leon,
Do you means that we should print informations only when debug in libraries?
I also saw some fprintf in other drivers, could you please give me some suggestions on how to tell the users when an error occurs?

providers/hns/hns_roce_u_hw_v1.c
eason-liuyx: > I didn't understand your problem and solution, but a couple of lines above you have hns_roce_wq_overflow() function which supposes to protect from overflow.
Hi, Leon, thanks. You are right, the commit message is uncorrect, I will rewrite it.

eason-liuyx: > I didn't understand your problem and solution, but a couple of lines above you have hns_roce_wq_overflow() function which supposes to protect from overflow.

Hi, Leon, thanks. You are right, the commit message is uncorrect, I will rewrite it.

eason-liuyx: > plus the "a %= b" construction is the better way to write it.

Hi Leon, do you mean: 
wqe_idx = (qp->sq.head + nreq)  %  qp->sq.wqe_cnt;

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: Hi Leon,
Do you means that we should print informations only when debug in libraries? Or it's not reasonable to print message here in inline mode?
I saw some fprintf in other drivers, should I replace all printf/fprintf in hns with a macro that only show message when debug?

Li-Weihang: > No fprints in libraries

Hi Leon,
Do you means that we should print informations only when debug in libraries? Or it's not reasonable to print message here in inline mode?
I saw some fprintf in other drivers, should I replace all printf/fprintf in hns with a macro that only show message when debug?

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: > ditto

Hi Leon,
I have modified code as suggested and pushed a new version, could you please review it again? Thank you :)

PR #588, opened 2019-10-02
kernel-boot: Tighten check if device is virtual

rleon:
Virtual devices like SIW or RXE don't set FW version because
they don't have one, use that fact to rely on having empty
fw_ver file to sense such virtual devices.

Such change is needed to ensure that virtual devices which are
attached to real hardware won't be renamed, because during
device attachment, user already supplied desired name.

### DIFF
--- a/kernel-boot/rdma_rename.c
+++ b/kernel-boot/rdma_rename.c
@@ -356,9 +356,12 @@ err_free:
 
 static int by_pci(struct data *d)
 {
+	bool is_virtual = false;
 	struct pci_info p = {};
 	char *subsystem;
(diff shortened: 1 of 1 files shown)


galpress:
Assuming that any device that doesn’t report its firmware version is a virtual device isn’t necessarily true, EFA is an example.

rleon:
> Assuming that any device that doesn’t report its firmware version is a virtual device isn’t necessarily true, EFA is an example.

You read my mind, I answered on the mailing list.

rleon:
I'll rewrite and reopen

PR #600, opened 2019-10-30
Pyverbs fixes

jgunthorpe:
Various fixes to the use of cython and the pyverbs test runner.

### DIFF
--- a/buildlib/pyverbs_functions.cmake
+++ b/buildlib/pyverbs_functions.cmake
@@ -14,6 +14,7 @@ function(rdma_cython_module PY_MODULE LINKER_FLAGS)
     include_directories(${PYTHON_INCLUDE_DIRS})
     add_custom_command(
       OUTPUT "${CFILE}"
+      MAIN_DEPENDENCY "${PYX}"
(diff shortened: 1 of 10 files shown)


buildlib/pyverbs_functions.cmake
```diff
@@ -9,6 +9,7 @@ function(rdma_cython_module PY_MODULE)
     include_directories(${PYTHON_INCLUDE_DIRS})
     add_custom_command(
       OUTPUT "${CFILE}"
+      MAIN_DEPENDENCY "${PYX}"
```
noaos: What does this change? I mean, cython will fail if you don't provide pyx files anyway.

noaos: I usually delete the build/ directory before I call ninja, so I wasn't aware.

jgunthorpe: Don't do things like that, it hides defects like this that effect the rest of us.

rleon:
Jason,
Can you please rebase so we will merge it?

PR #601, opened 2019-10-31
Build update

jgunthorpe:
Remove travis and add centos 8 to azp


nmorey:
@jgunthorpe Sorry. I missed your review request.

I'll pick this for v26.

Have you been able to setup a pipeline for your private repo ?
The basic setup seems easy enough but it seems to fails to get the distro images:
```
The pipeline is not valid. A service connection with name ucfconsort_registry could not be found.
The service connection does not exist or has not been authorized for use.
For authorization details, refer to https://aka.ms/yamlauthz.
A service connection with name ucfconsort_registry could not be found.
The service connection does not exist or has not been authorized for use. For authorization details, refer to https://aka.ms/yamlauthz.
```

nmorey:
It also seems that azure does not build stable branches (at least I cannot see this commit I just pushed on v26

jgunthorpe:
It is a missing trigger for the stable branches, we can fix it

Private repos.. These are possible, but it is quite an annoying process to authorize the service connection to the docker registry. My advice is to try to use the normal repo, otherwise we will have to go through the steps

nmorey:
OK. I'll do that. All branches are already dev/stable-XXX prefixed to it should not cause any issue

PR #602, opened 2019-11-01
libhns: Use syslog for debugging while no print by default

Li-Weihang:
There should be no fprintf/printf in libraries by default unless
debugging. So replace all fprintf/printf in libhns with a macro that is
controlled by HNS_ROCE_DEBUG.
This patch also standardizes all printtings to maintain a uniform style.

Signed-off-by: Weihang Li <liweihang@hisilicon.com>


Li-Weihang:
Will use debug file for printings like other providers and create a new PR.

PR #607, opened 2019-11-04
libhns: Use log file for debugging while no print by default

Li-Weihang:
There should be no fprintf/printf in libraries by default unless
debugging. So replace all fprintf/printf in libhns with a macro that is
controlled by HNS_ROCE_DEBUG.
Users can specify a log file for debugging by setting an environment
variable "HR_DEBUG_FILE", otherwice printings will be redirected to
stderr.

This patch also standardizes all printtings to maintain a uniform style.

Signed-off-by: Weihang Li <liweihang@hisilicon.com>

### DIFF
--- a/providers/hns/hns_roce_u.c
+++ b/providers/hns/hns_roce_u.c
@@ -83,6 +83,31 @@ static const struct verbs_context_ops hns_common_ops = {
 	.destroy_srq = hns_roce_u_destroy_srq,
 };
 
+static void open_debug_file(struct hns_roce_context *ctx)
+{
+	char *env;
+
+	env = getenv("HR_DEBUG_FILE");
+	if (!env) {
+		ctx->dbg_fp = stderr;
+		return;
+	}
+
+	ctx->dbg_fp = fopen(env, "aw+");
+	if (!ctx->dbg_fp) {
+		fprintf(stderr, "Failed to open debug file %s, using stderr\n",
+			env);
+		ctx->dbg_fp = stderr;
+		return;
+	}
+}
+
+static void close_debug_file(struct hns_roce_context *ctx)
+{
+	if (ctx->dbg_fp && ctx->dbg_fp != stderr)
+		fclose(ctx->dbg_fp);
+}
+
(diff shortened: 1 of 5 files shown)


prime-zeng:
What about taking linux syslog mechanism which is more general and flexible?

Li-Weihang:
> What about taking linux syslog mechanism which is more general and flexible?

Hi, thank you.
Leon suggest me to use a unified approach to see debug messages from rdma-core, so I modifed them from using syslog to using debug file controled by users.

prime-zeng:
what do you mean  "I modifed them from using syslog to using debug file controled by users. "?   i just 
 see you are using printf/fprintf now.

prime-zeng:
Refer to the link below, this is what i mean syslog but not printf/fprintf.    https://github.com/linux-rdma/rdma-core/blob/f5fb4c37b23431007a74345224ead825cd6f9a26/rdma-ndd/rdma-ndd.c#L87

Li-Weihang:
> what do you mean "I modifed them from using syslog to using debug file controled by users. "? i just
> see you are using printf/fprintf now.

Sorry for the misleading, I sent a PR which use syslog to replace printf/fprintf before, you can see it at:
https://github.com/linux-rdma/rdma-core/pull/602
Leon comments on it in mailing-list, you can refer to the link below:
https://patchwork.kernel.org/patch/11222269/
So I made a new PR which use files instead.

prime-zeng:
@rleon  i think what you expected is an unique solution from rdma-core's perspective (at least applicable to all providers), right?

rleon:
> @rleon i think what you expected is an unique solution from rdma-core's perspective (at least applicable to all providers), right?

Yes, this is what I wanted.
It is a little bit harsh to expect from users to configure and see debug output differently between providers.

rleon:
and custom mlx5/qedr debug variable can easily go, because it is for debug and not API.

Li-Weihang:
@prime-zeng @rleon I understand what you mean this time :) We need a unified API in rdma-core which is applicable for all providers for users to configure and see debug output.
I will work on it, thank you for your suggestions.

PR #611, opened 2019-11-07
mlx5: Add support for bulk counters in the DR API over root 

yishaih:
Flow counters can be allocated/freed one by one, or as a consecutive range of counters - bulk.

When the counters are allocated as a bulk, each of the counters can be assigned to a separate steering action. Members of this counter range are addressed by a base counter ID and an offset.

Add the above support to work over root/kernel flows when the DR flow counters API is used.


jgunthorpe:
This needs to be rebased

yishaih:
series was rebased.

PR #612, opened 2019-11-10
man: Fix return value for ibv_reg_dm_mr

noaos:
ibv_reg_dm_mr returns a struct ibv_mr pointer while man page defined
it as returning an int. Fix the description.

Fixes: e1cebbf50c262 ('verbs: Add device memory (DM) support in libibverbs')


noaos:
Done

PR #613, opened 2019-11-12
EFA RDMA read

galpress:
Support RDMA operations through the new verbs send API and report RDMA capabilities through the query device direct verb.

### DIFF
--- a/providers/efa/man/efadv_query_device.3.md
+++ b/providers/efa/man/efadv_query_device.3.md
@@ -37,6 +37,8 @@ struct efadv_device_attr {
 	uint16_t max_rq_sge;
 	uint16_t inline_buf_size;
 	uint8_t reserved[2];
+	uint32_t device_caps;
+	uint32_t max_rdma_size;
 };
 ```
 
@@ -60,6 +62,15 @@ struct efadv_device_attr {
 
 *inline_buf_size*
 :	Maximum inline buffer size.
+
+*device_caps*
+:	Bitmask of device capabilities:
+
+	EFADV_DEVICE_ATTR_CAPS_RDMA_READ:
+		RDMA read is supported.
+
+*max_rdma_size*
+:	Maximum RDMA transfer size in bytes.
 
 # RETURN VALUE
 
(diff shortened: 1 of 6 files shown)


jgunthorpe:
There are a couple checkpatch errors that should be fixed

providers/efa/man/efadv_query_device.3.md
```diff
@@ -61,6 +63,13 @@ struct efadv_device_attr {
 *inline_buf_size*
 :	Maximum inline buffer size.
 
+*max_wr_rdma_sge*
```
jgunthorpe: Isn't this a standard query already?

galpress: According to the documentation:

>         int                     max_sge_rd;             /* Maximum number of s/g per WR for RD QPs */


This isn't quite an RD QP.

galpress:
> There are a couple checkpatch errors that should be fixed

The two occurrences are line break warnings (a bit over 80), which I think will hurt the readability.
Your call if you prefer the line break 😃

providers/efa/man/efadv_query_device.3.md
jgunthorpe: max_sge is the max number of SGEs usable for all QPs types for everything but RDMA READ
max_sge_rd is the max number of SGEs usable for all QP types for RDMA READ

Where did you find that wrong comment?

jgunthorpe:
Those are just function decls, it is expected to wrap them. Touch the lines with clang-format and it will do a reasonable job.

providers/efa/man/efadv_query_device.3.md
galpress: Ah, shouldn't have trust anything but the man in the repo..
https://linux.die.net/man/3/ibv_query_device

Thanks, I will remove this field.

galpress:
> Those are just function decls, it is expected to wrap them. Touch the lines with clang-format and it will do a reasonable job.

Will do.

galpress:
Added RDMA read support report to the DV.

galpress:
Is this ready to merge?

rleon:
Gal, please update first patch and I'll merge it.

Thanks

galpress:
I assume you meant the second patch (kernel headers), done.
Thanks Leon

PR #614, opened 2019-11-13
Misc build updates

galpress:
Add EFA to redhat's udev rules and add an explicit udev trigger.
Cbuild is modified to support for Amazon Linux 1 & 2 packaging.

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -170,6 +170,40 @@ class centos7_epel(centos7):
         res.lines.append("RUN ln -s /usr/bin/cmake3 /usr/local/bin/cmake && ln -sf /usr/bin/python3.4 /usr/bin/python3");
         return res;
 
+class amazonlinux1(YumEnvironment):
+    docker_parent = "amazonlinux:1";
+    pkgs = {
+        'cmake',
+        'gcc',
+        'libnl3-devel',
+        'libudev-devel',
+        'make',
+        'pkgconfig',
+        'python',
+        'python-argparse',
+        'python27-docutils',
+        'rpm-build',
+        'valgrind-devel',
+    };
+    name = "amazonlinux1";
+    use_make = True;
+    pandoc = False;
+    build_pyverbs = False;
+    specfile = "buildlib/centos6.spec";
+    python_cmd = "python";
+    to_azp = False;
+
+class amazonlinux2(YumEnvironment):
+    docker_parent = "amazonlinux:2";
+    pkgs = centos7.pkgs;
+    name = "amazonlinux2";
+    use_make = True;
+    pandoc = False;
+    build_pyverbs = False;
+    specfile = "redhat/rdma-core.spec";
+    python_cmd = "python";
+    to_azp = False;
+
 class centos8(Environment):
     docker_parent = "centos:8";
     pkgs = {
@@ -451,6 +485,8 @@ environments = [centos6(),
                 centos7(),
                 centos7_epel(),
                 centos8(),
+                amazonlinux1(),
+                amazonlinux2(),
                 xenial(),
                 bionic(),
                 jessie(),
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -652,7 +652,8 @@ if "check_output" in dir(subprocess):
     for ln in subprocess.check_output(["rpmspec","-P",{tspec_file!r}]).splitlines():
        if ln.startswith(b"Source:"):
           tarfn = ln.strip().partition(b' ')[2].strip();
-    os.symlink({tarfn!r},os.path.join(b"SOURCES",tarfn));
+    if tarfn != {tarfn!r}:
+        os.symlink({tarfn!r},os.path.join(b"SOURCES",tarfn));
 """.format(passwd=":".join(str(I) for I in pwd.getpwuid(os.getuid())),
            group=":".join(str(I) for I in grp.getgrgid(os.getgid())),
            uid=os.getuid(),
--- a/redhat/rdma.udev-rules
+++ b/redhat/rdma.udev-rules
@@ -10,3 +10,4 @@ SUBSYSTEM=="module", KERNEL=="mlx*", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_
 SUBSYSTEM=="module", KERNEL=="iw_*", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
 SUBSYSTEM=="module", KERNEL=="be2net", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
 SUBSYSTEM=="module", KERNEL=="enic", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
+SUBSYSTEM=="module", KERNEL=="efa", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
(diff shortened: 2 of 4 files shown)


buildlib/cbuild
```diff
@@ -171,6 +171,65 @@ class centos7_epel(centos7):
         res.lines.append("RUN ln -s /usr/bin/cmake3 /usr/local/bin/cmake && ln -sf /usr/bin/python3.4 /usr/bin/python3");
         return res;
 
+class amazonlinux1(YumEnvironment):
+    docker_parent = "amazonlinux:1";
```
jgunthorpe: Do you really need all three? How different is this actually from Centos?

redhat/rdma.udev-rules
```diff
@@ -10,3 +10,4 @@ SUBSYSTEM=="module", KERNEL=="mlx*", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_
 SUBSYSTEM=="module", KERNEL=="iw_*", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
 SUBSYSTEM=="module", KERNEL=="be2net", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
 SUBSYSTEM=="module", KERNEL=="enic", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
+SUBSYSTEM=="module", KERNEL=="efa", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}="rdma.service"
```
jgunthorpe: Does efa also need the stuff in kernel-boot/ to be updated?

buildlib/cbuild
```diff
@@ -652,7 +652,9 @@ if "check_output" in dir(subprocess):
     for ln in subprocess.check_output(["rpmspec","-P",{tspec_file!r}]).splitlines():
        if ln.startswith(b"Source:"):
           tarfn = ln.strip().partition(b' ')[2].strip();
-    os.symlink({tarfn!r},os.path.join(b"SOURCES",tarfn));
+    dst = os.path.join(b"SOURCES",tarfn);
```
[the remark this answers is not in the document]
galpress: Happens when using the centos6 spec file:
`Source: rdma-core.tgz`

It names the tar rdma-core.tgz, and then the script tries to create a symlink with the same name.
Other spec files have the version/git_ver added to the tar name so they work fine.

buildlib/cbuild
galpress: I'm not part of the AmazonLinux distro team so I really can't say how different they are from CentOS.
Why not have all three? Is there a goal to reduce the number of images?

redhat/rdma.udev-rules
galpress: I don't think so.

jgunthorpe: I see, this  rules is just badly designed, it should look more like

ACTION=="remove", GOTO="rdma_ulp_modules_end"
SUBSYSTEM!="infiniband", GOTO="rdma_ulp_modules_end"TAG+="systemd", ENV{SYSTEMD_WANTS}+="rdma.service"

buildlib/cbuild
jgunthorpe: It takes a long time to build all these images for testing purposes, if they are pointless I prefer not to have them all. ie checking the same EPEL modified packages against amazon and c7 is probably redundant.

I'd suggest the latest image only..

buildlib/cbuild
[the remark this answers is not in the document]
jgunthorpe: centos6 doesn't even run this code since its python is so old it lacks 'check_output' 

It this because the aws1 image has a newer python and uses the centos6 spec file? The commit message should be better.

redhat/rdma.udev-rules
galpress: How do you prefer to handle it then?

buildlib/cbuild
galpress: OK, I'll remove amazonlinux2 and keep amazonlinux1 and amazonlix2_epel (which will be renamed to amazonlinux2).

buildlib/cbuild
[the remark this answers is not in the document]
galpress: Exactly, sorry for the confusion.
Do you prefer that I squash this fix with the AmazonLinux commit?

redhat/rdma.udev-rules
galpress: > I see, this rules is just badly designed, it should look more like
> 
> ACTION=="remove", GOTO="rdma_ulp_modules_end"
> SUBSYSTEM!="infiniband", GOTO="rdma_ulp_modules_end"TAG+="systemd", ENV{SYSTEMD_WANTS}+="rdma.service"

I'm not sure this will have the same functionality, as some of the modules listed in this file are part of the network stack.

buildlib/cbuild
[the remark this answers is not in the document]
galpress: Improved the commit message.

buildlib/cbuild
galpress: Ended up removing amazonlinux2_epel.

redhat/rdma.udev-rules
galpress: Left this change untouched for now.

jgunthorpe: The only reason to start rdma.service is if there is a rdma device, so doing it just because a net related module is loaded is already a bug.

RH really needs to fix this stuff. Fedora should be using kernel-boot and it is very unfortunate that RH8 shipped using this junky stuff. @dledford

buildlib/cbuild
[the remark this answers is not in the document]
jgunthorpe: This should be written more like

```
if tarfn != {tarfn!r}:
    os.symlink({tarfn!r}, dst)
```

buildlib/cbuild
```diff
@@ -170,6 +170,41 @@ class centos7_epel(centos7):
...
+        'cmake',
+        'gcc',
+        'libnl3-devel',
+        'libudev-devel',
+        'make',
+        'pkgconfig',
+        'python',
+        'python-argparse',
+        'python-docutils',
+        'python27-docutils',
```
jgunthorpe: Why both?

buildlib/cbuild
```diff
@@ -170,6 +170,41 @@ class centos7_epel(centos7):
...
+    use_make = True;
+    pandoc = False;
+    build_pyverbs = False;
+    specfile = "buildlib/centos6.spec";
+    python_cmd = "python";
+    to_azp = False;
+
+class amazonlinux2(YumEnvironment):
+    docker_parent = "amazonlinux:2";
+    pkgs = amazonlinux1.pkgs | {'systemd-devel'};
```
jgunthorpe: Can this be pkgs = centos7.pkgs ?

buildlib/cbuild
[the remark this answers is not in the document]
galpress: Ack.

redhat/rdma.udev-rules
galpress: Ok, I'll change this commit to your suggested fix.

buildlib/cbuild
galpress: `python-docutils` didn't work so I added the python27 one. I'll test without the former one and change accordingly.

redhat/rdma.udev-rules
jgunthorpe: I suggest to leave it broken and work with RH to get an acceptable outcome, ie use kernel boot or fixup this file.

buildlib/cbuild
galpress: Removed `python-docutils`.

buildlib/cbuild
galpress: Done

galpress:
Kind reminder

PR #619, opened 2019-11-17
pyverbs: Expose additional DV objects

noaos:
This series adds the needed infrastructure to pyverbs to allow users
to create mlx5 CQ and QP.

Tests using the extended CQ were added as well.

### DIFF
--- a/pyverbs/qp.pyx
+++ b/pyverbs/qp.pyx
@@ -853,7 +853,7 @@ cdef class QPAttr(PyverbsObject):
 
 cdef class QP(PyverbsCM):
     def __cinit__(self, object creator not None, object init_attr not None,
-                  QPAttr qp_attr=None):
+                  QPAttr qp_attr=None, **kwargs):
         """
         Initializes a QP object and performs state transitions according to
         user request.
@@ -873,11 +873,16 @@ cdef class QP(PyverbsCM):
                           using Context).
         :param qp_attr: Optional QPAttr object. Will be used for QP state
                         transitions after creation.
+        :param kwargs: Provider-specific QP creation attributes, meaning that
+                       the QP will be created by the provider.
         :return: An initialized QP object
         """
         cdef PD pd
         cdef Context ctx
         self.update_cqs(init_attr)
+        if len(kwargs) > 0:
+            # Leave QP initialization to the provider
+            return
         # In order to use cdef'd methods, a proper casting must be done, let's
         # infer the type.
         if issubclass(type(creator), Context):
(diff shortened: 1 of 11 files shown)


pyverbs/qp.pyx
```diff
@@ -803,11 +803,16 @@ cdef class QP(PyverbsCM):
...
         :param qp_attr: Optional QPAttr object. Will be used for QP state
                         transitions after creation.
+        :param kwargs: Provider-specific QP creation attributes, meaning that
+                       the QP will be created by the provider.
         :return: An initialized QP object
         """
         cdef PD pd
         cdef Context ctx
         self.update_cqs(init_attr)
+        if len(kwargs) > 0:
```
jgunthorpe: So why does this accept a kwargs if it must be 0 length?

noaos: For QP it must be zero. Mlx5QP uses the extra args, and you don't want the underlying C QP to be initialized via ibv_create_qp.

jgunthorpe: Then call the parameter 'do not call ibv_create_qp' and don't use some goofy kwargs

noaos: kwargs allows you to pass more than one extra parameter (we pass Mlx5DVQPInitAttr), why not leave this option for other providers?

jgunthorpe: Why are you passing a Mlx5DvQPInitAttr to the generic code? The signature of the super class constructor does not have to be the same as the subclass constructor.

noaos: If I don't, QP's constructor will go on initializing. We don't want that. We want the inner QP to be initialized by mlx5dv_create_qp. The parent class's constructor is being called implicitly before the child class's.

jgunthorpe: Ugh, no that is another problem. You should not use __cinit__ like that, the subclass should still define a python __init__ and it should still call super().__init__() with modified arguments. __cinit__ should only be used for things that must be done before python's __init__ is called. As the docs say cinit is kind of dangerous, you should avoid using it.

noaos: I'm not sure. Quoting from the docs: "The __cinit__() method is where you should perform basic C-level initialisation of the object, including allocation of any C data structures that your object will own. "

noaos: @jgunthorpe?

rleon: You need to fix.

Thanks

noaos: Please explain as the documentation suggests otherwise.

noaos: Just to make things clear - QP class doesn't know - or care - about Mlx5DVQPInitAttr. A user creates these attributes and passes them inside **kwargs so that QP class only gets a hint that the actual QP creation is done by another class.

jgunthorpe: That is only happening because your constructors are setup wrong. You need to fix the use of __cinit_ and missing __init__ so things work properly. I'm pretty sure when the docs say 'using cinit is dangerous' it means you shouldn't use it without a good reason.

PR #620, opened 2019-11-18
libqedr: Fix doorbell recording compatibility

mkalderon:
The create_qp response structure needs to be zeroed before receiving
data from kernel, otherwise, when running over an old kernel we can get
garbage for db_rec_addr and fail when attempting to map it.

Signed-off-by: Ariel Elior <ariel.elior@marvell.com>

### DIFF
--- a/providers/qedr/qelr_verbs.c
+++ b/providers/qedr/qelr_verbs.c
@@ -720,7 +720,7 @@ struct ibv_qp *qelr_create_qp(struct ibv_pd *pd,
 			      struct ibv_qp_init_attr *attrs)
 {
 	struct qelr_devctx *cxt = get_qelr_ctx(pd->context);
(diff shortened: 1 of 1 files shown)


galpress:
I was thinking about this as well, is this better than making the driver memset `udata->outlen` bytes to zero?

mkalderon:
I think it's a good idea to have the core memset udata->outlen, however, this specific fix is still required as it can run on drivers that don't memset the udata->outlen, so the compatibility fix in this case has to be in rdma-core.

jgunthorpe:
As past kernels already don't zero the fields we have no choice but to do this in userspace.

all the driver handling of udata should be revised to work more like the core code handling in uverbs_cmd.c which is careful to check and zero trailing data.

PR #629, opened 2019-11-28
pyverbs: return correct port number in QPAttr's AH property

noaos:
When using QPAttr's ah_attr getter, a new AHAttr object is created.
Port number wasn't passed to its constructor, so the default value (1)
was used.
Pass the current port number to AHAttr constructor.

Signed-off-by: Ido Kalir <idok@mellanox.com>
Reviewed-by: Noa Osherovich <noaos@mellanox.com>


rleon:
You have a typo in commit message: "AHAhhtr"

noaos:
Fixed, thanks.

rleon:
What about Fixes line?
"pyverbs: return .." -> "pyverbs: Return ..." for MLNX submissions.

noaos:
Also fixed, thanks.

PR #634, opened 2019-12-12
libhns: Refactor post send/recv

Li-Weihang:
These two patches were in previous series "Support UD on hip08"
(PR: https://github.com/linux-rdma/rdma-core/pull/587) which was closed because of the UD issue with hip08.

Although we have decided to not support UD, these two patches is still
significant to simplify codes of post_send/recv.

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -68,6 +68,9 @@
 #define HNS_ROCE_TPTR_OFFSET		0x1000
 #define HNS_ROCE_STATIC_RATE		3 /* Gbps */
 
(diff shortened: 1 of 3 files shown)


providers/hns/hns_roce_u.h
```diff
@@ -206,6 +209,13 @@ struct hns_roce_wq {
 	int				offset;
 };
 
+/* record the result of sge process */
+struct hns_roce_sge_info {
+	unsigned int		valid_num; /* sge length is not 0 */
```
rleon: Please don't do vertical space alignment in newly added code.

Thanks

Li-Weihang: OK, I have modified them and pushed a new version. Thank you.

PR #635, opened 2019-12-14
librdmacm: Use sched_yield instead of pthread_yield

kraj:
glibc redirects pthread_yield to sched_yield additionally we get it
working with musl on linux

Signed-off-by: Khem Raj <raj.khem@gmail.com>


rleon:
Thanks

PR #637, opened 2019-12-17
libhns: Add interface to configure lock-free

Li-Weihang:
In some scenarios, ULP can ensure that there is no concurrency when
processing io, thus lock free can be used to improve performance.


Li-Weihang:
OK, thank you.

PR #639, opened 2019-12-18
pyverbs: Code refactoring

noaos:
The following series refactors pyverbs' objects creation process and
cleans up some redundant tests:
The first patch removes its close_weakrefs() out of the base objects,
as it doesn't use 'self' and is called during teardown, when the
object might not be fully valid.

The second patch switches pyverbs to use Python's init method rather
thab Cython's cinit method, during which the Python object is not yet
valid. Calling rdma-core's objects creation methods can be done in
init just as well.

The last patch removes some unneeded tests and adapts QP creation
tests to consider TSO existance when setting QP caps.

### DIFF
--- a/pyverbs/addr.pyx
+++ b/pyverbs/addr.pyx
@@ -18,7 +18,8 @@ cdef class GID(PyverbsObject):
     """
     GID class represents ibv_gid. It enables user to query for GIDs values.
     """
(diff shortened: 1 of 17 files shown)


rleon:
It works for me.

Thanks

PR #640, opened 2019-12-20
fixes for redhat specific spec file

Honggang-LI:


redhat/rdma-core.spec
```diff
@@ -43,6 +43,7 @@ BuildRequires: perl-generators
 %endif
 
 Requires: dracut, kmod, systemd, pciutils
+Requires: libibverbs-utils%{?_isa} = %{version}-%{release}
```
jgunthorpe: Maybe we should move it to libibverbs-utils, or possibly just delete it? rxe_cfg horrible, broken and obsolete.

redhat/rdma-core.spec
```diff
@@ -108,6 +108,7 @@ RDMA core development libraries and headers.
 
 %package -n infiniband-diags
 Summary: InfiniBand Diagnostic Tools
+Requires: libibumad%{?_isa} = %{version}-%{release}
```
jgunthorpe: I don't understand. Why doesn't rpm build pickup the right shlib dependenciess automtically? 

We certainly don't have a = dependency like this.

Why would we ever need a specific package version requirement when we use symbol versions properly?

At worst should should just be >= some version at the start of the rdma-core era, but I think you'd need to justfiy that with some old libibumad that forces it.

redhat/rdma-core.spec
```diff
@@ -345,39 +345,38 @@ install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/
...
+%post -n rdma-core
+# we ship udev rules, so trigger an update.
+/sbin/udevadm trigger --subsystem-match=infiniband --action=change || true
+/sbin/udevadm trigger --subsystem-match=net --action=change || true
+/sbin/udevadm trigger --subsystem-match=infiniband_mad --action=change || true
+
 %post -n infiniband-diags -p /sbin/ldconfig
 %postun -n infiniband-diags -p /sbin/ldconfig
 
-# libibverbs
```
jgunthorpe: So strange that comments cause things to fail, are you sure? Seems nearly impossible.. No objection to removing them though

Honggang-LI: Yes, I am sure.

$ rpm -qp --scripts librdmacm-26.0-6.el8.x86_64.rpm
postinstall program: /sbin/ldconfig
postuninstall scriptlet (using /bin/sh):


# ibacm
^
Because of the heading '#', 'git commit' ignored the last line.

redhat/rdma-core.spec
Honggang-LI: > Maybe we should move it to libibverbs-utils, or possibly just delete it? rxe_cfg horrible, broken and obsolete.

We can't delete rxe_cfg, because it already get in rhel distros. Redhat has to support as it is in support life cycle. So, move it into libibverbs-utils looks a better choice.

redhat/rdma-core.spec
Honggang-LI: > I don't understand. Why doesn't rpm build pickup the right shlib dependenciess automtically?

rpm build do pickup the right shlib dependency. Because multiple versions libibumad provide identical list capabilities of provides, mixed version packages may be installed when multiple versions of packages are available in yum/dnf repositories. It is terrible for package building and customer support, when mixed version packages installed.

$ rpm -qpR infiniband-diags-28.0-1.fc32.x86_64.rpm | grep umad
libibumad.so.3()(64bit)
libibumad.so.3(IBUMAD_1.0)(64bit)
libibumad.so.3(IBUMAD_1.1)(64bit)

> Why would we ever need a specific package version requirement when we use symbol versions properly?
 
Please see my inline comment.

> At worst should should just be >= some version at the start of the rdma-core era, but I think you'd need to justfiy that with some old libibumad that forces it.

No, ">=' is helpless when multiple versions libibumad available in yum/dnf repository.

redhat/rdma-core.spec
rleon: I was under impression that RHEL manages its own version of spec and rdma-core. For example, what did you do with removed drivers (nes, cxgb3 e.t.c)?

Honggang-LI: Yes, you are right. We (Redhat) do have own version spec files for RHEL.

We also try our best to decrease the difference between upstream spec file with
Fedora/RHEL specific spec file. 

1) When there is some thing wrong with Fedora/RHEL rdma-core, I always test the
upstream version of rdma-core. As I know, most bugs are not Fedora/RHEL specific,
that means such bugs also impacted upstream rdma-core. The difference between 
upstream and Fedora/RHEL is really annoy, as I need pay extra attention to handle
them, when building/debugging with upstream rdma-core.

2) Redhat always build latest upstream release rdma-core for Fedora Rawhide, the
difference between upstream and Fedora may introduce regression issue because of
oversight when building rpm for Fedora Rawhide. Again, the difference may introduce
regression issue again, when backport Fedora version rdma-core for RHEL major release.

jgunthorpe: I thought rxe was tech preview at best in RHEL? Why is there a problem to delete rxe_cfg for tech preview suff?

redhat/rdma-core.spec
jgunthorpe: The shlib dependencies are supposed to be sufficient. If there is a historical problem with them then you need to identify exactly what it is and we can move forward properly. This = dependency is not correct.

redhat/rdma-core.spec
dledford: The actual historical path was that we took rxe to non-tech preview for one release cycle in rhel7 and then reversed that when we saw how many problems it had.  It is now tech preview in the latest rhel7 and rhel8.  Because of that, we can probably justify removing rxe_cfg and forcing an update of the iproute2 package to get the latest rdma tool, but we would need to identify a version of iproute2 that is "feature complete" as far as managing rxe devices is concerned, and then should ideally also create some systemd templates for autostarting rxe on top of tagged net interfaces.

redhat/rdma-core.spec
dledford: It is standard convention in rpm packaging that there be a hard = requirement between subpackages of a single source package.  Usually this is done in each subpackage by making it require the parent package be the same version.  But when you have a cross dependency like this, it makes sense to do it there as well.  That way if you do dnf upgrade rdma-core on the command line, the entire package set gets updated.  We rely on automatic library dependencies for external libraries, but not for same source package libraries.  There are reasons for that (although they are old and well established reasons that I have mostly long since forgotten).  As an example, a new build of the package may have changed the compile parameters in a way that materially effects the library without changing the symbol versions.

In any case, the typical situation is that *all* of the subpackages would have:

Requires: rdma-core${isa} = %{version}-%{release}

Then if a user tries to upgrade a single package like libibumad, it will trigger pulling in the updated rdma-core, which will then trigger all of the other subpackages to be upgraded too.  If we had set that up properly in the first place, this libibumad cross dependency would be redundant.  So it might be worth just fixing it the standard way instead of this way.

jgunthorpe: It certainly makes more sense to do this consistently everywhere if it is RPM policy. Otherwise the shlib packages are supposed to be versioned soley by symbol versions and no RPM version restriction should ever be needed. If this is not true it needs to be explained and delt with.

dledford: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_requiring_base_package

The official policy is that if there are dependencies between base and subpackages, then yes, an explicit requires of the base package in the subpackage on the entire release and version is expected.

jgunthorpe: I don't think that language applies here. The proposed depends is not a depends on the base but a depends on a different sub package (ie a library) which must be fully and correctly versioned using only the symver mechanism or things are fundamentally broken.

dledford: https://docs.fedoraproject.org/en-US/packaging-guidelines/#_explicit_requires

I believe we fall into this category in this instance, in which case we should have the core package requires.  The soname dependency fails in the specific instance that:

1) library is extended in backword compatible manner
2) application builds against later library and uses newer extensions
3) application is then installed on system with downrev library missing extensions

I saw the original Red Hat bug report that made Honggang write this patch, and I think it fell into that category, but I'm going from memory so I could be wrong.  In any case, we often extend our libraries without bumping the major sonumber, so it seems entirely appropriate to do the specific requires on the core package in all subpackages to make sure that our applications and our libraries stay in sync.

jgunthorpe: Like I keep saying, if symbol versions have been used wrongly in the past, then we need to actually know the actual details of that and can form a repair strategy.

Broken symver could effect any package, using libibmad, while adding an explicit requires here will only fix the one case of the package inside rdma-core. We need a proper fix, if that really is the problem.

jgunthorpe:
@Honggang-LI can you rework this PR to have the stuff without comments so we can move it ahead?

Honggang-LI:
> @Honggang-LI can you rework this PR to have the stuff without comments so we can move it ahead?

This PR includes 7 commits. As you requested, two needs update. So, I will split this PR.
These 5 commits  will be kept for this PR.

redhat: move all %post scripts into one section
redhat: Build pyverbs for Fedora greater than release 31
redhat: ExcludeArch arm32 platform
redhat: Requires pciutils
redhat: sub-packages should use architecture specific dependency

I will open a new PR for commit "redhat: Requires libibverbs-utils".

commit "redhat: infiniband-diags should require libibumad" will be removed. We will create
a fedora/RHEL specific patch for our internal Fedora/RHEL spec file.

redhat/rdma-core.spec
Honggang-LI: I will drop this commit.  We will create a fedora/RHEL specific patch for our internal Fedora/RHEL spec file.

jgunthorpe:
@Honggang-LI the point of having the spec files upstream is you can rely on our expertise, you shouldn't push stuff into the internal spec files if we already talked about it and found it is not the right solution. As I keep repeating if you can explain that actual problem we can probably solve it correctly for everyone.

PR #643, opened 2019-12-23
Debian 27.0-1 changes

bdrung:
This patch set contains the packaging changes from the Debian unstable 27.0-1 upload.


debian/control
```diff
@@ -16,8 +16,8 @@ Build-Depends: cmake (>= 2.8.11),
                ninja-build,
                pandoc,
                pkg-config,
-               python3-docutils,
```
jgunthorpe: We really need to put this wrap and sort in the CI.

bdrung: Feel free to do that. `wrap-and-sort` is part of the devscripts package: https://tracker.debian.org/pkg/devscripts

PR #644, opened 2019-12-23
Fix spelling mistakes in infiniband-diags man page

bdrung:
lintian found some spelling mistakes in the infiniband-diags man page.

### DIFF
--- a/infiniband-diags/man/infiniband-diags.8.in.rst
+++ b/infiniband-diags/man/infiniband-diags.8.in.rst
@@ -99,8 +99,8 @@ The following config files are common amongst many of the utilities.
 Utilities list
 ==============
 
-Basic fabric conectivity
-------------------------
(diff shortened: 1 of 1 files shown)


infiniband-diags/man/infiniband-diags.8.in.rst
```diff
@@ -99,8 +99,8 @@ The following config files are common amongst many of the utilities.
 Utilities list
 ==============
 
-Basic fabric conectivity
-------------------------
+Basic fabric connectivity
```
jgunthorpe: We should also run lintian from the CI...

bdrung: lintian operates on Debian packages. So it could run after a Debian package build. lintian learns new spelling mistakes from time to time.

PR #645, opened 2019-12-24
tests: Add a timeout mechanism

noaos:
Some tests, especially those involving traffic, might hang sometimes
due to dropped packets, bugs etc. During execution of run_tests.py,
such issues stuck the run indefinitely.
Python's unittest framework does not provide a builtin timeout
mechanism, but the signal module can be used to achieve such behavior.

Register a handler for SIGALRM that simply raises a PyverbsError if
it expires (the rest of the tests will continue to run as usual).
Provide a decorator to wrap a test method call with a setting of a
timer (3 seconds, as this is more than enough time for our tests).
Decorate traffic methods with the above decorator.

An execution with an expired timeout will look as follows:

$ python3 /usr/share/doc/rdma-core-28.0/tests/run_tests.py --verbose -k
test_cqex
test_rc_traffic_cq_ex (tests.test_cqex.CqExTestCase) ... ok
test_ud_traffic_cq_ex (tests.test_cqex.CqExTestCase) ... ERROR
test_xrc_traffic_cq_ex (tests.test_cqex.CqExTestCase) ... ok

ERROR: test_ud_traffic_cq_ex (tests.test_cqex.CqExTestCase)
  File "/usr/share/doc/rdma-core-28.0/tests/utils.py", line 46, in func_wrapper
    func(func_self)
  File "/usr/share/doc/rdma-core-28.0/tests/test_cqex.py", line 69, in test_ud_traffic_cq_ex
    is_cq_ex=True)
  File "/usr/share/doc/rdma-core-28.0/tests/utils.py", line 439, in traffic
    poll(server.cq)
  File "/usr/share/doc/rdma-core-28.0/tests/utils.py", line 373, in poll_cq_ex
    ret = cqex.start_poll(poll_attr)
  File "/usr/share/doc/rdma-core-28.0/tests/utils.py", line 37, in sig_handler
    raise PyverbsError('Timeout expired')
pyverbs.pyverbs_error.PyverbsError: Timeout expired

Ran 3 tests in 3.149s

FAILED (errors=1)

Reviewed-by: Edward Srouji <edwards@mellanox.com>

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -2,7 +2,20 @@
 # Copyright (c) 2019 Mellanox Technologies, Inc . All rights reserved. See COPYING file
 
 import importlib
+import signal
 import os
+
+from pyverbs.pyverbs_error import PyverbsError
+
+
+# Handler for timeouts to be used by the test methods
+def sig_handler(signum, frame):
+    raise PyverbsError('Timeout expired')
+
+
+# Register the above handler to catch SIGALRM
+signal.signal(signal.SIGALRM, sig_handler)
+
 
 # Load every test as a module in the system so that unittest's loader can find it
 def _load_tests():
--- a/tests/test_cqex.py
+++ b/tests/test_cqex.py
@@ -60,16 +60,19 @@ class CqExTestCase(RDMATestCase):
             server.pre_run(client.psn, client.qpn)
         return client, server
 
+    @u.timeout_decorator
     def test_ud_traffic_cq_ex(self):
         client, server = self.create_players('ud')
         u.traffic(client, server, self.iters, self.gid_index, self.ib_port,
                   is_cq_ex=True)
 
+    @u.timeout_decorator
     def test_rc_traffic_cq_ex(self):
         client, server = self.create_players('rc')
         u.traffic(client, server, self.iters, self.gid_index, self.ib_port,
                   is_cq_ex=True)
 
+    @u.timeout_decorator
     def test_xrc_traffic_cq_ex(self):
         client, server = self.create_players('xrc')
         u.xrc_traffic(client, server, is_cq_ex=True)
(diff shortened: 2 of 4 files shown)


tests/__init__.py
```diff
@@ -2,8 +2,21 @@
...
+from pyverbs.pyverbs_error import PyverbsError
+
+
+# Handler for timeouts to be used by the test methods
+def sig_handler(signum, frame):
+    raise PyverbsError('Timeout expired')
+
+
+# Register the above handler to catch SIGALRM
+signal.signal(signal.SIGALRM, sig_handler)
```
jgunthorpe: Hum. How this actually works is really tricky, I'm actually surprised it is useful.

What long operation are you expecting to break out of with this? Generally would be cleaner to add timers directly to that than to rely on EINTR working properly.

EdwardSro: There's no specific long test / operation which this was made for.
Its purpose is to handle hangouts in traffic cases, where the test could hang due to test/driver bug or any other reason that led to packet(s) drop.
(P.S: timeout decorator is also implemented in PyPI and I think it's ok to use (as long as there are no nested methods nor threads) )

jgunthorpe: I'm not convinced this works as you think, alarm is very tricky, and won't effect most verbs calls. If it were to effect them it would trigger a verbs exception throw from EINTR, not alarm. To make this work properly all system calls must check for EINTR and then call PyErr_CheckSignals(). This includes all the calls inside pyverbs cython code.

We may have this wrong anyhow, we should probably be using some error handling path that goes through PyErr_SetFromErrno() which handles this and more.

EdwardSro: I see the concern more clearly now.
Hm.. I guess we can drop this PR for now, I'll consider changing the infrastructure flow and the way we handle exceptions.

PR #651, opened 2019-12-31
pyverbs: Expose new post send API

noaos:
The following series adds pyverbs support for the new post send API.
The first 5 patches add unrelated but needed support (e.g. memory
window) and fixes for bugs that were found along the way.
These patches are followed by the feature itself, a documentation and
a test.


pyverbs/mr.pyx
```diff
@@ -109,6 +117,17 @@ cdef class MR(PyverbsCM):
...
 
+cdef class MWBindInfo(PyverbsCM):
+    def __init__(self, MR mr not None, addr, length, mw_access_flags):
+        super().__init__()
+        self.mr = mr
+        self.info.mr = mr.mr
+        self.info.addr = addr
+        self.info.length = length
+        self.info.mw_access_flags = mw_access_flags
+        mr.add_ref(self)
```
jgunthorpe: Why does this need to use add_ref?

EdwardSro: It doesn't need. Will remove.

EdwardSro:
Can you please abandon this PR ?
I've pushed a new, fixed, one. (since noaos will no longer work on this).

PR #653, opened 2020-01-01
tests: test_parent_domain: Fix exception when no IB device found

Kamalheib:
Avoid the following exception when no IB device found, by skipping the test.

ERROR: test_default_allocators (tests.test_parent_domain.ParentDomainTestCase)
  File "/home/kheib/git/upstream/rdma-core/tests/test_parent_domain.py", line 35, in setUp
    dev = d.get_device_list()[-1]
IndexError: list index out of range

Fixes: 4fec6737c9bd ("tests: Add a test for parent domain")

### DIFF
--- a/tests/test_parent_domain.py
+++ b/tests/test_parent_domain.py
@@ -8,8 +8,8 @@ from pyverbs.pyverbs_error import PyverbsRDMAError
 from pyverbs.srq import SrqAttr, SrqInitAttr, SRQ
 from pyverbs.qp import QPInitAttr, QP
 from tests.base import BaseResources
+from tests.base import RDMATestCase
 import pyverbs.mem_alloc as mem
-import pyverbs.device as d
 import pyverbs.enums as e
 from pyverbs.cq import CQ
 import tests.utils as u
@@ -25,15 +25,9 @@ class ParentDomainRes(BaseResources):
         self.parent_domain = None
 
 
-class ParentDomainTestCase(unittest.TestCase):
-    def __init__(self, methodName='runTest', dev_name=None):
-        super().__init__(methodName)
-        self.dev_name = dev_name
-
+class ParentDomainTestCase(RDMATestCase):
     def setUp(self):
-        if self.dev_name is None:
-            dev = d.get_device_list()[-1]
-            self.dev_name = dev.name.decode()
+        super().setUp()
         self.pd_res = ParentDomainRes(self.dev_name)
 
     def _create_parent_domain_with_allocators(self, alloc_func, free_func):


tests/test_parent_domain.py
```diff
@@ -32,7 +32,10 @@ def __init__(self, methodName='runTest', dev_name=None):
 
     def setUp(self):
         if self.dev_name is None:
-            dev = d.get_device_list()[-1]
+            device_list = d.get_device_list()
+            if len(device_list) == 0:
+                raise unittest.SkipTest("No IB devices found")
```
noaos: Please use single quotes.

Kamalheib: > Please use single quotes.

> Also, I'd change the title to
> tests: Fix exception when no IB device found


Sure, I'll push a fix soon.

jgunthorpe:
Something else is wrong here, the setUp function from RDMATestCase is supposed to guarentee dev_name is set

Surely if dev_name is None  we should throw the skip from the setUp function?

Kamalheib:
> Something else is wrong here, the setUp function from RDMATestCase is supposed to guarentee dev_name is set

Well, the ParentDomainTestCase class doesn't inherit from the RDMATestCase to guarantee that.

`class ParentDomainTestCase(unittest.TestCase):`

> Surely if dev_name is None we should throw the skip from the setUp function?

The skip exception is done after checking that there are no devices available, please see code after this change.

```
    def setUp(self):
        if self.dev_name is None:
            device_list = d.get_device_list()
            if len(device_list) == 0:
                raise unittest.SkipTest('No IB devices found')
            dev = device_list[-1]
            self.dev_name = dev.name.decode()
        self.pd_res = ParentDomainRes(self.dev_name)
...

```

jgunthorpe:
Well, that is the problem then, old tests need to be converted to the new methodology and rely on the test runner framework to provide and select the RDMA device to use.

Kamalheib:
> Well, that is the problem then, old tests need to be converted to the new methodology and rely on the test runner framework to provide and select the RDMA device to use.

OK, I will push a fix soon that change ParentDomainTestCase to inherit from RDMATestCase.

jgunthorpe:
This looks good to me, @EdwardSro ?

tests/test_parent_domain.py
```diff
@@ -8,6 +8,7 @@
 from pyverbs.srq import SrqAttr, SrqInitAttr, SRQ
 from pyverbs.qp import QPInitAttr, QP
 from tests.base import BaseResources
+from tests.base import RDMATestCase
 import pyverbs.mem_alloc as mem
 import pyverbs.device as d
```
EdwardSro: Please remove this import "import pyverbs.device as d", since it's no longer used after your last changes.

Kamalheib: I'll push a fix soon.

EdwardSro:
Now it looks good to me.

PR #655, opened 2020-01-03
libhns: Various cleanups

Li-Weihang:
This series include various cleanups, no code logic is changed.

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -187,7 +187,7 @@ struct hns_roce_srq {
 	pthread_spinlock_t		lock;
 	unsigned long			*wrid;
 	unsigned int			srqn;
-	int				max;
+	unsigned int			max_wqe;
 	unsigned int			max_gs;
 	unsigned int			wqe_shift;
 	int				head;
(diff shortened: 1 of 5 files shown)


providers/hns/hns_roce_u.h
```diff
@@ -113,6 +113,7 @@ struct hns_roce_buf {
 };
 
 #define BIT_CNT_PER_BYTE       8
+#define BIT_CNT_PER_LONG       (BIT_CNT_PER_BYTE * sizeof(uint64_t))
```
jgunthorpe: It isn't really a long. BIT_CNT_PER_U64? Kind of a silly thing to write. Why not just write 64?

providers/hns/hns_roce_u_verbs.c
```diff
@@ -270,14 +270,9 @@ static int align_qp_size(int req)
 	return nent;
 }
 
-static int align_srq_size(int req)
+static int align_queue_size(int req)
```
jgunthorpe: unsigned values should have unsigned types

Li-Weihang: OK, thanks for your comments :)
I have re-pushed a new version.

PR #659, opened 2020-01-08
tests: Avoid failures when features aren't supported

Kamalheib:
This series includes patches to avoid failures when certain features are not supported by some devices.

### DIFF
--- a/pyverbs/qp.pyx
+++ b/pyverbs/qp.pyx
@@ -891,6 +891,8 @@ cdef class QP(PyverbsCM):
             # let's infer the type.
             if issubclass(type(creator), Context):
                 self._create_qp_ex(creator, init_attr)
+                if self.qp == NULL:
+                    raise PyverbsRDMAErrno('Failed to create QP')
                 ctx = <Context>creator
                 self.context = ctx
                 ctx.add_ref(self)
@@ -902,14 +904,15 @@ cdef class QP(PyverbsCM):
                     xrcd = <XRCD>init_attr.xrcd
                     xrcd.add_ref(self)
                     self.xrcd = xrcd
+
             else:
                 self._create_qp(creator, init_attr)
+                if self.qp == NULL:
+                    raise PyverbsRDMAErrno('Failed to create QP')
                 pd = <PD>creator
                 self.pd = pd
                 pd.add_ref(self)
                 self.context = None
-            if self.qp == NULL:
-                raise PyverbsRDMAErrno('Failed to create QP')
 
         if qp_attr is not None:
             funcs = {e.IBV_QPT_RC: self.to_init, e.IBV_QPT_UC: self.to_init,
--- a/tests/test_addr.py
+++ b/tests/test_addr.py
@@ -2,9 +2,10 @@
 # Copyright (c) 2019 Mellanox Technologies, Inc. All rights reserved.  See COPYING file
 
 import unittest
+import errno
 
+from pyverbs.pyverbs_error import PyverbsError, PyverbsRDMAError
 from pyverbs.addr import GlobalRoute, AHAttr, AH
-from pyverbs.pyverbs_error import PyverbsError
 from tests.base import PyverbsAPITestCase
 import pyverbs.device as d
 import pyverbs.enums as e
@@ -29,8 +30,13 @@ class AHTest(PyverbsAPITestCase):
                     continue
                 gr = get_global_route(ctx, port_num=port_num)
                 ah_attr = AHAttr(gr=gr, is_global=1, port_num=port_num)
-                with AH(pd, attr=ah_attr):
-                    done += 1
+                try:
+                    with AH(pd, attr=ah_attr):
+                        done += 1
+                except PyverbsRDMAError as ex:
+                    if ex.error_code == errno.EOPNOTSUPP:
+                        raise unittest.SkipTest('Create AH is not supported')
+                    raise ex
         if done == 0:
             raise unittest.SkipTest('No port is up, can\'t create AH')
     # TODO: Test ibv_create_ah_from_wc once we have traffic
@@ -52,9 +58,11 @@ class AHTest(PyverbsAPITestCase):
                 ah_attr = AHAttr(is_global=0, port_num=port_num)
                 try:
                     ah = AH(pd, attr=ah_attr)
-                except PyverbsError as err:
-                    assert 'Failed to create AH' in err.args[0]
-                    done += 1
+                except PyverbsRDMAError as ex:
+                    if ex.error_code == errno.EOPNOTSUPP:
+                        raise unittest.SkipTest('Create AH is not supported')
+                    assert 'Failed to create AH' in str(ex)
+                    done +=1
                 else:
                     raise PyverbsError('Created a non-global AH on RoCE')
         if done == 0:
@@ -73,9 +81,14 @@ class AHTest(PyverbsAPITestCase):
                     continue
                 gr = get_global_route(ctx)
                 ah_attr = AHAttr(gr=gr, is_global=1, port_num=port_num)
-                with AH(pd, attr=ah_attr) as ah:
-                    ah.close()
-                    done += 1
+                try:
+                    with AH(pd, attr=ah_attr) as ah:
+                        ah.close()
+                        done += 1
+                except PyverbsRDMAError as ex:
+                    if ex.error_code == errno.EOPNOTSUPP:
+                        raise unittest.SkipTest('Create AH is not supported')
+                    raise ex
         if done == 0:
             raise unittest.SkipTest('No port is up, can\'t create AH')
 
--- a/tests/test_cq.py
+++ b/tests/test_cq.py
@@ -5,11 +5,13 @@ Test module for pyverbs' cq module.
 """
 import random
 
+from pyverbs.pyverbs_error import PyverbsError, PyverbsRDMAError
 from pyverbs.cq import CompChannel, CQ, CqInitAttrEx, CQEX
-from pyverbs.pyverbs_error import PyverbsError
 from tests.base import PyverbsAPITestCase
 import pyverbs.enums as e
 import tests.utils as u
+import unittest
+import errno
 
 
 class CQTest(PyverbsAPITestCase):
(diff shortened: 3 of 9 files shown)


pyverbs/qp.pyx
```diff
@@ -902,14 +902,17 @@ cdef class QP(PyverbsCM):
                     xrcd = <XRCD>init_attr.xrcd
                     xrcd.add_ref(self)
                     self.xrcd = xrcd
+                if self.qp == NULL:
+                    raise PyverbsRDMAErrno('Failed to create extended QP')
```
noaos: I don't think this is correct - we're still creating ibv_qp and not ibv_qp_ex.

tests/test_mr.py
```diff
@@ -161,35 +161,55 @@ def test_reg_mw_type1(self):
...
         Test ibv_dealloc_mw() for type 1 MW
         """
         for ctx, attr, attr_ex in self.devices:
             with PD(ctx) as pd:
-                with MW(pd, e.IBV_MW_TYPE_1) as mw:
-                    mw.close()
+                try:
+                    with MW(pd, e.IBV_MW_TYPE_1) as mw:
+                        mw.close()
+                except PyverbsRDMAError as ex:
```
noaos: So if MW dereg failed, you supress pyverbs' exception and not raising anything instead.

tests/test_mr.py
```diff
@@ -161,35 +161,55 @@ def test_reg_mw_type1(self):
...
         """
         for ctx, attr, attr_ex in self.devices:
             with PD(ctx) as pd:
-                with MW(pd, e.IBV_MW_TYPE_2) as mw:
-                    mw.close()
+                try:
+                    with MW(pd, e.IBV_MW_TYPE_2) as mw:
+                        mw.close()
+                except PyverbsRDMAError as ex:
+                    if 'allocate' in str(ex) and 'not implemented' in str(ex):
```
noaos: same

tests/test_cq.py
```diff
@@ -131,7 +132,9 @@ def test_create_cq_ex(self):
                 try:
                     with CQEX(ctx, cq_init_attrs_ex):
                         pass
-                except PyverbsError as ex:
+                except PyverbsRDMAError as ex:
+                    if 'not implemented' in str(ex):
+                        raise unittest.SkipTest('Create extended CQ is not supported')
```
noaos: I think it would be better to add the ibv_cq_init_attr_ex - isn't it possible that a device supports creation of extended CQ but some attributes are not supported? (same for all of this patch)

tests/test_qp.py
```diff
@@ -101,11 +103,21 @@ def test_create_qp_ex_no_attr_connected(self):
             with PD(ctx) as pd:
                 with CQ(ctx, 100, None, None, 0) as cq:
                     qia = get_qp_init_attr_ex(cq, pd, attr, attr_ex, e.IBV_QPT_RC)
-                    with QP(ctx, qia) as qp:
-                        assert qp.qp_state == e.IBV_QPS_RESET, 'RC QP should have been in RESET'
+                    try:
```
noaos: same comment about extended CQ - It might be more useful for users to get the exact attributes of the QP that failed to create.

tests/test_mr.py
Kamalheib: No, I'm raising the 'ex', please see line 198.

tests/test_mr.py
Kamalheib: Same. line 212.

tests/test_cq.py
Kamalheib: But this patch handles the case where the provider doesn't implement the create_cq_ex() at all.

pyverbs/qp.pyx
Kamalheib: Will that because the ibv_create_qp() and ibv_create_qp_ex() functions reutnrn "ibv_qp *".

noaos: I know, this makes your message inaccurate. Please change that.

tests/test_cq.py
noaos: I'm not asking for a complex logic. No point having a patch that only works properly for devices that do not support *at all* and wait for someone else to make it helpful to partial support.

pyverbs/qp.pyx
Kamalheib: OK, I'll drop this patch.

tests/test_cq.py
Kamalheib: Could you please add more info about how ibv_cq_init_attr_ex can help in partial support?

I'm thinking about extending the "if" statement to check for "not supported" to handle the partial support.

pyverbs/qp.pyx
jgunthorpe: It seems reasonable to throw when a create fails, but shouldn't this be right after the system call, why are we doing refcounting for something that doesn't need refcounting?

tests/test_mr.py
jgunthorpe: This should directly test for EOPNOTSUPP, not via a string compare.

tests/test_cq.py
jgunthorpe: ibv_create_qp_ex is the same as ibv_create_qp if a minimual set of attributes are requested. A provider can support or not any combination of the attributes.

tests/test_addr.py
```diff
@@ -52,9 +57,11 @@ def test_create_ah_roce(self):
                 ah_attr = AHAttr(is_global=0, port_num=port_num)
                 try:
                     ah = AH(pd, attr=ah_attr)
-                except PyverbsError as err:
-                    assert 'Failed to create AH' in err.args[0]
-                    done += 1
+                except PyverbsRDMAError as ex:
+                    if 'not implemented' in str(ex):
+                        raise unittest.SkipTest('Create AH is not supported')
```
juhlee-microsoft: 'AH creation was not supported

tests/test_addr.py
```diff
@@ -73,9 +80,14 @@ def test_destroy_ah(self):
...
-                with AH(pd, attr=ah_attr) as ah:
-                    ah.close()
-                    done += 1
+                try:
+                    with AH(pd, attr=ah_attr) as ah:
+                        ah.close()
+                        done += 1
+                except PyverbsRDMAError as ex:
+                    if 'create' in str(ex) and 'not implemented' in str(ex):
+                        raise unittest.SkipTest('Create AH is not supported')
```
juhlee-microsoft: Ditto - wording. Please keep the consistent style and past sentence like existing sentences.

pyverbs/qp.pyx
Kamalheib: OK, I'll change this patch to raise the exception right after create QP fail.

tests/test_cq.py
Kamalheib: Well, there are two failure cases here, The first one is that the provider doesn't implement the verb at all (This patch handle it), and the second one is that provider implements the verb partially, In this case, the only way to determine the error is to check errno value (I'll update the patch to check for "not supported").

@jgunthorpe and @EdwardSro : Thoughts?

tests/test_mr.py
Kamalheib: > This should directly test for EOPNOTSUPP, not via a string compare.

Well, the error in this case is `#define ENOSYS 88	/* Function not implemented */` - Please see below:

```
  File "/home/rdma-core/tests/test_mr.py", line 182, in test_dereg_mw_type1
    with MW(pd, e.IBV_MW_TYPE_1) as mw:
  File "mr.pyx", line 124, in pyverbs.mr.MW.__init__
pyverbs.pyverbs_error.PyverbsRDMAError: Failed to allocate MW. Errno: 38, Function not implemented
```

currently, the only way to test is via comparing the string from the exception, unless you meat to change the PyverbsError to store the errno value - so later in the code, I can do the following:

```
try:
    with MW(pd, e.IBV_MW_TYPE_1):
        pass
        if ex.error_code == errno.ENOSYS:
            raise unittest.SkipTest('Create memory window of type 1 is not supported')
        raise ex
```

jgunthorpe: The error code is supposed to EOPNOTSUPP, returning ENOSYS is a bug in libibverbs or the provider that we should fix.

jgunthorpe: And yes to changing to store the error code directly in the PyverbsRDMAError

EdwardSro: We already have a PyverbsRDMAErrno which returns PyverbsRDMAError with the errno in its message. So maybe it's better to re-define it as a class, inheriting from PyverbsRDMAError and storing the errno. (must check what is the consequences of this change - if any)

tests/test_addr.py
EdwardSro: > Please keep the consistent style and past sentence **like existing sentences**.
Where have you seen this (past sentence) style in the existing messages ? 
I don't think it's appropriate here to use past sentence...

tests/test_mr.py
Kamalheib: > We already have a PyverbsRDMAErrno which returns PyverbsRDMAError with the errno in its message. So maybe it's better to re-define it as a class, inheriting from PyverbsRDMAError and storing the errno. (must check what is the consequences of this change - if any)

Based on [1] reference Cython classes can not inherit from normal classes, which probably describes why PyverbsRDMAErrno is not defined as a class.

`Normal Python classes can inherit from cdef classes, but not the other way around. Cython requires to know the complete inheritance hierarchy in order to lay out their C structs, and restricts it to single inheritance. Normal Python classes, on the other hand, can inherit from any number of Python classes and extension types, both in Cython code and pure Python code.
`

I plan to store the errno value to be part of the PyverbsError base class, I think this change makes more sense, first because the "error_code" is used in the base class to build the msg string, second maybe in the future we want to use the errno code with PyverbsUserError?

@EdwardSro : Thoughts?

[1] - https://cython.readthedocs.io/en/latest/src/tutorial/cdef_classes.html

EdwardSro: Oh right. Then yes, I would go in that way.
Regarding PyverbsUserError, it should be used for errors made solely by pyverbs / pyverbs' users (e.g. passing wrong device name / an exception for feature that's partially supported **in pyverbs** etc.), so errno shouldn't be relevant for PyverbsUserError (if it is, then most likely PyverbsRDMAError or other exception should be used)

jgunthorpe: While this is OK, it seems a little off, and maybe a followup should revise how we handle this.

But basically, tests should only be skipped from decorators that delcare some required test capability. ie running the  test_reg_mw_wrong_type tests without MW support is just as wrong.

Probing the capability from a EOPNOTSUPP in a decorator requires some more fancy scheme, and requires the tests be moved to the new scheme where the test framework so each test runs on a single ibv_context and the decorator can then check its support.

PR #660, opened 2020-01-09
verbs: Relaxed ordering memory regions

yishaih:
This series exposes an IBV_ACCESS_RELAXED_ORDERING optional MR access flag.
This optional flag allows creation of relaxed ordering memory regions.
Access through such MRs can improve performance by allowing the system to reorder certain accesses.

The series uses the new ioctl command to get a device context, this command enables reading some core generic capabilities such as supporting an optional MR access flags by IB core and its related drivers.

This capability enables transparent masking of the optional flags in libibverbs when the kernel doesn't support the MR optional access mode.

The matching kernel series was sent to 'for-next'.


libibverbs/verbs.c
```diff
@@ -339,6 +341,17 @@ struct ibv_mr *ibv_reg_mr_iova(struct ibv_pd *pd, void *addr, size_t length,
 	return mr;
 }
 
+struct ibv_mr *ibv_reg_mr_iova2(struct ibv_pd *pd, void *addr, size_t length,
+				uint64_t iova, int access)
+{
+	struct verbs_device *device = verbs_get_device(pd->context->device);
+
+	if (!(device->core_support & IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS))
+		access &= ~(typeof(access))IBV_ACCESS_OPTIONAL_RANGE;
```
jgunthorpe: Don't do bit manipulations on signed types, and there is no need for typeof obfuscation here.
Change the type of access to unsigned int in the internal functions.

yishaih: PR was updated according to this note.

PR #664, opened 2020-01-13
Request python docutils package

rleon:


redhat/rdma-core.spec
```diff
@@ -35,6 +35,13 @@ BuildRequires: python3
 BuildRequires: python
 %endif
 %endif
+
+%if 0%{?rhel} >= 8 || 0%{?fedora} >= 30 || %{with_pyverbs}
+BuildRequires: python3-docutils
+%else
+BuildRequires: python-docutils
+%endif
```
jgunthorpe: Lets just put this inside the existing pandoc stanza and update the FC version to 30.

Not sure what the pyverbs is all about either??

README.md
```diff
@@ -52,21 +52,21 @@ only load from the system path.
 ### Debian Derived
 
 ```sh
-$ apt-get install build-essential cmake gcc libudev-dev libnl-3-dev libnl-route-3-dev ninja-build pkg-config valgrind python3-dev cython3
+$ apt-get install build-essential cmake gcc libudev-dev libnl-3-dev libnl-route-3-dev ninja-build pkg-config valgrind python3-dev cython3 python3-docutils
```
jgunthorpe: These should all mention pandoc as well

jgunthorpe:
The commit message is misleading, we only get that error if the user does not use the tar.gz we distribute, and in that case you need pandoc too.

redhat/rdma-core.spec
rleon: > Not sure what the pyverbs is all about either??
Are you talking about "|| %{with_pyverbs}"?
I added it, because with_pyverbs we will install python3 without any relation to fedora version.

rleon: > Lets just put this inside the existing pandoc stanza and update the FC version to 30.
> 

Because I need with_pyverbs check, putting inside pandoc looks extremely ugly.
Pandoc doesn't need with_pyverbs, while python3-docutils needs.

rleon:
1. Added pandoc to README file
2. Didn't change spec file

PR #666, opened 2020-01-13
efa: Use the correct barrier between BAR writes on post send flow

galpress:
The TX Low Latency Queue (LLQ) is stored on the device's BAR, hence
mmio_flush_writes should be used instead of udma_to_device_barrier
before ringing the doorbell.

Fixes: f7f275ff2b1a ("efa: Elastic Fabric Adapter (EFA) userspace RDMA provider")
Fixes: f189a2ac12d6 ("efa: Support send using extended QP API")

### DIFF
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -1191,7 +1191,7 @@ int efa_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 	struct efa_ah *ah;
 	int err = 0;
 
-	pthread_spin_lock(&qp->sq.wq.wqlock);
+	mmio_wc_spinlock(&qp->sq.wq.wqlock);
 	while (wr) {
 		err = efa_post_send_validate_wr(qp, wr);
 		if (err) {
@@ -1236,9 +1236,13 @@ int efa_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 	}
 
 ring_db:
-	udma_to_device_barrier();
+	mmio_flush_writes();
 	mmio_write32(qp->sq.db, qp->sq.wq.desc_idx);
 
+	/*
+	 * Not using mmio_wc_spinunlock as the doorbell write should be done
+	 * inside the lock.
+	 */
 	pthread_spin_unlock(&qp->sq.wq.wqlock);
 	return err;
 }
@@ -1452,7 +1456,7 @@ static void efa_send_wr_start(struct ibv_qp_ex *ibvqpx)
 {
 	struct efa_qp *qp = to_efa_qp_ex(ibvqpx);
 
-	pthread_spin_lock(&qp->sq.wq.wqlock);
+	mmio_wc_spinlock(&qp->sq.wq.wqlock);
 	qp->wr_session_err = 0;
 	qp->sq.num_wqe_pending = 0;
 	qp->sq.phase_rb = qp->sq.wq.phase;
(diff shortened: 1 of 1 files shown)


providers/efa/verbs.c
```diff
@@ -1236,7 +1236,7 @@ int efa_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 	}
 
 ring_db:
-	udma_to_device_barrier();
+	mmio_flush_writes();
 	mmio_write32(qp->sq.db, qp->sq.wq.desc_idx);
 
 	pthread_spin_unlock(&qp->sq.wq.wqlock);
```
jgunthorpe: Can you put the mmio_write32 after the spin_unlock? Then you can use the wc spinlock primitive, which I think you may already need as there is no mmio_wc_start()?

galpress: Why would we need mmio_wc_start? 
Isn't it needed in case there are mmio writes before the writes to the WC (LLQ) memory?

jgunthorpe: No, it makes the spinlock work. Normal spinlocks do not contain WC writes, so various collisions are possible. It is essentially a pairing to the flush_writes so that the CPU orders strongly with other CPUs that might have already issued WC writes.

In this case I think you care less about strict ordering of WC writes as this is a ring, not a doorbell, but there may be unexpected corner cases on some archs

galpress: The doorbell write should be inside the spinlock though, so the wc_spinlock cannot be used here.
I can add a mmio_wc_start after acquiring the spinlock, but I'm still not entirely convinced it's needed for this specific use case, maybe to make sure the previous doorbell is flushed?

galpress: @jgunthorpe ?

jgunthorpe: I thought you were respinning?

WC memory has tricky ordering rules, I prefer to see the wc spinlock used whenever possible, as it has defined semantics.

Your case seems to be a bit different, so maybe some new primitive and pattern is needed, but I'm really not sure it is OK as written. The biggest problem with WC memory is what happens when you write to the same address twice, for instance, because the ring buffer wrapped around.

Remember that mmio_write32 is not strongly ordered WRT to WC memory, only the flush_writes serves as a barrier, but any later WC writes can be hoisted to before the mmio_write32. This is part of what the wc_spinlock is designed to prevent.

mmio_flush_writes() is also not defined to include udma_to_device_barrier(), they are different things.

galpress: > Remember that mmio_write32 is not strongly ordered WRT to WC memory, only the flush_writes serves as a barrier, but any later WC writes can be hoisted to before the mmio_write32. This is part of what the wc_spinlock is designed to prevent.

Later WC writes == writes in the next post send invocation that will bypass the doorbell write?
Is it better to just add another mmio_flush_writes straight after ringing the doorbell?

> mmio_flush_writes() is also not defined to include udma_to_device_barrier(), they are different things.

udma_to_device_barrier is not needed in this case as there's no DMA memory involved here.
BTW, the mmio_wc_start does state that: "This barrier implies udma_to_device_barrier()", no?

jgunthorpe: The pattern is like a spinlock, a barrier above where writes can't leak 'backwards' and a barrier where writes can't leak 'forwards', I wouldn't add another barrier after the doorbell

jgunthorpe:
Well, it is weird, but probably OK, can you add a comment that the bottom bit as an open coded wc_unlock for a reason?

galpress:
Sure, will do.
Also, I've updated the commit message which is kinda hard to see here.

PR #667, opened 2020-01-13
Update MAINTAINERS

EdwardSro:
Update current pyverbs' maintainer to Edward.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -179,6 +179,6 @@ S:	Supported
 F:	providers/vmw_pvrdma/
 
 PYVERBS
-M:	Noa Osherovich <noaos@mellanox.com>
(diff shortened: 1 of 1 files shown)


MAINTAINERS
```diff
@@ -179,6 +179,7 @@ S:	Supported
 F:	providers/vmw_pvrdma/
 
 PYVERBS
-M:	Noa Osherovich <noaos@mellanox.com>
+M:	Edward Srouji <edwards@mellanox.com>
+H:	Noa Osherovich <noaos@mellanox.com>
```
rleon: there is no need in this line, because Noa's email is not active.

Thanks

EdwardSro: Removed

PR #669, opened 2020-01-15
Fix incorrect return code

Kamalheib:
The proper return code is EOPNOTSUPP when an operation is not supported.

### DIFF
--- a/libibverbs/man/ibv_advise_mr.3.md
+++ b/libibverbs/man/ibv_advise_mr.3.md
@@ -73,8 +73,9 @@ application performance.
 **ibv_advise_mr()** returns 0 when the call was successful, or the value
 		    of errno on failure (which indicates the failure reason).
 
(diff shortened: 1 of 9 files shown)


libibverbs/man/ibv_advise_mr.3.md
```diff
@@ -73,7 +73,7 @@ application performance.
 **ibv_advise_mr()** returns 0 when the call was successful, or the value
 		    of errno on failure (which indicates the failure reason).
 
-*ENOSYS*
+*EOPNOTSUPP*
```
jgunthorpe: The man pages should specify that ENOSYS may sometimes be returned by old versions of the library.

Kamalheib: Thanks for pointing that out, I'll post a fix soon.

PR #670, opened 2020-01-15
librdmacm: use header <infiniband/acm.h>

Honggang-LI:


jgunthorpe:
Oh that is weird, how did this break sparse?

jgunthorpe:
Ok, it is becasue ibacm is marked NO SPARSE so it does not get its header files installed.

Probably the right answer is that rdmacm should own the header file that defines the acm protocol, not ibacm.

Honggang-LI:
> Ok, it is becasue ibacm is marked NO SPARSE so it does not get its header files installed.

Maybe we need fix that?

> Probably the right answer is that rdmacm should own the header file that defines the acm protocol, not ibacm.

No, librdmacm does not have such header file that defines the acm protocol.

jgunthorpe:
Fixing sparce for ibacm looked supremely difficult last I checked

I'm saying rdmacm *should* define the ACM protocol in a header, that is the proper place for such a definition to live, not in ibacm. Move all the protocol defines there and make a new header

Honggang-LI:
> Fixing sparce for ibacm looked supremely difficult last I checked
> 
> I'm saying rdmacm _should_ define the ACM protocol in a header, that is the proper place for such a definition to live, not in ibacm. Move all the protocol defines there and make a new header

The ibacm is an implementation of ACM protocol. ibacm also includes ACM provider interface (infiniband/acm_prov.h). If we move the ACM protocol header, infiniband/acm.h, into rdmacm, the acm.h and acm_prov.h will live in different sub-packages of rdma-core. It looks weird for me.

Please feel free to close this PR as won't fix, in case it is too difficult to fix sparse for ibacm.

jgunthorpe:
ACM is a protocol that runs from librdmacm to some daemon. The part connected to librdmacm rightly belongs in librdmacm as there could be many implementations of the acm daemon.

The provider protocol is some internal detail of our ibacm daemon, so it should continue to live in ibacm.

Honggang-LI:
Ping? Any comment? thanks

jgunthorpe:
I thought you were going to respin this as requested

librdmacm/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ publish_headers(rdma
   )
 publish_headers(infiniband
   ib.h
+  acm.h
```
jgunthorpe: Keep sorted

jgunthorpe:
Oh, I see, it was fixed, it is actually hard to keep track with github on force pushes ..

Honggang-LI:
> Oh, I see, it was fixed, it is actually hard to keep track with github on force pushes ..
I had to force push again. Edit the file with github will create a unnecessary new patch. And the new patch failed CI test because of missing sign-off line.

PR #672, opened 2020-01-22
libhns: Use standard errnos instead of -1

Li-Weihang:
-1 can't reflect the detailed reason of error, so replace them with standard errnos.


rleon:
I see no harm in this patch but don't see any value too.
In order to enkoy from errno, you need to set it and return value is not enough.

Li-Weihang:
> I see no harm in this patch but don't see any value too.
> In order to enkoy from errno, you need to set it and return value is not enough.

Thank you Leon, we will do more work for this.

PR #677, opened 2020-01-28
tests: Fix exception when query GIDs

Kamalheib:
Avoid the following error by iterating over the availbe gid range.

  File "/home/rdma-core/tests/test_cq_events.py", line 26, in setUp
    super().setUp()
  File "/home/rdma-core/tests/base.py", line 122, in setUp
    self._add_gids_per_device(ctx, dev_name)
  File "/home/rdma-core/tests/base.py", line 149, in _add_gids_per_device
    self._add_gids_per_port(ctx, dev, port+1)
  File "/home/rdma-core/tests/base.py", line 138, in _add_gids_per_port
    gid = ctx.query_gid(port, idx)
  File "device.pyx", line 195, in pyverbs.device.Context.query_gid
pyverbs.pyverbs_error.PyverbsRDMAError: Failed to query gid 1 of port 1

Fixes: bb5958680256 ("tests: RDMATestCase")

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -133,15 +133,12 @@ class RDMATestCase(unittest.TestCase):
         # Don't add ports which are not active
         if ctx.query_port(port).state != e.IBV_PORT_ACTIVE:
             return
-        idx = 0
-        while True:
+
+        for idx in range(ctx.query_port(port).gid_tbl_len):
             gid = ctx.query_gid(port, idx)
-            if gid.gid[-19:] == self.ZERO_GID:
-                # No point iterating on
-                break
-            else:
+            # Avoid adding ZERO GIDs
+            if gid.gid[-19:] != self.ZERO_GID:
                 self.args.append([dev, port, idx])
-                idx += 1
 
     def _add_gids_per_device(self, ctx, dev):
         port_count = ctx.query_device().phys_port_cnt


tests/base.py
```diff
@@ -133,15 +133,14 @@ def _add_gids_per_port(self, ctx, dev, port):
         # Don't add ports which are not active
         if ctx.query_port(port).state != e.IBV_PORT_ACTIVE:
             return
-        idx = 0
-        while True:
+
+        for idx in range(ctx.query_port(port).gid_tbl_len):
             gid = ctx.query_gid(port, idx)
             if gid.gid[-19:] == self.ZERO_GID:
```
jgunthorpe: I'm not sure this existing logic is correct, @paravmellanox ? Can we have ZERO_GID holes in the gid table? I think so?

paravmellanox: @jgunthorpe 
@Kamalheib 
yes, gid table is constructed out of IPv4 and IPv6 addresses for physical port, vlan, macvlan, bond device etc for RoCE.

and IP addresses they can be added and deleted in any sequence. with DHCP also lease can expire, resulting into IP address removal and hence GID removal, that leads to holes.
so it should iterate over whole GID table.

Kamalheib: Thanks for reviewing the patch.
I'll update the patch to iterate over the whole GID table.

tests/base.py
```diff
@@ -133,15 +133,14 @@ def _add_gids_per_port(self, ctx, dev, port):
...
-                # No point iterating on
-                break
```
jgunthorpe: It is correct now, but just write it more simply as

if gid.gid[-19:] != self.ZERO_GID:
   self.args.append([dev, port, idx])

and why does this append an array not a tuple? Strange.

Kamalheib: > It is correct now, but just write it more simply as
> 
> if gid.gid[-19:] != self.ZERO_GID:
> self.args.append([dev, port, idx])
> 
OK, I'll update the patch.

> and why does this append an array not a tuple? Strange.

Yes, that is strange, I agree that this need to be a tuple to avoid confusion, but probably in a different patch.

PR #680, opened 2020-01-29
tests: Use error code to determine the failure

Kamalheib:
Instead of comare strings, use the stored error code in the
PyverbsRDMAError exception to determine the failure.


tests/test_mlx5_var.py
```diff
@@ -10,6 +10,7 @@
 from tests.base import BaseResources
```
EdwardSro: Please fix "compare*" typo in the commit message (maybe rewrite as "Instead of string comparison").

Kamalheib: > Please fix "compare*" typo in the commit message (maybe rewrite as "Instead of string comparison").

Thanks for your review, I'll push a fix soon.

PR #681, opened 2020-01-30
mlx5: Change flow_meter_parameter to be void *

yishaih:
Change flow_meter_parameter type from __be64[] to void *.
It allows to user to assign pre allocated buffer.

### DIFF
--- a/providers/mlx5/mlx5dv.h
+++ b/providers/mlx5/mlx5dv.h
@@ -1444,7 +1444,7 @@ struct mlx5dv_dr_flow_meter_attr {
 	uint8_t                 active;
 	uint8_t                 reg_c_index;
 	size_t			flow_meter_parameter_sz;
-	__be64			flow_meter_parameter[];
+	void			*flow_meter_parameter;
 };
 
 struct mlx5dv_dr_domain *


providers/mlx5/mlx5dv.h
```diff
@@ -1444,7 +1444,7 @@ struct mlx5dv_dr_flow_meter_attr {
 	uint8_t                 active;
 	uint8_t                 reg_c_index;
 	size_t			flow_meter_parameter_sz;
-	__be64			flow_meter_parameter[];
+	void			*flow_meter_parameter;
```
jgunthorpe: So what actually gets put in here? We don't have a pyverbs test for this?

Can you snip an example from the user?

yishaih: The usage as 'void *' can be seen in below links from DPDK upstream git.

Create meter:
http://git.dpdk.org/dpdk/tree/drivers/net/mlx5/mlx5_glue.c #l776

Prepare meter:
http://git.dpdk.org/dpdk/tree/drivers/net/mlx5/mlx5_flow_meter.c #l60

PR #683, opened 2020-02-03
ibacm: Do not open non Infiniband device

Honggang-LI:


ibacm/src/acm.c
```diff
@@ -2596,17 +2596,64 @@ acm_open_port(struct acmc_port *port, struct acmc_device *dev, uint8_t port_num)
...
 
+static bool dev_support_ib(struct ibv_device *ibdev)
+{
+	struct ibv_context *verbs;
+	struct ibv_device_attr attr;
+	struct ibv_port_attr port_attr;
+	int i, ret;
+	bool has_ib_port = false;
+
+	verbs = ibv_open_device(ibdev);
```
jgunthorpe: Let us try to avoid doing open_device twice?

Honggang-LI: I think that do open_device twice can keep the code simple. BTW, ibv_query_port also had been do twice. The function is not in critical path, so I think do twice is OK.

If you have any suggestion, please feel free to edit the patch.

Thanks

ibacm/src/acm.c
```diff
@@ -2596,17 +2596,64 @@ acm_open_port(struct acmc_port *port, struct acmc_device *dev, uint8_t port_num)
...
 	size_t size;
 	int i, ret;
-	unsigned int opened_ib_port_cnt = 0;
+	bool has_ib_port;
 
 	acm_log(1, "%s\n", ibdev->name);
+
+	has_ib_port = dev_support_ib(ibdev);
+	if (!has_ib_port) {
+		acm_log(0, "%s does not support Infiniband\n", ibdev->name);
```
Hakon-Bugge: Spelled InfiniBand (upper case "B").

ibacm/src/acm.c
```diff
@@ -2596,17 +2596,64 @@ acm_open_port(struct acmc_port *port, struct acmc_device *dev, uint8_t port_num)
...
 	struct ibv_context *verbs;
 	size_t size;
 	int i, ret;
-	unsigned int opened_ib_port_cnt = 0;
+	bool has_ib_port;
 
 	acm_log(1, "%s\n", ibdev->name);
+
+	has_ib_port = dev_support_ib(ibdev);
+	if (!has_ib_port) {
```
Hakon-Bugge: I guess you do not need this local bool has_ib_port? Why not 

if (dev_support_ib(ibdev)) {

ibacm/src/acm.c
```diff
@@ -2596,17 +2596,64 @@ acm_open_port(struct acmc_port *port, struct acmc_device *dev, uint8_t port_num)
 	port->state = IBV_PORT_DOWN;
 }
 
+static bool dev_support_ib(struct ibv_device *ibdev)
+{
+	struct ibv_context *verbs;
+	struct ibv_device_attr attr;
```
Hakon-Bugge: ... attr = {};

and you can remove the memset() below.

ibacm/src/acm.c
```diff
@@ -2596,17 +2596,64 @@ acm_open_port(struct acmc_port *port, struct acmc_device *dev, uint8_t port_num)
...
+	struct ibv_port_attr port_attr;
```
Hakon-Bugge: Move this decl down to the loop looping through the ports, and use = {} and remove the memset.

But why do you zero-initialize these device and port attrs anyway?

ibacm/src/acm.c
```diff
@@ -2596,17 +2596,64 @@ acm_open_port(struct acmc_port *port, struct acmc_device *dev, uint8_t port_num)
...
+	int i, ret;
+	bool has_ib_port = false;
+
+	verbs = ibv_open_device(ibdev);
+	if (verbs == NULL) {
```
Hakon-Bugge: prefer
if (!verbs) {

ibacm/src/acm.c
Honggang-LI: I'd like to use memset to initialize a struct with zero. It is a clear statement for me.

ibacm/src/acm.c
Honggang-LI: Declaration in for loop is C99 syntax. I'd like to put all declarations in the beginning of a function.

zero-initialize those struct to avoid garbage value, which may introduce bugs. It is my coding habit too.

ibacm/src/acm.c
```diff
@@ -2630,29 +2628,13 @@ static void acm_open_dev(struct ibv_device *ibdev)
...
-		acm_log(1, "%d InfiniBand %s opened for %s\n",
-				opened_ib_port_cnt,
-				opened_ib_port_cnt == 1 ? "port" : "ports",
-				ibdev->name);
-		return;
-	}
+	list_add(&dev_list, &dev->entry);
+
+	acm_log(1, "%s opened\n", ibdev->name);
+	return;
```
Hakon-Bugge: Did this revert have a conflict with d04d4660d114 ("ibacm: Fix a memory leak in an acm_open_dev() error path")?

The fix from d04d4660d114 is also reverted here. The free() below will never be executed.

Hakon-Bugge:
Nit: s/infiniband/InfiniBand/ in  $Subject

jgunthorpe:
The revert commits don't seem to be adding any readability, just mention them in your commit message and squash everything to one patch.

ibacm/src/acm.c
```diff
@@ -2620,6 +2620,27 @@ static void acm_open_dev(struct ibv_device *ibdev)
...
+				ibdev->name, i + 1, ret);
+			continue;
+		}
+
+		if (port_attr.link_layer == IBV_LINK_LAYER_INFINIBAND) {
+			acm_log(1, "%s port %d is an InfiniBand port\n", ibdev->name, i + 1);
+			has_ib_port = true;
+		} else {
+			acm_log(1, "%s port %d is not an InfiniBand port\n", ibdev->name, i + 1);
+		}
```
Hakon-Bugge: These six lines can of course be written as:

has_ib_port = port_attr.link_layer == IBV_LINK_LAYER_INFINIBAND;
acm_log(1, "%s port %d is %san InfiniBand port\n", ibdev->name, i + 1, has_ib_port ? "" : "not ");

but I am OK with this either way.

PR #684, opened 2020-02-03
rsockets: Add check for established connection

shefty:
Rsockets will use a thread to drive connection state. However,
checks against the rsocket state are made outside of locks in
send/recv/poll.  If the state is still listed as opening, 
rs_do_connect will be called, where the state will be checked
under lock.  Because a separate thread is driving the connection
state, the rsocket state may be connected at this point.

Update rs_do_connect to return 0 (success) if the rsocket
is already in the connected state.  This fixes failures using
rsockets in the JDK test set.

Signed-off-by: Sean Hefty <sean.hefty@intel.com>

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -1454,7 +1454,7 @@ connected:
 		rs->state = rs_connect_rdwr;
 		break;
 	default:
-		ret = ERR(EINVAL);
+		ret = (rs->state & rs_connected) ? 0 : ERR(EINVAL);
 		break;
 	}
 


librdmacm/rsocket.c
```diff
@@ -1454,7 +1454,7 @@ static int rs_do_connect(struct rsocket *rs)
 		rs->state = rs_connect_rdwr;
 		break;
 	default:
-		ret = ERR(EINVAL);
+		ret = (rs->state & rs_connected) ? 0 : ERR(EINVAL);
```
rleon: Isn't this code racy?
We can set rs_connected state inside rs_accept() and that function is not protected by lock.

shefty: This code is done under lock, so wouldn't say this change is racy.  It may be that other code is.

rs_accept() is a little different, in that the application does not have access to the rsocket being modified there until raccept returns.  rs_accept() is only called from the cm thread, and only that thread has access to the rsocket until after rsocket is placed on the accept queue.

shefty: To clarify, there are races in code like this (in rrecv):
```
	if (rs->state & rs_opening) {
		ret = rs_do_connect(rs);
```
The state is checked outside of a lock.  By the time rs_do_connect() runs, the rs->state has changed (to connected in this case) by the CM thread.  rs_do_connect repeats the state check, but under lock; however, it incorrectly assumes that the rsocket cannot already be connected.

rleon: Thanks for the clarification.

Indeed, I meant that race is between non-locked access to rs_connected and this line.

PR #688, opened 2020-02-04
ibacm: Fix bug in acm_get_ep()

Hakon-Bugge:
Commit 26e05f8304a5 ("ibacm: use ccan/list.h") introduced a bug in
acm_get_ep() by not incrementing inx.

Fixes: 26e05f8304a5 ("ibacm: use ccan/list.h")

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -898,6 +898,7 @@ static struct acmc_ep *acm_get_ep(int index, uint8_t port_num)
 			list_for_each(&dev->port[i].ep_list, ep, entry) {
 				if (index == inx)
 					return ep;
+				++inx;
 			}
 		}
 	}


ibacm/src/acm.c
```diff
@@ -896,7 +896,7 @@ static struct acmc_ep *acm_get_ep(int index, uint8_t port_num)
 			if (dev->port[i].state != IBV_PORT_ACTIVE)
 				continue;
 			list_for_each(&dev->port[i].ep_list, ep, entry) {
-				if (index == inx)
+				if (index == inx++)
```
rleon: Nice catch,
But please can put "inx++" after this "if (index) return ep;" so it will be less error prone.

Hakon-Bugge: That would look ugly, but resemble the original for loop better. But since the needless brackets are there, I can use them for something ;-)

Hakon-Bugge: v2 ready.

PR #694, opened 2020-02-09
efa: Fail context allocation in case create AH verb doesn't have udata

galpress:
The EFA provider cannot work correctly when there's no udata in
create_ah verb, as we must get the AH number from the kernel.

Without this patch, the provider assumes the AH number is zero for all
AHs, which results in unexpected behavior.
On context allocation, identify kernels that don't support udata and
exit.

### DIFF
--- a/providers/efa/efa.c
+++ b/providers/efa/efa.c
@@ -73,6 +73,10 @@ static struct verbs_context *efa_alloc_context(struct ibv_device *vdev,
 	ctx->max_llq_size = resp.max_llq_size;
 	pthread_spin_init(&ctx->qp_table_lock, PTHREAD_PROCESS_PRIVATE);
 
+	/* ah udata is mandatory for ah number retrieval */
+	if (!(ctx->cmds_supp_udata_mask & EFA_USER_CMDS_SUPP_UDATA_CREATE_AH))
+		goto err_free_spinlock;
+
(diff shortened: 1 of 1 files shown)


providers/efa/efa.c
```diff
@@ -73,6 +73,10 @@ static struct verbs_context *efa_alloc_context(struct ibv_device *vdev,
 	ctx->max_llq_size = resp.max_llq_size;
 	pthread_spin_init(&ctx->qp_table_lock, PTHREAD_PROCESS_PRIVATE);
 
+	/* ah udata is mandatory for ah number retrieval */
+	if (!(ctx->cmds_supp_udata_mask & EFA_USER_CMDS_SUPP_UDATA_CREATE_AH))
```
rleon: check_comp_mask()  for new code, please.

galpress: check_comp_mask isn't quite appropriate here, it's not a standard case where we fail due to unknown bit set. This case is making sure that the bit is set.
It could be written like:

```
	if (check_comp_mask(ctx->cmds_supp_udata_mask,
			    ~EFA_USER_CMDS_SUPP_UDATA_CREATE_AH))
		goto err_free_spinlock;
```

But it makes the code more cryptic, especially since the usual check is for if (!check_comp_mask), not if (check_comp_mask).

PR #696, opened 2020-02-11
Fix issues found by Coverity

Honggang-LI:


jgunthorpe:
Surprised gcc doesn't check the scanf format lengths like it does for snprintf/etc

PR #697, opened 2020-02-11
libhns: Update ibvqp->state in hns_roce_u_v2_modify_qp()

Li-Weihang:
The state of ibvqp will be updated after calling modify_qp(), because the
assignment of ibvqp->state is out of hns driver, it can't be guaranteed
that the configuration of ibvqp->state will take effect immediately. This
may cause hns_roce_u_v2_modify_qp() in post_send/post_recv process is
skipped.

Therefore, update value of ibvqp->state in modify_qp().

### DIFF
--- a/providers/hns/hns_roce_u_hw_v2.c
+++ b/providers/hns/hns_roce_u_hw_v2.c
@@ -1107,6 +1107,9 @@ static int hns_roce_u_v2_modify_qp(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 	if (ret)
 		return ret;
 
+	if (attr_mask & IBV_QP_STATE)
+		qp->state = attr->qp_state;
+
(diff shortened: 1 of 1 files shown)


providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -1091,6 +1091,9 @@ static int hns_roce_u_v2_modify_qp(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 	struct hns_roce_qp *hr_qp = to_hr_qp(qp);
 	bool flag = false; /* modify qp to error */
 
+	if (attr_mask & IBV_QP_STATE)
```
rleon: No way, that updating QP state before calling to modify_qp  is correct.

Li-Weihang: Thank you. I move it after modify_qp.

PR #699, opened 2020-02-13
libqedr: Fix user context allocation forward compatibility

mkalderon:
The user context alloc request structure introduced a new field
to indicate doorbell recovery is supported. This field was added so
that additional features could be added in the future by setting a
capability flag. However, the field wasn't zeroed, and was initialized
using "|=" instead of "=" leading to garbage in the other bits.
For forward compatability, we need to make sure all other bits are
zero.

Cc: stable@linux-rdma.org # v27 v28
Fixes: d9b2ba480af5 ("libqedr: Add support for Doorbell Overflow Recovery")
Signed-off-by: Ariel Elior <ariel.elior@marvell.com>

### DIFF
--- a/providers/qedr/qelr_main.c
+++ b/providers/qedr/qelr_main.c
@@ -181,7 +181,7 @@ static struct verbs_context *qelr_alloc_context(struct ibv_device *ibdev,
 	qelr_open_debug_file(ctx);
 	qelr_set_debug_mask();
 
-	cmd.context_flags |= QEDR_ALLOC_UCTX_DB_REC;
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Please add the cc stable stuff to the commit message, we want this fixed in v27 too

@nmorey

mkalderon:
> Please add the cc stable stuff to the commit message, we want this fixed in v27 too
> 
> @nmorey
added to commit message + pull request comments. 
I noticed there's already a v28, so added that as well. 
thanks,

PR #703, opened 2020-02-19
libhns: Not process return value of flushing cqe

Li-Weihang:
When a firmware reset occurred because of some errors, modify qp to error
in post_send() will be failed. The bad_wr will be returned to the user,
then the user won't wait for the corresponding wc. But as sq db has been
updated, the driver will return a software wc to the user, which may cause
unexpected behaviours.

Because the result of modifying qp is meaningless in this scenario, codes
to handle the return value are removed.

### DIFF
--- a/providers/hns/hns_roce_u_hw_v2.c
+++ b/providers/hns/hns_roce_u_hw_v2.c
@@ -916,9 +916,7 @@ out:
 		attr_mask = IBV_QP_STATE;
 		attr.qp_state = IBV_QPS_ERR;
 
-		ret = hns_roce_u_v2_modify_qp(ibvqp, &attr, attr_mask);
-		if (ret)
-			*bad_wr = wr;
+		hns_roce_u_v2_modify_qp(ibvqp, &attr, attr_mask);
 	}
 
 	return ret;
@@ -1019,9 +1017,7 @@ out:
 		attr_mask = IBV_QP_STATE;
 		attr.qp_state = IBV_QPS_ERR;
 
-		ret = hns_roce_u_v2_modify_qp(ibvqp, &attr, attr_mask);
-		if (ret)
-			*bad_wr = wr;
+		hns_roce_u_v2_modify_qp(ibvqp, &attr, attr_mask);
 	}
 
 	return ret;


providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -916,9 +916,7 @@ int hns_roce_u_v2_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 		attr_mask = IBV_QP_STATE;
 		attr.qp_state = IBV_QPS_ERR;
 
-		ret = hns_roce_u_v2_modify_qp(ibvqp, &attr, attr_mask);
-		if (ret)
-			*bad_wr = wr;
+		(void)hns_roce_u_v2_modify_qp(ibvqp, &attr, attr_mask);
```
rleon: Why do you need casting to void?
Do you see any compilation warnings which require you to put it?

Li-Weihang: Hi Leon, it is because of a warning reported by a inner static code analysis tool that all return values should be checked.
Is it ok to put the "(void)" here? If not, I will remove it. Thank you.

rleon: Yes, please remove.

Li-Weihang: OK, thank you.

Hakon-Bugge:
Not related to commit, but there are a handful of ret = EINVAL/ENOMEM (i.e. the positive values of the error code around here).

Hakon-Bugge:
> Not related to commit, but there are a handful of ret = EINVAL/ENOMEM (i.e. the positive values of the error code around here).

Which is correct since this is user space, so it is the negative ones that should be looked at.

Li-Weihang:
> > Not related to commit, but there are a handful of ret = EINVAL/ENOMEM (i.e. the positive values of the error code around here).
> 
> Which is correct since this is user space, so it is the negative ones that should be looked at.

Thanks for your reminder, will check and modify the negative return value in hns' codes in another patch :)

PR #705, opened 2020-02-20
mlx5: Expose packet pacing DV APIs

yishaih:
This series exposes packet pacing DV APIs, it enables allocating a device entry that can be used later on by DEVX commands.

The matching kernel part was sent into rdma-next.


tests/test_mlx5_pp.py
```diff
@@ -0,0 +1,57 @@
...
+        self.pp_res = Mlx5PPRes(self.dev_name)
+
+    def test_pp_alloc(self):
+        """
+        Allocate two packet pacing entries with the same configuration. One of
+        the entries is allocated with a dedicated index.
+        Then verify that the indexes are different and free the entries.
+        """
+        # An arbitrary valid rate limit value (in kbps)
+        rate_limit = (100).to_bytes(length=4, byteorder='big', signed=True)
```
jgunthorpe: Please use the struct module for this

EdwardSro: I'll use "rate_limit = struct.pack('>I', 100)" instead.
PR will be updated.

yishaih: PR was updated as asked including kernel-headers.

PR #707, opened 2020-02-20
rsockets: Do not overwrite connection failure error code

shefty:
In rs_do_connect, if the rsocket state is not connected, we set
errno = EINVAL and ret = -1.  Later in the code, we modify the
rsocket state to an error state and save errno with the rsocket.
Apps can obtain the error data through getsockopt.

This can result in overwriting real connection error data that
may have been saved with the rsocket.  The JDK catches this error
in a negative test case.  It reads the EINVAL error code from the
socket, when it expects ECONNREFUSED.

Update the code so that we only save the error with the socket if
it came from a connection error.  In other cases, simply set errno
and return a failure.  We add explicit state checks for the other
non-connected rsocket cases.

Signed-off-by: Sean Hefty <sean.hefty@intel.com>

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -1453,9 +1453,14 @@ connected:
 
 		rs->state = rs_connect_rdwr;
 		break;
+	case rs_connect_error:
+	case rs_disconnected:
+	case rs_error:
+		ret = ERR(ENOTCONN);
+		goto unlock;
 	default:
 		ret = (rs->state & rs_connected) ? 0 : ERR(EINVAL);
-		break;
+		goto unlock;
 	}
 
 	if (ret) {
@@ -1466,6 +1471,7 @@ connected:
 			rs->err = errno;
 		}
 	}
+unlock:
 	fastlock_release(&rs->slock);
 	return ret;
 }


librdmacm/rsocket.c
```diff
@@ -1453,9 +1453,14 @@ static int rs_do_connect(struct rsocket *rs)
...
 		break;
+	case rs_connect_error:
+	case rs_disconnected:
+	case rs_error:
+		ret = ERR(ENOTCONN);
+		goto unlock;
 	default:
 		ret = (rs->state & rs_connected) ? 0 : ERR(EINVAL);
-		break;
+		goto unlock;
```
jgunthorpe: Is it OK to not change the state to rs_connect_error on both of these unwinds?

shefty: I believe so.  In the added code, the socket is in the wrong state for processing, so the call just returns an error.    In the other cases, we've handled an event that's driving the socket state.  If in the processing of that event an error occurs, the connection failed (connect_error).

If we are in the disconnected or error state, then some other thread finished establishing the connection and tore it down before this thread could execute.  So it would be incorrect to change the state to report a connect_error.

I think the only viable state we can be in in the default case now is connected.

PR #708, opened 2020-02-23
ibdiag: Compare CA device names by using the maximum length between them

haimbo:
Compare CA device names by using the maximum length between them in order to
avoid matches of sub strings between CA device names that can match the
wrong device.

Example before the fix:
Input CA name: abc_11
List CA names: abc_1, def_1
When running : ibstat abc_11, the output will be the details of abc_1 even
though abc_11 does not exist.

After the fix, the output of this case will be:
'abc_11' IB device can't be found

### DIFF
--- a/infiniband-diags/ibstat.c
+++ b/infiniband-diags/ibstat.c
@@ -311,8 +311,7 @@ int main(int argc, char *argv[])
 
 	if (argc) {
 		for (node = device_list; node; node = node->next)
-			if (!strncmp(node->ca_name, argv[0],
-				     strlen(node->ca_name)))
+			if (!strcmp(node->ca_name, argv[0]))
 				break;
 		if (!node)
 			IBPANIC("'%s' IB device can't be found", argv[0]);


infiniband-diags/ibstat.c
```diff
@@ -310,10 +312,14 @@ int main(int argc, char *argv[])
 		IBPANIC("can't list IB device names");
 
 	if (argc) {
-		for (node = device_list; node; node = node->next)
-			if (!strncmp(node->ca_name, argv[0],
```
rleon: strcmp() instead of strncmp() will do the trick.

 15 int main()                                                                                                         
 16 {                                                                                                                 
 17         char str1[] = "abc";                                                                                      
 18         char str2[] = "abcd";                                                                                     
 19                                                                                                                   
 20         printf("%d\n", strcmp(str1, str2));                                                                       
 21         printf("%d\n", strcmp(str2, str1));                                                                       
 22                                                                                                                   
 23         return 0;                                                                                                 
 24 }

And add fixes line please.

PR #712, opened 2020-03-03
Clean rdmacm

rleon:


librdmacm/cma.c
```diff
@@ -1526,6 +1526,8 @@ static void ucma_copy_conn_param_to_kern(struct cma_id_private *id_priv,
 
 int rdma_connect(struct rdma_cm_id *id, struct rdma_conn_param *conn_param)
 {
+	uint32_t qp_num = (conn_param) ? conn_param->qp_num : 0;
+	uint8_t srq = (conn_param) ? conn_param->srq : 0;
```
Hakon-Bugge: Wy parentheses around conn_param?

librdmacm/cma.c
```diff
@@ -1547,18 +1549,13 @@ int rdma_connect(struct rdma_cm_id *id, struct rdma_conn_param *conn_param)
...
-					     (id->qp->srq != NULL));
-	} else if (conn_param) {
-		ucma_copy_conn_param_to_kern(id_priv, &cmd.conn_param,
-					     conn_param, conn_param->qp_num,
-					     conn_param->srq);
-	} else {
-		ucma_copy_conn_param_to_kern(id_priv, &cmd.conn_param,
-					     conn_param, 0, 0);
+		qp_num = id->qp->qp_num;
+		srq = id->qp->srq != NULL;
```
Hakon-Bugge: srq = !! id->qp->srq

librdmacm/cma.c
rleon: habbit, i can remove

librdmacm/cma.c
rleon: will change

rleon:
Fixed, thanks

rleon:
> Guess you had a small slip from your fingers.

Thanks, it is always the case when you need to do gazillion things in parallel.

rleon:
> Very nitty, but if you look at the old code, you can drive the assertion that if(!id->qp) then conn_param != NULL (since it being de-referenced in the call to ucma_copy_conn_param_to_kern().
> 
> Hence I wonder, can we simply have:
> 
> uint32_t qp_num = id->qp ? id->qp->qp_num : conn_param->qp_num;
> (ditto for srq).
> and avoid the "if (id->qp)" below?

It is only partially true, the original code 

        if (id->qp) {
.....
       } else if (conn_param) {
...
       } else {
.... no QP and no conn_param
}

Hakon-Bugge:
> > Very nitty, but if you look at the old code, you can drive the assertion that if(!id->qp) then conn_param != NULL (since it being de-referenced in the call to ucma_copy_conn_param_to_kern().
> > Hence I wonder, can we simply have:
> > uint32_t qp_num = id->qp ? id->qp->qp_num : conn_param->qp_num;
> > (ditto for srq).
> > and avoid the "if (id->qp)" below?
> 
> It is only partially true, the original code
> 
> ```
>     if (id->qp) {
> ```
> 
> .....
> } else if (conn_param) {
> ...
> } else {
> .... no QP and no conn_param
> }

In rdma_ connect(), yes. But in rdma_accept(), I only see if() .. else ..

rleon:
I did it symmetrically, do you want me to change?

Hakon-Bugge:
> I did it symmetrically, do you want me to change?

Yes. Keep it as simple as possible.

PR #713, opened 2020-03-03
libibverbs: Fix query_device_ex dummy function not to return EOPNOTSUPP

mkalderon:
Following thread on broken compatibility between v28 and older version:
https://www.spinics.net/lists/linux-rdma/msg89386.html

Implement query_device_ex dummy function for all providers.
Returning EOPNOTSUPP broke compatibility with apps compiled against
older rdma-core

Suggested-by: Jason Gunthorpe <jgg@mellanox.com>
Cc: stable@linux-rdma.org # v28
Fixes: commit c28410765bdf ("libibverbs: Fix incorrect return code")

### DIFF
--- a/libibverbs/dummy_ops.c
+++ b/libibverbs/dummy_ops.c
@@ -369,11 +369,22 @@ static int query_device(struct ibv_context *context,
 	return EOPNOTSUPP;
 }
 
+/* Provide a generic implementation for all providers that don't implement
+ * query_device_ex.
+ */
 static int query_device_ex(struct ibv_context *context,
 			   const struct ibv_query_device_ex_input *input,
 			   struct ibv_device_attr_ex *attr, size_t attr_size)
 {
-	return EOPNOTSUPP;
+	if (input && input->comp_mask)
+		return EINVAL;
+
+	if (attr_size < sizeof(attr->orig_attr))
+		return EOPNOTSUPP;
+
+	memset(&attr->orig_attr, 0, sizeof(attr->orig_attr));
+
+	return ibv_query_device(context, &attr->orig_attr);
 }
 
 static int query_port(struct ibv_context *context, uint8_t port_num,
--- a/libibverbs/man/ibv_query_device_ex.3
+++ b/libibverbs/man/ibv_query_device_ex.3
@@ -3,18 +3,23 @@
 .\"
 .TH IBV_QUERY_DEVICE_EX 3 2014-12-17 libibverbs "Libibverbs Programmer's Manual"
 .SH "NAME"
-ibv_query_device_ex \- query an RDMA device's attributes
+ibv_query_device_ex \- query an RDMA device's attributes including extended
+device properties.
 .SH "SYNOPSIS"
 .nf
 .B #include <infiniband/verbs.h>
 .sp
 .BI "int ibv_query_device_ex(struct ibv_context " "*context",
+.BI "                        struct ibv_query_device_ex_input " "*input",
 .BI "                        struct ibv_device_attr_ex " "*attr" );
 .fi
 .SH "DESCRIPTION"
 .B ibv_query_device_ex()
 returns the attributes of the device with context
 .I context\fR.
+The argument
+.I input
+is a pointer to an ibv_query_device_ex_input structure, used for future extensions
 The argument
 .I attr
 is a pointer to an ibv_device_attr_ex struct, as defined in <infiniband/verbs.h>.
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -2962,12 +2962,15 @@ ibv_query_device_ex(struct ibv_context *context,
 	struct verbs_context *vctx;
 	int ret;
 
+	if (input && input->comp_mask)
+		return EINVAL;
+
 	vctx = verbs_get_ctx_op(context, query_device_ex);
 	if (!vctx)
 		goto legacy;
 
 	ret = vctx->query_device_ex(context, input, attr, sizeof(*attr));
-	if (ret == EOPNOTSUPP)
+	if (ret == EOPNOTSUPP || ret == ENOSYS)
 		goto legacy;
 
 	return ret;


libibverbs/dummy_ops.c
```diff
@@ -369,11 +369,15 @@ static int query_device(struct ibv_context *context,
...
 
+/* Provide a generic implementation for all providers that don't implement
+ * query_device_ex.
+ */
 static int query_device_ex(struct ibv_context *context,
 			   const struct ibv_query_device_ex_input *input,
 			   struct ibv_device_attr_ex *attr, size_t attr_size)
 {
-	return EOPNOTSUPP;
+	memset(attr, 0, sizeof(*attr));
```
jgunthorpe: This has to be attr_size not sizeof(attr) right?

jgunthorpe: Or maybe it should be like

if (attr_size < sizeof(attr->orig_attr)) return EOPNOTSUPP;
memset(&attr->orig_attr, 0, sizeof(attr->orig_attr))
return ibv_query_device(..)

jgunthorpe: And also it has to check that the incoming comp mask is zero and fail.

mkalderon: yes, attr_size and not sizeof. 
attr_size  < sizeof(attr->orig_attr) will happen only if ibv_device_attr will change, but that's why _ex was added, can add the check, but then it should probably be added for all calls in providers that do implement the callback as well?

libibverbs/dummy_ops.c
```diff
@@ -369,11 +369,22 @@ static int query_device(struct ibv_context *context,
...
 
+/* Provide a generic implementation for all providers that don't implement
+ * query_device_ex.
+ */
 static int query_device_ex(struct ibv_context *context,
 			   const struct ibv_query_device_ex_input *input,
 			   struct ibv_device_attr_ex *attr, size_t attr_size)
 {
-	return EOPNOTSUPP;
+	if (input && input->comp_mask)
```
jgunthorpe: don't test for null, crashing is the right thing to do

jgunthorpe: Actually, I'm not sure about the comp_mask any more, why doesn't the wrapper test it?
This is one of those things that doesn't make any sense :(
Stick with copying the wrapper, no sense in making it more complicated if we ever want to use this thing.

mkalderon: Why is crashing the right thing to do? 

which wrapper are you referring to? currently the ibv_cmd_query_device_ex is checking comp_mask -> do you want to take it out and place it in the ibv_query_device_ex?

jgunthorpe: Sigh, this is a big mess it seems.

There is already a fallback wrapper for query_device_ex in verbs.h:

```
static inline int
ibv_query_device_ex(struct ibv_context *context,
                    const struct ibv_query_device_ex_input *input,
                    struct ibv_device_attr_ex *attr)
{
        struct verbs_context *vctx;
        int ret;

        vctx = verbs_get_ctx_op(context, query_device_ex);
        if (!vctx)
                goto legacy;

        ret = vctx->query_device_ex(context, input, attr, sizeof(*attr));
        if (ret == EOPNOTSUPP)
                goto legacy;

        return ret;

legacy:
        memset(attr, 0, sizeof(*attr));
        ret = ibv_query_device(context, &attr->orig_attr);

        return ret;
}
```

It became broken when the ret was changed - the inline wrappers really should rarely be changed.

So the logical step here is to make the op never NULL for new libibverbs and handle the fallback in verbs directly, as you are doing here.

But this wrapper doesn't test the input comp mask, while the cmd caller does, so it makes zero sense.

Let us go for worst case consistency then..

1) Fix ibv_query_device_ex inline wrapper to have

if (ret == EOPNOTSUPP || ret == ENOSYS)
    goto legacy;

This will allow new builds to work with old libibverbs again

2) Change the dummy query_device_ex to not fail and internally do to fallback. This will make old builds work with new libibverbs again

3) Make *both* the wrapper and the dummy match ibv_cmd_querty_device on the legacy path - check for input and comp_mask - we  can't have different flows with different requirements.

4) Continue to allow input to be NULL, I suppose this is ABI now.

5) Fix the man page for ibv_query_device_ex.3 which doesn't even have the right function signature or document really anything. Grr.

@Kamalheib your change stumbled into quite the mess here :(

Kamalheib: > @Kamalheib your change stumbled into quite the mess here :(

@jgunthorpe: Sorry to see that, well the change was suggested by you :/, please see the following:

https://github.com/linux-rdma/rdma-core/pull/659#discussion_r366458194

mkalderon: > Sigh, this is a big mess it seems.
> 
I agree, 

> 
> Let us go for worst case consistency then..
> 
> 1. Fix ibv_query_device_ex inline wrapper to have
> 
> if (ret == EOPNOTSUPP || ret == ENOSYS)
> goto legacy;
> 
> This will allow new builds to work with old libibverbs again
you'll still have new builds return "EOPNOTSUPP" for ibv_query_device_ex and old libibverbs
will check for ENOSYS -> failing the operation.

> 
> 1. Change the dummy query_device_ex to not fail and internally do to fallback. This will make old builds work with new libibverbs again
do the fallback inside the query_device_ex ? 
that's what I did in the current pull-request.

> 2. Make _both_ the wrapper and the dummy match ibv_cmd_querty_device on the legacy path - check for input and comp_mask - we  can't have different flows with different requirements.
I don't follow here -> take the check outside of cmd_query? into the wrapper ? or leave it in the query_device as well ? 

> 3. Continue to allow input to be NULL, I suppose this is ABI now.
> 4. Fix the man page for ibv_query_device_ex.3 which doesn't even have the right function signature or document really anything. Grr.
> 
> @Kamalheib your change stumbled into quite the mess here :(

jgunthorpe: #1 - The change in verbs.h must be as I said, ret == ENOSYS || ret == EOPNOTSUPP, can't delete that code.
#2 - Put it in all three places

mkalderon:
Jason, Sorry for the delay on this one... 
A few comments / questions
1. I made the changes as you requested
2. Am able to run old app against new rdma-core succesfully
3. I only modified the signature in man pages -> not sure what exactlly you would like me to elaborate about
4. While making the change something popped up - I noticed that throughout rdma-core providers and some core, in some cases error are returned negative and in some not, such as -EINVAL or EINVAL, whether return codes are negative or positive is inconsistent -> do you have a guideline for this ?

jgunthorpe:
errno's need to follow whatever is written in the man page

nmorey:
Picked for stable-v28.
Older branches did not get the change to ENOTSUPP as I guessed (right apparently) it would break compatibility

PR #715, opened 2020-03-04
tests: Add parameterizing to PyverbsAPITestCase

Kamalheib:
Add support for providing a device name parameter to the tests that
inherits from PyverbsAPITestCase, This is needed to give the users the
ability to run the tests on a specific device on systems with multiple
devices.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -33,6 +33,31 @@ TIMEOUT = 14
 
 
 class PyverbsAPITestCase(unittest.TestCase):
+    """
+    A base class for test cases which provides the option for user parameters.
+    These can be provided by manually adding the test case to the runner:
+    suite = unittest.TestSuite()
+    ... # Regular auto-detection of test cases, no parameters used.
+    # Now follows your manual addition of test cases e.g:
+    suite.addTest(PyverbsAPITestCase.parametrize(<TestCaseName>, dev_name='..'))
+    """
+    def __init__(self, methodName='runTest', dev_name=None):
+        super(PyverbsAPITestCase, self).__init__(methodName)
+        self.dev_name = dev_name
+
+    @staticmethod
+    def parametrize(testcase_klass, dev_name=None):
+        """
+        Create a test suite containing all the tests from the given subclass
+        with the given dev_name.
+        """
+        loader = unittest.TestLoader()
+        names = loader.getTestCaseNames(testcase_klass)
+        suite = unittest.TestSuite()
+        for n in names:
+            suite.addTest(testcase_klass(n, dev_name=dev_name))
+        return suite
+
     def setUp(self):
         """
         Opens the devices and queries them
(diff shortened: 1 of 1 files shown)


tests/base.py
```diff
@@ -33,6 +33,31 @@
 
 
 class PyverbsAPITestCase(unittest.TestCase):
+    """
+    A base class for test cases which provides the option for user parameters.
+    These can be provided by manually adding the test case to the runner:
+    suite = unittest.TestSuite()
+    ... # Regular auto-detection of test cases, no parameters used.
+    # Now follows your manual addition of test cases e.g:
+    suite.addTest(PyverbsAPITestCase.parametrize(<TestCaseName>, dev_name='..'))
```
jgunthorpe: I would really prefer to have the test configuration be done with a command line parameter rather than these sketchy directions to edit the test runner directly.

We should have very few test configuration options, mainly the default ib_device ?

EdwardSro: > I would really prefer to have the test configuration be done with a command line parameter rather than these sketchy directions to edit the test runner directly.

Agree. 
 
> We should have very few test configuration options, mainly the default ib_device ?

Currently, this "parametrize" method is used in RDMATestCase. In that case, there's ib_port, gid_index and pkey_index in addition to the ib_device.

jgunthorpe: I'm not totally sure of the value of parameterizing gid and pkey index. I can't think of many tests that are sensitive to the pkey, and the main gid sensitivity is gid type - so the tests should be designed to get the various gid types directly (and skip if not available).. But it is fine to keep going where we are and just move a few things to the command line.

Kamalheib: @jgunthorpe Sorry for the late reply..., when you said that moving things to the command line, you meant to override the flags that provided by unittest parser?

```
./build/bin/run_tests.py -h
usage: run_tests.py [-h] [-v] [-q] [--locals] [-f] [-c] [-b] [-k TESTNAMEPATTERNS] [tests [tests ...]]

positional arguments:
  tests                a list of any number of test modules, classes and test methods.

optional arguments:
  -h, --help           show this help message and exit
  -v, --verbose        Verbose output
  -q, --quiet          Quiet output
  --locals             Show local variables in tracebacks
  -f, --failfast       Stop on first fail or error
  -c, --catch          Catch Ctrl-C and display results so far
  -b, --buffer         Buffer stdout and stderr during tests
  -k TESTNAMEPATTERNS  Only run tests which match the given substring

Examples:
  run_tests.py                           - run default set of tests
  run_tests.py MyTestSuite               - run suite 'MyTestSuite'
  run_tests.py MyTestCase.testSomething  - run MyTestCase.testSomething
  run_tests.py MyTestCase                - run all 'test*' test methods
                                       in MyTestCase
```

jgunthorpe: Yes, I've done this in the past, but I can't recall how I did it right now...

Kamalheib: > Yes, I've done this in the past, but I can't recall how I did it right now...

Just to confirm, when I say override, I mean to add a new parser into run_test.py that will parse the command-line arguments that will be passed to run_test.py (e.g. the following).

./build/bin/run_tests.py -h
usage: run_tests.py [-h] [-d DEV_NAME]

optional arguments:
  -h, --help            show this help message and exit
  -d DEV_NAME, --dev_name DEV_NAME
                        RDMA device name to run the tests on 

I'm also wondering if it is ok to lose the functionality of running a specific set of tests using the -k flag?

rleon: > I'm also wondering if it is ok to lose the functionality of running a specific set of tests using the -k flag?

What do you mean?

Kamalheib: > > I'm also wondering if it is ok to lose the functionality of running a specific set of tests using the -k flag?
> 
> What do you mean?

I mean that if we are going to write a new parser in run_test.py, this means we are losing the parser that unittest is providing.

rleon: I'm using -k option all the time and will be happy if we preserve this option.

Kamalheib:
Closing this PS as the patch is now part of  #736

PR #719, opened 2020-03-11
libhns: Various cleanups

Li-Weihang:
These are small cleanups that won't affect the logic of code.


providers/hns/hns_roce_u_verbs.c
```diff
@@ -67,8 +67,8 @@ int hns_roce_u_query_device(struct ibv_context *context,
 	if (ret)
 		return ret;
 
-	major	   = (raw_fw_ver >> 32) & 0xffff;
-	minor	   = (raw_fw_ver >> 16) & 0xffff;
+	major	  = (raw_fw_ver >> 32) & 0xffff;
```
rleon: you didn't remove all space here and in the line below.

Li-Weihang: Oh, sorry, I thought it's a tab.

PR #720, opened 2020-03-12
azp: Fix azp image build

galpress:
For some reason the apt-get update command fails with the following
error message:
W: http://apt.llvm.org/bionic/dists/llvm-toolchain-bionic-8/InRelease: No system certificates available. Try installing ca-certificates.
W: http://apt.llvm.org/bionic/dists/llvm-toolchain-bionic-8/Release: No system certificates available. Try installing ca-certificates.
E: The repository 'http://apt.llvm.org/bionic llvm-toolchain-bionic-8 Release' does not have a Release file.

Which fails the entire build.
However, everything else works correctly if continuing instead of
exiting because of the &&.

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -257,7 +257,7 @@ class APTEnvironment(Environment):
     build_python = True;
     def get_docker_file(self,tmpdir):
         res = DockerFile(self.docker_parent);
-        res.lines.append("RUN apt-get update && apt-get install -y --no-install-recommends %s && apt-get clean && rm -rf /usr/share/doc/ /usr/lib/debug /var/lib/apt/lists/"%(
+        res.lines.append("RUN apt-get update; apt-get install -y --no-install-recommends %s && apt-get clean && rm -rf /usr/share/doc/ /usr/lib/debug /var/lib/apt/lists/"%(
             " ".join(sorted(self.pkgs))));
         return res;
 


galpress:
BTW, I tried installing ca-certificates as the message suggest, but it's not possible before calling apt-get update so it doesn't really matter.

jgunthorpe:
Sounds like the crazy apt-key magic didn't work for you? Are there other errors/warnings at the top of the run? I vaguely recall there were problems with old gpg versions.

I also fixed at least one problem that caused this recently (dd131810af36aa018030eaa3a700d991f37690a7) - related to umask troubles.

galpress:
The only thing that looks like an error is:
```
  Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. [IP: X.X.X.X 443]
```

Note that this was noticed on our CI environment, which always creates a new instance using the same image, which makes it even more strange that it started failing all of a sudden. Nothing really changed and it just stopped working.

jgunthorpe:
Hm, it worked for me here, so the base docker container or keys on the web are not invalid.

Although I think that apt.llvm.org changed to enforce https, so maybe that is having some impact?

This is why we don't build the images in the CI any more. A year doing that in travis showed it is not reliable.

PR #722, opened 2020-03-16
tests: Avoid RoCEv2 GIDs on unsupported devices

EdwardSro:
Avoid RoCEv2 GIDs on unsupported devices.
To be able to do that, ibv_query_gid_type was added to Pyverbs' infrastructure.


rleon:
Please delete changeid

EdwardSro:
> Please delete changeid

Removed

PR #727, opened 2020-03-23
mlx5: Introduce dynamic UAR allocation mode 

yishaih:
This series enables a dynamic allocation and management of a UAR.

Moving to that mode enables allocating a UAR only upon demand and drop the redundant static allocation of UARs upon context creation. In addition, it allows master and secondary processes that own the same command FD to allocate and manage UARs according to their needs, this can’t be achieved today.

As part of this option, QP & CQ creation flows were adapted to support this dynamic UAR mode once supported by the kernel.


providers/mlx5/verbs.c
```diff
@@ -4461,30 +4635,43 @@ struct mlx5dv_devx_uar *mlx5dv_devx_alloc_uar(struct ibv_context *context,
...
+	if (!bf) {
+		if (flags || !(ctx->flags & MLX5_CTX_FLAGS_NO_KERN_DYN_UAR))
+			return NULL;
 
-	fill_attr_in_uint32(cmd, MLX5_IB_ATTR_DEVX_QUERY_UAR_USER_IDX,
-			    bf->bfreg_dyn_index);
-	fill_attr_out_ptr(cmd, MLX5_IB_ATTR_DEVX_QUERY_UAR_DEV_IDX,
-		      &bf->devx_uar.dv_devx_uar.page_id);
+		/* fallback to legacy mode */
+		bf = mlx5_attach_dedicated_bf(context);
```
jgunthorpe: Why is this fallback here when the change to mlx5_alloc_td just switches mlx5_attach_dedicated_bf() to mlx5_attach_dedicated_uar() ?

providers/mlx5/mlx5.c
```diff
@@ -1386,6 +1386,10 @@ static struct verbs_context *mlx5_alloc_context(struct ibv_device *ibdev,
 		}
 	}
 
+	context->cq_uar = mlx5_attach_dedicated_uar(&v_ctx->context,
```
jgunthorpe: Why is it dedicated?

providers/mlx5/verbs.c
```diff
@@ -4572,6 +4664,16 @@ void clean_dyn_uars(struct ibv_context *context)
...
+		list_del(&bf->uar_entry);
+		mlx5_free_uar(context, bf);
+	}
+
+	list_for_each_safe(&ctx->dyn_uar_qp_shared_list, bf, tmp_bf, uar_entry) {
+		list_del(&bf->uar_entry);
+		mlx5_free_uar(context, bf);
+	}
+
 	if (ctx->cq_uar)
```
jgunthorpe: cq doesn't get dynamic too?

providers/mlx5/verbs.c
```diff
@@ -325,14 +325,97 @@ static void mlx5_insert_dyn_uuars(struct mlx5_context *ctx,
...
+	pthread_mutex_lock(&ctx->dyn_bfregs_mutex);
+
+retry_dedicated:
+	bf = list_pop(&ctx->dyn_uar_qp_dedicated_list, struct mlx5_bf, uar_entry);
+	if (!bf) {
+		if (ctx->qp_alloc_dedicated_uuars < ctx->qp_max_dedicated_uuars) {
+			if (mlx5_alloc_qp_uar(context, true))
+				goto end;
+
+			goto retry_dedicated;
```
jgunthorpe: Use a loop instead of these gotos

providers/mlx5/verbs.c
yishaih: The TD is moved to use mlx5_attach_dedicated_uar() in the next patch upon adapting legacy mode to new one, by that step the fallback is really done internally.

providers/mlx5/mlx5.c
yishaih: This UAR is dedicated for all CQs in the given context comparing other potential customers for such a UAR (QP, DEVX).

providers/mlx5/verbs.c
yishaih: It gets when kernel supports the new mode, otherwise the static one for CQs will be used.

providers/mlx5/verbs.c
yishaih: Done, thanks.
A new series with this change and up-to-date kernel-header patch was uploaded.

providers/mlx5/mlx5.c
jgunthorpe: CQ can share with anything except BF users...

providers/mlx5/verbs.c
jgunthorpe: So is it deleted in a later patch? It looks weird

yishaih: Yes,  I preferred this incremental mode as of the massive cleanup/refactor in the next patch for the legacy mode to clarify things.

providers/mlx5/mlx5.c
yishaih: Right, this can be some candidate for further optimization to share with QPs with NC mode for example, for now it follows current logic to be for CQs.

PR #729, opened 2020-03-25
mlx5: Add RDMA_TX flow table type

yishaih:
This series adds new RDMA_TX flow type, flow steering rules in this flow type are used to filter transmitted RDMA traffic.

The matching kernel part was sent to rdma-next.


yishaih:
Kernel-headers patch was updated to the accepted kernel stuff.

PR #730, opened 2020-03-25
ibacm: check provider file ends with .so extension

markhaywood:
acm_open_providers() reads filenames and checks to see if the filenames
contain the .so (shared object) extension via strstr(). But this does
not verify that the extension is found at the end of the filename - only
that .so is a substring somewhere in the filename. This means filenames
of the sort "libibacmp.so.org" will be matched and loaded as providers.
The check should be modified to verify that the extension is at the end
of the filename.

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -2832,6 +2832,17 @@ static void acm_load_prov_config(void)
 	}
 }
 
+static int acm_string_end_compare(const char *s1, const char *s2)
+{
+	size_t s1_len = strlen(s1);
+	size_t s2_len = strlen(s2);
+
+	if (s1_len < s2_len)
+		return -1;
+
+	return strcmp(s1 + s1_len - s2_len, s2);
+}
+
 static int acm_open_providers(void)
 {
 	DIR *shlib_dir;
@@ -2854,7 +2865,7 @@ static int acm_open_providers(void)
 	}
 
 	while ((dent = readdir(shlib_dir))) {
-		if (!strstr(dent->d_name, ".so"))
+		if (acm_string_end_compare(dent->d_name, ".so"))
 			continue;
 
 		if (!check_snprintf(file_name, sizeof(file_name), "%s/%s",


ibacm/src/acm.c
```diff
@@ -2832,6 +2832,17 @@ static void acm_load_prov_config(void)
...
+	size_t s1_len = strlen(s1);
+	size_t s2_len = strlen(s2);
+
+	if (s1_len < s2_len)
+		return -1;
+
+	return strcmp(s1 + s1_len - s2_len, s2);
+}
+
 static int acm_open_providers(void)
```
rleon: What does it do?
Do you expect for specific names that include ".so"?

markhaywood: Yes. It requires that provider names end with .so. I believe that was always the intent.

markhaywood: In more detail ...

ibacm reads the provider library directory looking for provider libraries. When it finds one, it loads it. The function that performs this task is acm_open_providers(). For each file that it reads from the provider library directory, ibacm performs some sanity checking on the file to see if it seems reasonable to assume that it is a provider library before attempting to load it. Currently, one of the first things that it is doing is checking to see of the filename contains ".so" and if it does not, then it does not load it:

if (!strstr(dent->d_name, ".so"))
        continue;

But this check allows names like "libibacmp.so.org" to pass the check. I believe the intent was that the filename would end in ".so" (e.g., "libibacmp.so").

The change I am proposing is to use he function acm_string_end_compare() verify that the filename ends with ".so". 

acm_string_end_compare() will return a zero if the string s1 ends with the substring s2. Otherwise it returns a non-zero value.

Does this answer your question?

markhaywood: Are you asking for a less generic function? One that only checks to see if the string ends in ".so"? Something more like:

static int acm_is_sharedlib(const char *name)
{
        char ext[] = ".so";
        size_t ext_len = sizeof(".so") - 1;
        size_t name_len = strlen(name);

        if (name_len > ext_len && strcmp(name + name_len - ext_len, ext) == 0)
                return 0;

        return 1;
}

rleon: I have no idea and asked because it is not really UNIX way to rely on specific extension.

jgunthorpe:
@markhaywood  Can you make sure the plugins run 'in place' from a build.sh build. I remember hearing that these checks are also failing for that case. Everything in rdma-core should be able to run completely without installation.

PR #731, opened 2020-03-29
tests: Fix check for unsupported RoCEv2 devices

Kamalheib:
Avoid the following exception when running RDMATestCase tests over none
Mellanox devices:

  File "/home/rdma-core/tests/test_cq_events.py", line 26, in setUp
    super().setUp()
  File "/home/rdma-core/tests/base.py", line 129, in setUp
    self._add_gids_per_device(ctx, dev_name)
  File "/home/rdma-core/tests/base.py", line 161, in _add_gids_per_device
    self._add_gids_per_port(ctx, dev, port+1)
  File "/home/rdma-core/tests/base.py", line 154, in _add_gids_per_port
    vendor_pid in ROCEV2_UNSUPPORTED_DEVS[dev_attrs.vendor_id]:
KeyError: 5348

Fixes: f65543c15ea4 ("tests: Avoid RoCEv2 GIDs on unsupported devices")

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -39,6 +39,10 @@ ROCEV2_UNSUPPORTED_DEVS = {MLNX_VENDOR_ID: [CX3Pro_MLNX_PART_ID,
                                             CX3_MLNX_PART_ID]}
 
 
+def has_roce_hw_bug(vendor_id, vendor_part_id):
+    return vendor_part_id in ROCEV2_UNSUPPORTED_DEVS.get(vendor_id, [])
+
+
 class PyverbsAPITestCase(unittest.TestCase):
     def setUp(self):
         """
@@ -142,6 +146,7 @@ class RDMATestCase(unittest.TestCase):
         if port_attrs.state != e.IBV_PORT_ACTIVE:
             return
         dev_attrs = ctx.query_device()
+        vendor_id = dev_attrs.vendor_id
         vendor_pid = dev_attrs.vendor_part_id
         for idx in range(port_attrs.gid_tbl_len):
             gid = ctx.query_gid(port, idx)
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Why on earth do we need this anyhow? @EdwardSro ? Devices that do not support rocve2 should not be creating v2 gids. Is this because there is no way to find the gid type?

EdwardSro:
> Why on earth do we need this anyhow? @EdwardSro ? Devices that do not support rocve2 should not be creating v2 gids. Is this because there is no way to find the gid type?

We can have a gid of rocev2 type, but we can't know whether the device supports rocev2 or not.
At least that's the case with the device that was added to the "ROCEV2_UNSUPPORTED_DEVS" list, and any other device with the same behavior should be added there. Unless you have a better suggestion.

jgunthorpe:
@EdwardSro oh this is that HW bug in older chips, it should be made clearer, this is not some general mechanism, it is blacklist for a single broken silicon.

EdwardSro:
Yup, it's due to that HW bug. I didn't want to make some comments in Pyverbs that tell stories related to some specific HW bugs and wanted to make the fix as general as possible. But maybe I should have made it more clear in the related commit message.

@Kamalheib Thanks for the fix :)
btw, you can use ".get(key, [])" which will return an empty array in case the key was not found.

jgunthorpe:
You should write it as 'def has_hw_bug' so people can understand what this is and not think it is some expected behavior from verbs.

Kamalheib:
@jgunthorpe & @EdwardSro please review the latest change.

tests/base.py
```diff
@@ -38,6 +38,8 @@
 ROCEV2_UNSUPPORTED_DEVS = {MLNX_VENDOR_ID: [CX3Pro_MLNX_PART_ID,
                                             CX3_MLNX_PART_ID]}
```
EdwardSro: please surround top-level functions and class definitions with two blank lines (PEP conventions)

tests/base.py
```diff
@@ -38,6 +38,8 @@
...
 
+def has_hw_bug(vendor_id, vendor_part_id):
```
EdwardSro: The function name is too generic, let's change it to something like "has_roce_hw_bug"

rleon: maybe has_roce_version_hw_bug() ?

tests/base.py
Kamalheib: OK, I will fix it.

tests/base.py
Kamalheib: I think has_roce_hw_bug() makes more sense, I'll update the patch.

PR #732, opened 2020-03-30
ibdiag: Set the right subnet prefix when using the -G option

haimbo:
Use local port subnet prefix when constructing PathRecord request
while resolving destination.

### DIFF
--- a/infiniband-diags/ibdiag_common.c
+++ b/infiniband-diags/ibdiag_common.c
@@ -658,7 +658,7 @@ static int resolve_guid(char *ca_name, uint8_t ca_port, ib_portid_t *portid,
 {
 	ib_portid_t tmp;
 	uint8_t buf[IB_SA_DATA_SIZE] = { 0 };
-	uint64_t prefix;
+	__be64 prefix;
 	ibmad_gid_t selfgid;
 
 	if (!sm_id) {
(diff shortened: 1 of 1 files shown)


infiniband-diags/ibdiag_common.c
```diff
@@ -670,10 +672,15 @@ static int resolve_guid(char *ca_name, uint8_t ca_port, ib_portid_t *portid,
...
-		mad_set_field64(portid->gid, 0, IB_GID_PREFIX_F,
-				IB_DEFAULT_SUBN_PREFIX);
+	memcpy(&prefix, selfgid, sizeof(prefix));
+	gid_prefix = htobe64(prefix);
+	prefix = IB_DEFAULT_SUBN_PREFIX;
+	if (gid_prefix)
+		p_gid_prefix = &gid_prefix;
+	else
+		p_gid_prefix = &prefix;
+	mad_encode_field(portid->gid, IB_GID_PREFIX_F,
```
jgunthorpe: What? Why is this like this? It should be:

```
__be64 prefix;

memcpy(&prefix, portid->gid, sizeof(prefix));
mad_set_field64(portid->gid, 0, IB_GID_PREFIX_F,
￼				prefix:be64toh(prefix)?IB_DEFAULT_SUBN_PREFIX);
```

haimbo: Fixed

PR #733, opened 2020-03-31
tests: Change DeviceTest to inherit from PyverbsAPITestCase

Kamalheib:
Modify the DeviceTest class to inherit from PyverbsAPITestCase.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -15,16 +15,16 @@ import pyverbs.device as d
 PAGE_SIZE = resource.getpagesize()
 
 
-class DeviceTest(unittest.TestCase):
+class DeviceTest(PyverbsAPITestCase):
     """
     Test various functionalities of the Device class.
     """
 
-    def test_dev_list(self):
-        """
-        Verify that it's possible to get IB devices list.
-        """
-        d.get_device_list()
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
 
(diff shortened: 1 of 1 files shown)


tests/test_device.py
```diff
@@ -15,16 +15,16 @@
...
-    def test_dev_list(self):
-        """
-        Verify that it's possible to get IB devices list.
-        """
-        d.get_device_list()
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
```
jgunthorpe: Why do these need pass?

Kamalheib: To avoid getting/closing the devices in PyverbsAPITestCase setUp()/tearDown function and do it inside the tests of DeviceTest.

jgunthorpe: But that is the only thing PyverbsAPITestCase actually does, so why subclass it?

Kamalheib: Yes, you are right, maybe this change was too early.

This change is done because DeviceTest class is the only test that doesn't inherit from PyverbsAPITestCase or RDMATestCase and in #715 I've added a constructor that will initialize the dev_name which will be used to run the tests on a specific device.

jgunthorpe: Yes, makes sense, change the order :)

Kamalheib:
Closing this PS as the patch is now part of  #736

PR #736, opened 2020-04-05
Add support for passing device name as command-line argument

Kamalheib:
Add support for passing the device name command-line arguments to the test cases, This is needed to give the users the ability to run the tests on a specific device on systems with multiple devices.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,6 +3,7 @@
 
 rdma_python_test(tests
   __init__.py
+  args_parser.py
   base.py
   rdmacm_utils.py
   test_addr.py
--- a/tests/args_parser.py
+++ b/tests/args_parser.py
@@ -1,0 +1,29 @@
+# SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
+# Copyright (c) 2020 Kamal Heib <kamalheib1@gmail.com>, All rights reserved.  See COPYING file
+
+import argparse
+import sys
+
+
+class ArgsParser(object):
+    def __init__(self):
+        self.args = None
+
+    def get_config(self):
+        return self.args
+
+    def parse_args(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--dev',
+                            help='RDMA device to run the tests on')
+        parser.add_argument('-v', '--verbose', dest='verbosity',
+                            action='store_const',
+                            const=2, help='Verbose output')
+        ns, args = parser.parse_known_args()
+        self.args = vars(ns)
+        if self.args['verbosity']:
+            args += ['--verbose']
+        sys.argv[1:] = args
+
+
+parser = ArgsParser()
--- a/tests/base.py
+++ b/tests/base.py
@@ -15,6 +15,7 @@ from pyverbs.addr import AHAttr, GlobalRoute
 from pyverbs.xrcd import XRCD, XRCDInitAttr
 from pyverbs.srq import SRQ, SrqInitAttrEx
 from pyverbs.device import Context
+from args_parser import parser
 import pyverbs.cm_enums as ce
 import pyverbs.device as d
 import pyverbs.enums as e
@@ -44,19 +45,32 @@ def has_roce_hw_bug(vendor_id, vendor_part_id):
 
 
 class PyverbsAPITestCase(unittest.TestCase):
+    def __init__(self, methodName='runTest'):
+        super().__init__(methodName)
+        # Hold the command line arguments
+        self.config = parser.get_config()
+
     def setUp(self):
         """
         Opens the devices and queries them
         """
-        lst = d.get_device_list()
         self.devices = []
-        if len(lst) == 0:
-            raise unittest.SkipTest('No IB devices found')
-        for dev in lst:
-            c = d.Context(name=dev.name.decode())
+
+        dev_name = self.config['dev']
+        if dev_name:
+            c = d.Context(name=dev_name)
             attr = c.query_device()
             attr_ex = c.query_device_ex()
             self.devices.append((c, attr, attr_ex))
+        else:
+            for dev in d.get_device_list():
+                c = d.Context(name=dev.name.decode())
+                attr = c.query_device()
+                attr_ex = c.query_device_ex()
+                self.devices.append((c, attr, attr_ex))
+
+        if len(self.devices) == 0:
+            raise unittest.SkipTest('No IB devices found')
 
     def tearDown(self):
         for tup in self.devices:
@@ -79,7 +93,10 @@ class RDMATestCase(unittest.TestCase):
     def __init__(self, methodName='runTest', dev_name=None, ib_port=None,
                  gid_index=None, pkey_index=None):
         super(RDMATestCase, self).__init__(methodName)
-        self.dev_name = dev_name
+        # Hold the command line arguments
+        self.config = parser.get_config()
+        dev = self.config['dev']
+        self.dev_name = dev_name if dev_name else dev
         self.ib_port = ib_port
         self.gid_index = gid_index
         self.pkey_index = pkey_index
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -15,23 +15,36 @@ import pyverbs.device as d
 PAGE_SIZE = resource.getpagesize()
 
 
-class DeviceTest(unittest.TestCase):
+class DeviceTest(PyverbsAPITestCase):
     """
     Test various functionalities of the Device class.
     """
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def get_device_list(self):
+        lst = d.get_device_list()
+        if len(lst) == 0:
+            raise unittest.SkipTest('No IB device found')
+        dev_name = self.config['dev']
+        if dev_name:
+            for dev in lst:
+                if dev.name.decode() == dev_name:
+                    lst = [dev]
+                    break
+            if len(lst) == 0:
+                raise PyverbsRDMAError(f'No IB device with name {dev_name} found')
+        return lst
 
     def test_dev_list(self):
         """
         Verify that it's possible to get IB devices list.
         """
-        d.get_device_list()
-
-    @staticmethod
-    def get_device_list():
-        lst = d.get_device_list()
-        if len(lst) == 0:
-            raise unittest.SkipTest('No IB device found')
-        return lst
+        self.get_device_list()
 
     def test_open_dev(self):
         """
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
 # Copyright (c) 2018, Mellanox Technologies. All rights reserved.  See COPYING file
 
+from args_parser import parser
 import unittest
 import os
 from importlib.machinery import SourceFileLoader
@@ -9,4 +10,5 @@ from importlib.machinery import SourceFileLoader
 
 module_path = os.path.join(os.path.dirname(__file__), '__init__.py')
 tests = SourceFileLoader('tests', module_path).load_module()
+parser.parse_args()
 unittest.main(module=tests)


tests/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 rdma_python_test(tests
   __init__.py
   base.py
+  args_parser.py
```
jgunthorpe, line 7: keep sorted

tests/args_parser.py
```diff
@@ -0,0 +1,20 @@
...
+class ArgsParser(object):
+    def __init__(self):
+        self.args = None
+
+    def parse_args(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('-d', '--dev_name', help='RDMA device to run the tests on')
+        ns, args = parser.parse_known_args()
+        self.args = vars(ns)
+        sys.argv[1:] = args
```
jgunthorpe, line 17: This seems pretty clever.

In this case lets not use short '-d' strings due to possible conflict

tests/base.py
```diff
@@ -54,7 +55,10 @@ class PyverbsAPITestCase(unittest.TestCase):
     """
     def __init__(self, methodName='runTest', dev_name=None):
         super(PyverbsAPITestCase, self).__init__(methodName)
-        self.dev_name = dev_name
+        if parser.args['dev_name']:
```
jgunthorpe, line 58: Why keep the dev_name argument at all?

'parser' should be a static class member of PyverbsAPITestCase, and give parser a better name, like 'config' or something.

tests/test_device.py
```diff
@@ -15,36 +15,50 @@
...
         lst = d.get_device_list()
         if len(lst) == 0:
             raise unittest.SkipTest('No IB device found')
+        if dev_name is not None:
+            for dev in lst:
+                if dev.name == dev_name:
+                    lst = [dev]
+                    break
+            if len(lst) == 0:
+                raise PyverbsRDMAError('No IB device with name {d} found'.\
```
jgunthorpe, line 40: Use f strings

tests/test_device.py
```diff
@@ -15,36 +15,50 @@
...
-        d.get_device_list()
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
 
     @staticmethod
-    def get_device_list():
+    def get_device_list(dev_name):
```
jgunthorpe, line 30: Why pass in dev_name?

tests/base.py
```diff
@@ -44,6 +44,31 @@ def has_roce_hw_bug(vendor_id, vendor_part_id):
 
 
 class PyverbsAPITestCase(unittest.TestCase):
+    """
+    A base class for test cases which provides the option for user parameters.
+    These can be provided by manually adding the test case to the runner:
+    suite = unittest.TestSuite()
+    ... # Regular auto-detection of test cases, no parameters used.
+    # Now follows your manual addition of test cases e.g:
+    suite.addTest(PyverbsAPITestCase.parametrize(<TestCaseName>, dev_name='..'))
```
jgunthorpe, line 54: This comment isn't relevant is it?

tests/base.py
```diff
@@ -44,6 +45,34 @@ def has_roce_hw_bug(vendor_id, vendor_part_id):
...
+    """
+    def __init__(self, methodName='runTest', dev_name=None):
+        super(PyverbsAPITestCase, self).__init__(methodName)
+        if parser.args['dev_name']:
+            self.dev_name = parser.args['dev_name']
+        else:
+            self.dev_name = dev_name
+
+    @staticmethod
+    def parametrize(testcase_klass, dev_name=None):
```
jgunthorpe, line 64: Is this ever called? Why is it being added?

tests/CMakeLists.txt
Kamalheib: fixed

tests/args_parser.py
Kamalheib: Fixed:
1. '-d' removed.
2. I've added --verbos per the request from @galpress.

tests/base.py
Kamalheib: Fixed

tests/test_device.py
Kamalheib: Fixed

tests/test_device.py
Kamalheib: Fixed.

tests/base.py
Kamalheib: This is related to the use of parametrize(), I've decided to drop it from this patchset in v2.
I think with the use of the new command-line parser we can drop parametrize() from RDMATestCase too?

tests/base.py
Kamalheib: This was added to give the user an API to call parametrize() when creating a TestSuite like what is done in RDMATestCase, but I decided to drop it in v2 of this patch set.

tests/args_parser.py
```diff
@@ -0,0 +1,29 @@
...
+class ArgsParser(object):
+    def __init__(self):
+        self.args = None
+
+    def get_args(self):
+        return self.args
+
+    def parse_args(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--dev_name',
```
galpress, line 17: nit: consider using a `-` separator (i.e `--dev-name`)

rleon: I'm second to that request, or simply "--dev".

Thanks

tests/args_parser.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
+# Copyright (c) 2020 Kamal Heib <kamalheib1@gmail.com>, Inc. All rights reserved.  See COPYING file
```
rleon, line 2: You have "Inc." in the copyright line, did you do in on purpose?

tests/args_parser.py
Kamalheib: I agree I'll change it to --dev in v3.

tests/args_parser.py
Kamalheib: > You have "Inc." in the copyright line, did you do in on purpose?

Maybe in the future :-). I'll fix it in v3.

tests/args_parser.py
```diff
@@ -0,0 +1,29 @@
...
+    def get_args(self):
+        return self.args
+
+    def parse_args(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--dev',
+                            help='RDMA device to run the tests on')
+        parser.add_argument('-v', '--verbose', dest='verbosity',
+                            action='store_const',
+                            const=2, help='Verbose output')
```
galpress, line 21: Why 2?

tests/args_parser.py
```diff
@@ -0,0 +1,29 @@
...
+        self.args = None
+
+    def get_args(self):
+        return self.args
+
+    def parse_args(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--dev',
+                            help='RDMA device to run the tests on')
+        parser.add_argument('-v', '--verbose', dest='verbosity',
```
galpress, line 19: What's the reason to rename it to 'verbosity'?

tests/args_parser.py
galpress: Thanks 😄

tests/test_device.py
```diff
@@ -15,24 +15,37 @@
...
-    @staticmethod
-    def get_device_list():
+    def tearDown(self):
+        pass
+
+    def get_device_list(self):
         lst = d.get_device_list()
         if len(lst) == 0:
             raise unittest.SkipTest('No IB device found')
+        dev_name = self.config['dev']
```
galpress, line 33: This is a weird behavior IMO. Why would a get_device_list function get a device name parameter? I would expect it to return a list of all devices.

tests/args_parser.py
Kamalheib: This is the same definition as in unittest.

tests/args_parser.py
Kamalheib: This is the same definition as in unittest.

tests/test_device.py
Kamalheib: The idea is to run the tests on a specific device on a system with multiple devices if the user passes a '--dev' parameter, otherwise the return value will be a list of all devices in the system.

tests/test_device.py
```diff
@@ -15,24 +15,37 @@
...
+            for dev in lst:
+                if dev.name.decode() == dev_name:
+                    lst = [dev]
+                    break
+            if len(lst) == 0:
+                raise PyverbsRDMAError(f'No IB device with name {dev_name} found')
         return lst
 
+    def test_dev_list(self):
+        """
```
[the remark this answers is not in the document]
Kamalheib: > I don't see an added value for this test case. I assume you added it to check the newly modified "get_device_list" method (?).
> Anyway, it's a very basic function that I would not add a dedicated case for it (test_open_dev is also a simple test and already cover it)

Actually this test case was from day zero as part of 5add7bd862a8 ("pyverbs: Unittest framework"), I can remove it, but I don't think that this is related to this PR.

tests/base.py
```diff
@@ -79,7 +98,12 @@ class RDMATestCase(unittest.TestCase):
     def __init__(self, methodName='runTest', dev_name=None, ib_port=None,
                  gid_index=None, pkey_index=None):
         super(RDMATestCase, self).__init__(methodName)
-        self.dev_name = dev_name
+        self.config = parser.get_args()
+        dev = self.config['dev']
+        if dev_name:
```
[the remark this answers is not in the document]
Kamalheib: I agree I'll change it.

tests/base.py
```diff
@@ -52,18 +60,29 @@ def setUp(self):
         self.devices = []
         if len(lst) == 0:
             raise unittest.SkipTest('No IB devices found')
-        for dev in lst:
-            c = d.Context(name=dev.name.decode())
+
+        dev_name = self.config['dev']
+        if dev_name:
+            c = d.Context(name=dev_name)
```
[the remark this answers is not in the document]
Kamalheib: I don't think so, because of the returned list from "lst = d.get_device_list()" is a list of "Device" objects and not strings.

tests/base.py
```diff
@@ -44,6 +45,13 @@ def has_roce_hw_bug(vendor_id, vendor_part_id):
 
 
 class PyverbsAPITestCase(unittest.TestCase):
+    # Hold the command line arguments
+    config = None
```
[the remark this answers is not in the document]
Kamalheib: Yep, I'll fix it.

tests/test_device.py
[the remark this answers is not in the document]
EdwardSro: Oh, right. I just noticed now that it's not new.

tests/base.py
```diff
@@ -44,6 +45,11 @@ def has_roce_hw_bug(vendor_id, vendor_part_id):
 
 
 class PyverbsAPITestCase(unittest.TestCase):
+    def __init__(self, methodName='runTest'):
+        super(PyverbsAPITestCase, self).__init__(methodName)
```
jgunthorpe, line 49: It is just super() these days, right?

tests/base.py
```diff
@@ -52,11 +58,19 @@ def setUp(self):
...
-            c = d.Context(name=dev.name.decode())
+
+        dev_name = self.config['dev']
+        if dev_name:
+            c = d.Context(name=dev_name)
             attr = c.query_device()
             attr_ex = c.query_device_ex()
             self.devices.append((c, attr, attr_ex))
+        else:
+            for dev in lst:
```
jgunthorpe, line 69: for dev in d.get_device_list():

and fail if self.devices is empty

I don't understand why there is a list of devices. A test should only run on a single device.

tests/base.py
```diff
@@ -79,7 +93,10 @@ class RDMATestCase(unittest.TestCase):
     def __init__(self, methodName='runTest', dev_name=None, ib_port=None,
                  gid_index=None, pkey_index=None):
         super(RDMATestCase, self).__init__(methodName)
-        self.dev_name = dev_name
+        # Hold the command line arguments
+        self.config = parser.get_args()
```
jgunthorpe, line 97: get_config() ?

tests/test_device.py
jgunthorpe: Test cases that do not use the dev_name and other fixed parameters should be inheriting from a different class.

These are the 'old style' tests that should be deleted or updated..

tests/base.py
Kamalheib: Yep.

tests/base.py
Kamalheib: > for dev in d.get_device_list():
> 
> and fail if self.devices is empty
> 
OK, I'll fix it.

> I don't understand why there is a list of devices. A test should only run on a single device.

Well, all the test cases that inherit from PyverbsAPITestCase start with a for loop over all the available devices, looks like this is by design...

tests/base.py
Kamalheib: Yes, This makes more sense.

jgunthorpe:
did you forget some changes? The super is still the old way at least?

Kamalheib:
> did you forget some changes? The super is still the old way at least?

Sorry, I didn't understand your request, Do you mean call super().__init__(methodName)?

jgunthorpe:
yes

Kamalheib:
> yes

OK, I'll fix it.
BTW, Looks like there is multiple places that use the old way...
```
  tests/base.py:95:        super(RDMATestCase, self).__init__(methodName)
  tests/base.py:317:        super(TrafficResources, self).__init__(dev_name=dev_name,
  tests/base.py:458:        super(XRCResources, self).__init__(dev_name, ib_port, gid_index)
  tests/base.py:546:        super(XRCResources, self).init_resources()

```

Kamalheib:
@jgunthorpe & @rleon: Anything blocking this PR from getting merged?

rleon:
I don't think so, just wanted to get some sort of feedback from people who actively reviewed your PR.
As long as you are not breaking my "-k" option, I'm fine :)

rleon:
ok, any changes will be as followups.

Thanks Kamal.

PR #738, opened 2020-04-07
libibumad: Alphabetically sorting CA device list in umad_get_ca_devic…

haimbo:
…e_list API

Implement insertion sort in umad_get_ca_device_list API.
The CA device list will be returned sorted in alphabetical order.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ endif()
 set(PACKAGE_NAME "RDMA")
 
 # See Documentation/versioning.md
-set(PACKAGE_VERSION "29.0")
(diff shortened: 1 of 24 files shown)


haimbo:
After we checked all the options, we decided that this is the best solution
to make sorting when we create the list.

jgunthorpe:
Seriously? The best option is bubble sort?

I still prefer this be a helper function, not built it.

rleon:
> Seriously? The best option is bubble sort?
> 
> I still prefer this be a helper function, not built it.

I said the same (helper function) in some offline discussion too.

Thanks

jgunthorpe:
Replaced by #746

PR #739, opened 2020-04-08
libhns: Optimize buffer calculating process

Li-Weihang:
Calculating process of WQE buffer is not clear enough to understand, so
optimize the code to improve the readability.

Changelog:
v3:
- Use align() directly instead of defining a new macro hr_hw_page_align().

v2:
- Remove meaningless judgment of count in #1 as Leon suggests.
- Rewrite commit message of both patch.

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -39,6 +39,7 @@
 
 #include <infiniband/driver.h>
 #include <util/udma_barrier.h>
+#include <util/util.h>
 #include <infiniband/verbs.h>
 #include <ccan/bitmap.h>
 #include <ccan/container_of.h>
@@ -48,6 +49,10 @@
 #define HNS_ROCE_HW_VER2		('h' << 24 | 'i' << 16 | '0' << 8 | '8')
 
 #define PFX				"hns: "
+
+/* The minimum page size is 4K for hardware */
+#define HNS_HW_PAGE_SHIFT 12
+#define HNS_HW_PAGE_SIZE (1 << HNS_HW_PAGE_SHIFT)
 
 #define HNS_ROCE_MAX_INLINE_DATA_LEN	32
 #define HNS_ROCE_MAX_CQ_NUM		0x10000
@@ -262,6 +267,15 @@ struct hns_roce_u_hw {
 	uint32_t hw_version;
 	struct verbs_context_ops hw_ops;
 };
+
+/*
+ * The entries's buffer should be aligned to a multiple of the hardware's
+ * minimum page size.
+ */
+static inline unsigned int to_hr_hem_entries_size(int count, int buf_shift)
+{
+	return align(count << buf_shift, HNS_HW_PAGE_SIZE);
+}
 
 static inline struct hns_roce_device *to_hr_dev(struct ibv_device *ibv_dev)
 {
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -119,7 +119,7 @@ struct hns_roce_buf {
 };
 
 #define BIT_CNT_PER_BYTE       8
-#define BIT_CNT_PER_U64		64
+#define BIT_CNT_PER_LONG       (BIT_CNT_PER_BYTE * sizeof(unsigned long))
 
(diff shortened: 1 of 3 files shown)


providers/hns/hns_roce_u.h
```diff
@@ -263,6 +268,20 @@ struct hns_roce_u_hw {
 	struct verbs_context_ops hw_ops;
 };
 
+/*
+ * The entries's buffer should be aligned to a multiple of the hardware's
+ * minimum page size.
+ */
+#define hr_hw_page_align(x)		align(x, HNS_HW_PAGE_SIZE)
+
+static inline unsigned int to_hr_hem_entries_size(int count, int buf_shift)
```
rleon, line 277: This function doesn't make any sense.

if count == 0, hr_hw_page_align will return 0

providers/hns/hns_roce_u_verbs.c
```diff
@@ -634,137 +618,153 @@ static int hns_roce_alloc_recv_inl_buf(struct ibv_qp_cap *cap,
...
-		qp->sq.wqe_cnt = align_qp_size(attr->cap.max_send_wr);
-		qp->rq.wqe_cnt = align_qp_size(attr->cap.max_recv_wr);
-	} else {
-		qp->sq.wqe_cnt = align_queue_size(attr->cap.max_send_wr);
-		qp->rq.wqe_cnt = align_queue_size(attr->cap.max_recv_wr);
-	}
+	/* RQ WQE */
+	qp->rq.max_gs = max(1U, attr->cap.max_recv_sge);
+	if (hr_dev->hw_version == HNS_ROCE_HW_VER1)
+		qp->rq.wqe_shift = hr_ilog32(sizeof(struct hns_roce_rc_rq_wqe));
```
rleon, line 731: I see that I missed the commit 3ebc62089441860367c0534fb40832a1c9d251b6.
All those hr_ilog32() should be converted to be like in this patch 69dbf7ccd18af7c7dd18a16779b03fbce59def45

providers/hns/hns_roce_u.h
```diff
@@ -119,7 +119,7 @@ struct hns_roce_buf {
 };
 
 #define BIT_CNT_PER_BYTE       8
-#define BIT_CNT_PER_U64		64
+#define BIT_CNT_PER_LONG       (BIT_CNT_PER_BYTE * sizeof(unsigned long))
```
rleon, line 122: The change doesn't align with commit message.
Please explain why are you doing what are you doing.
Why did you change variable types from clean uint32_t/uin64_t to long?
Why did you mix them?

providers/hns/hns_roce_u.h
Li-Weihang: I see, thanks for you comments. Will fix it.

providers/hns/hns_roce_u_verbs.c
Li-Weihang: Thanks for your reminder, but I have fixed it in 3ebc620 :)

providers/hns/hns_roce_u.h
Li-Weihang: This macro is for bitmap in srq, and it's defined in type of unsigned long, so I think PER_LONG is better than PER_U64. Is it ok?
Sorry for that I didn't mentioned it in the commit message, will modify it and pay more attention on the commit message.

providers/hns/hns_roce_u.h
```diff
@@ -263,6 +268,17 @@ struct hns_roce_u_hw {
...
 
+/*
+ * The entries's buffer should be aligned to a multiple of the hardware's
+ * minimum page size.
+ */
+#define hr_hw_page_align(x)		align(x, HNS_HW_PAGE_SIZE)
+
+static inline unsigned int to_hr_hem_entries_size(int count, int buf_shift)
+{
+	return hr_hw_page_align(count << buf_shift);
```
jgunthorpe, line 279: The only place the macro hr_hw_page_align is used is here, just call align() directly

Li-Weihang: Thank you, I have fixed it.

PR #745, opened 2020-04-20
Add Enhanced Connection Established (ECE)

rleon:
Enhanced Connection Established or ECE is new negotiation scheme
introduced in IBTA v1.4 to exchange extra information about nodes
capabilities and later negotiate them at the connection establishment
phase.

The RDMA-CM messages (REQ, REP, SIDR_REQ and SIDR_REP) were extended
to carry two fields, one new and another gained new functionality:
 * VendorID is a new field that indicates that common subset of vendor
   option bits are supported as indicated by that VendorID.
 * AttributeModifier already exists, but overloaded to indicate which
   vendor options are supported by this VendorID.

This is userspace part of such functionality which is responsible to get data
from librdmacm and properly create and handle RDMA-CM messages.

Thanks


rleon:
Updated with internal feedback from Yishai.
Mainly implemented ibv_set_ece adn ibv_query_ece to be like advise_mr.

rleon:
Returned to original implementation.

PR #746, opened 2020-04-23
Function sort ca device list

haimbo:


infiniband-diags/ibstat.c
```diff
@@ -278,6 +278,7 @@ int main(int argc, char *argv[])
 	struct umad_device_node *first_node;
 	struct umad_device_node *last_node;
 	int dev_port = -1;
+	int size_of_cas = 0;
```
jgunthorpe, line 281: Why have a variable that is always 0?

PR #749, opened 2020-04-30
RDMA-CM unique device identifier

rleon:
Provide a way to identify the device in unique way.


rleon:
Added Yishai's ROB.
No other changes

jgunthorpe:
Seems fine, don't forget to update the kernel-headers with the right commit id

rleon:
Update kernel-header with kernel commit 17793833f81c and will merge once that commit will be in for-next.

Thanks

PR #750, opened 2020-05-02
RDMA-CM disassociate support

rleon:
In current implementation of RDMA-CM, if IB device resets itself, the
application and librdmacm can't recognise that the device is different,
because device identification is done based on node_guid. That node_guid
is the same before and after reset, but the device is different from the
kernel perspective.
    
This disconnect between kernel and rdma-cm is the reason why users need
to actually restart applications.
    
In the following patches, we will connect kernel and rdma-cm differently.
his will allow them to be in-sync. Once device "disappears" from the kernel,
it will be removed from librdmacm too. The "addition" of new device, will
reflect in addition in librdmacm too.

### DIFF
--- a/debian/libibverbs1.symbols
+++ b/debian/libibverbs1.symbols
@@ -6,6 +6,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.6@IBVERBS_1.6 24
  IBVERBS_1.7@IBVERBS_1.7 25
  IBVERBS_1.8@IBVERBS_1.8 28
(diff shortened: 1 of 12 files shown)


rleon:
Based on https://github.com/linux-rdma/rdma-core/pull/749
I put it here to show full picture.

Thanks

rleon:
Added Yishai's ROB,
No other changes.

rleon:
BTW, it is ready to be merged

rleon:
Rebased on https://github.com/linux-rdma/rdma-core/pull/749 with updated kernel header.

Thanks

PR #753, opened 2020-05-06
verbs: Enable asynchronous event FD per object

yishaih:
This series enables applicable events objects (i.e. QP, SRQ, CQ, WQ) to be created with their own asynchronous event FD from KABI point of view.

Before this series any affiliated event on an object was reported on the first asynchronous event FD that was created on the context without the ability to create and use a dedicated FD for it.

With this series we enable granularity and control for the usage per object.

For example, a secondary process that uses the same command FD as of the master one, can create its own objects with its dedicated event FD to be able to get the events for them once occurred, this couldn't be done before this series. 

Further series on top of this one may expose some option to an application to ask for a dedicated FD for its usage instead of using the default one, this may enable achieving the above use case. 

To achieve the above, any 'create' method for the applicable objects was extended to get from rdma-core its optional event FD, if wasn't supplied, the default one from the context will be used as part of kernel side.

As we prefer to not extend the 'write' mode KABIs anymore and fully move to the 'ioct' mode, as part of this extension QP, SRQ and WQ create/destroy commands were introduced over 'ioctl', the CQ KABI was extended over its existing 'ioctl' create command.

As part of moving to 'ioctl' for the above objects the few bugs were found and fixed.

The matching kernel part was sent into rdma-next.


yishaih:

PR #747, opened 2020-04-26
tests: Delete code duplication from CMTestCase

Kamalheib:
Getting the device name is already handled by the RDMATestCase, so
remove it.

Signed-off-by: Kamal Heib <kamalheib1@gmail.com>

PR #754, opened 2020-05-06
tests: RDMACM tests infrastructure refactoring

EdwardSro:
RDMACM tests infrastructure refactoring which includes:
1. Add Async and Sync CMResources. These aggregation objects inherit
   from CMResources and implement resources creation accordingly.
2. Add CMConnection class which is responsible for CMResources
   initialization, connection establishment and traffic between two peers.
3. Separate RDMACM aggregation objects from Verbs objects for easier
   maintenance.

This allows easier adoption of rdmacm upcoming tests.

### DIFF
--- a/tests/test_rdmacm.py
+++ b/tests/test_rdmacm.py
@@ -2,9 +2,7 @@
 # Copyright (c) 2019 Mellanox Technologies, Inc. All rights reserved. See COPYING file
 
 import multiprocessing as mp
-import subprocess
 import unittest
-import json
 import os
 
 from tests.rdmacm_utils import  CMSyncConnection, CMAsyncConnection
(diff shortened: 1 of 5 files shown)


EdwardSro:
This includes #747 change

tests/test_rdmacm.py
```diff
@@ -1,32 +1,48 @@
...
 import json
+import os
+
 
 NUM_OF_PROCESSES = 2
 
 
 class CMTestCase(RDMATestCase):
     def setUp(self):
+        if self.dev_name is not None:
```
EdwardSro, line 20: #747 change breaks rdmacm tests on IB devices in two possible ways:
1. When there's multiple devices/ports but there's just one active device/port, but the latter doesn't have a net interface IP the test would fail on IndexError (since we're trying to get the IP when there's none).
2. In RDMATestCase base class it randomizes which device and args to use, so in case we have more than one device in an active state but just one of them has net IP, it might not be selected, hence some rdmacm tests would run properly but other get skipped (since they could use another device with no net interface)

@Kamalheib currently, I've reverted the change in this PR. But you're welcome to give any feedback/suggestion.

Kamalheib: @EdwardSro: I think the test needs to be fixed by skip the run if the selected interface doesn't have a configured ip address, as there is no guarantee that each interface has a configured ip address.

EdwardSro: This was the behavior before;
If the user explicitly passes the device to be used, and it doesn't have a net interface with configured IP then the test is skipped.
But in case the user doesn't pass the device arg, CMTestCases search and select a device with a configured IP interface, if none were found then the test is skipped.
If I understood you correctly, relying on RDMATestCase chosen device can result in skipping some rdmacm tests while they might have the ability to run on other device(s)

Kamalheib: @EdwardSro: Well I think the bug exists from day zero because there is no check for "IndexError" when the "self.dev_name" is set...

And yes, The CMTestCase needs to relay on the selected device (even when that is done randomly), probably we need to make sure that the selected device has a configured IP address in RDMATestCase to avoid doing what was done in test_rdmacm.py in new tests.

EdwardSro: > And yes, The CMTestCase needs to relay on the selected device (even when that is done randomly), probably we need to make sure that the selected device has a configured IP address in RDMATestCase to avoid doing what was done in test_rdmacm.py in new tests.

I've done so, will upload a new commit for this in this series.

PR #755, opened 2020-05-10
mlx5: Improve memory usage in the DR area

yishaih:
This series improves the memory usage in the DR area by enabling an option to reclaim device memory which is not in use.

In addition, the hot memory calculations to consider whether sync call is required was refactored to prevent redundant checks and improve performance.

### DIFF
--- a/debian/ibverbs-providers.symbols
+++ b/debian/ibverbs-providers.symbols
@@ -21,6 +21,7 @@ libmlx5.so.1 ibverbs-providers #MINVER#
  MLX5_1.11@MLX5_1.11 25
  MLX5_1.12@MLX5_1.12 28
  MLX5_1.13@MLX5_1.13 29
+ MLX5_1.14@MLX5_1.14 30
  mlx5dv_init_obj@MLX5_1.0 13
  mlx5dv_init_obj@MLX5_1.2 15
  mlx5dv_query_device@MLX5_1.0 13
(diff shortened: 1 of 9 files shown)


debian/ibverbs-providers.symbols
```diff
@@ -98,6 +99,7 @@ libmlx5.so.1 ibverbs-providers #MINVER#
  mlx5dv_free_var@MLX5_1.12 28
  mlx5dv_pp_alloc@MLX5_1.13 29
  mlx5dv_pp_free@MLX5_1.13 29
+ mlx5dv_dr_domain_set_reclaim_device_memory@MLX5_1.14 30
```
rleon, line 102: I think that it should be sorted.

providers/mlx5/mlx5dv_dr.h
```diff
@@ -1042,6 +1047,7 @@ struct dr_icm_buddy_mem {
 
 	/* This is the list of used chunks. HW may be accessing this memory */
 	struct list_head	used_list;
+	unsigned int		used_memory;
```
rleon, line 1050: Memory is size_t

yishaih: This value is per body which can reach 64M bytes and not exposed outside from mlx5, 32 bits is much larger than is really needed.

debian/ibverbs-providers.symbols
yishaih: It's sorted by version as expected.

providers/mlx5/mlx5dv_dr.h
rleon: The size_t usage is about correctness and not about "reach".

debian/ibverbs-providers.symbols
rleon: It should be sorted by name and not by version.

Thanks

providers/mlx5/mlx5dv_dr.h
yishaih: Updated accordingly.

PR #756, opened 2020-05-12
d/control: Don't build-depend on pandoc on i386

cpaelzer:
Ubuntu `i386` no more has `pandoc` and would stall on build dependencies.
This was done as an emergency unblock late in the 20.04 cycle,
but it turned out that `i386` still gets all the man pages that pandoc
would convert properly installed.
See [amd64](https://launchpadlibrarian.net/466565153/buildlog_ubuntu-focal-amd64.rdma-core_28.0-1ubuntu1_BUILDING.txt.gz) with pandoc vs [i386](https://launchpadlibrarian.net/466565266/buildlog_ubuntu-focal-i386.rdma-core_28.0-1ubuntu1_BUILDING.txt.gz) (without pandoc) and as an example
the man page `ibv_resize_cq.3`.

There are other options like dropping pandoc completely on all architectures if it really isn't needed.
Other man pages for actual binaries seem to use rst2man, we could also consider pushing all those that use pandoc into a arch-neutral doc package maybe?

TL;DR: Anything that helps to get rid of the pandoc dependency on i386 will help Ubuntu :-)
Let me know what you think and what you'd prefer.


jgunthorpe:
Technically this is against Debian policy as source packages are supposed to rebuild from source. To avoid the pandoc (and rst2man!) dependency the official source tar.gz include pre-built man pages that were generated from the original sources as part of the CI.

I don't have any notion on what to replace pandoc with - we selected it because it uses Markdown as an input which makes all the man page text viewable natively on github. The rst versions of the man pages are a hold over that can't be converted to markdown..

cpaelzer:
md -> man seems to be a bad topic for keeping dependencies small.

 - `pandoc` has only a few deps, but doesn't exist on Ubuntu i386
 - `ronn` brings in plenty of ruby which we have an issue as well
 - `marked-man` is similar again but with nodejs this time

It is a mess, I agree that per policy things need to be regenerated from source.
I haven't found a great alternative yet :-/

cpaelzer:
It might come down to a -doc package or such that can be build arch-neutral and dependent on by the -dev currently carrying the man pages that are generated via pandoc.
But before I go down that route I'd want to know what you'd think about it first.

jgunthorpe:
It seems reasonable to have a doc package @bdrung ?

cpaelzer:
As of today we have these packages containing man pages rendered with pandoc:
- infiniband-diags: `check_lft_balance dump_fts ibaddr ibcacheedit ibccconfig ibccquery ibfindnodesusing ibhosts ibidsverify iblinkinfo ibnetdiscover ibnodes ibping ibportstate ibqueryerrors ibroute ibrouters ibrouters ibstat ibstatus ibstatus ibswitches ibsysstat ibtracert infiniband-diags perfquery saquery sminfo smpdump smpquery vendstat`
- libibumad-dev: `umad_init`
- libibverbs-dev: `efadv_create_driver_qp efadv_create_driver_qp ibv_advise_mr ibv_alloc_null_mr ibv_attach_counters_point_flow ibv_attach_mcast ibv_create_counters ibv_create_flow_action ibv_event_type_str ibv_fork_init ibv_get_device_guid ibv_get_device_list ibv_get_device_name ibv_get_pkey_index ibv_get_srq_num ibv_inc_rkey ibv_query_gid ibv_query_pkey ibv_rate_to_mbps ibv_rate_to_mult ibv_read_counters ibv_req_notify_cq ibv_rereg_mr ibv_resize_cq ibv_wr_post efadv`
- libibverbs-dev also contains all the generated names e.g. _mldv_devx_subscribe_devx_event_ becoming _mlx5dv_devx_subscribe_devx_event_: `mldv_set_context_attr mldv_create_qp mldv_create_flow_action_modify_header mldv_alloc_dm mldv_devx_create_cmd_comp mldv_create_cq mldv_create_flow_matcher mldv_flow_action_esp mldv_open_device mldv_create_mkey mldv_create_flow mldv_create_flow_action_packet_reformat mldv_devx_alloc_uar mldv_dr_flow mldv_devx_obj_create mldv_devx_qp_modify mldv_alloc_var mldv_devx_umem_reg mldv_devx_subscribe_devx_event mldv_is_supported mldv_dump mldv_devx_query_eqn mldv_wr_post mldv_devx_get_event mldv_devx_create_event_channel`
- librdmacm-dev: `rdma_establish, rdma_init_qp_attr`

To move all these into a a single -doc package seems wrong to me :-/
We'd not want to group all of them, and having 4 more -doc packages seems like a nasty workaround as well.

But I also agree that the simpler approach I already suggested with the PR (to not re-render them on i386) would violate the policy to build everything from source.

There must be a better way out, but I'm not seeing it yet...

TBH since Ubuntu dropped most of i386 already, the only dependency left on that arch is openmpi:
```
$ reverse-depends --release=groovy --arch=i386  src:rdma-core
Reverse-Depends
* libopenmpi-dev                (for libibverbs-dev)
```

But right now (for groovy) neither `openmpi` nor `rdma-core` are part of the i386 build white-list.
So maybe I can just sync the most recent version of rdma-core.
It will then -not- build on i386 since it is missing in the whitelist and therefore should not trigger a problem?

I'm a bit unsure on that, but it won't be solved by this PR for now. I'm closing the PR but might refer to this post in some mails I have to write ...

Thanks for the discussion.

bdrung:
I agree with what @cpaelzer concluded: Moving the man pages into a single -doc package looks ugly and having 4 more -doc packages is an overkill and dirty workaround. I see no clean way to get rid of the pandoc dependency on build. Moving away from pandoc is not feasible as well. So I agree with closing this ticket.

PR #757, opened 2020-05-14
pyverbs: Support ODP MR prefetch

EdwardSro:
This series adds pyverbs support for advising MR.
It also includes a simple ODP MR prefetch test.


rleon:
Please rebase

PR #761, opened 2020-05-24
pyverbs: Extend RDMACM support and tests

EdwardSro:
This series brings more support and tests for RDMACM into Pyverbs,
including UDP and multicast traffic.
It includes some other extensions and fixes that are necessary for the
sake of RDMACM extended support.


rleon:
merged manually

PR #764, opened 2020-05-27
mlx5: Expose DV APIs to support lag mode

yishaih:
This series exposes DV APIs to support lag mode.
It enables an application querying QP's configured and active lag port and to modify the configured port num.

A pyverbs stuff and test were added to demonstrate the expected usage.


yishaih:
Rebased to solve a conflict in pyverbs.

PR #765, opened 2020-05-28
efa: Fix create QP command over ioctl interface

galpress:
The provider should not access the 'ibv_resp' parts of the response
structure, use the max_inline_data in ibv_qp_init_attr_ex instead.

This issue can be observed when using the ioctl interface for the
create/destroy QP verbs, which keeps the 'ibv_resp' part of the response
cleared.

Suggested-by: Jason Gunthorpe <jgg@ziepe.ca>

### DIFF
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0 OR BSD-2-Clause
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All rights reserved.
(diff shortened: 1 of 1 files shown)


rleon:
Please add Fixes line and add stable@ so @nmorey will be able to pick it to stables too.

Thanks

galpress:
Done

rleon:
@nmorey can you please pick this patch to the stable@?

Thanks

rleon:
> @rleon Picked for v24 -> v29
> 
> As long as stuff is properly tagged with 'Fixes' it'll get picked almost automatically by git-maintain.

Good to know.
Thanks

PR #767, opened 2020-06-01
libibverbs: Speed up MR registration when RDMAV_HUGEPAGES_SAFE is ena…

galpress:
…bled

When RDMAV_HUGEPAGES_SAFE is enabled, every MR registration has an extra
overhead as a result of searching for the appropriate page size backing
up the buffer.

Since looking for the page size by parsing /proc/pid/smaps takes a
considerable time, this patch speeds it up by calling madvise with the
various page sizes until one of the calls succeeds (with the correct
page size).

Tested by measuring average time to complete ibv_reg_mr:
Before this patch:
RDMAV_HUGEPAGES_SAFE unset: 93us
RDMAV_HUGEPAGES_SAFE set  : 5414us (58x)

After this patch:
RDMAV_HUGEPAGES_SAFE unset: 93us
RDMAV_HUGEPAGES_SAFE set  : 103us (1.1x)

The possible page sizes are queried once on ibv_fork_init from the
/sys/kernel/mm/hugepages directory.
The get_page_size function iterates the different page sizes and calls
madvise for each, first with DOFORK and then with DONTFORK.

This is done for two reasons:
First, it's correct to end up with the pages set to DONTFORK as the
rbtree will later handle the reference counting and set the advice for
these pages appropriately.

Second, the kernel caches the vma flags and doesn't check madvise's
address and length parameters if the flags did not change. This means
that if the same page is being madvise'd twice the second call will
always succeed regardless of the page size. In order to solve that the
initial DOFORK was added.

Suggested-by: Wei Zhang <wzam@amazon.com>

### DIFF
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -57,68 +57,99 @@ struct ibv_mem_node {
 	int			refcnt;
 };
 
+#define MAX_NUM_PAGE_SIZES 32
+
 static struct ibv_mem_node *mm_root;
 static pthread_mutex_t mm_mutex = PTHREAD_MUTEX_INITIALIZER;
-static int page_size;
 static int huge_page_enabled;
 static int too_late;
+static int num_page_sizes;
+static unsigned long page_sizes[MAX_NUM_PAGE_SIZES];
 
-static unsigned long smaps_page_size(FILE *file)
-{
-	int n;
-	unsigned long size = page_size;
-	char buf[1024];
-
-	while (fgets(buf, sizeof(buf), file) != NULL) {
-		if (!strstr(buf, "KernelPageSize:"))
-			continue;
-
-		n = sscanf(buf, "%*s %lu", &size);
-		if (n < 1)
-			continue;
-
-		/* page size is printed in Kb */
-		size = size * 1024;
-
-		break;
-	}
-
-	return size;
-}
+enum {
+	IBV_SC_PAGE_SIZE_IDX = 0,
+};
 
 static unsigned long get_page_size(void *base)
 {
-	unsigned long ret = page_size;
-	pid_t pid;
-	FILE *file;
-	char buf[1024];
+	unsigned long ret = page_sizes[IBV_SC_PAGE_SIZE_IDX];
+	unsigned long base_aligned;
+	unsigned long pgsz;
+	int err;
+	int i;
 
-	pid = getpid();
-	snprintf(buf, sizeof(buf), "/proc/%d/smaps", pid);
-
-	file = fopen(buf, "r" STREAM_CLOEXEC);
-	if (!file)
-		goto out;
-
-	while (fgets(buf, sizeof(buf), file) != NULL) {
-		int n;
-		uintptr_t range_start, range_end;
-
-		n = sscanf(buf, "%" SCNxPTR "-%" SCNxPTR, &range_start, &range_end);
-
-		if (n < 2)
+	for (i = 0; i < num_page_sizes; i++) {
+		pgsz = page_sizes[i];
+		base_aligned = (uintptr_t)base & ~(pgsz - 1);
+		/* DOFORK is required as the kernel caches the given flags,
+		 * so calling madvise with MADV_DONTFORK twice on the same vma
+		 * returns success without checking the address and length.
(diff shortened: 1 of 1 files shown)


galpress:
This patch feels less intuitive than the existing code, but I think it's worth considering it as it shows a significant improvement over the current state.

The DOFORK -> DONTFORK behavior might be a bit problematic, but it can be changed to MADV_SEQUENTIAL -> MADV_NORMAL for example.

libibverbs/memory.c
```diff
@@ -57,68 +57,99 @@ struct ibv_mem_node {
...
+	int i;
+
+	for (i = 0; i < num_page_sizes; i++) {
+		pgsz = page_sizes[i];
+		base_aligned = (uintptr_t)base & ~(pgsz - 1);
+		/* DOFORK is required as the kernel caches the given flags,
+		 * so calling madvise with MADV_DONTFORK twice on the same vma
+		 * returns success without checking the address and length.
+		 */
+		while ((err = madvise((void *)base_aligned, pgsz,
```
jgunthorpe, line 88: This doesn't look like it should work at all. madvise will split VMAs, so if you call this with 4k it will happily split the huge page VMA into sub VMAs and go ahead.

The entire point of this nonsense code is to prevent splitting the VMA, which can only be done by learning the actual extents of the VMA and acting accordingly.

If it is too slow then don't use the broken forking mode. It doesn't actually work reliably anyhow.

galpress: > This doesn't look like it should work at all. madvise will split VMAs, so if you call this with 4k it will happily split the huge page VMA into sub VMAs and go ahead.

I thought so as well, apparently it doesn't. It returns an EINVAL error.

> The entire point of this nonsense code is to prevent splitting the VMA, which can only be done by learning the actual extents of the VMA and acting accordingly.
> 
> If it is too slow then don't use the broken forking mode. It doesn't actually work reliably anyhow.

Didn't know there are forking modes? And that one of them is broken?

jgunthorpe: How does the kernel flow generate EINVAL? I don't see it, I see VMA splitting.

By forking mode I mean don't use fork, so don't use this madvise marking

galpress: I can't point the exact flow where this happens in the kernel, but running this sample app:
```
#include <stdio.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <unistd.h>
#include <stdint.h>
#include <errno.h>

#define PAGE_SIZE (4096)
#define HUGEPAGE_SIZE (1024 * 1024 * 2)

int main()
{
	void *buf;
	void *buf_aligned;
	int err;
	int pgsz;
	int sizes[2] = { PAGE_SIZE, HUGEPAGE_SIZE };
	int i;

	buf = mmap(NULL, 64, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB,
		   -1, 0);
	if (!buf) {
		printf("Couldn't allocate memory\n");
		return -1;
	}

	printf("buf %p\n", buf);

	for (i = 0; i < 2; i++) {
		pgsz = sizes[i];
		buf_aligned = (void *)((uintptr_t)buf & ~(pgsz - 1));
		err = madvise(buf_aligned, pgsz, MADV_DONTFORK);
		printf("madvise pgsz %d err %d errno %d\n", pgsz, err, err ? errno : 0);
	}

	munmap(buf, 64);
	return 0;
}
```

Outputs:
```
buf 0x7f0b0f800000
madvise pgsz 4096 err -1 errno 22
madvise pgsz 2097152 err 0 errno 0
```

Do you suggest that all libibverbs apps don't use fork?

jgunthorpe: We can't rely on some kernel behavior like this without understanding if it is guarenteed or not, since I see code that is doing VMA splitting in madvise you need to figure out why it is isn't activting and if that is part of the API contract, then document all of this.

And yes, verbs apps using fork is very problematic, it should be avoided.

galpress: Thanks, I agree we shouldn't rely on this behavior without a better understanding.
Let's drop this patch for now.

PR #770, opened 2020-06-03
Add conditionals for systemd in rdma-core.spec file

galpress:
Add conditionals to build rdma-core packages for a system which does not have systemd.
Amazon Linux 1 cbuild packaging is modified to use rdma-core.spec file as well.

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -189,7 +189,7 @@ class amazonlinux1(YumEnvironment):
     use_make = True;
     pandoc = False;
     build_pyverbs = False;
-    specfile = "buildlib/centos6.spec";
+    specfile = "redhat/rdma-core.spec";
     python_cmd = "python";
     to_azp = False;
 
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -13,6 +13,9 @@ Source: rdma-core-%{version}.tgz
 # Do not build static libs by default.
 %define with_static %{?_with_static: 1} %{?!_with_static: 0}
 
+# Build with systemd by default.
+%define with_systemd %{?_with_systemd: %_with_systemd} %{?!_with_systemd: 1}
+
 # 32-bit arm is missing required arch-specific memory barriers,
 ExcludeArch: %{arm}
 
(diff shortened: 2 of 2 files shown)


rleon:
Maybe I'm mistaken here, but in early days of this repo and spec file, we got a decision do not support non-systemd systems. This is why centos6.spec was added later.

galpress:
Thanks Leon,
This patch keeps the default behavior for all packaging, so it shouldn't affect existing packaging.
Is there a reason to explicitly not allow it?

The commit that introduced centos6.spec says:
```
commit 9d6fe4eab695c6c76eb0e29e0acfd0e58623bc34
Author: Jason Gunthorpe <jgg@ziepe.ca>
Date:   Fri Jul 19 11:41:05 2019 -0300

    build/azp: Add centos6 to the test distributions

    We keep having small defects related to these old glibc's, run a build
    here too. The spec file is based on the old pre-packaging rdma-core spec
    file.

```

Looks like it's related to glibc stuff?

galpress:
BTW, the second patch misses the actual setting of _with_systemd to zero, I'll fix that.

rleon:
> Thanks Leon,
> This patch keeps the default behavior for all packaging, so it shouldn't affect existing packaging.
> Is there a reason to explicitly not allow it?
> 
> The commit that introduced centos6.spec says:
> 
> ```
> commit 9d6fe4eab695c6c76eb0e29e0acfd0e58623bc34
> Author: Jason Gunthorpe <jgg@ziepe.ca>
> Date:   Fri Jul 19 11:41:05 2019 -0300
> 
>     build/azp: Add centos6 to the test distributions
> 
>     We keep having small defects related to these old glibc's, run a build
>     here too. The spec file is based on the old pre-packaging rdma-core spec
>     file.
> 
>     Signed-off-by: Jason Gunthorpe <jgg@mellanox.com>
> ```
> 
> Looks like it's related to glibc stuff?

We still support compilation, so this is why centos6 was added for azp. From packaging perspective and from official support, we don't want to deal with bug reports against systems without systemd.

commit 5ecc975da872c7b0712edfbba1dd77314c823ace
Author: Leon Romanovsky <leonro@mellanox.com>
Date:   Tue Nov 27 20:55:41 2018 +0200

    cbuild: Disable packaging of CentOS 6

jgunthorpe:
I think it is necessary to have a proper different spec file, there are many differences that need to be handled before rdma-core would work correctly on a C6 era distro.

Replacements for all the systemd only functionality would be needed.

In your case it might be reasonable to have som AL1  spec file that only packages the shared libraries and none of the boot or daemon scripts?

rwespetal:
Sure, adding an additional spec file for AL1 will work. We'll open another PR for it.

PR #774, opened 2020-06-09
EFA summer cleanups

galpress:
A few cleanups to the EFA provider.
No functional changes in this series.

### DIFF
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -669,12 +669,12 @@ static void efa_qp_init_indices(struct efa_qp *qp)
 {
 	qp->sq.wq.wqe_posted = 0;
 	qp->sq.wq.wqe_completed = 0;
-	qp->sq.wq.desc_idx = 0;
+	qp->sq.wq.pc = 0;
 	qp->sq.wq.wrid_idx_pool_next = 0;
 
 	qp->rq.wq.wqe_posted = 0;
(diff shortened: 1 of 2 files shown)


providers/efa/verbs.c
```diff
@@ -1118,6 +1118,24 @@ static void efa_sq_advance_post_idx(struct efa_qp *qp)
...
+	udma_to_device_barrier();
+	efa_wq_ring_doorbell(&rq->wq, desc_idx);
+}
+
+static inline void efa_sq_ring_doorbell(struct efa_sq *sq, uint16_t desc_idx)
+{
+	mmio_flush_writes();
+	efa_wq_ring_doorbell(&sq->wq, desc_idx);
+}
+
```
rleon, line 1145: Let's do not add extra obfuscation. You added more lines than deleted and it is clear indication to me that extra functions are not needed.

galpress: I don't think the primitive of ringing the doorbell should be hard coded more than once in different flows.
What do you think about dropping efa_wq_ring_doorbell and use the mmio_write32 directly in the RQ/SQ functions?

rleon: If you can't leave without it, let's do it.

galpress: Done

PR #775, opened 2020-06-15
EFA ucontext attributes

galpress:
Keep the provider specific attributes in EFA context instead of device.
This allows for faster access as the context is frequently accessed
(opposed to the device), consistent struct to store all device
attributes, and better fits the design of the provider.


galpress:
I've dropped the kernel changes, kept the commit that moves the attributes to the context struct only.

PR #776, opened 2020-06-17
verbs: Introduce import verbs for device, PD, MR

yishaih:
This series Introduces import verbs for device, PD, MR which enables processes to share
their ibv_context and then share PD(s) and MR(s) that are associated with.

This functionality enables utilizing  and optimizing some application flows, few
examples below.
 
Any solution which is a single business logic based on multi-process design
needs this. Example include NGINX, with TCP load balancing, sharing the RSS
indirection table with RQ per process. HPC frameworks with multi-rank (process)
solution on single hosts. UCX can share IB resources using the shared PD and
can help dispatch data to multiple processes/MR's in single RDMA operation.
Also, there are use cases when a primary processes registered a large shared
memory range, and each worker process spawned will create a private QP on the
shared PD, and use the shared MR to save the registration time per-process.
 
As part of this series was added also some pyverbs stuff to support and demonstrate some usage of the APIs.

The verbs APIs were introduced in the mailing list by the below RFC [1], the matching kernel series was sent to rdma-next,

[1] https://patchwork.kernel.org/patch/11540665/

### DIFF
--- a/Documentation/pyverbs.md
+++ b/Documentation/pyverbs.md
@@ -626,3 +626,43 @@ ctx = Context(name='rocep0s8f0')
 uar = Mlx5UAR(ctx)
 uar.close()
 ```
+
+##### Import device, PD and MR
+Importing a device, PD and MR enables processes to share their context and then
+share PDs and MRs that is associated with.
+A process creates a device and then uses some of the Linux systems calls to dup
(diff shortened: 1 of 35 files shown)


yishaih:

PR #777, opened 2020-06-18
verbs: Introduce verbs_cq for extended CQ operations

galpress:
Extended QP implementation has a nice verbs_qp abstraction which does
not exist for extended CQs, this patch modifies CQ to be similar.
The verbs_cq struct is a union of both CQ and extended CQ, which allows
access for both. In addition, converting a CQ to the provider's CQ can
now be done using container_of, without explicit casts.

### DIFF
--- a/libibverbs/driver.h
+++ b/libibverbs/driver.h
@@ -102,6 +102,13 @@ struct verbs_qp {
 };
 static_assert(offsetof(struct ibv_qp_ex, qp_base) == 0, "Invalid qp layout");
 
+struct verbs_cq {
+	union {
+		struct ibv_cq cq;
+		struct ibv_cq_ex cq_ex;
+	};
+};
+
 enum ibv_flow_action_type {
 	IBV_FLOW_ACTION_UNSPECIFIED,
 	IBV_FLOW_ACTION_ESP = 1,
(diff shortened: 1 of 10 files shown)


libibverbs/verbs.h
```diff
@@ -1445,16 +1445,7 @@ struct ibv_wc_tm_info {
...
-	struct ibv_context     *context;
-	struct ibv_comp_channel *channel;
-	void		       *cq_context;
-	uint32_t		handle;
-	int			cqe;
-
-	pthread_mutex_t		mutex;
-	pthread_cond_t		cond;
-	uint32_t		comp_events_completed;
-	uint32_t		async_events_completed;
```
yishaih, line 1457: We should prevent changing a structure that is exposed from verbs.h , there might be some application that was written with direct usage of one on applicable/exposed fields (e.g. channel, cq_context, etc.) that post of such change might fail to compile.

ibv_qp_ex was introduced from day one with qp_base comparing this one.

libibverbs/driver.h
```diff
@@ -102,6 +102,15 @@ struct verbs_qp {
 };
 static_assert(offsetof(struct ibv_qp_ex, qp_base) == 0, "Invalid qp layout");
 
+struct verbs_cq {
+	union {
+		struct ibv_cq cq;
+		struct ibv_cq_ex cq_ex;
+	};
+};
```
yishaih, line 110: What is the use case that you come to solve with this patch ? is it only some preparation for future work ? we usually prefer having some series with some added value / usage and not only preparations.

libibverbs/verbs.h
galpress: You're definitely right, missed that.

libibverbs/driver.h
galpress: This is a cleanup patch, preventing the need for explicit casts. The use case is the providers supporting extended CQ, mlx4/5 and any future providers.

galpress:
Kind reminder.

PR #781, opened 2020-06-29
srp_daemon: srp_daemon.service should wait for network.target

markhaywood:
srp_daemon.service runs the start_on_all_ports script. This script
enumerates files under /sys/class/infiniband/*/ports/* to start an
srp_daemon_port service for each port. The files under
/sys/class/infiniband/*/ports/* are created during boot and do not
exist until after network.target. When started at boottime, srp_daemon
is not starting any port services as expected. For srp_daemon to work
as designed, it should wait for network.target.

Fixes: 1c7fe513e3e9 ("srp_daemon: One systemd service per port")

### DIFF
--- a/srp_daemon/srp_daemon.service.in
+++ b/srp_daemon/srp_daemon.service.in
@@ -3,6 +3,7 @@ Description=Daemon that discovers and logs in to SRP target systems
 Documentation=man:srp_daemon file:/etc/srp_daemon.conf
 DefaultDependencies=false
 Conflicts=emergency.target emergency.service
+After=network.target
 Before=remote-fs-pre.target
 
 [Service]


jgunthorpe:
network.target doesn't prove that the rdma ports have been hot plugged in yet

If I recall this script is some legacy thing ? The rule in srp_daemon.rules should auto start on all parts inherently.

markhaywood:
It is true that the that all ports will start anyway. And I had not considered hotplug. Given both of these facts, is there
any reason for srp_daemon.service to be running the start_on_all_ports service at all?

bvanassche:
srp_daemon.sh and srpd are the legacy scripts. srp_daemon.service is the recommended way to start the SRP daemon.

jgunthorpe:
The reason to have the start_on_all_ports script was to make the experiance more similar to the old scripts where you could start/stop and get back to the same place.

What are you trying to fix?

markhaywood:
I see. The reason that srp_daemon.service executes the script is for the case where the service has been manually stopped and needs to be restarted. OK.

The reason that I looked at what srp_daemon.service.in was doing at boot was because it was failing on our systems. From what I could tell by trial of elimination, "PrivateTmp=yes" in srp_daemon.service causes the service to fail at boot with:

Active: failed (Result: resources)

I noticed that this failure could be worked around, if the service waited longer in the boot process. In the meantime, I had noticed that start_on_all_ports wasn't configuring anything (and in fact, has a scripting bug in it when no ports exist that I'll be fixing in another patch). Looking back at the history, I saw that srp_daemon.service used to (prior to 1c7fe51 ("srp_daemon: One systemd service per port")) wait until after network.target and assumed that no one had noticed that start_on_all_ports wasn't configuring anything at boot due to the srp_daemon.rules that you mentioned earlier.

So, now I am back to the service fails at boot whenever "PrivateTmp=yes". Quite honestly, I'm not sure of the value of the option for srp_daemon since I don't know that it uses /tmp. But if it is set, we can't have it causing the service to fail at boot.

jgunthorpe:
PrivateTmp failing sounds like a completely unrelated problem.. Something else must be funny with the system

markhaywood:
Possibly the same problem that you and Honggang Li discussed here:

https://patchwork.kernel.org/patch/10136967/

Sounds like he resolved it with another version of systemd.

jgunthorpe:
PrivateTmp is set because it is part of the standard hardening package - but since this is no longer running a daemon it could probably go away - but really the systemd bug you are hitting should be fixed in the distro side ...

aron-silverton:
My guess is that the systemd issue is:

[d71f050599df](https://github.com/systemd/systemd/commit/d71f050599dfd63a53ef3e91a8ef1835a8985b92) ("core: implicitly order units with PrivateTmp= after systemd-tmpfiles-setup.service")

which was added in v233.

jgunthorpe:
Most probably, seems like a simple enough thing to test it?

aron-silverton:
> Most probably, seems like a simple enough thing to test it?

Never heard anybody say that about systemd before... ;-)

I should have an update late next week - trying to find somebody with cycles to build a patched systemd but lots of vacations going on at the moment including mine.

jgunthorpe:
Just add the depends to the service file and verify it fixes it

markhaywood:
Right. Adding "After=systemd-tmpfiles-setup.service" to srp_daemon.service fixes the issue. I think that is basically what the patch to systemd would do.

aron-silverton:
> Just add the depends to the service file and verify it fixes it

Of course. Sorry, did not have my thinking had on at the time.

PR #789, opened 2020-07-09
Add support for 0xefa1 device

galpress:
Kernel submission:
v1: https://lore.kernel.org/linux-rdma/20200709083630.21377-1-galpress@amazon.com/
v2: https://lore.kernel.org/linux-rdma/20200720080113.13055-1-galpress@amazon.com/
v3: https://lore.kernel.org/linux-rdma/20200721133049.74349-1-galpress@amazon.com/
v4: https://lore.kernel.org/linux-rdma/20200722140312.3651-1-galpress@amazon.com/

### DIFF
--- a/providers/efa/efa.c
+++ b/providers/efa/efa.c
@@ -58,6 +58,7 @@ static struct verbs_context *efa_alloc_context(struct ibv_device *vdev,
 	int err;
 
 	cmd.comp_mask |= EFA_ALLOC_UCONTEXT_CMD_COMP_TX_BATCH;
+	cmd.comp_mask |= EFA_ALLOC_UCONTEXT_CMD_COMP_MIN_SQ_WR;
 
 	ctx = verbs_init_and_alloc_context(vdev, cmd_fd, ctx, ibvctx,
 					   RDMA_DRIVER_EFA);
@@ -74,6 +75,7 @@ static struct verbs_context *efa_alloc_context(struct ibv_device *vdev,
 	ctx->inline_buf_size = resp.inline_buf_size;
 	ctx->max_llq_size = resp.max_llq_size;
 	ctx->max_tx_batch = resp.max_tx_batch;
+	ctx->min_sq_wr = resp.min_sq_wr;
 	pthread_spin_init(&ctx->qp_table_lock, PTHREAD_PROCESS_PRIVATE);
 
 	/* ah udata is mandatory for ah number retrieval */
--- a/providers/efa/efa.c
+++ b/providers/efa/efa.c
@@ -20,6 +20,7 @@ static void efa_free_context(struct ibv_context *ibvctx);
 static const struct verbs_match_ent efa_table[] = {
 	VERBS_DRIVER_ID(RDMA_DRIVER_EFA),
 	VERBS_PCI_MATCH(PCI_VENDOR_ID_AMAZON, 0xefa0, NULL),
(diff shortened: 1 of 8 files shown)


providers/efa/efa.c
```diff
@@ -20,6 +20,7 @@ static void efa_free_context(struct ibv_context *ibvctx);
 static const struct verbs_match_ent efa_table[] = {
 	VERBS_DRIVER_ID(RDMA_DRIVER_EFA),
 	VERBS_PCI_MATCH(PCI_VENDOR_ID_AMAZON, 0xefa0, NULL),
+	VERBS_PCI_MATCH(PCI_VENDOR_ID_AMAZON, 0xefa1, NULL),
```
jgunthorpe, line 23: No more updates to pci driver tables. VERBS_DRIVER_ID covers everything.

jgunthorpe:
Keep in mind the provider today will load on the new driver. Is it Ok or do you need more compatability negotaition during alloc_ucontext?

galpress:
> Keep in mind the provider today will load on the new driver. Is it Ok or do you need more compatability negotaition during alloc_ucontext?

No, that's not going to work with today's provider.
What's the motivation to stop using pci ids? Isn't preventing the provider to load on this id better than a ucontext handshake?

jgunthorpe:
All the providers are expected to negotiate with the kernel if the right combination stuff is supported.

Will this upgraded provider work on old kernels? That would be the rare exception.

Generally the model is provider/kernel binding is only by DRIVER_ID and any variance beyond that should be fully negotiated as part of get_context.

galpress:
> All the providers are expected to negotiate with the kernel if the right combination stuff is supported.
> 
> Will this upgraded provider work on old kernels? That would be the rare exception.

Probably, how old is old?

> Generally the model is provider/kernel binding is only by DRIVER_ID and any variance beyond that should be fully negotiated as part of get_context.

jgunthorpe:
How Old? All prior kernels

galpress:
Well, the provider is expected to work with prior kernels as well.
I don't think I understand what you mean?

jgunthorpe:
Well, lets see a patch that doesn't use PCI_ID then maybe it will be clear

galpress:
Done.
The pciid patch is still here because I had some issues without it, let's discuss it.

jgunthorpe:
What are some issues without it?

galpress:
The driver id binding didn't catch on one of the kernels I tested.

galpress:
I assume you replied on the wrong thread:

> How did such an old kernel have support for this new EFA device?

An older kernel with backported 0xefa1 support.

rleon:
> I assume you replied on the wrong thread:
> 
> > How did such an old kernel have support for this new EFA device?
> 
> An older kernel with backported 0xefa1 support.

Is this upstream kernel?
We rarely care about other cases.

galpress:
It's not an upstream kernel, but specific distros might backport these patches to support new devices.
I get the upstream kernel point, but we can't have these new devices only working on latest upstream kernel..
Do you strongly object keeping the pciid single liner? I can't see a way we keep shipping our customers upstream rdma-core without it :\\.

jgunthorpe:
I'd rather we not have forks of rdma-core for such a trivial thing. It seems we will have to wait a bit before going to the no new pci_id's.

galpress:
Updated the kernel headers commit, let me know if anything else is needed.

galpress:
Kind reminder.

galpress:
Is there anything stopping this PR from being merged?

PR #790, opened 2020-07-12
tests: Skip tests with unsupported MR access flags

galpress:
The shared PD test triggers a flow that tries to create an MR through
QpExRCRDMAWrite class with access flags that might not be supported on
some devices, skip the test in that case.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -4,6 +4,7 @@
 Provide some useful helper function for pyverbs' tests.
 """
 from itertools import combinations as com
+import errno
 import unittest
 import random
 import socket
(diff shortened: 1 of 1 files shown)


tests/test_qpex.py
```diff
@@ -105,7 +105,13 @@ def create_qp(self):
         self.qp = create_qp_ex(self, e.IBV_QPT_RC, e.IBV_QP_EX_WITH_RDMA_WRITE)
 
     def create_mr(self):
-        self.mr = u.create_custom_mr(self, e.IBV_ACCESS_REMOTE_WRITE)
+        try:
+            access_flags = e.IBV_ACCESS_REMOTE_WRITE
+            self.mr = u.create_custom_mr(self, access_flags)
```
EdwardSro, line 110: hm.. why not to include this try-catch check inside the create_custom_mr itself?
This will cover all current/future similar scenarios that call create_custom_mr.

galpress: Done.

PR #793, opened 2020-07-18
add ${NL_LIBRARY_DIRS} to link_directories()

chuckcranor:
Handles case where nl-route-3 and nl-3 libs are not installed in
a standard location.  pkg_check_modules() puts the location it
got from pkgconfig in the NL_LIBRARY_DIRS variable.

For example, I have a test copy of nl-route-3 and nl-3 in /tmp/test/lib.  There is no other version of these libs installed on my system.   I can configure rdma-core with:
<pre>
cmake -DCMAKE_PREFIX_PATH=/tmp/test -DCMAKE_INSTALL_PATH=/tmp/test -DPKG_CONFIG_USE_CMAKE_PREFIX_PATH=YES ../rdma-core
</pre>

Without the fix, the compile will fail with:
<pre>
[  7%] Linking C executable ../bin/rdma_rename
/usr/bin/ld: cannot find -lnl-route-3
/usr/bin/ld: cannot find -lnl-3
collect2: error: ld returned 1 exit status
</pre>

because the "-L" linking information returned by pkg_check_modules() in ${NL_LIBRARY_DIRS}  is ignored by the rdma-core CMakeLists.txt.  This is easy to fix by adding a call to link_directories().

(an alternate solution if you are using cmake 3.6 or newer to use an imported target created by pkg_check_modules() with the "IMPORTED_TARGET" argument that got added in cmake 3.6)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -442,6 +442,7 @@ if (ENABLE_RESOLVE_NEIGH)
   # FIXME use of pkgconfig is discouraged
   pkg_check_modules(NL libnl-3.0 libnl-route-3.0 REQUIRED)
   include_directories(${NL_INCLUDE_DIRS})
+  link_directories(${NL_LIBRARY_DIRS})
   set(NL_KIND 3)
 else()
   set(NL_KIND 0)


rleon:
Yes, it seems like the most simple solution.
I wonder if the better variant will be to convert pkg_check_modules() should be converted to find_library()/find_package().

jgunthorpe:
How did such an old kernel have support for this new EFA device?

chuckcranor:
> Yes, it seems like the most simple solution.
> I wonder if the better variant will be to convert pkg_check_modules() should be converted to find_library()/find_package().

That's easy to do too, and then you can stop using pkgconfig as noted in the "FIXME" comment.  If you did that, you'd use CMAKE_PREFIX_PATH cmake variable to point at the desired libnl3 to use rather than pkgconfig's PKG_CONFIG_PATH environment variable.  I just tested the following and it seems to work.  You won't need the link_directories() call anymore because find_library() returns the full path to the requested library (so no need for a "-L").  if this works better you can use the diff below or i can do another pullup request if you want.

<pre>
index edab9b0..7cfa8a2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,6 +167,7 @@ include(CMakeParseArguments)
 include(CheckCCompilerFlag)
 include(CheckIncludeFile)
 include(CheckTypeSize)
+include(FindPackageHandleStandardArgs)
 include(RDMA_EnableCStd)
 include(RDMA_Sparse)
 include(RDMA_BuildType)
@@ -432,17 +433,21 @@ FIND_PACKAGE(rst2man)
 # pthread
 FIND_PACKAGE (Threads REQUIRED)
 
-FIND_PACKAGE(PkgConfig REQUIRED)
-
 # libnl
 if (NOT DEFINED ENABLE_RESOLVE_NEIGH)
   set(ENABLE_RESOLVE_NEIGH "ON" CACHE BOOL "Enable internal resolution of neigh
bours for Etherent")
 endif()
 if (ENABLE_RESOLVE_NEIGH)
-  # FIXME use of pkgconfig is discouraged
-  pkg_check_modules(NL libnl-3.0 libnl-route-3.0 REQUIRED)
+  find_path(NL_INCLUDE_DIRS netlink/netlink.h PATH_SUFFIXES libnl3)
+  find_library(NL_LIBNL3 nl-3)
+  find_library(NL_LIBNL_ROUTE3 nl-route-3)
+  find_package_handle_standard_args(NL DEFAULT_MSG NL_INCLUDE_DIRS
+    NL_LIBNL3 NL_LIBNL_ROUTE3)
   include_directories(${NL_INCLUDE_DIRS})
-  link_directories(${NL_LIBRARY_DIRS})
+  if (NOT NL_INCLUDE_DIRS OR NOT NL_LIBNL3 OR NOT NL_LIBNL_ROUTE3)
+    message(FATAL_ERROR "libnl-3 not found: disable it or fix the prefix path")
+  endif()
+  set(NL_LIBRARIES ${NL_LIBNL_ROUTE3} ${NL_LIBNL3})
   set(NL_KIND 3)
 else()
   set(NL_KIND 0)
</pre>

rleon:
I would love to see that.
@jgunthorpe are you ok with this change?

jgunthorpe:
It turned out we needed the pkgconfig to make static linking working, IIRC

HuangShumin:
I find out rdma-core built with cmake -DENABLE_STATIC=1, libnl-3 and libnl-route-3 still cannot be staticly linked to libibverbs and librdmacm.

anyone who notices about this?

rleon:
> I find out rdma-core built with cmake -DENABLE_STATIC=1, libnl-3 and libnl-route-3 still cannot be staticly linked to libibverbs and librdmacm.
> 
> anyone who notices about this?

We didn't get any reports like this. Are you sure that libnl-3 and libnl-route-3 are compiled as static? Do they have *.a files?

HuangShumin:
yes, there does exist static netlink library files.
```
ls /usr/lib/x86_64-linux-gnu/libnl*.a
/usr/lib/x86_64-linux-gnu/libnl-3.a  /usr/lib/x86_64-linux-gnu/libnl-route-3.a
```
And in CMakeCache.txt:
```
//Path to a library.
pkgcfg_lib_NL_nl-3:FILEPATH=/usr/lib/x86_64-linux-gnu/libnl-3.a

//Path to a library.
pkgcfg_lib_NL_nl-route-3:FILEPATH=/usr/lib/x86_64-linux-gnu/libnl-route-3.a
```

rleon:
and what is the error when you try to build rdma-core?

HuangShumin:
there's no error during building rdma-core, only when other apps try to staticly link libibverbs.a and librdmacm.a, errors as below:
```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/librdmacm.a(cma.c.o): in function `check_abi_version_nl_cb':
(.text+0x4fb): undefined reference to `nlmsg_hdr'
/usr/bin/ld: (.text+0x514): undefined reference to `nlmsg_parse'
/usr/bin/ld: (.text+0x54c): undefined reference to `nla_get_u64'
/usr/bin/ld: (.text+0x562): undefined reference to `nla_get_string'
/usr/bin/ld: (.text+0x594): undefined reference to `nla_get_u64'
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/../../../x86_64-linux-gnu/librdmacm.a(cma.c.o): in function `ucma_query_route':
(.text+0x1255): undefined reference to `ibv_copy_path_rec_from_kern'
...
```
It seems rdma-core won't staticly link libnl-3.a and libnl-route-3.a

rleon:
Try to build with -DENABLE_RESOLVE_NEIGH=1

HuangShumin:
Same result.
Do you have the way to link libnl-3.a/libnl-route-3.a to librdmacm.a statically?

HuangShumin:
or can we put netlink source code to rdma-core src tree, so that rdma-core can be built-in with netlink code?

jgunthorpe:
This is fully expected, it is how static linking works. There is no expectation that static libraries include all of their dependencies (which might not even be static!)

The apps could to use pkgconfig to learn any additional libraries they require when static linking. If I recall we annotated that correctly.

HuangShumin:
crystally clear.

PR #800, opened 2020-07-31
allow install to complete even if no prebuilt man pages are available

chuckcranor:
allow install to complete even if no prebuilt man pages are available
    
If pandoc and rst2man are available, they are used to create man
page sources.  If not, then the build looks for prebuilt man pages
in ${CMAKE_SOURCE_DIR}/buildlib/pandoc-prebuilt.   If that directory
is not present (e.g. because you are compiling from a fresh checkout
of master from github), then rdma-core will compile fine but the
"make install" operation will fail.
    
This patch adds a new cmake cache variable NO_MAN_PAGES (default=0).
If set, it disables man page processing and allows "make install" to
succeed (without man pages) when neither pandoc/rst2man nor pandoc-prebuilt
are present.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,10 @@
 #   -DENABLE_IBDIAGS_COMPAT=True (default False)
 #      Include obsolete scripts. These scripts are replaced by C programs with
 #      a different interface now.
+#   -DNO_MAN_PAGES=1 (default 0, build/install the man pages)
+#      Disable man pages. Allows rdma-core to be built and installed
+#      (without man pages) when neither pandoc/rst2man nor the pandoc-prebuilt
+#      directory are available.
 
 cmake_minimum_required(VERSION 2.8.11 FATAL_ERROR)
 project(rdma-core C)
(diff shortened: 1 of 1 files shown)


jgunthorpe:
I'm really against the idea that rdma-core should 'make install' something that is defective.

I could agree to some '-DNO_MAN_PAGES' option to cmake that just skipped all man pages?

chuckcranor:
No problem, I will revise the pull to do it with something like a cmake -DNO_MAN_PAGES option.

Seems like with that in place, the failure should occur at "cmake" configuration time rather than at "make install" time?

chuckcranor:
I've revised the patch to use -DNO_MAN_PAGES.

CMakeLists.txt
```diff
@@ -703,18 +718,22 @@ else()
     message(STATUS " netlink/route/link.h and net/if.h NOT co-includable (old headers)")
   endif()
 endif()
-if (NOT PANDOC_FOUND)
-  if (NOT EXISTS "${CMAKE_SOURCE_DIR}/buildlib/pandoc-prebuilt")
-    message(STATUS " pandoc NOT found and NO prebuilt man pages. 'install' disabled")
-  else()
-    message(STATUS " pandoc NOT found (using prebuilt man pages)")
+if (NO_MAN_PAGES)
+  message(STATUS " man pages NOT enabled")
```
rleon, line 722: NOT enabled -> NOT built

rleon:
And I don't know how to add review comment about commit message, but you have places with double and tripple spaces between sentences.

Everything else is excellent +1.

CMakeLists.txt
chuckcranor: I updated the code with the suggested wording and removed the spaces from the commit msg.

PR #801, opened 2020-08-03
Add a dot to kernel headers update script commit message

galpress:
Commit messages usually end with a dot, do that in the script as well.


rleon:
Thanks

PR #803, opened 2020-08-09
pyverbs: Delete all tests that generates mlx5 CQE errors

rleon:
Library shouldn't print anything to stdout and mlx5 shouldn't be
exclusion here. Delete all pyverbs tests that generate CQE errors
in mlx5, till the developers will fix it.


galpress:
Are you deleting generic tests because of an mlx5 issue?

rleon:
> Are you deleting generic tests because of an mlx5 issue?

Yes, I don't want to release rdma-core with CQE junk prints.
@EdwardSro will work to add them.

rleon:
Starting from now, all new pyverbs tests should have -v output attached on supported and not supported devices.

Thanks

PR #808, opened 2020-08-18
rdma.init-kernel: Do not load blacklisted modules

rixon:
Add -b option to modprobe so blacklisted modules are not loaded.  This
makes it easier to prevent loading modules for certain Ethernet cards
with iWarp when only InfiniBand should be used for RDMA.


rleon:
Duplicate of https://github.com/linux-rdma/rdma-core/pull/797

PR #810, opened 2020-08-19
tests: Split QP tests based on QP types

Kamalheib:
As not all providers support all QP types, split the tests to be based
on the QP type, this is needed to make sure that all QP types that are
supported by a specific provider are tested.

### DIFF
--- a/tests/test_qp.py
+++ b/tests/test_qp.py
@@ -1,5 +1,7 @@
 # SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
 # Copyright (c) 2019 Mellanox Technologies, Inc. All rights reserved. See COPYING file
+# Copyright (c) 2020 Kamal Heib <kamalheib1@gmail.com>, All rights reserved.  See COPYING file
+
 """
 Test module for pyverbs' qp module.
 """
@@ -11,6 +13,7 @@ import os
 from pyverbs.pyverbs_error import PyverbsRDMAError
 from pyverbs.qp import QPInitAttr, QPAttr, QP
 from tests.base import PyverbsAPITestCase
+import pyverbs.utils as pu
 import pyverbs.enums as e
 from pyverbs.pd import PD
 from pyverbs.cq import CQ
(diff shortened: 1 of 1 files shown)


tests/test_qp.py
```diff
@@ -22,12 +22,11 @@ class QPTest(PyverbsAPITestCase):
...
     """
 
-    def test_create_qp_no_attr_connected(self):
+    def test_create_rc_qp_no_attr(self):
         """
-        Test QP creation via ibv_create_qp without a QPAttr object proivded.
-        Checked QP types are RC and UC.
+        Test RC QP creation via ibv_create_qp without a QPAttr object proivded.
         """
-        for ctx, attr, attr_ex in self.devices:
```
EdwardSro, line 30: consider using one "config" (one device, instead of iterating on all devices. splitting the tests and running on all devices will increase the run time).

tests/test_qp.py
```diff
@@ -37,48 +36,68 @@ def test_create_qp_no_attr_connected(self):
...
+                    try:
+                        with QP(pd, qia) as qp:
+                            assert qp.qp_state == e.IBV_QPS_RESET, 'UD QP should have been in RESET'
+                    except PyverbsRDMAError as ex:
+                        if ex.error_code == errno.EOPNOTSUPP:
+                            raise unittest.SkipTest('Create UD QP without attrs is not supported')
+                        raise ex
 
-    def test_create_qp_no_attr(self):
+    def test_create_raw_qp_no_attr(self):
```
EdwardSro, line 76: you can use the "@requires_root_on_eth" decorator (from utils) for raw_qp tests (in case you're using one device/port)

tests/test_qp.py
```diff
@@ -139,138 +177,253 @@ def test_create_qp_ex_no_attr_connected(self):
...
+                            raise unittest.SkipTest('Create UD QP is not supported')
+                        raise ex
+                    qp_attr, qp_init_attr = qp.query(e.IBV_QP_STATE | e.IBV_QP_CAP)
+                    verify_qp_attrs(caps, e.IBV_QPS_RESET, qp_init_attr, qp_attr)
+
+    def test_query_raw_qp(self):
         """
-        Queries a QP after creation. Verifies that its properties are as
-        expected.
+        Queries an RAW Packet QP after creation. Verifies that its properties are as expected.
```
EdwardSro, line 396: keep docstrings cut at 80c (72c PEP recommendation)
(apply on the other relevant places)

tests/test_qp.py
```diff
@@ -286,24 +439,29 @@ def test_modify_qp(self):
...
                     qa.qkey = 0x123
                     qp.to_init(qa)
-                    qp_attr, qp_iattr = qp.query(e.IBV_QP_QKEY)
+                    qp_attr, _ = qp.query(e.IBV_QP_QKEY)
                     assert qp_attr.qkey == qa.qkey, 'Extended QP, QKey is not as expected'
                     qp.to_rtr(qa)
                     qa.sq_psn = 0x45
                     qp.to_rts(qa)
-                    qp_attr, qp_iattr = qp.query(e.IBV_QP_SQ_PSN)
+                    qp_attr, _ = qp.query(e.IBV_QP_SQ_PSN)
```
EdwardSro, line 452: you can remove it in lines 468 and 473

tests/test_qp.py
```diff
@@ -139,138 +177,253 @@ def test_create_qp_ex_no_attr_connected(self):
...
+                            try:
+                                qp = QP(pd, qia)
+                            except PyverbsRDMAError as ex:
+                                if ex.error_code == errno.EOPNOTSUPP:
+                                    raise unittest.SkipTest('Create RAW Packet QP is not supported')
+                                raise ex
+                            qp_attr, qp_init_attr = qp.query(e.IBV_QP_STATE | e.IBV_QP_CAP)
+                            verify_qp_attrs(caps, e.IBV_QPS_RESET, qp_init_attr, qp_attr)
 
     def test_modify_qp(self):
```
EdwardSro, line 428: can you please change the name to test_modify_ud_qp? (and update the docstring accordingly)

tests/test_qp.py
```diff
@@ -22,12 +22,11 @@ class QPTest(PyverbsAPITestCase):
     Test various functionalities of the QP class.
     """
```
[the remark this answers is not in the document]
Kamalheib: I agree I'll fix it in v2.

tests/test_qp.py
Kamalheib: This is already handled by the PyverbsAPITestCase() class under the setUp() function.

tests/test_qp.py
Kamalheib: yes, I'll give it a try.

tests/test_qp.py
Kamalheib: I'll fix it in v2.

tests/test_qp.py
Kamalheib: I'll fix it in v2.

tests/test_qp.py
Kamalheib: Yes, I'll fix it in v2.

tests/test_qp.py
EdwardSro: Nope. In case the user did not pass --dev, PyverbsAPITestCase() will add all devices (from ibv_get_device_list) to self.devices.
Which will cause each one of these tests to be run over each device/port.
On the other hand, these tests are basic, which should not take long (after I checked, all of them finish in up to a second even with your split).

Kamalheib: Isn't this the whole idea from all the tests that inherits from PyverbsAPITestCase(), I mean to go over all the available devices and run tests unless the user pass --dev.

Do you suggest a new design?

EdwardSro: My concern was that the tests start to take much longer because of multi-device runs.
BTW all traffic tests (which inherits from RDMATestCase) randomly choose one "config".

But I'll take this back, you can keep it as is for now. Initially, I thought that test_qp takes longer than the actual results.

tests/test_qp.py
```diff
@@ -371,3 +301,30 @@ def get_qp_init_attr_ex(cq, pd, attr, attr_ex, qpt):
...
+        e.IBV_QPS_RESET : 'RESET',
+        e.IBV_QPS_INIT : 'INIT',
+        e.IBV_QPS_RTR : 'RTR',
+        e.IBV_QPS_RTS : 'RTS',
+        e.IBV_QPS_SQD : 'SQD',
+        e.IBV_QPS_SQE : 'SQE',
+        e.IBV_QPS_ERR : 'ERR',
+        e.IBV_QPS_UNKNOWN : 'UNKNOWN'
+    }
+    return qps_d.get(qp_state)
```
EdwardSro, line 330: You can use qp_type_to_str and qp_state_to_str respectively from pyverbs.utils

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
+    def test_create_ud_qp_with_attr(self):
+        """
+        Test UD QP creation via ibv_create_qp with a QPAttr object proivded.
+        """
+        self.create_qp_common_test(e.IBV_QPT_UD, e.IBV_QPS_RTS, False, True)
+
+    def test_create_raw_qp_with_attr(self):
+        """
+        Test RAW Packet QP creation via ibv_create_qp with a QPAttr object
+        proivded.
```
EdwardSro, line 108: type: provided*
(apply on the other docstrings)

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
-        Checked QP types are Raw Packet and UD. Raw Packet is skipped for
-        non-root users / Infiniband link layer.
-        """
+    def create_qp(self, creator, qp_init_attr, is_ex, with_attr):
+        try:
+            qp_attr = (None, QPAttr())[with_attr]
+            return QP(creator, qp_init_attr, qp_attr)
+        except PyverbsRDMAError as ex:
+            if ex.error_code == errno.EOPNOTSUPP:
+                with_str = ('with', 'without')[with_attr] + ('', ' extended')[is_ex]
```
EdwardSro, line 33: reversed: ('without', 'with')*

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
+                raise unittest.SkipTest(f'Create {get_qp_type_str(qp_init_attr.qp_type)} QP {with_str} attrs is not supported')
```
EdwardSro, line 34: please assign "get_qp_type_str(qp_init_attr.qp_type)" to a temp var. (128c length is just too much)

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
-
-    def test_create_qp_with_attr_connected(self):
-        """
-        Test QP creation via ibv_create_qp without a QPAttr object proivded.
-        Checked QP types are RC and UC.
+                        qia.qp_type = qp_type
+                        creator = pd
+
+                    qp = self.create_qp(creator, qia, is_ex, with_attr)
+                    assert qp.qp_state == qp_state , f'{get_qp_type_str(qp_type)} QP should have been in {get_qp_state_str(qp_state)}'
```
EdwardSro, line 59: please assign "get_qp_type_str(qp_init_attr.qp_type)" to a temp var. (128c length is just too much).
remove white space after qp_state

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
                 with CQ(ctx, 100, None, None, 0) as cq:
-                    for i in range(1, attr.phys_port_cnt + 1):
+                    if qp_type == e.IBV_QPT_RAW_PACKET:
+                        eth_port = 0
+                        for i in range(1, attr.phys_port_cnt + 1):
+                            if u.is_eth(ctx, i) and u.is_root():
+                                eth_port = i
+                                break
+                        if eth_port == 0:
+                            raise unittest.SkipTest('To Create RAW QP must done by root on Ethernet link layer')
```
EdwardSro, line 48: must be* done

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
-    """
-    qpts = [e.IBV_QPT_RC, e.IBV_QPT_UC, e.IBV_QPT_UD]
-    if u.is_eth(ctx, port_num) and u.is_root():
-        qpts.append(e.IBV_QPT_RAW_PACKET)
-    return qpts
-
-
 def verify_qp_attrs(orig_cap, state, init_attr, attr):
     assert state == attr.cur_qp_state
     assert orig_cap.max_send_wr <= init_attr.cap.max_send_wr
```
EdwardSro, line 268: btw you can move this to QPTest class, so you can use self.assertEqual instead.
i.e: self.assertEqual(state , attr.cur_qp_state).
This will print both values in case of failure (which assert doesn't do).

EdwardSro (changes requested):
V2 looks much better. Thanks!
There a few more comments to address.
+ All query tests fail. I guess because you're passing an attribute to the QP(), which it automatically modifies the state of the QP (already in the init). Please check that. (if that's the case then either skip passing an attr - by calling your create_qp auxiliary function with_attr false or changing the expected qp state assertion)

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
-                        raise ex
-
-
-    def test_create_qp_no_attr(self):
-        """
-        Test QP creation via ibv_create_qp without a QPAttr object proivded.
-        Checked QP types are Raw Packet and UD. Raw Packet is skipped for
-        non-root users / Infiniband link layer.
-        """
+    def create_qp(self, creator, qp_init_attr, is_ex, with_attr):
```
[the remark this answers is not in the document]
Kamalheib: Fixed.

tests/test_qp.py
```diff
@@ -22,317 +24,245 @@ class QPTest(PyverbsAPITestCase):
...
+        try:
+            qp_attr = (None, QPAttr())[with_attr]
+            return QP(creator, qp_init_attr, qp_attr)
+        except PyverbsRDMAError as ex:
+            if ex.error_code == errno.EOPNOTSUPP:
+                with_str = ('with', 'without')[with_attr] + ('', ' extended')[is_ex]
+                raise unittest.SkipTest(f'Create {get_qp_type_str(qp_init_attr.qp_type)} QP {with_str} attrs is not supported')
+            raise ex
+
+    def create_qp_common_test(self, qp_type, qp_state, is_ex, with_attr):
```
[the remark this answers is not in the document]
Kamalheib: Fixed.

tests/test_qp.py
Kamalheib: I missed these two functions, Fixed.

tests/test_qp.py
Kamalheib: Fixed.

tests/test_qp.py
Kamalheib: Fixed.

tests/test_qp.py
Kamalheib: Done.

tests/test_qp.py
Kamalheib: Done.

tests/test_qp.py
Kamalheib: Fixed.

tests/test_qp.py
Kamalheib: Ack, moved to the QPTest class.

Kamalheib:
> V2 looks much better. Thanks!
> There are a few more comments to address.
> BTW, now all query tests fail. I guess because you're passing an attribute to the QP(), which it automatically modifies the state of the QP (already in the init). Please check that. (if that's the case then either skip passing an attr - by calling your create_qp auxiliary function with_attr false or changing the expected qp state assertion)

@EdwardSro  Thanks a lot for your valuable review!

The query tests should be fixed by now, this was my bad by mixing between the is_ex and with_attr flags.

EdwardSro (approved):
Looks okay to me. Thanks.

PR #816, opened 2020-09-01
verbs: Set attributes to zero if query_device_ex() is not supported

Li-Weihang:
All of structrue ibv_device_attr_ex should be set to zero if a provider don't support query_device_ex(), or the fields which are not included in attr->orig_attr will get a random value.

Fixes: 58ccb638b540 ("libibverbs: Fix query_device_ex dummy function not to return EOPNOTSUPP")

### DIFF
--- a/libibverbs/dummy_ops.c
+++ b/libibverbs/dummy_ops.c
@@ -396,7 +396,7 @@ static int query_device_ex(struct ibv_context *context,
 	if (attr_size < sizeof(attr->orig_attr))
 		return EOPNOTSUPP;
 
-	memset(&attr->orig_attr, 0, sizeof(attr->orig_attr));
+	memset(attr, 0, attr_size);
 
 	return ibv_query_device(context, &attr->orig_attr);
 }
--- a/libibverbs/examples/devinfo.c
+++ b/libibverbs/examples/devinfo.c
@@ -495,7 +495,7 @@ static void print_raw_packet_caps(uint32_t raw_packet_caps)
 static int print_hca_cap(struct ibv_device *ib_dev, uint8_t ib_port)
 {
 	struct ibv_context *ctx;
-	struct ibv_device_attr_ex device_attr;
+	struct ibv_device_attr_ex device_attr = {};
 	struct ibv_port_attr port_attr;
 	int rc = 0;
 	uint8_t port;


rleon (commented):
It won't return garbage but will leave struct provided by the user unchanged.

Li-Weihang:
> It won't return garbage but will leave struct provided by the user unchanged.

Thanks for your comments. If query_device_ex() in dummy_ops.c is called, it means that the provider don't support ibv_query_device_ex(), all fields in structrue ibv_device_attr_ex except orig_attr means nothing to the users. I think we should set other fields to zero in core codes because the user maybe don't know that they have to initialize structrue ibv_device_attr_ex to zero at first. For example, if I run ibv_devinfo on hns devices, I will get some random value because the function print_hca_cap() in devinfo.c don't inlitialize device_attr to zero. Am I right? Or I should just fix the initialization issue in devinfo.c?

rleon:
> > It won't return garbage but will leave struct provided by the user unchanged.
> 
> Thanks for your comments. If query_device_ex() in dummy_ops.c is called, it means that the provider don't support ibv_query_device_ex(), all fields in structrue ibv_device_attr_ex except orig_attr means nothing to the users. I think we should set other fields to zero in core codes because the user maybe don't know that they have to initialize structrue ibv_device_attr_ex to zero at first. For example, if I run ibv_devinfo on hns devices, I will get some random value because the function print_hca_cap() in devinfo.c don't inlitialize device_attr to zero. Am I right? Or I should just fix the initialization issue in devinfo.c?

Let's do both.
1. After you will add Fixes line, I'll merge it.
2. Please fix ibv_devinfo too.

Thanks

Li-Weihang:
> > > It won't return garbage but will leave struct provided by the user unchanged.
> > 
> > 
> > Thanks for your comments. If query_device_ex() in dummy_ops.c is called, it means that the provider don't support ibv_query_device_ex(), all fields in structrue ibv_device_attr_ex except orig_attr means nothing to the users. I think we should set other fields to zero in core codes because the user maybe don't know that they have to initialize structrue ibv_device_attr_ex to zero at first. For example, if I run ibv_devinfo on hns devices, I will get some random value because the function print_hca_cap() in devinfo.c don't inlitialize device_attr to zero. Am I right? Or I should just fix the initialization issue in devinfo.c?
> 
> Let's do both.
> 
> 1. After you will add Fixes line, I'll merge it.
> 2. Please fix ibv_devinfo too.
> 
> Thanks

OK, thank you :)

libibverbs/dummy_ops.c
```diff
@@ -396,7 +396,7 @@ static int query_device_ex(struct ibv_context *context,
 	if (attr_size < sizeof(attr->orig_attr))
 		return EOPNOTSUPP;
 
-	memset(&attr->orig_attr, 0, sizeof(attr->orig_attr));
+	memset(attr, 0, sizeof(*attr));
```
jgunthorpe, line 399: This is not right, it should be 

memset(attr, 0, attr_size);

rleon: Why not? We are cleaning all struct without relation to provided size.

jgunthorpe: That will smash the stack, attr_size can be less than sizeof(*attr)

Li-Weihang: OK, thank you.

PR #818, opened 2020-09-02
kernel-boot: Load mlx5_ib and mlx4_ib also for InfiniBand only cards

alaahl:
Usually the initramfs is built with the device driver modules which are
the mlx4_core and mlx5_core for Mellanox HW. The mlx4_ib and mlx5_ib are
usually not packed inside the initramfs.

Hence, when an InfiniBand only card is installed on a system, the current
udev rules will not work and the needed RDMA driver is not loaded, that is
because we do not have an ethernet netdev, so we don't have an interface
that can match on the ID_NET_DRIVER check, that can work only when we have
a device that loads an ethernet netdev by default.

Note that both of mlx4 and mlx5 drivers call request_module_nowait() to
load their corresponding RDMA module, but that cannot work if only the
core module exists inside the initramfs.

Fix this by looking at the pci subsystem events and load the needed RDMA
module based on the driver owning the current pci device.

Signed-off-by: Alaa Hleihel <alaa@nvidia.com>
CC: Honggang Li <honli@redhat.com>

### DIFF
--- a/kernel-boot/rdma-hw-modules.rules
+++ b/kernel-boot/rdma-hw-modules.rules
@@ -1,11 +1,12 @@
 ACTION=="remove", GOTO="rdma_hw_modules_end"
-SUBSYSTEM!="net", GOTO="rdma_hw_modules_end"
 
+SUBSYSTEM!="net", GOTO="rdma_hw_modules_net_end"
+# For Ethernet cards with RoCE support
 # Automatically load RDMA specific kernel modules when a multi-function device is installed
-
 # These drivers autoload an ethernet driver based on hardware detection and
 # need userspace to load the module that has their RDMA component to turn on
 # RDMA.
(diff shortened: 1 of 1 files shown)


Honggang-LI:
Acked-by: Honggang Li <honli@redhat.com>

Thanks

aron-silverton:
Should " rdma-description.rules" set ID_RDMA_INFINIBAND based on mlxN_ib rather than _core? It is currently:

```
DRIVERS=="mlx4_core", ENV{ID_RDMA_INFINIBAND}="1"
DRIVERS=="mlx5_core", ENV{ID_RDMA_INFINIBAND}="1"
```

jgunthorpe:
It has to match the module that binds to the PCI device, which is the mlx_core modules

PR #821, opened 2020-09-03
 Add support for user mode XRC-SRQ's

mkalderon:
rdma-core part of the user mode XRC-SRQ's support
Kernel patch: https://patchwork.kernel.org/patch/11678337/

Signed-off-by: Yuval Basson <ybason@marvell.com>
Signed-off-by: Michal Kalderon <mkalderon@marvell.com>


mkalderon:
@jgunthorpe @rleon 
Kernel side of this was applied: https://lore.kernel.org/linux-rdma/20200722102339.30104-1-ybason@marvell.com/
thanks,
Michal

providers/qedr/qelr_hsi_rdma.h
```diff
@@ -163,6 +168,7 @@ enum rdma_cqe_type
 	RDMA_CQE_TYPE_REQUESTER,
 	RDMA_CQE_TYPE_RESPONDER_RQ,
 	RDMA_CQE_TYPE_RESPONDER_SRQ,
+	RDMA_CQE_TYPE_RESPONDER_XRC_SRQ,
```
rleon, line 171: Are you sure that addition here won't break anything?
You moved RDMA_CQE_TYPE_INVALID from the previous value.

mkalderon: @rleon yes, this enum is hsi with FW. Before support for XRC_SRQ rdma-core treated a cqe_type
of this type as invalid, now with support it is a valid CQE. thanks

rleon: Last question, won't you have a compatibility problem with old FW and new rdma-core?

Thanks

mkalderon: @rleon good point. I just checked with FW team and apparently CQE_TYPE_INVALID is never used and was never used. So also with old-FW there is no flow that can reach a cqe with this type. thanks!

rleon:
Thanks

PR #824, opened 2020-09-06
verbs: Make sure VM_DONTCOPY removed

yishaih:
If the interval passed to madvise() covers both mapped and unmapped address ranges error will be returned but mapped address ranges still will be processed.

Remove the VM_DONTCOPY flag if MADV_DOFORK has failed by iterating the suspicious region page-by-page.

### DIFF
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -587,6 +587,28 @@ static struct ibv_mem_node *undo_node(struct ibv_mem_node *node,
 	return node;
 }
 
+static int do_madvise(void *addr, size_t length, int advice,
+		      unsigned long range_page_size)
+{
+	int ret;
+	void *p;
+
+	ret = madvise(addr, length, advice);
+
+	if (!ret || advice == MADV_DONTFORK)
+		return ret;
+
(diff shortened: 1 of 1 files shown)


zhngaj (commented):
It might be not completely related to this patch. 
Since this patch also involves handling madvise on unmapped region, I'm wondering what could be the case where madvise() covers a unmaped region, calling free() before ibv_dereg_mr()?
If that's the case, instead of TRYING to recover by iterating the suspicious region page-by-page, why not returning the error code of ibv_dofork_range (e.g. https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/verbs.c#L461) and notify the upper layer?

yishaih:
@zhngaj  Yes it may be caused by calling free() before ibv_dereg_mr(), if free() internally will trigger brk() this might leave some area unmapped.
Returning error in release flow (i.e. dereg_mr) won’t help upper layer to recover and pages in application's heap will be left marked DONTCOPY.
Further malloc() might return the previous memory marked as DONTCOPY which won't be available to a child process post some fork() call.

PR #832, opened 2020-09-17
tests: Skip test if multicast is not supported

Kamalheib:
Avoid test failure by skipping the test if multicast is not supported.

### DIFF
--- a/tests/test_rdmacm.py
+++ b/tests/test_rdmacm.py
@@ -7,6 +7,7 @@ import os
 
 from tests.rdmacm_utils import  CMSyncConnection, CMAsyncConnection
 from pyverbs.pyverbs_error import PyverbsError
+from tests.utils import requires_mcast_support
 from tests.base import RDMATestCase
 import pyverbs.cm_enums as ce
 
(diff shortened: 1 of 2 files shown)


tests/test_rdmacm.py
```diff
@@ -149,11 +149,15 @@ def test_rdmacm_async_traffic(self):
         self.two_nodes_rdmacm_traffic(CMAsyncConnection, self.rdmacm_traffic)
 
     def test_rdmacm_async_multicast_traffic(self):
+        if not self.is_mcast_supported():
```
EdwardSro, line 152: Can you use a decorator instead?
Then you should move the is_mcast_supported under "# Decorators" section in tests/utils.py. You can take a look at "requires_root_on_eth" as an example.
It would look something like this:

def requires_mcast_support():
    def outer(func):
        def inner(instance):
            ctx = d.Context(name=instance.dev_name)
            print(ctx.query_device().max_mcast_grp)
            if ctx.query_device().max_mcast_grp == 0:
                raise unittest.SkipTest('Multicast is not supported on this device')
            return func(instance)
        return inner
    return outer

Kamalheib: ACK, I'll send a v2 with your suggestion.

EdwardSro (approved):

PR #833, opened 2020-09-22
tests: Make sure to set the port_num when creating RAW packet QP

Kamalheib:
When trying to create a RAW packet QP over devices with multiple ports
(like mlx4) make sure to pass the port number to the QPAttr() to avoid
creating RAW packet QP over an IB port.

### DIFF
--- a/tests/test_qp.py
+++ b/tests/test_qp.py
@@ -25,12 +25,12 @@ class QPTest(PyverbsAPITestCase):
     Test various functionalities of the QP class.
     """
 
-    def create_qp(self, creator, qp_init_attr, is_ex, with_attr):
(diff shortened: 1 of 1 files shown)


tests/test_qp.py
```diff
@@ -25,12 +25,12 @@ class QPTest(PyverbsAPITestCase):
     Test various functionalities of the QP class.
     """
 
-    def create_qp(self, creator, qp_init_attr, is_ex, with_attr):
+    def create_qp(self, creator, qp_init_attr, is_ex, with_attr, port_num):
```
EdwardSro, line 28: Is there a reason why you did not assign 1 as the default port_num value?

Kamalheib: I'll fix it in v2.

EdwardSro (approved):

PR #835, opened 2020-09-23
Migrate Fedora based distros to use the standard udev scripts

Honggang-LI:


redhat/rdma.conf
```diff
@@ -1,18 +1,27 @@
...
-ISERT_LOAD=yes
-# Load RDS (Reliable Datagram Service) network protocol
-RDS_LOAD=no
-# Load NFSoRDMA client transport module
-XPRTRDMA_LOAD=yes
-# Load NFSoRDMA server transport module
-SVCRDMA_LOAD=no
-# Load Tech Preview device driver modules
-TECH_PREVIEW_LOAD=no
+# These modules are loaded by the system if any RDMA devices is installed
```
rleon, line 1: Why do you copy this file from kernel-boot/ to redhat/?

Honggang-LI: For RHEL system with RDMA device, we always load ib_isert, ib_srpt and ib_umad modules. But they are unavailable in kernel-boot/modules/rdma.conf .

jgunthorpe: IB umad is already autoloaded properly only for IB and OPA, not needed in other cases.

PR #842, opened 2020-10-01
libqedr: Add additional device ids supported by this lib

mkalderon:
No changes are required for rdma-core either to support the
newly added device ids.


jgunthorpe:
Please fix the checkpatch warnings.

Do you have a good reason to do this? With modern kernels this table is not used.

mkalderon:
> Please fix the checkpatch warnings.
fixed
> 
> Do you have a good reason to do this? With modern kernels this table is not used.
Can you elaborate on this where was the change introduced? 
we still need the change for older kernels/distros.

Thanks

jgunthorpe:
Older distros is fine if you really are updating the driver on old distros

PR #843, opened 2020-10-06
pyverbs: Add mlx5 DC support

EdwardSro:
This series adds support to mlx5 DC QPs traffic usage.
In addition, two new DC traffic tests were added (Send and RDMA Write).

### DIFF
--- a/pyverbs/qp.pyx
+++ b/pyverbs/qp.pyx
@@ -1260,9 +1260,12 @@ cdef class QPEx(QP):
         :return: An initialized QPEx object
         """
         super().__init__(creator, init_attr, qp_attr)
-        self.qp_ex = v.ibv_qp_to_qp_ex(self.qp)
-        if self.qp_ex == NULL:
-            raise PyverbsRDMAErrno('Failed to create extended QP')
+        if init_attr.comp_mask & v.IBV_QP_INIT_ATTR_SEND_OPS_FLAGS:
+            self.qp_ex = v.ibv_qp_to_qp_ex(self.qp)
+            if self.qp_ex == NULL:
+                raise PyverbsRDMAErrno('Failed to create extended QP')
+        else:
+            self.logger.debug('qp_ex is not accessible since IBV_QP_INIT_ATTR_SEND_OPS_FLAGS was not passed.')
 
     @property
     def comp_mask(self):
--- a/tests/base.py
+++ b/tests/base.py
@@ -35,6 +35,7 @@ TIMEOUT = 14
 MLNX_VENDOR_ID = 0x02c9
 CX3_MLNX_PART_ID = 4099
 CX3Pro_MLNX_PART_ID = 4103
+DCT_KEY = 0xbadc0de
 # Dictionary: vendor_id -> array of part_ids of devices that lack RoCEv2 support
 ROCEV2_UNSUPPORTED_DEVS = {MLNX_VENDOR_ID: [CX3Pro_MLNX_PART_ID,
                                             CX3_MLNX_PART_ID]}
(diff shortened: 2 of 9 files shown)


tests/base.py
```diff
@@ -529,3 +539,84 @@ def pre_run(self, rpsns, rqps_num):
         self.rqps_num = rqps_num
         self.rpsns = rpsns
         self.to_rts()
+
+
+class Mlx5DcResources(TrafficResources):
```
galpress, line 544: Wouldn't it be better to keep it in a separate file (e.g. base_mlx5 or base_dc?)

tests/utils.py
```diff
@@ -391,6 +391,9 @@ def post_send_ex(agr_obj, send_object, gid_index, port, send_op=None, qp_idx=0):
         qp.wr_set_ud_addr(ah, agr_obj.rqps_num[qp_idx], agr_obj.UD_QKEY)
     if qp_type == e.IBV_QPT_XRC_SEND:
         qp.wr_set_xrc_srqn(agr_obj.remote_srqn)
+    if qp_type == e.IBV_QPT_DRIVER:
```
galpress, line 394: IBV_QPT_DRIVER isn't necessarily DC.

tests/base.py
EdwardSro: I'll consider doing so (base_mlx5) in v2.
I thought of separating provider-specific tests into: tests/providers/mlx5 tests/providers/efa etc.. then each provider can have its own utils/base if needed. But that may come later, not as part of this series

tests/utils.py
EdwardSro: Right, will fix in v2.
Thanks.

tests/base.py
galpress: Makes sense, we plan to push EFA traffic tests soon too.

pyverbs/qp.pyx
```diff
@@ -1260,9 +1260,10 @@ cdef class QPEx(QP):
         :return: An initialized QPEx object
         """
         super().__init__(creator, init_attr, qp_attr)
-        self.qp_ex = v.ibv_qp_to_qp_ex(self.qp)
-        if self.qp_ex == NULL:
-            raise PyverbsRDMAErrno('Failed to create extended QP')
+        if init_attr.comp_mask & v.IBV_QP_INIT_ATTR_SEND_OPS_FLAGS:
+            self.qp_ex = v.ibv_qp_to_qp_ex(self.qp)
+            if self.qp_ex == NULL:
+                raise PyverbsRDMAErrno('Failed to create extended QP')
```
Kamalheib, line 1266: @EdwardSro Could you please tell what will happen if someone tries to create a QPEx object without setting IBV_QP_INIT_ATTR_SEND_OPS_FLAGS bit in the comp_mask?

tests/base.py
Kamalheib: I think that this provider(s) test(s)/resource(s) need to be in a specific provider file.

Kamalheib (changes requested):

pyverbs/qp.pyx
EdwardSro: I think I should revert that in V2.
You can take a look at the QP object. ibv_create_qp takes a PD while ibv_create_qp_ex takes a context instead, but both return an ibv_qp.
QP class has a creator parameter, which can be either PD or Context instance, upon that we decide if we are creating a qp or a qp_ex. Since both of them return ibv_qp in C, we also have the same QP class in python.

Back to your question, he'll have a qp (which actually is an extended qp) but he won't have access to the new send APIs, if he tries to do so he can fail on segfault since self.qp_ex is not defined...

tests/base.py
EdwardSro: > I think that this provider(s) test(s)/resource(s) need to be in a specific provider file.

I agree. As I said, will do it in a separate series.

pyverbs/qp.pyx
EdwardSro: Eventually, I could not remove this check, since each provider-specific that wants to inherit from QPEX but doesn't always need the new post send API, would fail otherwise. I added a debug print in case the user doesn't pass IBV_QP_INIT_ATTR_SEND_OPS_FLAGS.

tests/utils.py
```diff
@@ -391,6 +392,9 @@ def post_send_ex(agr_obj, send_object, gid_index, port, send_op=None, qp_idx=0):
         qp.wr_set_ud_addr(ah, agr_obj.rqps_num[qp_idx], agr_obj.UD_QKEY)
     if qp_type == e.IBV_QPT_XRC_SEND:
         qp.wr_set_xrc_srqn(agr_obj.remote_srqn)
+    if isinstance(agr_obj, Mlx5DcResources):
```
galpress, line 395: What about old post_send? Or is DC only supported on extended QPs?

EdwardSro: It's only supported in extended QPs

PR #844, opened 2020-10-06
tests: Add a dry run option

EdwardSro:
Add a --dry_run command-line argument that prints a list of the full
test names that are loaded by default and exit without running them.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -4,6 +4,8 @@
 import importlib
 import os
 
+from args_parser import parser
+
 # Load every test as a module in the system so that unittest's loader can find it
 def _load_tests():
     res = []
@@ -20,8 +22,23 @@ __test_modules__ = _load_tests()
 # a single test.
 tests = importlib.import_module(".", __name__)
 
+
+def _show_tests_and_exit(loader, standard_tests, pattern):
+    """
+    Prints the full test names that are loaded with the current modules via
+    loadTestsFromModule protocol, without modifying standard_tests.
+    """
+    for mod in __test_modules__:
+        for test in loader.loadTestsFromModule(mod, pattern):
+            for test_case in test:
+                print(test_case.id())
+    return standard_tests
+
+
 def load_tests(loader, standard_tests, pattern):
     """Implement the loadTestsFromModule protocol"""
+    if parser.args['list_tests']:
+        return _show_tests_and_exit(loader, standard_tests, pattern)
     for mod in __test_modules__:
         standard_tests.addTests(loader.loadTestsFromModule(mod, pattern))
     return standard_tests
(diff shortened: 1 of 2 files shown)


Kamalheib (commented):
@EdwardSro Could you please tell me why you will need this option?

rleon:
> @EdwardSro Could you please tell me why you will need this option?

I can share my use case for this option.
During development, I'm running tests with "-k ..." option, but because rdma-core is updated from time to time, I need to run with "-v" first to make sure that I didn't miss anything. So instead of doing that, I'll be able to get list of tests without running tests.

Edward has different use case in mind.

EdwardSro:
> @EdwardSro Could you please tell me why you will need this option?

I got multiple requests to add it, and I also use it. The base of each different usage is similar to Leon's use case: a need to see/get what tests do we have in an easy way.
Some practical use cases:
- A developer who wants to run a mini regression, but wants to dynamically run/exclude a specific set of tests. He/she can use the dry_run option and instantly get full names of the tests which can be easily parsed/collected into a set that can be used later with -k option to run them.
- I guess at some point there'll be "too many" tests to run for a first-time test user, especially if he wants to only check specific coverage. So it would be easier if he sees the full name and path of each test that can give a general image of the underlying content.

jgunthorpe:
I think I would call this --list-tests not dry run

rleon:
> I think I would call this --list-tests not dry run

I asked for a --dry-run like all git commands, and not --dry_run.

Kamalheib:
> I think I would call this --list-tests not dry run

I agree with @jgunthorpe --list-tests or --get-tests-list make more sense than "dry run".

EdwardSro:
> I think I would call this --list-tests not dry run

At first place it was "--show_tests", but Leon suggested changing it to --dry-run. Actually, as it's implemented, the run_tests literally goes and loads the modules as it's going to run them but stops by printing (not manually printing the file names or tests). So dry-run made sense to me.

@rleon Regarding why I chose --dry_run instead of --dry-run, from python argparse docs:
"For optional argument actions, the value of dest is normally inferred from the option strings" .... "**Any internal - characters will be converted to _ characters** to make sure the string is a valid attribute name."

I don't mind to change it to --dry-run / --list-tests / --show-tests.
Let's agree on what's better... (If the majority here are suggesting --list-tests instead, then maybe it would make more sense to use it)

jgunthorpe:
Right, it does nothing like a simulated run of the tests, it just prints the test names. for - vs _ just be consistent with the other options

EdwardSro:
> Right, it does nothing like a simulated run of the tests, it just prints the test names. for - vs _ just be consistent with the other options

Changed it to --list-tests

Kamalheib (approved):
LGTM

PR #848, opened 2020-10-13
Revert "ibacm: check provider file ends with .so extension"

Honggang-LI:
This reverts commit ad5d934d688911149d795aee1d3b9fa06bf171a9.

The OPA opa-address-resolution provider's name does not ends with
'*.so'. It has version number in the end, for example,
'libdsap.so.1.0.0'.

Because of commit ad5d934d688911149d795aee1d3b9fa06bf171a9,
the provider libdsap.so.1.0.0 was not opened/used for address resolution
for OPA device. It is a regression issue, so revert it.

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -2834,17 +2834,6 @@ static void acm_load_prov_config(void)
 	}
 }
 
-static int acm_string_end_compare(const char *s1, const char *s2)
-{
-	size_t s1_len = strlen(s1);
-	size_t s2_len = strlen(s2);
-
-	if (s1_len < s2_len)
-		return -1;
-
-	return strcmp(s1 + s1_len - s2_len, s2);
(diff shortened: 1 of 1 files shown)


weiny2:
Normally library installations add a softlink <name>.so -> <name>.so.<ver>  Perhaps the fix is in the dsap plugin?

Honggang-LI:
> Normally library installations add a softlink
> 
> [name].so =>> [name].so.[ver]
> 
> Perhaps the fix is in the dsap plugin?

OK. I will update the opa-address-resolution package to create soft link for libdsap.so.XXX .

jgunthorpe:
We don't support out of tree ACM plugins anyhow. It is just a mess I haven't been bothered to fixup yet. You should encourage that stuff to be merged into rdma-core where it can be built properly.

PR #850, opened 2020-10-18
build: Add code coverage option to CMakeLists.txt

galpress:
Allow to compile the library with code coverage enabled through the
-DENABLE_CODE_COVERAGE=1 flag in CMakeLists.txt.

By default, the library is compiled with code coverage disabled.


jgunthorpe:
I'm a little reluctant to add things that can be handled by pre-specfiying CFLAGS during the cmake invokation? Is there something special about coverage that needs this?

galpress:
Not really, seems nicer to have an explicit option to compile the library with code coverage enabled.
Anyway, closing..

PR #851, opened 2020-10-18
Skip query gid tests on unsupported devices

galpress:
Skip test_query_gid_table and test_query_gid_ex when the device link layer is unspecified.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -106,6 +106,8 @@ class DeviceTest(PyverbsAPITestCase):
         """
         devs = self.get_device_list()
         with d.Context(name=devs[0].name.decode()) as ctx:
+            if u.is_unspecified(ctx, port_num=1):
+                raise unittest.SkipTest('ibv_query_gid_ex is not supported on this device')
             try:
                 ctx.query_gid_ex(port_num=1, gid_index=0)
             except PyverbsRDMAError as ex:
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -89,6 +89,8 @@ class DeviceTest(PyverbsAPITestCase):
         with d.Context(name=devs[0].name.decode()) as ctx:
             device_attr = ctx.query_device()
             port_attr = ctx.query_port(1)
+            if port_attr.link_layer == e.IBV_LINK_LAYER_UNSPECIFIED:
+                raise unittest.SkipTest('ibv_query_gid_table is not supported on this device')
             max_entries = device_attr.phys_port_cnt * port_attr.gid_tbl_len
             try:
                 ctx.query_gid_table(max_entries)
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -794,5 +794,15 @@ def is_eth(ctx, port_num):
     return ctx.query_port(port_num).link_layer == e.IBV_LINK_LAYER_ETHERNET
 
 
+def is_unspecified(ctx, port_num):
+    """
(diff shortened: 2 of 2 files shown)


jgunthorpe:
What about iwarp too?

tests/test_device.py
```diff
@@ -89,6 +89,8 @@ def test_query_gid_table(self):
         with d.Context(name=devs[0].name.decode()) as ctx:
             device_attr = ctx.query_device()
             port_attr = ctx.query_port(1)
+            if port_attr.link_layer == e.IBV_LINK_LAYER_UNSPECIFIED:
```
EdwardSro, line 92: Why can you assume that ibv_query_gid_table is not supported on IBV_LINK_LAYER_UNSPECIFIED? AFAIK, this type of link layer is a legacy value used to indicate that the link layer is IB/iWARP

galpress: I assume legacy code would return IBV_GID_TYPE_SYSFS_IB_ROCE_V1, which returns EINVAL in case of IBV_LINK_LAYER_UNSPECIFIED (in query_sysfs_gid_entry()).

galpress:
> What about iwarp too?

I don't know? What about it?
Does it support these verbs?

jgunthorpe:
No gid table on iwarp. We should detect gid table support directly not overload other ill defined things like link_layer

galpress:
That's how libibverbs detects it in query_sysfs_gid_entry(), do you have an alternative suggestion?

jgunthorpe:
That is something different, that is how to interprite the ambigous gid_type ROCE_V1

In latest kernels you should call UVERBS_METHOD_QUERY_GID_ENTRY and there is no GID table if it returns EOPNOTSUPP. Kernel side might need a fix to make that true.

galpress:
This test isn't running on latest kernels only, do you want it fail on older kernels?

Another alternative is to catch EINVAL in addition to EOPNOTSUPP for skipping the test.

galpress:
Any thoughts?

jgunthorpe:
I think suggesting tests should work Ok on old kernels is somewhat problematic, the tests are supposed to find bugs and they get fixed, I don't want to see tests carry code to detect the bug is actually just not fixed yet.

However, this is the same discussion as on the kernel side, so can the test the gid table length for != 0 before running this test instead?

rleon:
Hi,

I'm waiting for the resolution till Thursday on this PR for the release.

Thanks

galpress:
Thanks

@jgunthorpe but EFA has a gid table length == 1, the problem is the gid type.

jgunthorpe:
What is the problem?

galpress:
We don't have a gid type (unspecified link layer) so the function returns EINVAL.

jgunthorpe:
I thought the function returned EINVAL because of the kernel bug? Where is the gid type check that is causing problems?

galpress:
Looks like now the there's an inconsistency between the ioctl and sysfs paths.
The ioctl works fine, but the sysfs still fails due to:
https://github.com/linux-rdma/rdma-core/blob/c63fa797b1c71903e99647be299b3ff0f0643707/libibverbs/cmd_device.c#L359

Should I just remove the else clause?

jgunthorpe:
It needs to set the gid_type to something, but yes that looks like the right thing to do. Set the gid type to TYPE_IB? That is what it would have been before right?

galpress:
Right, dropping this PR and will open a new one with the libibverbs patch.

PR #857, opened 2020-10-27
Fix cmd_fd leak in mlx5_init_context

nzmsv:
The `alloc_context` function is supposed to take ownership of `cmd_fd` and close it if it fails and returns `NULL`.  This was not done in early exit error paths from `mlx5_init_context` (called from `mlx5_alloc_context`).

### DIFF
--- a/providers/mlx5/mlx5.c
+++ b/providers/mlx5/mlx5.c
@@ -1334,24 +1334,13 @@ static void mlx5_uninit_context(struct mlx5_context *context)
 }
 
 static struct mlx5_context *mlx5_init_context(struct ibv_device *ibdev,
-						int cmd_fd,
-						void *private_data)
+						int cmd_fd)
 {
-	struct mlx5dv_context_attr *ctx_attr = private_data;
 	struct mlx5_device *mdev = to_mdev(ibdev);
 	struct mlx5_context *context;
 	int low_lat_uuars;
 	int tot_uuars;
 	int ret;
-
-	if (ctx_attr && ctx_attr->comp_mask) {
-		errno = EINVAL;
-		return NULL;
-	}
-
-	ret = get_uar_info(mdev, &tot_uuars, &low_lat_uuars);
-	if (ret)
-		return NULL;
 
 	context = verbs_init_and_alloc_context(ibdev, cmd_fd, context, ibv_ctx,
 					       RDMA_DRIVER_MLX5);
@@ -1365,6 +1354,12 @@ static struct mlx5_context *mlx5_init_context(struct ibv_device *ibdev,
 		strcpy(context->hostname, "host_unknown");
 
 	mlx5_single_threaded = single_threaded_app();
+
+	ret = get_uar_info(mdev, &tot_uuars, &low_lat_uuars);
+	if (ret) {
+		mlx5_uninit_context(context);
+		return NULL;
+	}
 	context->tot_uuars = tot_uuars;
 	context->low_lat_uuars = low_lat_uuars;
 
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Drivers are supposed to call verbs_init_and_alloc_context() as their first action, so lets re-organize things so that is the case. There is some cruft here too. Can you try this?

```
diff --git a/providers/mlx5/mlx5.c b/providers/mlx5/mlx5.c
index c419750939963e..5f0ef0ba9ad3fa 100644
--- a/providers/mlx5/mlx5.c
+++ b/providers/mlx5/mlx5.c
@@ -1333,25 +1333,14 @@ static void mlx5_uninit_context(struct mlx5_context *context)
 }
 
 static struct mlx5_context *mlx5_init_context(struct ibv_device *ibdev,
-						int cmd_fd,
-						void *private_data)
+					      int cmd_fd)
 {
-	struct mlx5dv_context_attr *ctx_attr = private_data;
 	struct mlx5_device *mdev = to_mdev(ibdev);
 	struct mlx5_context *context;
 	int low_lat_uuars;
 	int tot_uuars;
 	int ret;
 
-	if (ctx_attr && ctx_attr->comp_mask) {
-		errno = EINVAL;
-		return NULL;
-	}
-
-	ret = get_uar_info(mdev, &tot_uuars, &low_lat_uuars);
-	if (ret)
-		return NULL;
-
 	context = verbs_init_and_alloc_context(ibdev, cmd_fd, context, ibv_ctx,
 					       RDMA_DRIVER_MLX5);
 	if (!context)
@@ -1364,6 +1353,12 @@ static struct mlx5_context *mlx5_init_context(struct ibv_device *ibdev,
 		strcpy(context->hostname, "host_unknown");
 
 	mlx5_single_threaded = single_threaded_app();
+
+	ret = get_uar_info(mdev, &tot_uuars, &low_lat_uuars);
+	if (ret) {
+		mlx5_uninit_context(context);
+		return NULL;
+	}
 	context->tot_uuars = tot_uuars;
 	context->low_lat_uuars = low_lat_uuars;
 
@@ -1568,7 +1563,7 @@ static struct verbs_context *mlx5_alloc_context(struct ibv_device *ibdev,
 	bool				always_devx = false;
 	int ret;
 
-	context = mlx5_init_context(ibdev, cmd_fd, NULL);
+	context = mlx5_init_context(ibdev, cmd_fd);
 	if (!context)
 		return NULL;
 
@@ -1626,7 +1621,7 @@ static struct verbs_context *mlx5_import_context(struct ibv_device *ibdev,
 	struct mlx5_context *mctx;
 	int ret;
 
-	mctx = mlx5_init_context(ibdev, cmd_fd, NULL);
+	mctx = mlx5_init_context(ibdev, cmd_fd);
 	if (!mctx)
 		return NULL;
 
```

nzmsv:
Actually, it looks like there may have been a bug introduced in #776: the check for `ctx_attr->comp_mask` was not a no-op prior to that change and it is now (the function `mlx5_init_context` is now always called with `NULL` for `private_data` and so this turned into dead code, but this wasn't the case prior to cb3be404fadc879c941e29c36c4550394e7039a3).  @yishaih: was this intentional?

yishaih:
The suggestion from Jason seems right, just need also to add the below code [1] to mlx5_alloc_context() where ctx_attr may be not NULL, post calling to mlx5_init_context(), this check comes to enforce zero comp_mask as for now any other value is not supported.

Generally saying,
The idea was to call verbs_init_and_alloc_context() post some input validation checks to prevent init/uninit, but this really miss the closing of the cmd_fd upon an error. The issue was first introduced before the series that you pointed on but with commit ID 7b3c1405d90e8c1440aec7e84114fb5eb0fc2327.

@nzmsv 
Do you plan to send a fixed patch accordingly or that you want me to handle  ?

[1]
if (ctx_attr && ctx_attr->comp_mask) {
		errno = EINVAL;
		goto err;
	}

nzmsv:
Thank you @yishaih!  Easy enough to make the change, so done.

yishaih:
Please add Fixes line, rather than that it seems fine.

yishaih:
@nzmsv please see my note regarding adding Fixes line, can you update the PR with ?

PR #862, opened 2020-11-01
tests: Don't use destroyed address handle

galpress:
When submitting a work request, its address handle must remain valid
until the corresponding completion is polled.

Move the AH allocation outside of the send function and make sure it's
not destroyed before we're finished using it.

Signed-off-by: Chen Brasch <cbrasch@amazon.com>

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -350,15 +350,15 @@ def get_global_route(ctx, gid_index=0, port_num=1):
     return gr
 
 
-def xrc_post_send(agr_obj, qp_num, send_object, gid_index, port, send_op=None):
+def xrc_post_send(agr_obj, qp_num, send_object, send_op=None):
     agr_obj.qps = agr_obj.sqp_lst
     if send_op:
-        post_send_ex(agr_obj, send_object, gid_index, port, send_op)
+        post_send_ex(agr_obj, send_object, send_op)
     else:
-        post_send(agr_obj, send_object, gid_index, port)
+        post_send(agr_obj, send_object)
 
 
-def post_send_ex(agr_obj, send_object, gid_index, port, send_op=None, qp_idx=0):
+def post_send_ex(agr_obj, send_object, send_op=None, qp_idx=0, ah=None):
     qp = agr_obj.qps[qp_idx]
     qp_type = qp.qp_type
     qp.wr_start()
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -399,7 +398,7 @@ def post_send_ex(agr_obj, send_object, gid_index, port, send_op=None, qp_idx=0):
     qp.wr_complete()
 
 
-def post_send(agr_obj, send_wr, gid_index, port, qp_idx=0):
+def post_send(agr_obj, send_wr, gid_index, port, qp_idx=0, ah=None):
```
EdwardSro, line 401: remove gid_index and port as they're not in use anymore (callers should be recursively updated)

tests/utils.py
```diff
@@ -413,7 +412,6 @@ def post_send(agr_obj, send_wr, gid_index, port, qp_idx=0):
     """
     qp_type = agr_obj.qp.qp_type
     if qp_type == e.IBV_QPT_UD:
-        ah = get_global_ah(agr_obj, gid_index, port)
```
EdwardSro, line 416: removing this will break current direct post_send calls on UD QPs such as in test_odp_ud_traffic (in test_odp). please add AH in such tests (I think it's the only one)

EdwardSro (changes requested):

tests/utils.py
galpress: Ack

tests/utils.py
galpress: Missed that, will fix

galpress:
We'll also change the DC AH creation in:
https://github.com/linux-rdma/rdma-core/blob/6983e50fec06c5090b491a09db73eb7f5b8a8399/tests/utils.py#L396

As it has the same issue as UD.

rleon:
> We'll also change the DC AH creation in:
> 
> https://github.com/linux-rdma/rdma-core/blob/6983e50fec06c5090b491a09db73eb7f5b8a8399/tests/utils.py#L396
> 
> As it has the same issue as UD.

Gal, 
Where are we standing with this PR?

Thanks

galpress:
Sorry got sidetracked, will update the PR soon.

EdwardSro (approved):

PR #863, opened 2020-11-02
backport some spec file commits from Fedora repo

Honggang-LI:
We have issue to build rpm from upstream rdma-core repo with Fedora 32+ .
This PR backport some commits from Fedora repo to address such issue.

### DIFF
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -66,8 +66,13 @@ Conflicts: infiniband-diags <= 1.6.7
 # Ninja was introduced in FC23
 BuildRequires: ninja-build
 %define CMAKE_FLAGS -GNinja
+%if 0%{?fedora} >= 33
+%define make_jobs ninja-build -C %{_vpath_builddir} -v %{?_smp_mflags}
+%define cmake_install DESTDIR=%{buildroot} ninja-build -C %{_vpath_builddir} install
+%else
 %define make_jobs ninja-build -v %{?_smp_mflags}
 %define cmake_install DESTDIR=%{buildroot} ninja-build install
+%endif
 %else
 # Fallback to make otherwise
 BuildRequires: make
@@ -346,8 +351,9 @@ rm -f %{buildroot}%{_sysconfdir}/rdma/modules/rdma.conf
 install -D -m0644 redhat/rdma.conf %{buildroot}%{_sysconfdir}/rdma/modules/rdma.conf
 
 # ibacm
-bin/ib_acme -D . -O
-install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/
+(if [ -d %{__cmake_builddir} ]; then cd %{__cmake_builddir}; fi
+ ./bin/ib_acme -D . -O &&
+ install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/)
 
 # Delete the package's init.d scripts
 rm -rf %{buildroot}/%{_initrddir}/
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -27,7 +27,7 @@ BuildRequires: /usr/bin/rst2man
 BuildRequires: valgrind-devel
 BuildRequires: systemd
 BuildRequires: systemd-devel
-%if 0%{?fedora} >= 32
+%if 0%{?fedora} >= 32 || 0%{?rhel} >= 8
 %define with_pyverbs %{?_with_pyverbs: 1} %{?!_with_pyverbs: %{?!_without_pyverbs: 1} %{?_without_pyverbs: 0}}
 %else
 %define with_pyverbs %{?_with_pyverbs: 1} %{?!_with_pyverbs: 0}
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -274,6 +274,13 @@ easy, object-oriented access to IB verbs.
 %setup
 
 %build
+# This package uses top level ASM constructs which are incompatible with LTO.
+# Top level ASMs are often used to implement symbol versioning.  gcc-10
+# introduces a new mechanism for symbol versioning which works with LTO.
+# Converting packages to use that mechanism instead of toplevel ASMs is
+# recommended.
+# Disable LTO
+%define _lto_cflags %{nil}
 
 # New RPM defines _rundir, usually as /run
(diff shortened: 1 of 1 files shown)


redhat/rdma-core.spec
```diff
@@ -53,7 +53,7 @@ BuildRequires: python-docutils
 BuildRequires: perl-generators
 %endif
 
-Requires: dracut, kmod, systemd, pciutils
```
jgunthorpe, line 56: If you are going to get rid of the requires on what provides /sbin/udevadm then the %post needs to check if /sbin/udevadm exists before trying to run it, right?

redhat/rdma-core.spec
```diff
@@ -274,6 +274,13 @@ easy, object-oriented access to IB verbs.
 %setup
 
 %build
+# This package uses top level ASM constructs which are incompatible with LTO.
```
jgunthorpe, line 282: In the commit message what is 'cherry picked from commit abad9da95aadc756bb7cc348da6e13509f499fdb'? Proper commit message plase

redhat/rdma-core.spec
```diff
@@ -354,10 +354,11 @@ rm -rf %{buildroot}/%{_initrddir}/
 rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
 
 %post -n rdma-core
-# we ship udev rules, so trigger an update.
```
jgunthorpe, line 350: Oh there it is, please squash this into the earlier patch

redhat/rdma-core.spec
```diff
@@ -354,10 +354,11 @@ rm -rf %{buildroot}/%{_initrddir}/
 rm -rf %{buildroot}/%{_sbindir}/srp_daemon.sh
```
jgunthorpe, line 363: Here is another cherry picked from ???

redhat/rdma-core.spec
```diff
@@ -346,7 +351,11 @@ rm -f %{buildroot}%{_sysconfdir}/rdma/modules/rdma.conf
 install -D -m0644 redhat/rdma.conf %{buildroot}%{_sysconfdir}/rdma/modules/rdma.conf
 
 # ibacm
+%if 0%{?fedora} >= 33
+%{buildroot}%{_bindir}/ib_acme -D . -O
+%else
 bin/ib_acme -D . -O
```
jgunthorpe, line 357: Why can't we use buildroot everywhere? This looks kind of weird too, IIRC ib_acme writes a file, so under the 'out of source builds' concept it should be something more like

```
(cd %{buildroot} &&
 %{_bindir}/ib_acme -D . -O &&
 install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/)
```

jgunthorpe (changes requested):
Good stuff

redhat/rdma-core.spec
Honggang-LI: Yes, you are right. I will squash next commit into this to fix this.

redhat/rdma-core.spec
Honggang-LI: Got it. I will update the commit message.

redhat/rdma-core.spec
Honggang-LI: OK.

redhat/rdma-core.spec
Honggang-LI: I will replace it with something like this:

```
(cd %{buildroot} &&
 ./%{_bindir}/ib_acme -D ./%{_sysconfdir}/rdma/ -O &&
 chmod 0644 ./%{_sysconfdir}/rdma/ibacm_opts.cfg)
```

Honggang-LI: Can't change directory, the rpmbuild tool 'mock' is unhappy about this.

```
+ cd /builddir/build/BUILDROOT/rdma-core-33.0-1.el8.x86_64
+ .//usr/bin/ib_acme -D .//etc/rdma/ -O
.//usr/bin/ib_acme: error while loading shared libraries: libibverbs.so.1: cannot open shared object file: No such file or directory
```

jgunthorpe: Oh, it can't be _bindir, it has to be ./bin/ib_acme  - it needs the compiled but not yet installed version to get the rpath to resolve the yet to be installed library.

Honggang-LI: It seems it does not works for epel-8.
```
+ cd /builddir/build/BUILDROOT/rdma-core-33.0-1.el8.x86_64//usr
+ ./bin/ib_acme -D /builddir/build/BUILDROOT/rdma-core-33.0-1.el8.x86_64//etc/rdma/ -O
./bin/ib_acme: error while loading shared libraries: libibverbs.so.1: cannot open shared object file: No such file or directory
```

jgunthorpe: That is the wrong path, //usr is the installation direction, you need to cd to the directory that holds the .o files

Honggang-LI:
```
(if [ -d %{__cmake_builddir} ]; then cd %{__cmake_builddir}; fi
 ./bin/ib_acme -D . -O &&
 install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/)
```

jgunthorpe: That seems like it could work, do you mean it does?

Honggang-LI: Solution A:
```
 # ibacm
 %if 0%{?fedora} >= 33
 %{buildroot}%{_bindir}/ib_acme -D . -O
 %else
 bin/ib_acme -D . -O
 install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/
```

Solution B:
```
# ibacm
(if [ -d %{__cmake_builddir} ]; then cd %{__cmake_builddir}; fi
 ./bin/ib_acme -D . -O &&
 install -D -m0644 ibacm_opts.cfg %{buildroot}%{_sysconfdir}/rdma/)
```

Both A and B works for me. Which one do you want?

jgunthorpe:
Looks fine @Honggang-LI, can you please resolve the minor CI warnings about the commit messages? Particularly the signed-off-by notes

Honggang-LI:
> Looks fine @Honggang-LI, can you please resolve the minor CI warnings about the commit messages? Particularly the signed-off-by notes

I added signed-off-by tags. Thanks

PR #864, opened 2020-11-03
Add masked atomic opcodes

Li-Weihang:
Add Masked Cmp & Swap and Masked Fetch & Add opcodes for IBV_WC* and IBV_WR*, than use them in libhns.

### DIFF
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -502,6 +502,8 @@ enum ibv_wc_opcode {
 	IBV_WC_BIND_MW,
 	IBV_WC_LOCAL_INV,
 	IBV_WC_TSO,
+	IBV_WC_MASKED_COMP_SWAP,
+	IBV_WC_MASKED_FETCH_ADD,
 /*
  * Set value of IBV_WC_RECV so consumers can test if a completion is a
  * receive by testing (opcode & IBV_WC_RECV).
@@ -1076,6 +1078,8 @@ enum ibv_wr_opcode {
 	IBV_WR_BIND_MW,
 	IBV_WR_SEND_WITH_INV,
 	IBV_WR_TSO,
+	IBV_WR_MASKED_ATOMIC_CMP_AND_SWP,
+	IBV_WR_MASKED_ATOMIC_FETCH_AND_ADD,
 	IBV_WR_DRIVER1,
 };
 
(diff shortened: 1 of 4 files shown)


libibverbs/verbs.h
```diff
@@ -1076,6 +1078,8 @@ enum ibv_wr_opcode {
 	IBV_WR_BIND_MW,
 	IBV_WR_SEND_WITH_INV,
 	IBV_WR_TSO,
+	IBV_WR_MASKED_ATOMIC_CMP_AND_SWP,
+	IBV_WR_MASKED_ATOMIC_FETCH_AND_ADD,
```
jgunthorpe, line 1082: Changing the value of IBV_WR_DRIVER1 is ABI breaking

libibverbs/verbs.h
```diff
@@ -502,6 +502,8 @@ enum ibv_wc_opcode {
 	IBV_WC_BIND_MW,
 	IBV_WC_LOCAL_INV,
 	IBV_WC_TSO,
+	IBV_WC_MASKED_COMP_SWAP,
```
jgunthorpe, line 505: Any new ABI needs man page updates, and the man page updates need to also be posted to the mailing lists.

Is this IBTA specified?

jgunthorpe (changes requested):

libibverbs/verbs.h
Li-Weihang: Thanks for the reminder, I will update the man page.

They're not IBTA specified, but they are discribed in RFC 7306. These opcodes allow users to pass in a mask which specifies the field boundary and then perform a bitwise logical AND operation on Add/Swap/Compare Data with the mask. That makes the ATOMIC operations more flexible.

They have been added in kernel(commit 5e80ba8ff0bd "IB/core: Add support for masked atomic operations") and supported by mlx4 and mlx5.

libibverbs/verbs.h
Li-Weihang: Thank you, I'll do more work on this.

jgunthorpe:
Also I'm asking everyone adding new API to verbs to also provide pyverbs tests for the new API, or a very good reason why not. I think this should not be so hard here so can you please try?

Li-Weihang:
> Also I'm asking everyone adding new API to verbs to also provide pyverbs tests for the new API, or a very good reason why not. I think this should not be so hard here so can you please try?

Sorry, we didn't know we had to do this before. Thanks for your reminder, we will try it.

Li-Weihang:
May take some time to prepare the man pages and pyverbs, so close this MR at first.

PR #867, opened 2020-11-04
efa: Flush write combining writes before writing to the LLQ

galpress:
An mmio_wc_start() is needed before writing to the LLQ memory in order
to prevent the WQEs copy from being reordered relative to other mmio
writes, such as tx doorbells.

Placing the mmio_wc_start() before each WQE memcpy is wasteful, as the
doorbell write doesn't happen on each iteration. Instead, it is placed
right after the doorbell write, as a preparation for the next batch of
WQEs.

This prevents provider to issue more than max_tx_batch LLQ writes
between two doorbells. This is especially relevant when the user calls
the _post API with more WQEs than max_tx_batch.

Fixes: 7aad28d ("efa: Respect maximum TX doorbell batch")
Signed-off-by: Shadi Ammouri <sammouri@amazon.com>

### DIFF
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -1395,7 +1395,6 @@ static inline void efa_rq_ring_doorbell(struct efa_rq *rq, uint16_t pc)
 
 static inline void efa_sq_ring_doorbell(struct efa_sq *sq, uint16_t pc)
 {
-	mmio_flush_writes();
 	mmio_write32(sq->wq.db, pc);
 }
 
@@ -1516,15 +1515,19 @@ int efa_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 
 		if (curbatch == qp->sq.max_batch_wr) {
 			curbatch = 0;
+			mmio_flush_writes();
 			efa_sq_ring_doorbell(&qp->sq, qp->sq.wq.pc);
+			mmio_wc_start();
 		}
 
 		wr = wr->next;
 	}
 
 ring_db:
-	if (curbatch)
+	if (curbatch) {
+		mmio_flush_writes();
 		efa_sq_ring_doorbell(&qp->sq, qp->sq.wq.pc);
+	}
 
 	/*
 	 * Not using mmio_wc_spinunlock as the doorbell write should be done
@@ -1780,6 +1783,7 @@ static int efa_send_wr_complete(struct ibv_qp_ex *ibvqpx)
 	pc = qp->sq.wq.pc - qp->sq.num_wqe_pending;
 	sq_desc_idx = pc & qp->sq.wq.desc_mask;
 
+	/* mmio_wc_start() comes from efa_send_wr_start() */
 	while (qp->sq.num_wqe_pending) {
 		num_wqe_to_copy = min3(qp->sq.num_wqe_pending,
 				       qp->sq.wq.wqe_cnt - sq_desc_idx,
@@ -1798,13 +1802,17 @@ static int efa_send_wr_complete(struct ibv_qp_ex *ibvqpx)
 			      qp->sq.wq.desc_mask;
 
 		if (curbatch == max_txbatch) {
+			mmio_flush_writes();
 			efa_sq_ring_doorbell(&qp->sq, pc);
 			curbatch = 0;
+			mmio_wc_start();
 		}
 	}
 
-	if (curbatch)
+	if (curbatch) {
+		mmio_flush_writes();
 		efa_sq_ring_doorbell(&qp->sq, qp->sq.wq.pc);
+	}
 out:
 	/*
 	 * Not using mmio_wc_spinunlock as the doorbell write should be done


providers/efa/verbs.c
```diff
@@ -1397,6 +1397,11 @@ static inline void efa_sq_ring_doorbell(struct efa_sq *sq, uint16_t pc)
 {
 	mmio_flush_writes();
 	mmio_write32(sq->wq.db, pc);
+	/* Make sure that no further writes to LLQ BAR are performed before
+	 * the DB write is done. Otherwise we might get more than max tx batch
+	 * writes to LLQ between two doorbells.
+	 */
+	mmio_flush_writes();
```
jgunthorpe, line 1404: flush_writes should not be used, use the proper barrier for whatever this is doing.

jgunthorpe (changes requested):

providers/efa/verbs.c
galpress: Why is that not the proper barrier?
We need to make sure the doorbell write is done before we continue with other writes to the LLQ BAR.

jgunthorpe: I think you want to use mmio_wc_start() and mmio_flush_writes() at the end

galpress: Sorry for the long delay, will fix, thanks.

galpress:
I thought about calling mmio_wc_start() right before the mmio_memcpy_x64(), but that adds unnecessary flushes so I kept it in the doorbell function to act as a preparation for subsequent WC writes.

jgunthorpe:
That's complete backwards, it is an API, if you are using WC memory, you have to follow othe API. If you think the flushes are too strong then fix them. EFA does not look special at all here to justify doing anything different

galpress:
The flushes aren't too strong, they're just not needed that frequently.
The WQEs themselves are already ordered by mmio_memcpy_x64(), there's no reason to flush every time a new WQE is copied. The flush is needed to order multiple WQEs with regards to a single doorbell.

jgunthorpe:
Is wq.db in WC memory?

galpress:
No, the LLQ/WQEs are in WC memory, and doorbells are NC.

jgunthorpe:
Can you reogranize this so the WC barrier wrappers the actual memory accesses that are WC then please? This seems too obtuse

galpress:
That's what I tried to do, but the flow goes like:
```
While session not done:
    1. Prepare WQE
    2. Copy to LLQ (WC)
    3. If max_tx_batch were copied:
            3.1. mmio_flush_writes()
            3.2. write DB (NC)
```

Theoretically the WC barrier could be placed between 1-2, so it's clear it wraps the WC memory access, but then it would be needed conditionally (just in case max_tx_batch is reached and a doorbell is required, as that's the only case both memory types are involved).
But we already have a check if max_tx_batch is reached, and adding the barrier at the end of step 3 is the same as doing it at the beginning of the iteration + get the batch check for free.
Also, placing the barrier between 1-2 would mean a double barrier on the first iteration, unless we also remove the mmio_wc_spinlock() at the beginning.

jgunthorpe:
I would expect something like this
```
diff --git a/providers/efa/verbs.c b/providers/efa/verbs.c
index 54d208a82fd582..4e1d3435efd33b 100644
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -1395,7 +1395,6 @@ static inline void efa_rq_ring_doorbell(struct efa_rq *rq, uint16_t pc)
 
 static inline void efa_sq_ring_doorbell(struct efa_sq *sq, uint16_t pc)
 {
-	mmio_flush_writes();
 	mmio_write32(sq->wq.db, pc);
 }
 
@@ -1516,15 +1515,19 @@ int efa_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 
 		if (curbatch == qp->sq.max_batch_wr) {
 			curbatch = 0;
+			mmio_flush_writes();
 			efa_sq_ring_doorbell(&qp->sq, qp->sq.wq.pc);
+			mmio_wc_start();
 		}
 
 		wr = wr->next;
 	}
 
 ring_db:
-	if (curbatch)
+	if (curbatch) {
+		mmio_flush_writes();
 		efa_sq_ring_doorbell(&qp->sq, qp->sq.wq.pc);
+	}
 
 	/*
 	 * Not using mmio_wc_spinunlock as the doorbell write should be done
@@ -1780,6 +1783,7 @@ static int efa_send_wr_complete(struct ibv_qp_ex *ibvqpx)
 	pc = qp->sq.wq.pc - qp->sq.num_wqe_pending;
 	sq_desc_idx = pc & qp->sq.wq.desc_mask;
 
+	/* wc_start comes from efa_send_wr_start() */
 	while (qp->sq.num_wqe_pending) {
 		num_wqe_to_copy = min3(qp->sq.num_wqe_pending,
 				       qp->sq.wq.wqe_cnt - sq_desc_idx,
@@ -1798,13 +1802,17 @@ static int efa_send_wr_complete(struct ibv_qp_ex *ibvqpx)
 			      qp->sq.wq.desc_mask;
 
 		if (curbatch == max_txbatch) {
+			mmio_flush_writes();
 			efa_sq_ring_doorbell(&qp->sq, pc);
 			curbatch = 0;
+			mmio_wc_start();
 		}
 	}
 
-	if (curbatch)
+	if (curbatch) {
+		mmio_flush_writes();
 		efa_sq_ring_doorbell(&qp->sq, qp->sq.wq.pc);
+	}
 out:
 	/*
 	 * Not using mmio_wc_spinunlock as the doorbell write should be done
```

galpress:
Yea, that's perfectly fine with me, thanks.
Will give it a spin in our CI to be on the safe side and will update the PR.

galpress:
@jgunthorpe please let me know if this PR can be merged.

galpress:
Thanks.
@nmorey @jgunthorpe what is the stable backporting process? This fix is very important for us, what's the best way we can get a release that includes it?

nmorey:
@galpress It's being pushed to stable-v3[12] branches.
I don't have a planned dates for releases.

The "process" is either to wait for the next stable release (most of the time happens roughly at the same time as new releases) or email me about it.

galpress:
OK, thanks Nicolas.

nmorey:
Do you need a new release for v31 and/or v32 then ?
Or can you work from the stable branches themselves ?

galpress:
Yes, I checked but we can't work with stable branches unless it's tagged as a release.
Would it be possible to create one?

Thanks

nmorey:
Do you need v31, v32 or both ?

galpress:
v31 would be great.

nmorey:
I have a patch pending but I'll make one for v31 tomorrow

galpress:
Thanks a lot!

PR #873, opened 2020-11-10
libqedr: Fix return code when qelr_create_qp_ex() isn't supported

Kamalheib:
As libqedr support both RoCE and iWarp, make sure to return EOPNOTSUPP
from qelr_create_qp_ex() when running in iWarp mode.

Fixes: cae4a99ae679 ("libqedr: add support for XRC-SRQ's.")

### DIFF
--- a/providers/qedr/qelr_main.c
+++ b/providers/qedr/qelr_main.c
@@ -113,6 +113,14 @@ static const struct verbs_context_ops qelr_ctx_ops = {
 	.free_context = qelr_free_context,
 };
 
+static const struct verbs_context_ops qelr_ctx_roce_ops = {
+	.close_xrcd = qelr_close_xrcd,
+	.create_qp_ex = qelr_create_qp_ex,
+	.create_srq_ex = qelr_create_srq_ex,
+	.get_srq_num = qelr_get_srq_num,
+	.open_xrcd = qelr_open_xrcd,
+};
+
 static void qelr_uninit_device(struct verbs_device *verbs_device)
 {
 	struct qelr_device *dev = get_qelr_dev(&verbs_device->device);
@@ -171,7 +179,6 @@ static struct verbs_context *qelr_alloc_context(struct ibv_device *ibdev,
 						int cmd_fd,
 						void *private_data)
 {
-	struct verbs_context *v_ctx;
 	struct qelr_devctx *ctx;
 	struct qelr_alloc_context cmd = {};
 	struct qelr_alloc_context_resp resp;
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Please post this to the mailing list to get Michal and Ariel's attention, I don't think they are on github?

rleon:
@mkalderon ???

mkalderon:
@Kamalheib sorry for the late response was on vacation. 
How did you get to the create_qp_ex call ? ( creating xrcd / srq_ex would fail before) 
anyhow, i think the fix should be more complete in qelr_alloc_context not initialize the XRC cb
functions for iwarp at all.

Kamalheib:
> @Kamalheib sorry for the late response was on vacation.
> How did you get to the create_qp_ex call ? ( creating xrcd / srq_ex would fail before)

I found this while running pyverbs tests.

> anyhow, i think the fix should be more complete in qelr_alloc_context not initialize the XRC cb
> functions for iwarp at all.

ACK, I'll send v2.

providers/qedr/qelr_main.c
```diff
@@ -252,11 +252,13 @@ static struct verbs_context *qelr_alloc_context(struct ibv_device *ibdev,
...
-	v_ctx->close_xrcd = qelr_close_xrcd;
-	v_ctx->create_srq_ex = qelr_create_srq_ex;
-	v_ctx->get_srq_num = qelr_get_srq_num;
+	if (IS_ROCE(ibdev)) {
+		v_ctx->create_qp_ex = qelr_create_qp_ex;
+		v_ctx->open_xrcd = qelr_open_xrcd;
+		v_ctx->close_xrcd = qelr_close_xrcd;
+		v_ctx->create_srq_ex = qelr_create_srq_ex;
+		v_ctx->get_srq_num = qelr_get_srq_num;
+	}
```
jgunthorpe, line 261: This open coded assignment to the ops is wrong, everything is supposed to go through verbs_context_ops and verbs_set_ops()

you need to add a new ops group that is qib_roce_ops and then conditionally call verbs_set_ops on it.

Kamalheib: Done

providers/qedr/qelr_main.c
```diff
@@ -113,6 +113,14 @@ static const struct verbs_context_ops qelr_ctx_ops = {
 	.free_context = qelr_free_context,
 };
 
+static const struct verbs_context_ops qelr_ctx_roce_ops = {
+	.create_qp_ex = qelr_create_qp_ex,
+	.open_xrcd = qelr_open_xrcd,
+	.close_xrcd = qelr_close_xrcd,
+	.create_srq_ex = qelr_create_srq_ex,
+	.get_srq_num = qelr_get_srq_num,
+};
```
jgunthorpe, line 122: Keep lists sorted

Kamalheib: > Keep lists sorted

Well..., In almost all providers this isn't kept sorted :-), someone forgot to do that ;-).

Anyway, fixed.

PR #875, opened 2020-11-12
libhns: Add new PCI device matching for HIP09

Li-Weihang:
The 200G device has a new device ID 0xA228, add it to the PCI table.

Signed-off-by: Weihang Li <liweihang@huawei.com>


rleon:
@jgunthorpe I'm looking at your commit 0c3cecfe04d5 ("verbs: Use CHARDEV info from netlink to bind drivers") and you wrote that hns needs those PCI bumps, is it still relevant?

Thanks

PR #877, opened 2020-11-12
Network type

rpearsonhpe:
Cleanup rxe provider to support kernel commit referenced below.

Reference: e0d696d201dd ("RDMA/rxe: Move the definitions for rxe_av.network_type to uAPI")


rpearsonhpe:
oops duplicate to Jason's

PR #878, opened 2020-11-12
rxe extended verbs support

rpearsonhpe:
A series of 5 commits to implement the extended verbs APIs
    - ibv_query_device_ex
    - ibv_create_cq_ex
    - ibv_create_qp_ex

### DIFF
--- a/kernel-headers/rdma/rdma_user_rxe.h
+++ b/kernel-headers/rdma/rdma_user_rxe.h
@@ -181,4 +181,25 @@ struct rxe_modify_srq_cmd {
 	__aligned_u64 mmap_info_addr;
 };
 
+/* This data structure is stored at the base of work and
+ * completion queues shared between user space and kernel space.
+ * It contains the producer and consumer indices. Is also
+ * contains a copy of the queue size parameters for user space
+ * to use but the kernel must use the parameters in the
+ * rxe_queue struct. For performance reasons arrange to have
+ * producer and consumer indices in separate cache lines
+ * the kernel should always mask the indices to avoid accessing
+ * memory outside of the data area
+ */
+struct rxe_queue_buf {
+	__u32			log2_elem_size;
+	__u32			index_mask;
+	__u32			pad_1[30];
+	__u32			producer_index;
+	__u32			pad_2[31];
+	__u32			consumer_index;
+	__u32			pad_3[31];
+	__u8			data[];
+};
+
 #endif /* RDMA_USER_RXE_H */
--- a/providers/rxe/rxe.c
+++ b/providers/rxe/rxe.c
@@ -162,12 +162,135 @@ static int rxe_dereg_mr(struct verbs_mr *vmr)
 	return 0;
 }
 
+static int cq_start_poll(struct ibv_cq_ex *current,
+			 struct ibv_poll_cq_attr *attr)
+{
+	struct rxe_cq *cq = container_of(current, struct rxe_cq, vcq.cq_ex);
+
+	pthread_spin_lock(&cq->lock);
+
+	cq->cur_index = load_consumer_index(cq->queue);
+
+	if (check_cq_queue_empty(cq)) {
+		pthread_spin_unlock(&cq->lock);
+		errno = ENOENT;
+		return errno;
+	}
+
+	cq->wc = addr_from_index(cq->queue, cq->cur_index);
+	cq->vcq.cq_ex.status = cq->wc->status;
+	cq->vcq.cq_ex.wr_id = cq->wc->wr_id;
+
+	return 0;
+}
+
+static int cq_next_poll(struct ibv_cq_ex *current)
+{
+	struct rxe_cq *cq = container_of(current, struct rxe_cq, vcq.cq_ex);
+
+	advance_cq_cur_index(cq);
+
+	if (check_cq_queue_empty(cq)) {
+		store_consumer_index(cq->queue, cq->cur_index);
+		pthread_spin_unlock(&cq->lock);
+		errno = ENOENT;
+		return errno;
+	}
+
+	cq->wc = addr_from_index(cq->queue, cq->cur_index);
+	cq->vcq.cq_ex.status = cq->wc->status;
+	cq->vcq.cq_ex.wr_id = cq->wc->wr_id;
+
+	return 0;
+}
+
+static void cq_end_poll(struct ibv_cq_ex *current)
+{
+	struct rxe_cq *cq = container_of(current, struct rxe_cq, vcq.cq_ex);
+
+	advance_cq_cur_index(cq);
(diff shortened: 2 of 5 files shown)


rpearsonhpe:
per a conversation with Jason removed the capability check on create_cq_ex and create_qp_ex.

providers/rxe/rxe.c
```diff
@@ -694,88 +694,655 @@ static int rxe_post_srq_recv(struct ibv_srq *ibvsrq,
...
+		case IBV_QPT_UD:
+			if (attr->send_ops_flags & ~RXE_SUP_UD_QP_SEND_OPS_FLAGS)
+				goto err;
+			break;
+		case IBV_QPT_RAW_PACKET:
+			goto err;
+		case IBV_QPT_XRC_SEND:
+			if (attr->send_ops_flags & ~RXE_SUP_XRC_QP_SEND_OPS_FLAGS)
+				goto err;
+			break;
```
jgunthorpe, line 1218: I thought there was no xrc support at all? why doesn't this return error instead of break? Why implement the dummy wr_set_xrc_srqn() ? Just leave the pointer as null, not allowed to call it unless a XRC QP type is requested which should fail.

kernel-headers/rdma/rdma_user_rxe.h
```diff
@@ -158,8 +158,54 @@ struct rxe_recv_wqe {
...
+struct rxe_alloc_context_cmd {
+	__aligned_u64		provider_cap;
+};
+
+struct rxe_alloc_context_resp {
+	__aligned_u64		driver_cap;
+};
+
+struct rxe_create_cq_cmd {
+	__u8 is_extended;	/* called from ibv_create_cq_ex */
```
jgunthorpe, line 201: The call path can't matter, we have three call paths for most commands, the driver data must be uniform across all of them.

Here is where it needs to be feature based. In this case I suggest you pass into the kernel the sizeof(rxe_uverbs_wc) so that the kernel knows how to format the completion queue entries. 0 means sizeof (ib_uverbs_wc).

kernel-headers/rdma/rdma_user_rxe.h
```diff
@@ -158,8 +158,54 @@ struct rxe_recv_wqe {
...
+	__u8			reserved;
+
+	/* any extras go here */
+	__aligned_u64		timestamp;
+	__aligned_u64		realtime;
+};
+
+enum rxe_capabilities {
+	RXE_CAP_NONE		= 0,
+	RXE_CAP_CMD_EX		= 1ULL << 0,
```
jgunthorpe, line 189: This should be called RXE_CAP_CREATE_CQ_UDATA like EFA/mlx5. The kernel sets it when it knows how to check that the unsupported parts of the udata are zero. Current kernel accepts any udata on input, this needs to be fixed.

kernel-headers/rdma/rdma_user_rxe.h
jgunthorpe: The standard pattern is to add a 

 u64 comp_mask;
 u64 flags;

Then use the flags as you need. The comp_mask is 0 (and checked to be 0 in the kernel) for future use.

jgunthorpe: It might be wise to take the timestamp part out of here, then this series requires no more work. Add timestamp to user/kernel as another series, it will be easir to follow the compatability logic

jgunthorpe (changes requested):

rpearsonhpe:
per request from Jason dropped any extensions requiring changes to the kernel driver or headers. Can add these later. This is now down to two commits.

providers/rxe/rxe_queue.h
```diff
@@ -86,28 +88,58 @@ static inline void advance_producer(struct rxe_queue *q)
...
+	return atomic_load_explicit(&q->producer_index,
+				    memory_order_relaxed);
+}
+
+/* Must hold producer_index lock */
+static inline void store_producer_index(struct rxe_queue *q, uint32_t index)
+{
+	/* flush writes to work queue before moving index */
+	atomic_thread_fence(memory_order_release);
+	atomic_store(&q->producer_index, index);
```
jgunthorpe, line 113: This combination is just atomic_store_explicit(... memory_order_release);

providers/rxe/rxe_queue.h
```diff
@@ -86,28 +88,58 @@ static inline void advance_producer(struct rxe_queue *q)
...
+}
+
+/* Must hold consumer_index lock */
+static inline uint32_t load_consumer_index(struct rxe_queue *q)
+{
+	return atomic_load_explicit(&q->consumer_index,
+				    memory_order_relaxed);
```
jgunthorpe, line 120: I suspect this should be memory_order_acquire

providers/rxe/rxe_queue.h
```diff
@@ -86,28 +88,58 @@ static inline void advance_producer(struct rxe_queue *q)
...
+}
+
+/* Must hold consumer_index lock */
+static inline void store_consumer_index(struct rxe_queue *q, uint32_t index)
+{
+	/* flush writes to work queue before moving index */
+	atomic_thread_fence(memory_order_release);
+	atomic_store(&q->consumer_index, index);
```
jgunthorpe, line 128: atomic_store_explicit

providers/rxe/rxe.c
```diff
@@ -162,20 +162,144 @@ static int rxe_dereg_mr(struct verbs_mr *vmr)
...
 }
 
+static int cq_start_poll(struct ibv_cq_ex *current,
+			 struct ibv_poll_cq_attr *attr)
+{
+	struct rxe_cq *cq = container_of(current, struct rxe_cq, vcq.cq_ex);
+
+	pthread_spin_lock(&cq->lock);
+
+	atomic_thread_fence(memory_order_acquire);
```
jgunthorpe, line 172: atomic_thread_fence/atomic_XX pairs should always be merged into an atomic_XX_explicit operation for performance

providers/rxe/rxe.c
```diff
@@ -185,15 +309,118 @@ static struct ibv_cq *rxe_create_cq(struct ibv_context *context, int cqe,
...
+		cq->vcq.cq_ex.read_dlid_path_bits
+			= cq_read_dlid_path_bits;
+
+	return &cq->vcq.cq_ex;
+
+err_unmap:
+	if (cq->mmap_info.size)
+		munmap(cq->queue, cq->mmap_info.size);
+err_destroy:
+	(void)ibv_cmd_destroy_cq(&cq->vcq.cq);
```
jgunthorpe, line 419: No (void) please

jgunthorpe (changes requested):

PR #879, opened 2020-11-13
Update AZP to use FC33

jgunthorpe:
And a few other minor fixes

PR #880, opened 2020-11-13
Fix LTO support

jgunthorpe:
Now that gcc has defined attribute symver we can make LTO support work on compilers that support the new attribute.

@nmorey @Honggang-LI

### DIFF
--- a/buildlib/FindLDSymVer.cmake
+++ b/buildlib/FindLDSymVer.cmake
@@ -27,15 +27,25 @@ else()
 endif()
 
 # And matching source, this also checks that .symver asm works
-check_c_source_compiles("
-void ibv_get_device_list_1(void);
-void ibv_get_device_list_1(void){}
-asm(\".symver ibv_get_device_list_1, ibv_get_device_list@IBVERBS_1.1\");
-void ibv_get_device_list_0(void);
-void ibv_get_device_list_0(void){}
-asm(\".symver ibv_get_device_list_0, ibv_get_device_list@@IBVERBS_1.0\");
-
-int main(int argc,const char *argv[]){return 0;}" _LDSYMVER_SUCCESS)
+if (HAVE_FUNC_ATTRIBUTE_SYMVER)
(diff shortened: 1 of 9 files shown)


buildlib/FindLDSymVer.cmake
```diff
@@ -27,15 +27,25 @@ else()
...
-check_c_source_compiles("
-void ibv_get_device_list_1(void);
-void ibv_get_device_list_1(void){}
-asm(\".symver ibv_get_device_list_1, ibv_get_device_list@IBVERBS_1.1\");
-void ibv_get_device_list_0(void);
-void ibv_get_device_list_0(void){}
-asm(\".symver ibv_get_device_list_0, ibv_get_device_list@@IBVERBS_1.0\");
-
-int main(int argc,const char *argv[]){return 0;}" _LDSYMVER_SUCCESS)
+if (HAVE_FUNC_ATTRIBUTE_SYMVER)
```
rleon, line 30: Why do we need if-else here?
Isn't HAVE_FUNC_ATTRIBUTE_SYMVER already checked that symver work?

Honggang-LI:
"linux-rdma.rdma-core (Build Test Build RPMs for fedora31)" is build task for fedora33. As I wrote in https://github.com/linux-rdma/rdma-core/pull/879 , please update buildlib/azure-pipelines.yml .

I confirmed the "symver" attribute usage with epel7, epel8, fedora31, fedoar32, fedora33 and fedora34/rawhide. It works as expected.

The PR looks good for me. Thanks

buildlib/FindLDSymVer.cmake
jgunthorpe: ATTRIBUTE_SYMVER is checking if the C compiler attribute is supported, this is testing if the linker option works. We need to use the attribute version or this test fails on LTO linkers.

nmorey:
@jgunthorpe Seems to work well on Tumbleweed

PR #881, opened 2020-11-15
Fix issues reported by Coverity

Honggang-LI:


providers/qedr/qelr_verbs.c
```diff
@@ -2675,10 +2675,10 @@ qelr_create_xrc_srq(struct ibv_context *context,
 
 err1:
 	qelr_destroy_srq_buffers(ibv_srq);
-	free(srq);
 err0:
 	DP_ERR(cxt->dbg_fp,
 	       "create srq: failed to create %p. rc=%d\n", srq, rc);
+	free(srq);
```
jgunthorpe, line 2681: Just delete the srq argument to the printer, it is useless, nothing else has printed this pointer yet.

jgunthorpe (changes requested):

providers/qedr/qelr_verbs.c
Honggang-LI: OK

PR #882, opened 2020-11-15
librdmacm: close XRC domain before close the protection domain

Honggang-LI:
Before open an eXtended Reliable Connection (XRC) domain, a protection
domain should be established. So, close the XRC domain before close
protection domain.

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -298,10 +298,10 @@ static void remove_cma_dev(struct cma_device *cma_dev)
 		return;
 	}
 
-	if (cma_dev->pd)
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Is this fixing some bug? The XRCD and PD are unrelated objects, so I don't think this ordering should matter?

Honggang-LI:
> Is this fixing some bug? The XRCD and PD are unrelated objects, so I don't think this ordering should matter?

No bug. I think the objects should be delete in reverse order of they created. Please feel free to close it.

rleon:
I'm big fan of symmetrical flows.

PR #883, opened 2020-11-15
verbs: Introduce ibv_is_fork_initialized verb

galpress:
Add ibv_is_fork_initialized verb which allows the app to query whether
fork support (ibv_fork_init) is enabled/disabled.

The verb returns an enum with the disabled/enabled value, and could
later be extended to return a "not required" value if in the future a call
to ibv_fork_init() will not be required.

Motivation and more information:
https://lists.openfabrics.org/pipermail/ofiwg/2020-October/001969.html
https://lists.openfabrics.org/pipermail/ofiwg/2020-October/001978.html

### DIFF
--- a/debian/libibverbs1.symbols
+++ b/debian/libibverbs1.symbols
@@ -10,6 +10,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.10@IBVERBS_1.10 31
  IBVERBS_1.11@IBVERBS_1.11 32
  IBVERBS_1.12@IBVERBS_1.12 34
+ IBVERBS_1.13@IBVERBS_1.13 35
  (symver)IBVERBS_PRIVATE_34 34
  _ibv_query_gid_ex@IBVERBS_1.11 32
  _ibv_query_gid_table@IBVERBS_1.11 32
@@ -75,6 +76,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  ibv_import_mr@IBVERBS_1.10 31
  ibv_import_pd@IBVERBS_1.10 31
  ibv_init_ah_from_wc@IBVERBS_1.1 1.1.6
+ ibv_is_fork_initialized@IBVERBS_1.13 35
  ibv_modify_qp@IBVERBS_1.0 1.1.6
  ibv_modify_qp@IBVERBS_1.1 1.1.6
  ibv_modify_srq@IBVERBS_1.0 1.1.6
--- a/libibverbs/CMakeLists.txt
+++ b/libibverbs/CMakeLists.txt
@@ -21,7 +21,7 @@ configure_file("libibverbs.map.in"
 
 rdma_library(ibverbs "${CMAKE_CURRENT_BINARY_DIR}/libibverbs.map"
   # See Documentation/versioning.md
-  1 1.12.${PACKAGE_VERSION}
+  1 1.13.${PACKAGE_VERSION}
   all_providers.c
   cmd.c
   cmd_ah.c
--- a/libibverbs/libibverbs.map.in
+++ b/libibverbs/libibverbs.map.in
@@ -153,6 +153,11 @@ IBVERBS_1.12 {
 		ibv_reg_dmabuf_mr;
 } IBVERBS_1.11;
 
+IBVERBS_1.13 {
+	global:
+		ibv_is_fork_initialized;
+} IBVERBS_1.12;
+
 /* If any symbols in this stanza change ABI then the entire staza gets a new symbol
    version. See the top level CMakeLists.txt for this setting. */
 
--- a/libibverbs/man/CMakeLists.txt
+++ b/libibverbs/man/CMakeLists.txt
@@ -41,6 +41,7 @@ rdma_man_pages(
   ibv_import_mr.3.md
   ibv_import_pd.3.md
   ibv_inc_rkey.3.md
+  ibv_is_fork_initialized.3.md
   ibv_modify_qp.3
   ibv_modify_qp_rate_limit.3
   ibv_modify_srq.3
--- a/libibverbs/man/ibv_is_fork_initialized.3.md
+++ b/libibverbs/man/ibv_is_fork_initialized.3.md
@@ -1,0 +1,48 @@
+---
+date: 2020-10-09
+footer: libibverbs
+header: "Libibverbs Programmer's Manual"
+layout: page
+license: 'Licensed under the OpenIB.org BSD license (FreeBSD Variant) - See COPYING.md'
+section: 3
+title: IBV_IS_FORK_INITIALIZED
+---
+
+# NAME
+
+ibv_is_fork_initialized - check if fork support (ibv_fork_init) is enabled
+
+# SYNOPSYS
+
+```c
+#include <infiniband/verbs.h>
+
+enum ibv_fork_status {
+	IBV_FORK_DISABLED,
+	IBV_FORK_ENABLED,
+	IBV_FORK_UNNEEDED,
+};
+
+enum ibv_fork_status ibv_is_fork_initialized(void);
+```
+
+# DESCRIPTION
+
+**ibv_is_fork_initialized()** checks whether libibverbs **fork()** support was
+enabled through the **ibv_fork_init()** verb.
+
+# RETURN VALUE
+
+**ibv_is_fork_initialized()** returns IBV_FORK_DISABLED if fork support is
+disabled, or IBV_FORK_ENABLED if enabled. IBV_FORK_UNNEEDED return value
+indicates that the kernel copies DMA pages on fork, hence a call to
+**ibv_fork_init()** is unneeded.
+
+# SEE ALSO
+
+**fork**(2),
+**ibv_fork_init**(3)
+
+# AUTHOR
+
+Gal Pressman <galpress@amazon.com>
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -174,6 +174,11 @@ int ibv_fork_init(void)
 	return 0;
 }
 
+enum ibv_fork_status ibv_is_fork_initialized(void)
+{
+	return mm_root ? IBV_FORK_ENABLED : IBV_FORK_DISABLED;
+}
+
 static struct ibv_mem_node *__mm_prev(struct ibv_mem_node *node)
 {
 	if (node->left) {
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -139,6 +139,12 @@ enum ibv_device_cap_flags {
 	IBV_DEVICE_MANAGED_FLOW_STEERING = 1 << 29
 };
 
+enum ibv_fork_status {
+	IBV_FORK_DISABLED,
+	IBV_FORK_ENABLED,
+	IBV_FORK_UNNEEDED,
+};
+
 /*
  * Can't extended above ibv_device_cap_flags enum as in some systems/compilers
  * enum range is limited to 4 bytes.
@@ -3354,6 +3360,12 @@ int ibv_detach_mcast(struct ibv_qp *qp, const union ibv_gid *gid, uint16_t lid);
  * effect of an application calling fork() is undefined.
  */
 int ibv_fork_init(void);
+
+/**
+ * ibv_is_fork_initialized - Check if fork support
+ * (ibv_fork_init) was enabled.
+ */
+enum ibv_fork_status ibv_is_fork_initialized(void);
 
 /**
  * ibv_node_type_str - Return string describing node_type enum value


jgunthorpe:
I don't want to revise this later, can you confirm that the new kernel has no need of this? If yes lets add the 3rd option from day one.

galpress:
Do you want to add an unused value? Even if it works there's currently no way for userspace to tell it's not needed.

How can this be verified to work? Do you have in mind a simple test?

jgunthorpe:
I'd expect to add the kernel part too. I'm not sure of a test, someone should know since it seems to be a big deal..

galpress:
That's a bit out of scope for me right now.
I can change the patch so it returns a boolean,  and future extensions can come in the form of new functions (e.g. ibv_is_fork_needed()).

jgunthorpe:
If you can't test the fork support then why do you care about this at all? If you can confirm the new kernel support works I'll do all the plumbing

galpress:
I'm not very familiar with the kernel changes in that area other than what's posted in the cited thread.
Let me do some research and see how we can test new kernels.

jgunthorpe:
Just disable ibv_fork_init() and run the forking workloads under a v5.10 kernel. On 5.9 you should see random memory corruption triggered failures, on 5.10 it should always work

wzamazon:
@galpress @jgunthorpe 

I ran the osu_latency_mp test (which is a work load with fork()) on linux kernel 5.10 (supplied by @galpress) without RDMAV_FORK_SAFE, but I am still seeing memory corruption failure. Meanwhile, if I enable RDMAV_FORK_SAFE the corruption failure was gone.

The following is the the detail:

osu_latency_mp is part of the OSU microbenchmark, which is designed to test MPI with fork(). This test will fork multiple processes and send/receive data between them.

I ran this this using Open MPI 4.10 /libfaric efa provider master branch/rdma-core 31.2 on linux kernel module 5.10 (@galpress supplied the kernel). What I am seeing is that if I ran without RDMAV_FORK_SAFE set, I got data corruption. Specifically, I got the following error:
```
mpirun -x FI_EFA_ENABLE_SHM_TRANSFER=0 -np 2 -machinefile nodelist ./omb-5.7/libexec/osu-micro-benchmarks/mpi/pt2pt/osu_latency_mp -t 4:4
# OSU MPI Multi-process Latency Test v5.7
# Number of forked processes in sender: 4
# Number of forked processes in receiver: 4
# Size Latency (us)
osu_latency_mp: prov/efa/src/rxr/rxr_pkt_cmd.c:952: rxr_pkt_handle_recv_completion: Assertion `0 && "invalid control pkt type"' failed.
[ip-172-31-9-71:36220] *** Process received signal ***
```
Note currently the EFA provider will abort if it detects fork() has been called, I had to modify the EFA provider to disable that for the test to proceed.

Which means receiver received a packet with packet type ID being 0, which should not happen because packet type ID starts from 0.

If I ran the same test with RDMAV_FORK_SAFE to 1 and also disable MR cache (libfabric's MR cache is not compatible with rdma-core's fork support), I do not get such error.

Please let me know if there are further test you want to do on this issue.

galpress:
@jgunthorpe thoughts?

jgunthorpe:
It is likely there is something wrong with this test, eg the child is using MRs or umems from the parent without re-creating them. Why is the failure deep inside EFA and not around a user MR? Doesn't that suggest that EFA didn't recreate its queues or something?

galpress:
> It is likely there is something wrong with this test, eg the child is using MRs or umems from the parent without re-creating them.

Wouldn't the test just segfault if the child uses its parent MR (on older kernels with DONTFORK enabled)?

> Why is the failure deep inside EFA and not around a user MR?

The error is around user MR, the packet type id is part of the memory region.

> Doesn't that suggest that EFA didn't recreate its queues or something?

You mean EFA libfabric provider? Why would it recreate the queues?

jgunthorpe:
If you create queues in the parent then fork don't expect them to work in the child.

Similarily if you created a MR in the parent cached it, and then forked, don't expect the MR to work in the child.

At least you should test by commenting out the don't fork and keep everything else the same. If that still breaks I would suggest to disable THP, 4k pages was well tested but there might yet be a bug in THP (though even that was lightly tested).

galpress:
@wzamazon can you please share some details about the test? Does the child process use QPs/MRs created by the parent process?

I'm pretty sure THP is not enabled, but we're setting aside huge pages for libfabric's use.
Maybe we can try to run the test without huge pages at all?

galpress:
@jgunthorpe in case the test actually does something wrong, wouldn't it fail on older kernels as well?

jgunthorpe:
This is why I said you shouldn't change the test so much, just get rid of the don't fork and only the don't fork. Enable/disable MR cache may cause bugs

wzamazon:
@galpress @jgunthorpe 
I have done the test requested.

First, I want to make a correction. Earlier I said:

```
osu_latency_mp is part of the OSU microbenchmark, which is designed to test MPI with fork(). This test will fork multiple processes and send/receive data between them.
```

The second part is wrong. What the test does is:

```
1. We initially have two processes, 0 and 1
2. Each processes fork a few child processes (in my test, each process forked 4 child process)
3. Child processes does nothing but sleep.
4. Parent process 0 send data 1 through MPI_Send/Recv, which use libfbaric which uses rdma-core
```
Now to answer @galpress question: "Does the child process use QPs/MRs created by the parent process?"

The answer is no. As mentioned before child process does nothing.

I also did what @jgunthorpe suggested: keep RDMAV_FORK_SAFE environment variable on,  but do not call ```ibv_dontfork_range```. I still get the data validation error.

jgunthorpe:
It is hard to see how you can have the patched kernel and get failures with such a simple test. I would say turn off Transparent Huge Pages and try again, assuming there is some bug in that area.

galpress:
@wzamazon are you using transparent huge pages?
To be on the safe side about the kernel, I'll recompile a clean 5.10 kernel and share the AMI with you again.

jgunthorpe:
You need kernel commit 70e806e4e645, and looking at it, I don't see THP support in there, so that would be my guess

wzamazon:
@galpress @jgunthorpe 

I was able to narrow down the problem to be with explicit huge pages. 

A little background: the libfabric EFA provider uses bounce buffers to send/receive data. By default, these bounce buffers are backed by explicit huge pages (allocated by ```mmap(MAP_HUGETLB)```). This is a way to make the bounce buffer to be backed by regular pages too.

What I found was on kernel 5.10-rc6 and 5.11-rc4:

If the bounce buffer is backed huge page, the osu_latency_mp test would fail, unless we enable rdma-core's fork support.

If the bounce buffer is backed by regular page, the osu_latency_mp test would pass. (rdma-core's fork support is not needed).

So my conclusion is that there is a bug in 5.10/5.11 that caused explicit huge page member (allocated by ```mmap(MAP_HUGETLB)``` still need fork support.

I am writing a lower level test (that only use rdma-core) to demonstrate the issue better. Meanwhile, please let me know if there is any more test I can do.

wzamazon:
@galpress @jgunthorpe 

I was able to write a to demonstrate the issue, see attached 
[sre.tar.gz](https://github.com/linux-rdma/rdma-core/files/5899975/sre.tar.gz). This test only uses rdma-core.

Compile this package, there will be two executables: ```send``` and ```recv```. 

```recv``` will receive 1 message, and expect it to be "Hello, world!"

```send``` will send a message of "Hello, world!".

```send``` has two options: 

```-f``` which will call ```fork``` to create a child process (child process will not do anything). ```fork``` will be called after memory registration, and before the data assignment.

```-h``` which will use huge pages as send buffer. huge page was allocated by ```mmap(NULL, SEND_BUFF_SIZE, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB, -1, 0)```.

My test show:

1. on kernel 4.14 ```send``` with ```-f``` will cause data corruption. Run ```send``` with ```RDMA_FORK_SAFE=1``` can fix the data corruption.

2. on kernel 5.10-rc6 and 5.11-rc4 ```send``` with ```-f``` will not cause data corruption, indicating that regular pages does not need fork support any more (which is expected).

3. on kernel 5.10-rc6 and 5.11-rc4 ```send``` with ```-f``` and ```-h``` still has data corruption, indicating that explicitly allocated huge page still need fork support (which is unexpected).

**Comment by wzamazon** [comment]
@galpress @jgunthorpe 

I did some debugging, and I think I know the reason why hugetlb pages still need fork support in kernel 5.10.

```fork``` will call ```copy_page_range``` for each VMA.

Commit 70e806e4e645 introduced a function copy_present_page, which will copy pinned pages. This function was called by ```copy_page_range```.

The problem was for hugetlb pages (allocated by mmap(MAP_HUGETLB)),  ```copy_page_range``` will call function ```copy_hugetlb_page_range```, thus bypassing commit 70e806e4e645. This happened at https://github.com/torvalds/linux/blob/1048ba83fb1c00cd24172e23e8263972f6b5d9ac/mm/memory.c#L1150.

Does this make sense to you?

**Comment by jgunthorpe** [comment]
Yes, I belive you are completely correct, I pointed at this when reviewing Peter's patches but it didn't get tackled

**Comment by galpress** [comment]
Thanks Wei and Jason!

Jason, do you want to report this issue to Peter or should we do it?
How would you like to proceed with this PR? Looks like the "unneeded" value can't be used correctly on any kernel yet. Should I keep the patch as is (enabled/disabled) only?

**Comment by jgunthorpe** [comment]
I think it would be good if you bring it up, shows wider interest

**Comment by galpress** [comment]
@jgunthorpe Peter's fix solved the issue.
How would you like to proceed with this PR?

**Comment by jgunthorpe** [comment]
If you belive the kernel is fixed to not require MADV_DONTFORK at all with Peter's patch then lets ensure the 3rd option is ready to go.

**Comment by galpress** [comment]
@wzamazon did you try to run a full test suite or just the specific test case? Let's make sure there's no regression.
@jgunthorpe we're still missing a kernel capability report to the userspace.

Can we push this PR with three values, with the UNNEEDED value temporarily unused so we can unblock libfabric changes? We can later add the capability bit to indicate when it's not needed.

**Comment by jgunthorpe** [comment]
The kernel cap will be something like returning a new attribute from the get_context verb and plumbing it throughout. When the kernel is working I can draft a patch

As long as the patch includes the unneeded concept it can go forward (note it need rebasing anyhow)

**Comment by galpress** [comment]
> The kernel cap will be something like returning a new attribute from the get_context verb and plumbing it throughout. When the kernel is working I can draft a patch

Hmm, there might be a small issue since the ibv_fork_init()/ibv_is_fork_initialized() calls are "global" library calls, unrelated to device/context. Having the cap bit in get_context assumes at least one context has been created.

> As long as the patch includes the unneeded concept it can go forward (note it need rebasing anyhow)

Thanks, I will update the patch.

**Comment by wzamazon** [comment]
> did you try to run a full test suite or just the specific test case? Let's make sure there's no regression.

I only ran the specific fork related test. Regarding the regression test, you mean would Peter's patch cause any regression?

**Comment by galpress** [comment]
> > did you try to run a full test suite or just the specific test case? Let's make sure there's no regression.
> 
> 
> I only ran the specific fork related test. Regarding the regression test, you mean would Peter's patch cause any regression?

Yes, we should verify that we can run all of our tests with Peter's patches and fork support disabled.

@jgunthorpe I updated the patch.

**Comment by wzamazon** [comment]
@jgunthorpe Are you waiting for my test result to merge the patch?

**Comment by jgunthorpe** [comment]
Yes please

**Comment by shijin-aws** [comment]
@jgunthorpe @galpress . We verified the fork test with huge page enabled passed on the kernel with Peter's patches. We also run regression tests suggested by @wzamazon on the kernel with Peter's patches and fork support disabled. And we did not observe any regression with Peter's patches by comparing with the results from kernel without Peter's patches.

**Comment by galpress** [comment]
Thanks, I rebased the patch to v35.

**Comment by jgunthorpe** [comment]
So this is all confirmed and done now? I saw Peter's patches went into v5.12-rc already?

**Comment by galpress** [comment]
Yes, although this doesn't handle the fork not needed case.
Hopefully this patch can be merged first in order to unblock libfabric, meanwhile we can figure out how the not needed value should be used.

**Comment by galpress** [comment]
Thanks
@jgunthorpe how do you think the kernel capability report should be done?
You suggested to return a new attribute from the get_context flow, but that implies that a context have been created prior to calling the ibv_is_fork_initialized verb, is that your preferred solution?

**Comment by galpress** [comment]
I guess that a simple kernel patch (that still has the context allocation assumption) could look something like:
```
diff --git a/drivers/infiniband/core/uverbs_std_types_device.c b/drivers/infiniband/core/uverbs_std_types_device.c
index 9ec6971056fa..60464a0b4e97 100644
--- a/drivers/infiniband/core/uverbs_std_types_device.c
+++ b/drivers/infiniband/core/uverbs_std_types_device.c
@@ -212,7 +212,8 @@ static int UVERBS_HANDLER(UVERBS_METHOD_GET_CONTEXT)(
 	struct uverbs_attr_bundle *attrs)
 {
 	u32 num_comp = attrs->ufile->device->num_comp_vectors;
-	u64 core_support = IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS;
+	u64 core_support = IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS |
+			   IB_UVERBS_CORE_SUPPORT_COPY_ON_FORK;
 	int ret;
 
 	ret = uverbs_copy_to(attrs, UVERBS_ATTR_GET_CONTEXT_NUM_COMP_VECTORS,
@@ -240,7 +241,8 @@ static int UVERBS_HANDLER(UVERBS_METHOD_GET_CONTEXT)(
 static int UVERBS_HANDLER(UVERBS_METHOD_QUERY_CONTEXT)(
 	struct uverbs_attr_bundle *attrs)
 {
-	u64 core_support = IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS;
+	u64 core_support = IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS |
+			   IB_UVERBS_CORE_SUPPORT_COPY_ON_FORK;
 	struct ib_ucontext *ucontext;
 	struct ib_device *ib_dev;
 	u32 num_comp;
diff --git a/include/uapi/rdma/ib_user_ioctl_verbs.h b/include/uapi/rdma/ib_user_ioctl_verbs.h
index 22483799cd07..36eeefd2fd88 100644
--- a/include/uapi/rdma/ib_user_ioctl_verbs.h
+++ b/include/uapi/rdma/ib_user_ioctl_verbs.h
@@ -46,6 +46,7 @@
 
 enum ib_uverbs_core_support {
 	IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS = 1 << 0,
+	IB_UVERBS_CORE_SUPPORT_COPY_ON_FORK = 1 << 1,
 };
 
 enum ib_uverbs_access_flags {
```

But I'm pretty sure that the new verb would be called way before a context is actually allocated?

jgunthorpe:
I had imainged doing it in the context like this, but if you need to so early you could do it with a sysfs or netlink I suppose?

galpress:
OK, any reasons to choose one over the other?

jgunthorpe:
What do apps do? I think you might be right that they trigger the don't fork very early, so probably a netlink is the way to do it?

galpress:
I meant sysfs over netlink?

jgunthorpe:
We've been trying to use netlink because sysfs doesn't work very well in containers

galpress:
I see, thanks.

Do you think it should be added as a new netlink (e.g. RDMA_NLDEV_CMD_COPY_ON_FORK), or as an attribute to an existing one (RDMA_NLDEV_CMD_SYS_GET I guess?)?

jgunthorpe:
@rleon usualy helps figure out how to fit things into netlink, possibly it is not a good fit and we should use sysfs anyhow

galpress:
@rleon your help would be appreciated 😄 .

rleon:
It is hard to say.
Till now, we used netlink to get/set device/port properties of struct ib_device, while COPY_ON_FORK doesn't really fit.
It doesn't fit to _SYS_ too, because that section is used for global ib_core parameters - like module parameters for whole subsystem.

Can this COPY_ON_FORK applicable to one device and not applicable to another device at the same time?

galpress:
Hmm, not really, it's either supported by the kernel or it isn't, that's why I think sys might be a good fit.
The actual implementation of the function is likely always going to return 1/true (supported).

BTW, I verified that this verb would be used from containers as well so netlink is preferable.

rleon:
So yes, global == RDMA_NLDEV_CMD_SYS_GET

However, when will you return false? What will be the in-kernel test?

galpress:
From my understanding there is no in-kernel test, the function would always return true as the kernel "supports" it.

It's kinda like how IB_UVERBS_CORE_SUPPORT_OPTIONAL_MR_ACCESS is always turned on in UVERBS_HANDLER(UVERBS_METHOD_GET_CONTEXT), it's there so the userspace can detect older vs newer kernels.

jgunthorpe:
Kernel will never return false.

The test is if the kernel has the patch or not.

rleon:
It is problematic, such kernel uapi patch has all chances to be backported. It is very shaky idea to rely on it without actual in-kernel test.

jgunthorpe:
It is all we can do. No other option. Patch will have to have a good commit message

rleon:
How will it work for distros who take whole subsystem?

jgunthorpe:
QA will have to catch it if they make a mistake, there is no alternative I can see, and we need this functionality to be exposed.

Gal, how do you feel about adding some kind of test for this to run_tests.py? It seems quite doable.

galpress:
Which test do you have in mind? API test that calls ibv_is_fork_initialized?

PR #884, opened 2020-11-15
libibverbs: Fix reporting link_layer for efa devices

Kamalheib:
Avoid reporting link_layer "InfiniBand" for efa devices.

Before:

$ ibv_devinfo
hca_id:	efa_0
	transport:			unspecified (4)
	fw_ver:				0.0.0.0
	node_guid:			0000:0000:0000:0000
	sys_image_guid:			0000:0000:0000:0000
	vendor_id:			0x1d0f
	vendor_part_id:			61344
	hw_ver:				0xEFA0
	phys_port_cnt:			1
		port:	1
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		4096 (5)
			sm_lid:			0
			port_lid:		0
			port_lmc:		0x01
			link_layer:		InfiniBand

After:

$ ibv_devinfo
hca_id:	efa_0
	transport:			unspecified (4)
	fw_ver:				0.0.0.0
	node_guid:			0000:0000:0000:0000
	sys_image_guid:			0000:0000:0000:0000
	vendor_id:			0x1d0f
	vendor_part_id:			61344
	hw_ver:				0xEFA0
	phys_port_cnt:			1
		port:	1
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		4096 (5)
			sm_lid:			0
			port_lid:		0
			port_lmc:		0x01
			link_layer:		Unknown

### DIFF
--- a/libibverbs/examples/devinfo.c
+++ b/libibverbs/examples/devinfo.c
@@ -231,6 +231,7 @@ static const char *link_layer_str(uint8_t link_layer)
 {
 	switch (link_layer) {
 	case IBV_LINK_LAYER_UNSPECIFIED:
+		return "Unspecified";
 	case IBV_LINK_LAYER_INFINIBAND:
 		return "InfiniBand";
 	case IBV_LINK_LAYER_ETHERNET:


galpress:
Thanks Kamal!
I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.

Kamalheib:
> Thanks Kamal!
> I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.

I was trying to match between sysfs and ibv_devinfo reported info...
Anyway, I agree I'll send patches that will fix both kernel and rdma-core.

$ cat /sys/class/infiniband/efa_0/ports/1/link_layer 
Unknown

rleon:
> > Thanks Kamal!
> > I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.
> 
> I was trying to match between sysfs and ibv_devinfo reported info...
> Anyway, I agree I'll send patches that will fix both kernel and rdma-core.
> 
> $ cat /sys/class/infiniband/efa_0/ports/1/link_layer
> Unknown

What exactly will you fix in the kernel?
If you plan to change "Uknown" to be "Unspecified" you can't do it without breaking half of the world.
From kernel perspective, we have only three types of links: ethernet, ib and unknown.

Kamalheib:
> > > Thanks Kamal!
> > > I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.
> > 
> > 
> > I was trying to match between sysfs and ibv_devinfo reported info...
> > Anyway, I agree I'll send patches that will fix both kernel and rdma-core.
> > $ cat /sys/class/infiniband/efa_0/ports/1/link_layer
> > Unknown
> 
> What exactly will you fix in the kernel?
> If you plan to change "Uknown" to be "Unspecified" you can't do it without breaking half of the world.
> From kernel perspective, we have only three types of links: ethernet, ib and unknown.

I plan to do the following change:

```
diff --git a/drivers/infiniband/core/sysfs.c b/drivers/infiniband/core/sysfs.c
index b8abb30f80df..6c03156fbe72 100644
--- a/drivers/infiniband/core/sysfs.c
+++ b/drivers/infiniband/core/sysfs.c
@@ -334,6 +334,9 @@ static ssize_t link_layer_show(struct ib_port *p, struct port_attribute *unused,
        const char *output;
 
        switch (rdma_port_get_link_layer(p->ibdev, p->port_num)) {
+       case IB_LINK_LAYER_UNSPECIFIED:
+               output = "Unspecified";
+               break;
        case IB_LINK_LAYER_INFINIBAND:
                output = "InfiniBand";
                break;
```

rleon:
> > > > Thanks Kamal!
> > > > I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.
> > > 
> > > 
> > > I was trying to match between sysfs and ibv_devinfo reported info...
> > > Anyway, I agree I'll send patches that will fix both kernel and rdma-core.
> > > $ cat /sys/class/infiniband/efa_0/ports/1/link_layer
> > > Unknown
> > 
> > 
> > What exactly will you fix in the kernel?
> > If you plan to change "Uknown" to be "Unspecified" you can't do it without breaking half of the world.
> > From kernel perspective, we have only three types of links: ethernet, ib and unknown.
> 
> I plan to do the following change:
> 
> ```
> diff --git a/drivers/infiniband/core/sysfs.c b/drivers/infiniband/core/sysfs.c
> index b8abb30f80df..6c03156fbe72 100644
> --- a/drivers/infiniband/core/sysfs.c
> +++ b/drivers/infiniband/core/sysfs.c
> @@ -334,6 +334,9 @@ static ssize_t link_layer_show(struct ib_port *p, struct port_attribute *unused,
>         const char *output;
>  
>         switch (rdma_port_get_link_layer(p->ibdev, p->port_num)) {
> +       case IB_LINK_LAYER_UNSPECIFIED:
> +               output = "Unspecified";
> +               break;
>         case IB_LINK_LAYER_INFINIBAND:
>                 output = "InfiniBand";
>                 break;
> ```

As I wrote above, you can't do it.

Thanks

Kamalheib:
> > > > > Thanks Kamal!
> > > > > I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.
> > > > 
> > > > 
> > > > I was trying to match between sysfs and ibv_devinfo reported info...
> > > > Anyway, I agree I'll send patches that will fix both kernel and rdma-core.
> > > > $ cat /sys/class/infiniband/efa_0/ports/1/link_layer
> > > > Unknown
> > > 
> > > 
> > > What exactly will you fix in the kernel?
> > > If you plan to change "Uknown" to be "Unspecified" you can't do it without breaking half of the world.
> > > From kernel perspective, we have only three types of links: ethernet, ib and unknown.
> > 
> > 
> > I plan to do the following change:
> > ```
> > diff --git a/drivers/infiniband/core/sysfs.c b/drivers/infiniband/core/sysfs.c
> > index b8abb30f80df..6c03156fbe72 100644
> > --- a/drivers/infiniband/core/sysfs.c
> > +++ b/drivers/infiniband/core/sysfs.c
> > @@ -334,6 +334,9 @@ static ssize_t link_layer_show(struct ib_port *p, struct port_attribute *unused,
> >         const char *output;
> >  
> >         switch (rdma_port_get_link_layer(p->ibdev, p->port_num)) {
> > +       case IB_LINK_LAYER_UNSPECIFIED:
> > +               output = "Unspecified";
> > +               break;
> >         case IB_LINK_LAYER_INFINIBAND:
> >                 output = "InfiniBand";
> >                 break;
> > ```
> 
> As I wrote above, you can't do it.
> 
> Thanks

As you can see in the change I don't plan to change "Uknown" to be "Unspecified"...
Could you please tell me why reporting "Unspecified" could be problematic?! and what this will harm/break?

jgunthorpe:
What is the point of changing the kernel? Nothing is supposed to be reading that anyhow these days..

rleon:
> > > > > > Thanks Kamal!
> > > > > > I think it's better to add an explicit case for "Unspecified" instead of falling back to the "Unknown" case.
> > > > > 
> > > > > 
> > > > > I was trying to match between sysfs and ibv_devinfo reported info...
> > > > > Anyway, I agree I'll send patches that will fix both kernel and rdma-core.
> > > > > $ cat /sys/class/infiniband/efa_0/ports/1/link_layer
> > > > > Unknown
> > > > 
> > > > 
> > > > What exactly will you fix in the kernel?
> > > > If you plan to change "Uknown" to be "Unspecified" you can't do it without breaking half of the world.
> > > > From kernel perspective, we have only three types of links: ethernet, ib and unknown.
> > > 
> > > 
> > > I plan to do the following change:
> > > ```
> > > diff --git a/drivers/infiniband/core/sysfs.c b/drivers/infiniband/core/sysfs.c
> > > index b8abb30f80df..6c03156fbe72 100644
> > > --- a/drivers/infiniband/core/sysfs.c
> > > +++ b/drivers/infiniband/core/sysfs.c
> > > @@ -334,6 +334,9 @@ static ssize_t link_layer_show(struct ib_port *p, struct port_attribute *unused,
> > >         const char *output;
> > >  
> > >         switch (rdma_port_get_link_layer(p->ibdev, p->port_num)) {
> > > +       case IB_LINK_LAYER_UNSPECIFIED:
> > > +               output = "Unspecified";
> > > +               break;
> > >         case IB_LINK_LAYER_INFINIBAND:
> > >                 output = "InfiniBand";
> > >                 break;
> > > ```
> > 
> > 
> > As I wrote above, you can't do it.
> > Thanks
> 
> As you can see in the change I don't plan to change "Uknown" to be "Unspecified"...

However you changed.

> Could you please tell me why reporting "Unspecified" could be problematic?! and what this will harm/break?

IB puppet scripts returns link_layer as is, you have potential to break users of those scripts.

Thanks

Kamalheib:
> IB puppet scripts returns link_layer as is, you have potential to break users of those scripts.
> 
As of today, the only provider that returns link_layer = IB_LINK_LAYER_UNSPECIFIED is the EFA driver..., And you think that it should stay reported wrongly as "Unkown" because you think there are some scripts that could break?! (Even that I don't know how they could break if the value is reported only for the EFA driver) At least could you please give the name of one script that you know about?

Thanks

rleon:
> > IB puppet scripts returns link_layer as is, you have potential to break users of those scripts.
> 
> As of today, the only provider that returns link_layer = IB_LINK_LAYER_UNSPECIFIED is the EFA driver..., And you think that it should stay reported wrongly as "Unkown" because you think there are some scripts that could break?! (Even that I don't know how they could break if the value is reported only for the EFA driver) At least could you please give the name of one script that you know about?
> 
> Thanks

It is not me who are changing sysfs entry introduced more than ten years ago and it is upto you to support your claim that nothing is going to be broken.

And yes, from the kernel perspective, this field is "unknown".

Thanks

Kamalheib:
> > > IB puppet scripts returns link_layer as is, you have potential to break users of those scripts.
> > 
> > 
> > As of today, the only provider that returns link_layer = IB_LINK_LAYER_UNSPECIFIED is the EFA driver..., And you think that it should stay reported wrongly as "Unkown" because you think there are some scripts that could break?! (Even that I don't know how they could break if the value is reported only for the EFA driver) At least could you please give the name of one script that you know about?
> > Thanks
> 
> It is not me who are changing sysfs entry introduced more than ten years ago and it is upto you to support your claim that nothing is going to be broken.
> 
For now, you are the only person that talking about **Unknown** scripts that could break ...

******At least could you please give the name of one script that you know about?******

> And yes, from the kernel perspective, this field is "unknown".
> 
> Thanks

galpress:
Adding new sysfs values by itself isn't breaking anything, if a new link layer is added then a new value should be added to the sysfs, and it shouldn't affect scripts who look for older values.
This case is a bit different since we're changing the output of an existing value, but if EFA is the only driver setting the link layer to unspecified then isn't it the only driver that "breaks"? All other drivers remain unchanged.
I find it hard to believe that there is any code out there that's running over EFA and looks at this field.

Is there a legacy story behind this field where old drivers with infiniband link layer would report unspecified or something? Why was unspecified considered Infiniband in ibv_devinfo in the first place?

rleon:
> > > > IB puppet scripts returns link_layer as is, you have potential to break users of those scripts.
> > > 
> > > 
> > > As of today, the only provider that returns link_layer = IB_LINK_LAYER_UNSPECIFIED is the EFA driver..., And you think that it should stay reported wrongly as "Unkown" because you think there are some scripts that could break?! (Even that I don't know how they could break if the value is reported only for the EFA driver) At least could you please give the name of one script that you know about?
> > > Thanks
> > 
> > 
> > It is not me who are changing sysfs entry introduced more than ten years ago and it is upto you to support your claim that nothing is going to be broken.
> 
> For now, you are the only person that talking about **Unknown** scripts that could break ...
> 
> ******At least could you please give the name of one script that you know about?******
> 
> > And yes, from the kernel perspective, this field is "unknown".
> > Thanks

Will it help you if I use same language as Jason?
"Let's don't change legacy sysfs entries. no one is supposed to use them."

And regarding scripts, I changed my scripts a long time ago.

Thanks

rleon:
> Adding new sysfs values by itself isn't breaking anything, if a new link layer is added then a new value should be added to the sysfs, and it shouldn't affect scripts who look for older values.

Agree with this part.

> This case is a bit different since we're changing the output of an existing value, but if EFA is the only driver setting the link layer to unspecified then isn't it the only driver that "breaks"? All other drivers remain unchanged.

We are not talking about drivers, but the consumers of this sysfs.

> I find it hard to believe that there is any code out there that's running over EFA and looks at this field.
> 
> Is there a legacy story behind this field where old drivers with infiniband link layer would report unspecified or something? Why was unspecified considered Infiniband in ibv_devinfo in the first place?

I don't think so, the commit 8ad330a002ff ("IB/core: Add link layer type information to sysfs") from 2010 specifies that we had two types of link layers from the beginning.

galpress:
But the only sysfs consumers affected are the ones running over an EFA device, the rest will see the same value as they did before.

jgunthorpe:
Guys, if nobody is reading the value then there is no reason to change it, it is obsolete. If something is reading it then we need to understand what it is and check it. Seems high risk to me and I don't see the purpose??

nmorey:
Agreed with Jason. This risks breaking any scripts that might have been written in the last 10y for a very small upside.

rleon:
> Agreed with Jason. This risks breaking any scripts that might have been written in the last 10y for a very small upside.

Thanks Nicolas, I said the same.

Kamalheib:
> Guys, if nobody is reading the value then there is no reason to change it, it is obsolete. If something is reading it then we need to understand what it is and check it. Seems high risk to me and I don't see the purpose??

As of today the reported link layer from the kernel sysfs for the EFA provider is "Unknown", while the get_link_layer() EFA driver kernel callback return "IB_LINK_LAYER_UNSPECIFIED", also the reported link_layer from userspace ibv_devinfo app is "InfiniBand". so the purpose of this change is to avoid confusing the EFA users. 

Now that there is pushback for touching the kernel sysfs, what do you guys suggest to do:

0- Drop this PR and leave the reported info as is.
1- Report "Unkown" from ibv_devinfo (match the kernel sysfs value).
2- Report "Unspecified" from ibv_devinfo (regardless of what is reported from the kernel sysfs).

```
$ cat /sys/class/infiniband/efa_0/ports/1/link_layer 
Unknown
```

```
enum rdma_link_layer efa_port_link_layer(struct ib_device *ibdev,
                                         u8 port_num)
{
        return IB_LINK_LAYER_UNSPECIFIED;
}
```


```
$ ibv_devinfo 
hca_id:	efa_0
	transport:			unspecified (4)
	fw_ver:				0.0.0.0
	node_guid:			0000:0000:0000:0000
	sys_image_guid:			0000:0000:0000:0000
	vendor_id:			0x1d0f
	vendor_part_id:			61344
	hw_ver:				0xEFA0
	phys_port_cnt:			1
		port:	1
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		4096 (5)
			sm_lid:			0
			port_lid:		0
			port_lmc:		0x01
			link_layer:		InfiniBand
```

nmorey:
IMHO the PR makes sense so I'd say number 1
2 adds some code that I don't think necessary. I doubt anyone would be confused by seeing unspecified vs unknown.

galpress:
I would go with 2, it is inconsistent with the sysfs value, but it matches the value returned from ibv_query_port which might be more useful.

jgunthorpe:
There is no reason to refer to the sysfs at all. It is dead legacy stuff. This patch here is fine as is, no need for a kernel patch.

rleon:
As discussed, ibv_devinfo change is ok, kernel change is not.

Thanks

PR #893, opened 2020-11-22
mlx5: DR, Enhance dynamic flex parser support

yishaih:
This series enhances the dynamic flex parser support to a non-native network protocol parser.
This can be achieved by using DEVX to configure the parser then use its ID as part of rule insertion.

In addition,
Add support to match on 'tunnel geneve option' and improve ICMP and MPLS to rely on parser id capabilities.


yishaih:
Kernel headers patch was updated to match accepted kernel code.

PR #895, opened 2020-11-23
Add user space dmabuf support

j-xiong:
This is the user space counter-part of the kernel patch set to add 
dma-buf support to the RDMA subsystem. This patch series adds user 
space API for registering dma-buf based memory regions, updates 
pyverbs with the new API, and adds new tests. 
 
This series consists of six patches. The first patch updates the 
kernel headers for dma-buf support. Patch 2 adds the new API function 
and updates the man pages. Patch 3 implements the new API in the mlx5 
provider. Patch 4 adds new class definitions to pyverbs for the new API. 
Patch 5 adds a set of new tests for the new API. Patch 6 fixes bug in 
the utility code of the tests.

### DIFF
--- a/debian/libibverbs1.symbols
+++ b/debian/libibverbs1.symbols
@@ -9,6 +9,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.9@IBVERBS_1.9 30
  IBVERBS_1.10@IBVERBS_1.10 31
  IBVERBS_1.11@IBVERBS_1.11 32
+ IBVERBS_1.12@IBVERBS_1.12 34
  (symver)IBVERBS_PRIVATE_34 34
  _ibv_query_gid_ex@IBVERBS_1.11 32
  _ibv_query_gid_table@IBVERBS_1.11 32
@@ -99,6 +100,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  ibv_rate_to_mbps@IBVERBS_1.1 1.1.8
  ibv_rate_to_mult@IBVERBS_1.0 1.1.6
  ibv_read_sysfs_file@IBVERBS_1.0 1.1.6
+ ibv_reg_dmabuf_mr@IBVERBS_1.12 34
  ibv_reg_mr@IBVERBS_1.0 1.1.6
  ibv_reg_mr@IBVERBS_1.1 1.1.6
  ibv_reg_mr_iova@IBVERBS_1.7 25
--- a/libibverbs/dummy_ops.c
+++ b/libibverbs/dummy_ops.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2017 Mellanox Technologies, Inc.  All rights reserved.
+ * Copyright (c) 2020 Intel Corporation.  All rights reserved.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
@@ -435,6 +436,14 @@ static struct ibv_mr *reg_mr(struct ibv_pd *pd, void *addr, size_t length,
 	return NULL;
 }
 
+static struct ibv_mr *reg_dmabuf_mr(struct ibv_pd *pd, uint64_t offset,
+				    size_t length, uint64_t iova,
+				    int fd, int access)
+{
+	errno = EOPNOTSUPP;
+	return NULL;
+}
+
 static int req_notify_cq(struct ibv_cq *cq, int solicited_only)
 {
 	return EOPNOTSUPP;
@@ -542,6 +551,7 @@ const struct verbs_context_ops verbs_dummy_ops = {
 	query_srq,
 	read_counters,
 	reg_dm_mr,
+	reg_dmabuf_mr,
 	reg_mr,
 	req_notify_cq,
 	rereg_mr,
@@ -663,6 +673,7 @@ void verbs_set_ops(struct verbs_context *vctx,
 	SET_OP(vctx, read_counters);
 	SET_PRIV_OP(ctx, query_srq);
 	SET_OP(vctx, reg_dm_mr);
+	SET_PRIV_OP_IC(vctx, reg_dmabuf_mr);
 	SET_PRIV_OP(ctx, reg_mr);
 	SET_OP(ctx, req_notify_cq);
 	SET_PRIV_OP(ctx, rereg_mr);
--- a/libibverbs/verbs.c
+++ b/libibverbs/verbs.c
@@ -1,6 +1,7 @@
 /*
  * Copyright (c) 2005 Topspin Communications.  All rights reserved.
  * Copyright (c) 2006, 2007 Cisco Systems, Inc.  All rights reserved.
+ * Copyright (c) 2020 Intel Corperation.  All rights reserved.
  *
  * This software is available to you under a choice of one of two
  * licenses.  You may choose to be licensed under the terms of the GNU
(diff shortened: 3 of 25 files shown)


pyverbs/dmabuf.pyx
```diff
@@ -0,0 +1,85 @@
...
+        close_weakrefs([self.dmabuf_mrs])
+
+        # destroy a dumb scanout buffer:
+        #     struct drm_mode_destroy_dumb {
+        #         __u32 handle;
+        #     };
+        destroy_dumb = bytearray(4)
+        pack_into('=i', destroy_dumb, 0, self.handle)
+        ioctl(self.dri_fd, DRM_IOCTL_MODE_DESTROY_DUMB, destroy_dumb)
+        close(self.dri_fd)
```
EdwardSro, line 80: move this implementation to a cpdef close method. then call the self.close method from the __dealloc__ (as in all other places in pyverbs).
This will give the python user the ability to manually close this object (and not wait for the GC)

EdwardSro, line 80: before destroying please check if self.dri_fd is NULL. and after closing assign the self.dri_fd to be NULL (to prevent double deallocation - in case the GC done that twice or both the user and GC try to close. it shouldn't happen, but unfortunately sometimes cython has its "magic")

pyverbs/mr.pyx
```diff
@@ -348,6 +350,77 @@ cdef class DMMR(MR):
...
+        self.logger = logging.getLogger(self.__class__.__name__)
+        if dmabuf is None:
+            dmabuf = DmaBuf(length + offset)
+        self.mr = v.ibv_reg_dmabuf_mr(pd.pd, offset, length, dmabuf.fd, access)
+        if self.mr == NULL:
+            raise PyverbsRDMAErrno('Failed to register a dma-buf MR. length: {len}, access flags: {flags}'.
+                                   format(len=length, flags=access,))
+        super().__init__(pd, length, access)
+        self.pd = pd
+        self.dmabuf = dmabuf
```
EdwardSro, line 374: you should add a close + __dealloc__ and make sure to assign those to NULL upon closure. (this will reduce the ref-counts for python, and can hint the garbage collector to close those objects later)

EdwardSro (changes requested):

tests/test_mr.py
```diff
@@ -366,3 +367,130 @@ def test_destroy_dm_mr(self):
                         dm_mr = DMMR(pd, dm_mr_len, e.IBV_ACCESS_ZERO_BASED,
                                      dm=dm, offset=dm_mr_offset)
                         dm_mr.close()
+
```
EdwardSro, line 370: Some general comments:
- In case there is no dri card those tests will fail, and they can't. you should not run/skip those tests in such cases (you can check it's done in other places by try-except and raising unittest.SkipTest(..) )
- It's preferred not to use random value. we don't have/maintain a random seed, so in case of failures because of some specific value it may not be possible for the user to reproduce it. If you see that it's important for some value to be randomized and has an added value for the test(s), then at least print those attrs and their values in case of failure (so the user can reproduce it manually / have an idea of what's happening)

EdwardSro (changes requested):

j-xiong:
PR has been updated to address all the existing review comments.

debian/libibverbs1.symbols
```diff
@@ -9,6 +9,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.9@IBVERBS_1.9 30
  IBVERBS_1.10@IBVERBS_1.10 31
  IBVERBS_1.11@IBVERBS_1.11 32
+ IBVERBS_1.12@IBVERBS_1.12 33
```
jgunthorpe, line 12: 34 not 33 now

debian/libibverbs1.symbols
```diff
@@ -99,6 +100,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  ibv_rate_to_mbps@IBVERBS_1.1 1.1.8
  ibv_rate_to_mult@IBVERBS_1.0 1.1.6
  ibv_read_sysfs_file@IBVERBS_1.0 1.1.6
+ ibv_reg_dmabuf_mr@IBVERBS_1.12 33
```
jgunthorpe, line 103: 34 not 33

libibverbs/dummy_ops.c
```diff
@@ -670,6 +680,7 @@ void verbs_set_ops(struct verbs_context *vctx,
 	SET_PRIV_OP_IC(vctx, set_ece);
 	SET_PRIV_OP_IC(vctx, unimport_mr);
 	SET_PRIV_OP_IC(vctx, unimport_pd);
+	SET_OP(ctx, reg_dmabuf_mr);
```
jgunthorpe, line 683: Keep sorted

libibverbs/verbs.c
```diff
@@ -370,6 +371,23 @@ void ibv_unimport_mr(struct ibv_mr *mr)
...
+	struct ibv_mr *mr;
+
+	mr = get_ops(pd->context)->reg_dmabuf_mr(pd, offset, length, iova,
+						 fd, access);
+	if (mr) {
+		mr->context = pd->context;
+		mr->pd      = pd;
+		mr->addr    = (void *)offset;
+		mr->length  = length;
+	}
```
jgunthorpe, line 387: No horizontal white space

Success oriented flow, write as:
```C
  if (!mr)
       return NULL;
 mr->.. = ...
 return mr 
```

libibverbs/verbs.h
```diff
@@ -1962,6 +1963,10 @@ struct ibv_context_ops {
 	void *(*_compat_attach_mcast)(void);
 	void *(*_compat_detach_mcast)(void);
 	void *(*_compat_async_event)(void);
+	struct ibv_mr *		(*reg_dmabuf_mr)(struct ibv_pd *pd,
```
jgunthorpe, line 1966: Do not change this structure.

jgunthorpe:
Also fix all of the CI errors and warnings - click on checks and follow it through

pyverbs/dmabuf_alloc.c
```diff
@@ -0,0 +1,278 @@
...
+
+struct drm {
+	int fd;
+	int (*alloc)(struct drm *drm, uint64_t size, uint32_t *handle, int gtt);
+	int (*mmap_offset)(struct drm *drm, uint32_t handle, uint64_t *offset);
+};
+
+static int i915_alloc(struct drm *drm, uint64_t size, uint32_t *handle, int gtt)
+{
+	struct drm_i915_gem_create gem_create = {0};
```
jgunthorpe, line 35: Do not use {0}, the idiom is {}, the CI should throw warnings from clang for this usage.

pyverbs/dmabuf_alloc.c
```diff
@@ -0,0 +1,278 @@
...
+	struct drm_version version = {0};
+	char name[16] = {0};
+	int err;
+	struct drm *drm;
+
+	drm = malloc(sizeof(*drm));
+	if (!drm)
+		return NULL;
+
+	sprintf(path, "/dev/dri/renderD%d", unit + 128);
```
jgunthorpe, line 110: snprintf

jgunthorpe (changes requested):

j-xiong:
PR updated to address review comments from @jgunthorpe.

rleon:
Please fix conflct, symbols and update kernel-headers with kernel-header/update script so we will be able to merge this PR.

Thanks

j-xiong:
PR rebased to resolve conflicts and symbols. Re-ran kernel-header/update against linux-rdma/for-next.

PR #896, opened 2020-11-26
libhns: Support owner mode doorbell

Li-Weihang:
The doorbell needs to store PI information into QPC, so the RoCEE should
wait for the results of storing, that is, it needs two bus operations to
complete a doorbell. When ROCEE is in SDI mode, multiple doorbells may be
interlocked because the RoCEE can only handle bus operations serially. So a
flag to mark if HIP09 is working in SDI mode is added. When the SDI flag is
set, the ROCEE will ignore the PI information of the doorbell, continue to
fetch wqe and verify its validity by it's owner_bit.

Signed-off-by: Weihang Li <liweihang@huawei.com>


providers/hns/hns_roce_u_hw_v2.h
```diff
@@ -40,9 +40,10 @@
 
 #define HNS_ROCE_CMDSN_MASK			0x3
 
-enum {
-	HNS_ROCE_SUPPORT_RQ_RECORD_DB = 1 << 0,
-	HNS_ROCE_SUPPORT_SQ_RECORD_DB = 1 << 1,
+enum hns_roce_qp_flags {
```
rleon, line 43: Why aren't those flags in hns-abi.h? They exist in the kernel and taken from there.

Li-Weihang: Thanks for your reminder. These flag should be in hns-abi.h, I will modify the code.

PR #901, opened 2020-12-07
Coule of pyverbs tests fixes

galpress:


tests/test_cq.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
 # Copyright (c) 2019 Mellanox Technologies, Inc. All rights reserved. See COPYING file
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All rights reserved.
```
juhlee-microsoft, line 3: Is it possible to have 2 copyright statements?

tests/test_device.py
```diff
@@ -155,8 +156,8 @@ def test_query_gid_ex_bad_flow(self):
                              f'Got {os.strerror(ex.error_code)} but '
                              f'Expected {os.strerror(errno.ENODATA)}')
         else:
-            raise PyverbsRDMAError(f'Successfully queried '
-                                   'non-existent gid index {i}')
+            raise PyverbsRDMAError('Successfully queried '
+                                   f'non-existent gid index {no_gid_index}')
```
juhlee-microsoft, line 160: Any chance of exception raised from 134 ~ 142?

tests/test_cq.py
rleon: Yes, the copyright statements are not limited to one company/author. Everyone who adds/changes code can claim their copyright on it.

Of course in this particular case where Gal changed two lines and requested copyright for that, I'm feeling sad, it has zero value outside of the micro-world of enterprise lawyers.

Thanks

tests/test_device.py
galpress: I don't think so.

EdwardSro (approved):

PR #905, opened 2020-12-15
libhns: Add support of UD for HIP09

Li-Weihang:
Add service type of Unreliable Datagram support for HIP09 by achieving Address Handle related interfaces and adding a branch in hns_roce_post_send().

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -44,11 +44,13 @@
 #include <ccan/array_size.h>
 #include <ccan/bitmap.h>
 #include <ccan/container_of.h>
+#include <linux/if_ether.h>
 #include "hns_roce_u_abi.h"
 
 #define HNS_ROCE_HW_VER1		('h' << 24 | 'i' << 16 | '0' << 8 | '6')
 
 #define HNS_ROCE_HW_VER2		('h' << 24 | 'i' << 16 | '0' << 8 | '8')
+#define HNS_ROCE_HW_VER3		0x130
 
 #define PFX				"hns: "
 
@@ -269,6 +271,23 @@ struct hns_roce_qp {
 	unsigned long			flags;
 };
 
+struct hns_roce_av {
+	uint8_t port;
+	uint8_t gid_index;
+	uint8_t hop_limit;
+	uint32_t flowlabel;
+	uint16_t udp_sport;
+	uint8_t sl;
+	uint8_t tclass;
+	uint8_t dgid[HNS_ROCE_GID_SIZE];
+	uint8_t mac[ETH_ALEN];
+};
+
+struct hns_roce_ah {
+	struct ibv_ah ibv_ah;
+	struct hns_roce_av av;
+};
+
 struct hns_roce_u_hw {
 	uint32_t hw_version;
 	struct verbs_context_ops hw_ops;
@@ -316,6 +335,11 @@ static inline struct  hns_roce_qp *to_hr_qp(struct ibv_qp *ibv_qp)
 	return container_of(ibv_qp, struct hns_roce_qp, ibv_qp);
 }
 
+static inline struct hns_roce_ah *to_hr_ah(struct ibv_ah *ibv_ah)
+{
+	return container_of(ibv_ah, struct hns_roce_ah, ibv_ah);
+}
+
 int hns_roce_u_query_device(struct ibv_context *context,
 			    const struct ibv_query_device_ex_input *input,
 			    struct ibv_device_attr_ex *attr, size_t attr_size);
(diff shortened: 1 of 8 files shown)


providers/hns/hns_roce_u.h
```diff
@@ -124,6 +126,9 @@ struct hns_roce_buf {
 #define BIT_CNT_PER_BYTE       8
 #define BIT_CNT_PER_LONG       (BIT_CNT_PER_BYTE * sizeof(unsigned long))
 
+#define GENMASK(h, l) \
```
rleon, line 129: EFA already has this define.
Please reuse it.

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -766,6 +766,122 @@ static __le32 get_immtdata(enum ibv_wr_opcode opcode, const struct ibv_send_wr *
 	}
 }
 
+static int check_ud_opcode(struct hns_roce_ud_sq_wqe *ud_sq_wqe,
+			   const struct ibv_send_wr *wr)
+{
+	uint32_t ib_op = wr->opcode;
+
+	if (ib_op != IBV_WR_SEND && ib_op != IBV_WR_SEND_WITH_IMM)
+		return EINVAL;
```
rleon, line 775: Error returned from the functions should be < 0. I see that you are returning > 0 already, but it doesn't look right.

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -766,6 +766,122 @@ static __le32 get_immtdata(enum ibv_wr_opcode opcode, const struct ibv_send_wr *
...
+		       UD_SQ_WQE_HOPLIMIT_S, ah->av.hop_limit);
+
+	roce_set_field(ud_sq_wqe->lbi_flow_label, UD_SQ_WQE_FLOW_LABEL_M,
+		       UD_SQ_WQE_FLOW_LABEL_S, ah->av.flowlabel);
+
+	roce_set_field(ud_sq_wqe->udpspn_rsv, UD_SQ_WQE_UDP_SPN_M,
+		       UD_SQ_WQE_UDP_SPN_S, ah->av.udp_sport);
+
+	if (unlikely(ah->av.sl > MAX_SERVICE_LEVEL))
+		return EINVAL;
```
rleon, line 804: Why don't you check it at the beginning of the function?

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -766,6 +766,122 @@ static __le32 get_immtdata(enum ibv_wr_opcode opcode, const struct ibv_send_wr *
...
+	memcpy(ud_sq_wqe->dgid, ah->av.dgid, HNS_ROCE_GID_SIZE);
+
+	return 0;
+}
+
+static int fill_ud_data_seg(struct hns_roce_ud_sq_wqe *ud_sq_wqe,
+			    struct hns_roce_qp *qp, struct ibv_send_wr *wr,
+			    struct hns_roce_sge_info *sge_info)
+{
+	int ret = 0;
```
rleon, line 820: It is not used

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -766,6 +766,122 @@ static __le32 get_immtdata(enum ibv_wr_opcode opcode, const struct ibv_send_wr *
...
+		       sge_info->start_idx & (qp->ex_sge.sge_cnt - 1));
+
+	set_sge((struct hns_roce_v2_wqe_data_seg *)ud_sq_wqe, qp, wr, sge_info);
+
+	ud_sq_wqe->msg_len = htole32(sge_info->total_len);
+
+	roce_set_field(ud_sq_wqe->sge_num_pd, UD_SQ_WQE_SGE_NUM_M,
+		       UD_SQ_WQE_SGE_NUM_S, sge_info->valid_num);
+
+	return ret;
```
rleon, line 834: This is always 0.

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: Thanks, I will put it at the beginning.

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: Thanks, will remove it.

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: Thanks, will remove it.

providers/hns/hns_roce_u.h
Li-Weihang: OK, will put it in ccan/. Thank you.

providers/hns/hns_roce_u_hw_v2.c
Li-Weihang: I use positive return values because I found the following disscusion before: https://www.spinics.net/lists/linux-rdma/msg89507.html, and according to the man page, ibv_post_send() should return a positive errno. Did I misunderstand it?
Thank you.

providers/hns/hns_roce_u.h
Li-Weihang: I don't find GENMASK in CCAN, so I will achieve it in util.h.

providers/hns/hns_roce_u_hw_v2.c
rleon: Jason talks about "errno = ...", which is returned to the user, while here we are talking about internal return value from internal functions. 

For example, this is why librdmacm has ERR() macro.
https://github.com/linux-rdma/rdma-core/blob/master/librdmacm/cma.h#L89
It separates "errno" and "ret".

The current hns code is correct, but misleading.

Li-Weihang: Thanks for your explanation, we will fix this later.

PR #908, opened 2020-12-22
librdmacm: Don't overwrite errno returned from libibverbs

rleon:
Some functions reports fixed ENOMEM when getting any failure, so
it's hard for user to know which actual error happens on them.

Fixes: 663098bfc3ac ("Rename librdmacm")
Signed-off-by: Xiao Yang <yangx.jy@cn.fujitsu.com>


rleon:
Fixes line is not right, but we don't care about it because the error was even before rdma-core.

PR #912, opened 2020-12-28
tests: Fix test_query_gid_ex_bad_flow

Kamalheib:
This change fixes two issues:
1. Avoid the failure of query_gid_table() by calculating the max_entries
    for all ports when running the test over devices with multiple ports.
2. Avoid the failure of successfully queried non-existing gid index when
    the gid_tbl_len=1, This happen because the possible_indices was wrongly
    calculated to query the only exist gid over port one.

Fixes: cbb04cdfaff1 ("tests: Add query_gid_table and query_gid_ex bad flow")

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -135,11 +135,14 @@ class DeviceTest(PyverbsAPITestCase):
         ctx, device_attr, _ = self.devices[0]
         try:
             port_attr = ctx.query_port(1)
-            max_entries = device_attr.phys_port_cnt * port_attr.gid_tbl_len
+            max_entries = 0
+            for port_num in range(1, device_attr.phys_port_cnt + 1):
+                attr = ctx.query_port(port_num)
+                max_entries += attr.gid_tbl_len
             gid_indices = {gid_entry.gid_index for gid_entry in
                            ctx.query_gid_table(max_entries) if gid_entry.port_num == 1}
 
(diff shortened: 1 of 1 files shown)


tests/test_device.py
```diff
@@ -135,11 +135,14 @@ def test_query_gid_ex_bad_flow(self):
...
-            max_entries = device_attr.phys_port_cnt * port_attr.gid_tbl_len
+            max_entries = 0
+            for port_num in range(1, device_attr.phys_port_cnt + 1):
+                attr = ctx.query_port(port_num)
+                max_entries += attr.gid_tbl_len
             gid_indices = {gid_entry.gid_index for gid_entry in
                            ctx.query_gid_table(max_entries) if gid_entry.port_num == 1}
 
-            possible_indices = set(range(port_attr.gid_tbl_len))
+            possible_indices = set(range(port_attr.gid_tbl_len)) if port_attr.gid_tbl_len > 1 else set()
```
EdwardSro, line 145: Not sure why we need this change after your first fix.
Assuming we have only one gid (gid_tbl_len == 1), the no_gid_index will pop from an empty set and the test would be skipped, wouldn't it?

Kamalheib: No, because on devices with two ports and each one of them has only one gid entry the test will fail, because the no_gid_index was calculated wrongly to query the gid under index 0 (which is valid).

EdwardSro (approved):

PR #914, opened 2020-12-30
efa: EFA_GET/SET() fixes

galpress:
This commit matches the EFA_GET/SET() macros with their kernel
counterparts.
This change clears the mask before assigning the new value, and makes
use of FIELD_GET/PREP() which makes the various *_SHIFT definitions
redundant.


rleon:
I prefer that everything borrowed from general kernel code will be placed in common files, e.g. utils.h/c.

Thanks

galpress:
Thanks, will do.

PR #915, opened 2020-12-30
tests: Port command line option

Kamalheib:
This PR adds the '--port' command line option that will be used to run the pyverbs on a specific device port, This is needed when there are multiple ports supported by one device, so the tests will run on the specified device port instead of on all of them.

### DIFF
--- a/tests/args_parser.py
+++ b/tests/args_parser.py
@@ -16,6 +16,9 @@ class ArgsParser(object):
         parser = argparse.ArgumentParser()
         parser.add_argument('--dev',
                             help='RDMA device to run the tests on')
+        parser.add_argument('--port',
+                            help='Use port <port> of RDMA device', type=int,
+                            default=1)
         parser.add_argument('-v', '--verbose', dest='verbosity',
                             action='store_const',
                             const=2, help='Verbose output')
--- a/tests/base.py
+++ b/tests/base.py
@@ -68,6 +68,7 @@ class PyverbsAPITestCase(unittest.TestCase):
         Opens the devices and queries them
         """
         self.devices = []
+        self.ib_port = self.config['port']
 
         dev_name = self.config['dev']
         if dev_name:
@@ -118,7 +119,7 @@ class RDMATestCase(unittest.TestCase):
         self.config = parser.get_config()
         dev = self.config['dev']
         self.dev_name = dev_name if dev_name else dev
-        self.ib_port = ib_port
+        self.ib_port = ib_port if ib_port else self.config['port']
         self.gid_index = gid_index
         self.pkey_index = pkey_index
         self.gid_type = gid_type if gid_index is None else None
@@ -250,9 +251,7 @@ class RDMATestCase(unittest.TestCase):
                 self.args.append([dev, port, idx, ip_addr])
 
     def _add_gids_per_device(self, ctx, dev):
-        port_count = ctx.query_device().phys_port_cnt
-        for port in range(port_count):
-            self._add_gids_per_port(ctx, dev, port+1)
+        self._add_gids_per_port(ctx, dev, self.ib_port)
 
     def _select_config(self):
         args_with_inet_ip = []
--- a/tests/test_addr.py
+++ b/tests/test_addr.py
@@ -5,8 +5,9 @@ import unittest
 import errno
 
 from pyverbs.pyverbs_error import PyverbsError, PyverbsRDMAError
-from pyverbs.addr import GlobalRoute, AHAttr, AH
 from tests.base import PyverbsAPITestCase
+from pyverbs.addr import AHAttr, AH
+import pyverbs.device as d
 import pyverbs.enums as e
 from pyverbs.pd import PD
 import tests.utils as u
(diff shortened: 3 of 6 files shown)


rleon:
Why do we need such functionality?
When will you need to run API checks on specific port and it will give you different results based on port number?
Except mlx4, I'm not aware of any.

Kamalheib:
> Why do we need such functionality?
> When will you need to run API checks on specific port and it will give you different results based on port number?
> Except mlx4, I'm not aware of any.

As you can see in PR that some of the tests ran only over port 1 and others run over all the available ports.

We use the pyverbs tests in the regression system that we have at Redhat and It's important for us to know over which port the test is running because in some of the setups with multiple ports (not only mlx4) only one port is connected (could be port 1 or 2) the other port is down. So with this functionality we 100% sure that we run the test over the port that we want.

rleon:
> > Why do we need such functionality?
> > When will you need to run API checks on specific port and it will give you different results based on port number?
> > Except mlx4, I'm not aware of any.
> 
> As you can see in PR that some of the tests ran only over port 1 and others run over all the available ports.

I would say that this is a bug in the test, all should run over one (single) port.

> 
> We use the pyverbs tests in the regression system that we have at Redhat and It's important for us to know over which port the test is running because in some of the setups with multiple ports (not only mlx4) only one port is connected (could be port 1 or 2) the other port is down. So with this functionality we 100% sure that we run the test over the port that we want.

Ahh, ok, thanks

tests/test_addr.py
```diff
@@ -16,77 +17,69 @@ class AHTest(PyverbsAPITestCase):
...
-                    raise ex
-        if done == 0:
-            raise unittest.SkipTest('No port is up, can\'t create AH')
+            try:
+                with AH(pd, attr=ah_attr) as ah:
+                    ah.close()
+            except PyverbsRDMAError as ex:
+                if ex.error_code == errno.EOPNOTSUPP:
+                    raise unittest.SkipTest('Create AH is not supported')
+                raise ex
```
EdwardSro, line 85: Keep a blank line at EOF

tests/test_addr.py
```diff
@@ -16,77 +17,69 @@ class AHTest(PyverbsAPITestCase):
...
 
     def test_destroy_ah(self):
         """
         Test ibv_destroy_ah.
         """
-        done = 0
-        for ctx, attr, attr_ex in self.devices:
+        for ctx, _, _ in self.devices:
+            self.verify_state(ctx)
+            gr = u.get_global_route(ctx)
```
EdwardSro, line 76: I think you should pass the port_num here (looks like it was a bug also before your changes)

EdwardSro (changes requested):

EdwardSro:
> I would say that this is a bug in the test, all should run over one (single) port.

Not really a bug. Part of the old tests (basic API tests) was running on all devices and ports of the setup. But yes, I tend to agree that they should run with a single "config" (device, port, etc.)
Newer tests that inherit from the new infrastructure doesn't do that, and we were gradually (on demand) adapting and changing the behaviour of the old tests when we were adding new cases to them.

EdwardSro:
@Kamalheib 
Pay attention, that after your changes, the port cmd arg would be ignored if --dev was not passed, and would have no effect. I would document (at least in the help) or prevent the user from doing so (passing --port without --dev) to eliminate any confusion.

Kamalheib:
> @Kamalheib
> Pay attention, that after your changes, the port cmd arg would be ignored if --dev was not passed, and would have no effect. I would document that (at least in the help) or prevent the user from doing so (passing --port without --dev) to eliminate any confusion.

@EdwardSro 
Correct me If I'm wrong, you mean only the tests that inherit from RDMATestCase (under _add_gids_per_device())?
I suggest to change it too to use the passed --port parameter? thoughts?

EdwardSro:
> Correct me If I'm wrong, you mean only the tests that inherit from RDMATestCase (under _add_gids_per_device())?
Correct.
> I suggest to change it too to use the passed --port parameter? thoughts?
Do you mean that even if the user didn't explicitly pass the device (using --dev) that the test would still use the --port with the chosen device by RDMATestCase?
That's also ok. Anyway, the user has to be responsible when using --port if he/she has multiple devices while some of them have 1 and others have 2+ ports.

Kamalheib:
> > Correct me If I'm wrong, you mean only the tests that inherit from RDMATestCase (under _add_gids_per_device())?
> 
> Correct.
> 
> > I suggest to change it too to use the passed --port parameter? thoughts?
> 
> Do you mean that even if the user didn't explicitly pass the device (using --dev) that the test would still use the --port with the chosen device by RDMATestCase?

Yes, This what I meant.

> That's also ok. Anyway, the user has to be responsible when using --port if he/she has multiple devices while some of them have 1 and others have 2+ ports.

Correct, This is the goal of this PR, I'll update the PR to address your comment, and I'll push v2.

Thanks,
Kamal

tests/test_addr.py
Kamalheib: Agree, Fixed.

tests/test_addr.py
Kamalheib: Fixed.

tests/base.py
```diff
@@ -118,7 +119,7 @@ def __init__(self, methodName='runTest', dev_name=None, ib_port=None,
         self.config = parser.get_config()
         dev = self.config['dev']
         self.dev_name = dev_name if dev_name else dev
-        self.ib_port = ib_port
+        self.ib_port = self.config['port']
```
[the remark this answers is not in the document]
Kamalheib: Yes, Fixed.

EdwardSro (approved):

tests/args_parser.py
```diff
@@ -16,6 +16,9 @@ def parse_args(self):
         parser = argparse.ArgumentParser()
         parser.add_argument('--dev',
                             help='RDMA device to run the tests on')
+        parser.add_argument('--port',
+                            help='Use port <port> of RDMA device', type=int,
+                            default=1, choices=[1, 2])
```
rleon, line 21: choices=[1, 2] <- chellsio supports 4 ports and Mellanox has such multiport device too.
It is better to be based on max supported by the device.

So I suggest to simply drop this "choices=[]" for now.

Kamalheib: This is needed to validate that the user is passing a valid port number, So instead of dropping the choices option, what about changing it to [1,2,3,4]?

rleon: If user provided bad port, pyverbs test will crash.
Why isn't good enough?

Kamalheib: > If user provided bad port, pyverbs test will crash.
> Why isn't good enough?

This is a very bad user experience :-), running the whole pyverbs tests and later to understand the user passed a wrong port number...

rleon: We are talking about pyverbs tests, even today with choices=[1,2], it will crash if i try to run over single port device.

IMHO, it just doesn't worth to make it "clean".

EdwardSro: I think Leon is right... I would just drop the choices.
As he mentioned, even now passing port num 2 on a single port device would fail the tests. The querying for the max ports option would complicate things and is not worth it right now.

Kamalheib: done.

rleon:
Thanks

PR #916, opened 2020-12-31
pyverbs: Add a new active speed translation

EdwardSro:
Prepare support for a new IB speed rate (NDR).
A translation was added to the active speed attribute.

### DIFF
--- a/pyverbs/device.pyx
+++ b/pyverbs/device.pyx
@@ -1078,7 +1078,8 @@ def width_to_str(width):
 
 def speed_to_str(speed):
     l = {0: '0.0 Gbps', 1: '2.5 Gbps', 2: '5.0 Gbps', 4: '5.0 Gbps',
-         8: '10.0 Gbps', 16: '14.0 Gbps', 32: '25.0 Gbps', 64: '50.0 Gbps'}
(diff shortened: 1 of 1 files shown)


pyverbs/device.pyx
```diff
@@ -1078,7 +1078,8 @@ def width_to_str(width):
 
 def speed_to_str(speed):
     l = {0: '0.0 Gbps', 1: '2.5 Gbps', 2: '5.0 Gbps', 4: '5.0 Gbps',
-         8: '10.0 Gbps', 16: '14.0 Gbps', 32: '25.0 Gbps', 64: '50.0 Gbps'}
+         8: '10.0 Gbps', 16: '14.0 Gbps', 32: '25.0 Gbps', 64: '50.0 Gbps',
+         128: '100.0 Gbps'}
```
juhlee-microsoft, line 1082: You will need to add new value in devinfo.c also

EdwardSro: > You will need to add new value in devinfo.c also

I believe that this change should come with the libibverbs update (updating the rates of ibv_devinfo which should come soon - not by me).
I guess I could have waited with this PR, but didn't want it to be missed out.

PR #919, opened 2021-01-06
More SRD pyverbs testing

galpress:
The first patch adds SRD RDMA read test, and the next two patches add send and send with immediate testing on SRD extended QPs with old send API.


tests/utils.py
```diff
@@ -401,7 +401,7 @@ def post_send_ex(agr_obj, send_object, send_op=None, qp_idx=0, ah=None):
     qp.wr_complete()
 
 
-def post_send(agr_obj, send_wr, qp_idx=0, ah=None):
+def post_send(agr_obj, send_wr, qp_idx=0, ah=None, is_imm=False):
     """
```
EdwardSro, line 405: please update the docstring (just noticed that ah is also missing)

galpress: Done

EdwardSro (approved):

PR #921, opened 2021-01-09
libhns: Support extended verbs

Li-Weihang:
Support ibv_create_qp_ex on libhns.

### DIFF
--- a/providers/hns/hns_roce_u_verbs.c
+++ b/providers/hns/hns_roce_u_verbs.c
@@ -541,34 +541,68 @@ int hns_roce_u_destroy_srq(struct ibv_srq *srq)
 	return 0;
 }
 
-static int hns_roce_verify_qp(struct ibv_qp_init_attr *attr,
-			      struct hns_roce_context *context)
+enum {
+	CREATE_QP_SUP_CREATE_FLAGS = 0,
+};
+
+enum {
+	CREATE_QP_SUP_COMP_MASK = IBV_QP_INIT_ATTR_PD |
+				  IBV_QP_INIT_ATTR_CREATE_FLAGS,
+};
+
+static int check_qp_create_mask(struct ibv_qp_init_attr_ex *attr)
 {
-	struct hns_roce_device *hr_dev =
-		to_hr_dev(context->ibv_ctx.context.device);
-	uint32_t min_wqe_num = hr_dev->hw_version == HNS_ROCE_HW_VER1 ?
-			       HNS_ROCE_V1_MIN_WQE_NUM :
-			       HNS_ROCE_V2_MIN_WQE_NUM;
+	if (!check_comp_mask(attr->comp_mask, CREATE_QP_SUP_COMP_MASK))
+		return -EOPNOTSUPP;
 
-	if (!attr->cap.max_send_wr ||
-	    attr->cap.max_send_wr > context->max_qp_wr ||
-	    attr->cap.max_recv_wr > context->max_qp_wr ||
-	    attr->cap.max_send_sge > context->max_sge  ||
-	    attr->cap.max_recv_sge > context->max_sge)
-		return EINVAL;
+	if (attr->comp_mask & IBV_QP_INIT_ATTR_CREATE_FLAGS &&
+	    !check_comp_mask(attr->create_flags, CREATE_QP_SUP_CREATE_FLAGS))
+		return -EOPNOTSUPP;
 
-	if (attr->cap.max_send_wr < min_wqe_num)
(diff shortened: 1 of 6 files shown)


providers/hns/hns_roce_u_verbs.c
```diff
@@ -251,12 +251,33 @@ int hns_roce_u_dealloc_mw(struct ibv_mw *mw)
...
+
+enum {
+	CREATE_CQ_SUPPORTED_COMP_MASK = IBV_CQ_INIT_ATTR_MASK_FLAGS,
+};
+
+static int verify_cq_create_attr(struct ibv_cq_init_attr_ex *attr,
+				 struct hns_roce_context *context)
 {
-	if (*cqe < 1 || *cqe > context->max_cqe)
+	if (attr->comp_mask & ~CREATE_CQ_SUPPORTED_COMP_MASK)
```
jgunthorpe, line 269: This should be check_comp_mask() - same for all other places

providers/hns/hns_roce_u_verbs.c
```diff
@@ -272,25 +293,42 @@ static int hns_roce_alloc_cq_buf(struct hns_roce_cq *cq)
...
 
-	ret = ibv_cmd_create_cq(context, cqe, channel, comp_vector,
-				&cq->ibv_cq, &cmd.ibv_cmd, sizeof(cmd),
-				&resp.ibv_resp, sizeof(resp));
+	if (use_ex)
+		ret = ibv_cmd_create_cq_ex(context, attr, &cq->verbs_cq,
+					   &cmd_ex.ibv_cmd, sizeof(cmd_ex),
+					   &resp_ex.ibv_resp, sizeof(resp_ex));
+	else
+		ret = ibv_cmd_create_cq(context, attr->cqe, attr->channel,
```
jgunthorpe, line 327: As we just discussed on mlx5, simply always call ibv_cmd_create_cq_ex, it now internally does this logic.

providers/hns/hns_roce_u_verbs.c
```diff
@@ -251,12 +251,33 @@ int hns_roce_u_dealloc_mw(struct ibv_mw *mw)
...
-	if (*cqe < 1 || *cqe > context->max_cqe)
+	if (attr->comp_mask & ~CREATE_CQ_SUPPORTED_COMP_MASK)
+		return -EINVAL;
+
+	if (attr->comp_mask & IBV_CQ_INIT_ATTR_MASK_FLAGS &&
+	    attr->flags & ~CREATE_CQ_SUPPORTED_FLAGS)
+		return -EINVAL;
+
+	if (!attr->cqe || attr->cqe > context->max_cqe)
 		return -EINVAL;
```
jgunthorpe, line 278: Does this pass the pyverbs test suite? I don't think you can support ibv_create_qp_ex without also supporting wc_flags and ibv_start_poll()/etc.

jgunthorpe (changes requested):

providers/hns/hns_roce_u_verbs.c
Li-Weihang: Thanks for you reminder, this commit failed on test_cqex. We'll look into the details of cq_ex.

Li-Weihang: The patch to support cq_ex is dropped from this PR because it may take some time to figure out how to use ibv_create_cq_ex() correctly.

providers/hns/hns_roce_u_verbs.c
```diff
@@ -853,15 +877,30 @@ static int qp_exec_create_cmd(struct ibv_pd *pd,
 	cmd.log_sq_stride = qp->sq.wqe_shift;
 	cmd.log_sq_bb_count = hr_ilog32(qp->sq.wqe_cnt);
 
-	ret = ibv_cmd_create_qp(pd, &qp->ibv_qp, attr, &cmd.ibv_cmd,
-				sizeof(cmd), &resp.ibv_resp, sizeof(resp));
+	if (attr->comp_mask & CREATE_QP_EX2_COMP_MASK) {
```
jgunthorpe, line 880: Thre is no reason to have this if, just always use ibv_cmd_create_qp_ex2() and always decleare the ex cmd structure

Li-Weihang: Hi Jason, I have updated the patch. Is there anything that needs to be modified?

jgunthorpe:
I'll look after we make this rdma-core release

Li-Weihang:
This patch will be merged with a patchset to support XRC on HIP09, so I close this PR, thanks for all the comments.

PR #922, opened 2021-01-11
srp_daemon: Fix systemd dependency

nmorey:
remote-fs-pre.target is a passive target that is not loaded without someone
actively requiring it. Even is remote-fs.target is active.
This means that srp_daemon will not get started at boot
unless another service explicitely requires remote-fs-pre.target.

This solves the issue by having the srp_daemon service itself actively
asking for remote-fs-pre.target to be loaded.

Fixes: 08f689176497 (Revise systemd dependencies for all units)

### DIFF
--- a/srp_daemon/srp_daemon.service.in
+++ b/srp_daemon/srp_daemon.service.in
@@ -16,4 +16,4 @@ ProtectKernelModules=yes
 RestrictRealtime=yes
 
 [Install]
-WantedBy=remote-fs-pre.target
+WantedBy=multi-user.target


nmorey:
Fixed culprit commit

nmorey:
The fix doesn't actually work (I'm guessing systemd doesn't like this Before/Want cycle) but the issue does exists

jgunthorpe:
This is all correct, if something does need ordering with remotes-fs-pre.target then srp_daemon will properly order. It is just a before, not a requirement.

srp_daemon runs on boot because the post-install scripts in the packaging are supposed to install it into systemd, eg with %systemd_post in rpm land.

nmorey:
It is installed in systemd but unless there's a dependency path to it, it won't get loaded at boot, and right now there are none.

This would have done the trick
```
[Install]
WantedBy=remote-fs-pre.target
```
But remote-fs-pre.target is a passive target and does not get loaded unless someone explicitely requires it.
Having something like
`WantedBy=multi.user` would work as this target is an active one.

jgunthorpe:
Yes, some install section seems like the right thing, but then again, we don't necessarily want it running all the time either, do we?

nmorey:
Well there's still the systemd enabled/disabled flag
By default the service will be disabled and it won't get loaded anyway.

If we use a new target for the WantedBy, it'll get loaded only if enabled by the admin (at least that's the goal)

jgunthorpe:
That seems reasonable

nmorey:
After some fiddling around, changing WantedBy to multi-user.target seems like the best solution.
It doesn't change anything for people who have not manually enabled the service and will properly start if for those who have

nmorey:
@jgunthorpe Can we get this in if you're OK with this latest version ?

jgunthorpe:
I'm not sure. It was remote-fs-pre because you need to start srp daemon if the user plans to mount any filesystems linked to a SRP scsi device, you got rid of that completely. Should it be:
 Wants=multi-user.target
 Before=remote-fs-pre.target

?

nmorey:
The former WantedBy=remote-fs-pre has 0 effect.
Either remote-fs-pre had no service with dependencies to it, then it would only be a passive target so none of the service it "Wants" will be loaded
Either it had a dependency to it, and in this case the Before=remote-fs-pre would already ensure that srp_daemon is loaded realy enough. So the WantedBy= has no additional value.

In the new setup:
- srp_daemon will always be loaded (if enabled) has the WantedBy=multi-user adds a dependency to it
- if remote-fs-pre is required by someone, we still have the Before=remote-fs-pre make sure srp_daemon gets loaded early enough

PR #923, opened 2021-01-13
ibacm: acm.c load plugin while it is soft link

Honggang-LI:
Because of commit ad5d934d688911149d795aee1d3b9fa06bf171a9,
the provider libdsap.so.1.0.0 was not opened/used for address resolution
for OPA device.

As discussed in this closed PR:
https://github.com/linux-rdma/rdma-core/pull/848

I create a soft link for libdsap.so => libdsap.so.1.0.0 . The soft link
was ignored becuase it is a not regular file.

### DIFF
--- a/ibacm/src/acm.c
+++ b/ibacm/src/acm.c
@@ -2879,7 +2879,7 @@ static int acm_open_providers(void)
 			acm_log(0, "Error - could not stat: %s\n", file_name);
 			continue;
 		}
-		if (!S_ISREG(buf.st_mode))
+		if (!(S_ISREG(buf.st_mode) || S_ISLNK(buf.st_mode)))
 			continue;
 
 		acm_log(2, "Loading provider %s...\n", file_name);


ibacm/src/acm.c
```diff
@@ -2879,7 +2879,7 @@ static int acm_open_providers(void)
 			acm_log(0, "Error - could not stat: %s\n", file_name);
 			continue;
 		}
-		if (!S_ISREG(buf.st_mode))
+		if (!S_ISREG(buf.st_mode) && !S_ISLNK(buf.st_mode))
```
Hakon-Bugge, line 2882: A nit. For me,
        if (!(S_ISREG(buf.st_mode) || S_ISLNK(buf.st_mode)))
is easier to read.

Hakon-Bugge (approved):

jgunthorpe:
As before, we don't support out of tree ACM plugins and they are just being built wrong to install a *versioned* shared library. Plugins are not versioned like that.

ibacm/src/acm.c
Honggang-LI: Fixed. Thanks

Honggang-LI:
> As before, we don't support out of tree ACM plugins and they are just being built wrong to install a _versioned_ shared library. Plugins are not versioned like that.

We talked with OPA provider. It is up to OPA provider to merge DSAP plugin to rdma-core or not.

I'd like to get a properly *versioned* shared library. When libraries are not versioned, it is really easy get compatibility issues.

jgunthorpe:
"properly" for a plugin does not mean the libtool 1.0.0 junk, that is useless.

Honggang-LI:
> "properly" for a plugin does not mean the libtool 1.0.0 junk, that is useless.

We still want versioned plugin. Will this PR be rejected or merged?
thanks

jgunthorpe:
You do not need a versioned pluging, I'm not merging things to support wrongly built out of tree plugins. Get the plugin into rdma-core, or fix it to track the rdma-core requirements for building them. That means don't use libtool.

PR #931, opened 2021-01-17
tests: Add device memory tests

EdwardSro:
The series extends the write method of DMMR with an offset parameter and adds new device memory tests that include control and data path.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -4,6 +4,7 @@
 """
 Test module for pyverbs' device module.
 """
+from multiprocessing import Process, Queue
 import unittest
 import resource
 import random
@@ -245,14 +246,17 @@ class DMTest(PyverbsAPITestCase):
     """
     Test various functionalities of the DM class.
     """
+    def setUp(self):
+        super().setUp()
+        _, _, attr_ex = self.devices[0]
+        if attr_ex.max_dm_size == 0:
+            raise unittest.SkipTest('Device memory is not supported')
 
     def test_create_dm(self):
         """
         test ibv_alloc_dm()
         """
         for ctx, attr, attr_ex in self.devices:
-            if attr_ex.max_dm_size == 0:
-                raise unittest.SkipTest('Device memory is not supported')
             dm_len = random.randrange(u.MIN_DM_SIZE, attr_ex.max_dm_size/2,
                                       u.DM_ALIGNMENT)
             dm_attrs = u.get_dm_attrs(dm_len)
@@ -264,8 +268,6 @@ class DMTest(PyverbsAPITestCase):
         test ibv_free_dm()
         """
         for ctx, attr, attr_ex in self.devices:
-            if attr_ex.max_dm_size == 0:
-                raise unittest.SkipTest('Device memory is not supported')
             dm_len = random.randrange(u.MIN_DM_SIZE, attr_ex.max_dm_size/2,
                                       u.DM_ALIGNMENT)
             dm_attrs = u.get_dm_attrs(dm_len)
@@ -277,8 +279,6 @@ class DMTest(PyverbsAPITestCase):
         test ibv_alloc_dm() with an illegal size and comp mask
         """
         for ctx, attr, attr_ex in self.devices:
-            if attr_ex.max_dm_size == 0:
-                raise unittest.SkipTest('Device memory is not supported')
             dm_len = attr_ex.max_dm_size + 1
             dm_attrs = u.get_dm_attrs(dm_len)
             try:
@@ -306,8 +306,6 @@ class DMTest(PyverbsAPITestCase):
         Test calling ibv_free_dm() twice
         """
         for ctx, attr, attr_ex in self.devices:
-            if attr_ex.max_dm_size == 0:
-                raise unittest.SkipTest('Device memory is not supported')
             dm_len = random.randrange(u.MIN_DM_SIZE, attr_ex.max_dm_size/2,
                                       u.DM_ALIGNMENT)
             dm_attrs = u.get_dm_attrs(dm_len)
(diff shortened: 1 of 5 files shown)


tests/test_device.py
```diff
@@ -247,14 +248,17 @@ class DMTest(PyverbsAPITestCase):
     """
     Test various functionalities of the DM class.
     """
+    def setUp(self):
+        super().setUp()
+        _, _, attr_ex = self.devices[0]
```
Kamalheib, line 253: @EdwardSro Any reason why the test needs to run only over the first device in the list?

tests/test_device.py
```diff
@@ -376,3 +368,40 @@ def test_dm_read(self):
...
+    def alloc_dm(self, res_queue, size):
+        """
+        Alloc device memory. Used by multiple processes that allocate DMs in
+        parallel.
+        :param res_queue: Result Queue to return the result to the parent
+                            process.
+        :param size: The DM allocation size.
+        :return: None
+        """
+        ctx, _, _ = self.devices[0]
```
Kamalheib, line 381: same here.

tests/test_device.py
```diff
@@ -376,3 +368,40 @@ def test_dm_read(self):
...
+            dm = d.DM(ctx, d.AllocDmAttr(length=size))
+        except PyverbsError as err:
+            res_queue.put(err.error_code)
+        res_queue.put(0)
+
+    def test_multi_process_alloc_dm(self):
+        """
+        Several processes try to allocate device memory simultaneously.
+        """
+        _, _, attr_ex = self.devices[0]
```
Kamalheib, line 392: same here.

tests/test_mr.py
```diff
@@ -326,43 +356,146 @@ def test_reg_mw_wrong_type(self):
...
 
-class DMMRTest(PyverbsAPITestCase):
+class DeviceMemoryAPITest(PyverbsAPITestCase):
     """
-    Test various functionalities of the DMMR class.
+    Test various API usages of the DMMR class.
     """
+    def setUp(self):
+        super().setUp()
+        _, _, attr_ex = self.devices[0]
```
Kamalheib, line 365: same here.

tests/test_mr.py
```diff
@@ -326,43 +356,146 @@ def test_reg_mw_wrong_type(self):
...
+        if attr_ex.max_dm_size == 0:
+            raise unittest.SkipTest('Device memory is not supported')
+
     def test_create_dm_mr(self):
+        ctx, _, attr_ex = self.devices[0]
```
Kamalheib, line 370: same here.

tests/test_mr.py
```diff
@@ -326,43 +356,146 @@ def test_reg_mw_wrong_type(self):
...
+                with DM(ctx, AllocDmAttr(length=dm_size)) as dm:
+                    DMMR(PD(ctx), dmmr_size, dm_access, dm, 0)
+
+    def test_dm_bad_access(self):
+        """
+        Test multiple types of bad access to the Device Memory. Device memory
+        access requests a 4B alignment. The test tries to access the DM
+        with bad alignment or outside of the allocated memory.
+        """
+        ctx, _, _ = self.devices[0]
```
Kamalheib, line 387: same here.

tests/test_mr.py
```diff
@@ -326,43 +356,146 @@ def test_reg_mw_wrong_type(self):
...
-                for i in range(10):
-                    dm_len = random.randrange(u.MIN_DM_SIZE, attr_ex.max_dm_size/2,
-                                              u.DM_ALIGNMENT)
-                    dm_attrs = u.get_dm_attrs(dm_len)
-                    with d.DM(ctx, dm_attrs) as dm:
-                        dm_mr_len = random.randint(1, dm_len)
-                        dm_mr_offset = random.randint(0, (dm_len - dm_mr_len))
-                        DMMR(pd, dm_mr_len, e.IBV_ACCESS_ZERO_BASED, dm=dm,
-                             offset=dm_mr_offset)
+        ctx, _, _ = self.devices[0]
```
Kamalheib, line 408: same here.

tests/test_device.py
EdwardSro: It doesn't "need" to run on the first device. But running on one device only would save testing time.
Do you use the option of running tests on all devices (by not passing --dev on purpose)?

Kamalheib: I wonder why not you guys use the --dev option? - because if a user would like to run the tests on all devices on the system this will be a bug...

rleon: > I wonder why not you guys use the --dev option? - because if a user would like to run the tests on all devices on the system this will be a bug...

Agree, --dev sets to the correct device

EdwardSro: We do use --dev. But when the user doesn't use it, we choose an arbitrary device (which is the first device).
This is not the case just here. If you look at the API tests of test_cqex, test_mr, test_device, and others, we do the same thing.
Tests that inherit from RDMATestCase though, randomly select the device (if --dev was not passed).
So I have the following options in mind:
1. Keep things as is
2. Change tests inheriting from PyverbsAPITestCase to run on all devices
3. Change  PyverbsAPITestCase behaviour to randomly choose one device (same as RDMATestCase)

I would like to hear your opinions based on your usage.

rleon: if --dev is not provided, the runner (upper logic) should chose default, internal test logic should assume that device is already chosen.

EdwardSro: That's what I meant in the 3rd option, which seems the most reasonable to me.  (I can change RDMATestCase to choose the first device by default instead of randomizing).
I will do it in a different PR, I need to apply this change on the remaining tests.

rleon:
@EdwardSro can you please rebase?

EdwardSro:
> @EdwardSro can you please rebase?

Done

PR #933, opened 2021-01-19
redhat: Remove base package dependency from all sub-packages

Honggang-LI:
Depends on base package is useful for RDMA users. Unfortunately,
it will pull unwanted packages for Cloud or Edge users who may
don't have RDMA hardware.

### DIFF
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -92,7 +92,6 @@ scripts, dracut rules, and the rdma-ndd utility.
 
 %package devel
 Summary: RDMA core development libraries and headers
-Requires: %{name}%{?_isa} = %{version}-%{release}
 Requires: libibverbs%{?_isa} = %{version}-%{release}
 Provides: libibverbs-devel = %{version}-%{release}
 Obsoletes: libibverbs-devel < %{version}-%{release}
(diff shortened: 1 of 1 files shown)


rleon:
Can you please remind me why they are installing rdma-core if they don't have RDMA hardware?

Honggang-LI:
> Can you please remind me why they are installing rdma-core if they don't have RDMA hardware?

For example, an application may support socket and verbs at the same time, which means the application
may be compiled with libibverbs. Then the application depends on sub-package libibverbs. When the sub-package
libibverbs depends on base package rdma-core, the rdma-core package will be installed as dependency of libibverbs.

rleon:
> > Can you please remind me why they are installing rdma-core if they don't have RDMA hardware?
> 
> For example, an application may support socket and verbs at the same time, which means the application
> may be compiled with libibverbs. Then the application depends on sub-package libibverbs. When the sub-package
> libibverbs depends on base package rdma-core, the rdma-core package will be installed as dependency of libibverbs.

Thanks

jgunthorpe:
rdma-core is just the boot support in RHEL, why is installing it a problem? I agree libraries and devel shouldn't pull it in, but stuff like ibacm should

Honggang-LI:
> rdma-core is just the boot support in RHEL, why is installing it a problem? 

rdma-core will pull packages it depends on for Cloud and Edge users who don't have RDMA hardware and want minimal image.

> I agree libraries and devel shouldn't pull it in, but stuff like ibacm should

Let the RDMA user explicitly install rdma-core when they do want RDMA support.

jgunthorpe:
I'm not as worried about this as I once was.. We have a kernel solution for the module loading problem now, and verbs also triggers module autoloading. The missing autoloading peice from a library perspective is umad. Otherwise it is just the kernel ULPs that need the boot support: srp, nfs and so on.

PR #934, opened 2021-01-19
libhns: Add support for UD inline

Li-Weihang:
HIP08_C supports ud inline up to size of 1024 Bytes. When data size is smaller than 8 bytes, they will be stored in sqwqe. Otherwise, the data will be filled into extended sges.


Li-Weihang:
The dependency on cap_flags in hns-abi.h has been deleted according to comments from the mailing list, so the 'needs-kernel-patch' label can be removed, thanks.

PR #937, opened 2021-01-22
Add NAME_FIXED option for rdma_rename

edwardsp:
usage:  ```rdma_rename <old-name> NAME_FIXED <new-name>```

### DIFF
--- a/kernel-boot/rdma_rename.c
+++ b/kernel-boot/rdma_rename.c
@@ -27,6 +27,7 @@
  * NAME_PCI - based on PCI/slot/function location
  * NAME_GUID - based on node GUID
  * NAME_ONBOARD - based on-board device index
+ * NAME_FIXED - rename the device to the fixed named in the next argument
  *
  * The stable names are combination of device type technology and rename mode.
  * Infiniband - ib*
@@ -483,6 +484,19 @@ out:
 	return 0;
 }
 
+static int set_fixed_name(struct data *d, char *name)
+{
+	int ret;
+
+	ret = asprintf(&d->name, "%s", name);
+	if (ret == -1) {
+		d->name = NULL;
+		return -ENOMEM;
+	}
+
+	return 0;
+}
+
 static int device_rename(struct nl_sock *nl, struct data *d)
 {
 	struct nlmsghdr *hdr;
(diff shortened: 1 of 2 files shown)


rleon (changes requested):
Please add this mode in the first block of this file where all possible modes are listed.
Also please add description to the commit message too.

It will be also nice to add commented example line to our udev rule.

Thanks, it is very welcomed change.

kernel-boot/rdma_rename.c
```diff
@@ -483,6 +483,19 @@ static int by_guid(struct data *d)
 	return 0;
 }
 
+static int set_fixed_name(struct data *d, char *name)
+{
+	int ret = -1;
```
rleon, line 488: No need in assignment, you are overwriting it.

kernel-boot/rdma_rename.c
```diff
@@ -629,6 +650,8 @@ int main(int argc, char **argv)
 		ret = by_pci(&d);
 	if (ret && (np & NAME_GUID))
 		ret = by_guid(&d);
+	if (ret && (np & NAME_FIXED))
```
rleon, line 653: It is not correct if (..), if everything above failed, we will execute this section which is not what we want.

You need to place NAME_FIXED outside of this if .. block.

edwardsp: I was just trying to keep with the convention of the current code.  NAME_FIXED can only be set if the other options aren't set (in the current implementation).  By "and"ing with the return value we are just setting an order of precedence if this ever were to change.  Would you prefer it to be, `if (np & NAME_FIXED)`?  (it should be checked for the `if(ret) goto out;` part)

kernel-boot/rdma_rename.c
edwardsp: thanks, I've updated it

kernel-boot/rdma_rename.c
rleon: Please squash all commits into one and push with "-f" to this topic.

The "ret ... " scheme is needed for the fall-back, but you are right, we can leave it here.
Sorry for my mistake.

edwardsp: I've squashed the commits.  Thanks

rleon: 1. Your commit message is still empty.
2. Update and add an example to kernel-boot/rdma-persistent-naming.rules as a comment, something like this:
#ACTION=="add", SUBSYSTEM=="infiniband", PROGRAM="rdma_rename %k NAME_FIXED static_name"

edwardsp: I've updated both.  Thanks!

rleon:
Thanks

PR #941, opened 2021-02-04
verbs: Fix possible port loop overflow

yishaih:
This series includes two fixes to prevent a possible port loop overflow when calling ibv_query_port().

### DIFF
--- a/libibverbs/examples/devinfo.c
+++ b/libibverbs/examples/devinfo.c
@@ -198,7 +198,7 @@ static void print_formated_gid(union ibv_gid *gid, int i,
 
 static int print_all_port_gids(struct ibv_context *ctx,
 			       struct ibv_port_attr *port_attr,
-			       uint8_t port_num)
+			       uint32_t port_num)
 {
 	enum ibv_gid_type_sysfs type;
 	union ibv_gid gid;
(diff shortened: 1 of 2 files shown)


Hakon-Bugge:
-Wextra detects this:

qq.c: In function ‘main’:
qq.c:6:5: warning: comparison is always true due to limited range of data type [-Wtype-limits]
     for (p = 1; p <= 255; p++)
     ^

Should we aim at compiling with -Wextra?

That sai,

LGTM.

mark-bloch:
> -Wextra detects this:
> 
> qq.c: In function ‘main’:
> qq.c:6:5: warning: comparison is always true due to limited range of data type [-Wtype-limits]
> for (p = 1; p <= 255; p++)
> ^
> 
> Should we aim at compiling with -Wextra?
> 
> That said,
> 
> LGTM.

We already set it: https://github.com/linux-rdma/rdma-core/blob/master/CMakeLists.txt#L224
the issue is that we use phys_port_cnt and not the hardcoded value of 255.
it seems gcc can't/doesn't check based on max value of the variable, maybe there is an option for that?

Hakon-Bugge:
> We already set it: https://github.com/linux-rdma/rdma-core/blob/master/CMakeLists.txt#L224
> the issue is that we use phys_port_cnt and not the hardcoded value of 255.
> it seems gcc can't/doesn't check based on max value of the variable, maybe there is an option for that?

I did not see any option for that. A min/max analysis should have detected this.

libibverbs/examples/devinfo.c
```diff
@@ -608,7 +608,7 @@ static int print_hca_cap(struct ibv_device *ib_dev, uint8_t ib_port)
 			      device_attr.max_dm_size);
 	}
 
-	for (port = 1; port <= device_attr.orig_attr.phys_port_cnt; ++port) {
+	for (port = 0; port++ < device_attr.orig_attr.phys_port_cnt;) {
```
jgunthorpe, line 611: Don't do gross stuff like this, just make the port unsigned int.

yishaih: OK, the PR was updated accordingly.

PR #952, opened 2021-03-01
Add options to conditionalize building the docs

bstinsonmhk:
We want to do this in case the OS does not ship pandoc.

Signed-off-by: Brian Stinson <bstinson@redhat.com>

### DIFF
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -1,3 +1,15 @@
+%if 0%{?fedora} >= 25 || 0%{?rhel} == 8
+# Fedora and RHEL 8 have pandoc, so we can build documentation
+%bcond_without builddocs
+%else
+# RHEL > 8 does not ship pandoc, so we don't want to try to build documentation directly
+%bcond_with builddocs
+%endif
+
+# Flip this to 1 to if you're building with prebuilt docs. If you want to do
+# this on Fedora you probably want to pass --without builddocs to rpmbuild
+%define prebuilt_docs 0
+
 Name: rdma-core
 Version: 34.0
 Release: 1%{?dist}
@@ -80,8 +92,10 @@ BuildRequires: make
 %define cmake_install DESTDIR=%{buildroot} make install
 %endif
 
-%if 0%{?fedora} >= 25 || 0%{?rhel} >= 8
-# pandoc was introduced in FC25, Centos8
+%if %{with builddocs}
+# pandoc was introduced in FC25, Building on Fedora defaults to building with
+# docs. Building on Enterprise Linux defaults to building without docs.
+
 BuildRequires: pandoc
 %endif
 
@@ -301,6 +315,9 @@ easy, object-oriented access to IB verbs.
          -DCMAKE_INSTALL_UDEV_RULESDIR:PATH=%{_udevrulesdir} \
          -DCMAKE_INSTALL_PERLDIR:PATH=%{perl_vendorlib} \
          -DENABLE_IBDIAGS_COMPAT:BOOL=True \
+%if ! %{with builddocs} 
+         -DNO_MAN_PAGES=1 \
+%endif
 %if %{with_static}
          -DENABLE_STATIC=1 \
 %endif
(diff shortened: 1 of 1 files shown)


redhat/rdma-core.spec
```diff
@@ -301,6 +315,9 @@ easy, object-oriented access to IB verbs.
          -DCMAKE_INSTALL_UDEV_RULESDIR:PATH=%{_udevrulesdir} \
          -DCMAKE_INSTALL_PERLDIR:PATH=%{perl_vendorlib} \
          -DENABLE_IBDIAGS_COMPAT:BOOL=True \
+%if ! %{with builddocs} 
+         -DNO_MAN_PAGES=1 \
```
jgunthorpe, line 319: No, package build scripts can never include NO_MAN_PAGES - shipping rdma-core without man pages is completely unacceptable.

jgunthorpe (changes requested):

redhat/rdma-core.spec
bstinsonmhk: What if we switched the bcondition at the top to always build docs by default? The problem I'm trying to solve here is to not require pandoc in downstream specs (which are already shipping prebuilt docs anyways), while allowing us to do test builds against the upstream spec too.

jgunthorpe: The cmake automatically uses the prebuilt docs, there is no need to change the specfile at all.

bstinsonmhk: I'm trying to remove the BuildRequires on pandoc from the spec. So to ask my question another way: is there some way I can use this upstream spec to build a non-production build on an OS that does not ship pandoc?

jgunthorpe: Remove the buildrequires and everything will work if your source tar has the correct pre-built man pages in it.

Honggang-LI: > Remove the buildrequires and everything will work if your source tar has the correct pre-built man pages in it.

**_NO_**, this won't work for source tarball generated from upstream git repo. We **always** have to test patch against latest
upstream repo.

nmorey: For SUSE we use the git repo but pre-build the doc before submitting to our build system.
It adds a step but solved the pandoc dependency issue for us.

jgunthorpe: @Honggang-LI if it doesn't work then the source tgz you are using doesn't have docs prebuilt and, again, complete NO to distros shipping without docs. You have to use tgz with prebuilt doc, and if you do then no changes to the spec file are required. See Nichola's note for how suse does it.

nmorey: @Honggang-LI The spec file and the gen-pandoc.sh script can be found here if you want to have a look:
https://build.opensuse.org/package/show/science:HPC/rdma-core

Honggang-LI:
@rleon 

We need a script to build a cannonical distribution tar from pure source. As the PR is closed now, is the script available for us?

nmorey:
@Honggang-LI The release script from AZP is available in `buildlib/azure-pipelines-release.yml`

rleon:
I closed it because this PR is not going to be merged.

Honggang-LI:
> @Honggang-LI The release script from AZP is available in `buildlib/azure-pipelines-release.yml`

Well, I don't have an Azure account to run this pipeline script. We need a docker script.

nmorey:
@Honggang-LI Can't you just extract the script part and package it however you need it ?
```
$ python3 -c "import yaml; print(yaml.safe_load(open('buildlib/azure-pipelines-release.yml'))['stages'][0]['jobs'][0]['steps'][1]['bash'])"
set -e
mkdir build-pandoc artifacts
cd build-pandoc
CC=gcc-10 cmake -GNinja ..
ninja docs

cd ..
python3 buildlib/cbuild make-dist-tar build-pandoc

```

Honggang-LI:
> @Honggang-LI Can't you just extract the script part and package it however you need it ?

We need docker container script. Your suggestion will not work for RHEL9 users, as pandoc will not available for RHEL9.

nmorey:
@Honggang-LI I get that.
But can't you use something like the script above to build a "release" tarball from any commit in the git (in a CI tool for example, on any distro that has pandoc). And then use this tarball to run whatever you need to on a RHEL9.

PR #965, opened 2021-03-20
pyverbs: Add support for null memory regions.

TaranovK:
The patch adds support for null MRs for pyverbs ~~as well as for rdmacm~~.
~~The patch also adds man pages for rdmacm.~~

### DIFF
--- a/pyverbs/mr.pyx
+++ b/pyverbs/mr.pyx
@@ -33,7 +33,7 @@ cdef class MR(PyverbsCM):
     it is done in close().
     """
     def __init__(self, creator not None, length=0, access=0, address=None,
-                 implicit=False, **kwargs):
+                 implicit=False, null_mr=False, **kwargs):
         """
         Allocate a user-level buffer of length <length> and register a Memory
         Region of the given length and access flags.
@@ -47,6 +47,7 @@ cdef class MR(PyverbsCM):
                         provided, a memory will be allocated in the class
                         initialization.
         :param implicit: Implicit the MR address.
+        :param null_mr: Allocate a null memory region.
         :param kwargs: Arguments:
             * *handle*
                 A valid kernel handle for a MR object in the given PD (creator).
@@ -64,6 +65,10 @@ cdef class MR(PyverbsCM):
             # In order to safely cast addr to void*, it is firstly cast to uintptr_t.
             self.buf = <void*><uintptr_t>address
 
+        if null_mr:
+            self.is_user_addr = True
+            self.buf = NULL
+
         mr_handle = kwargs.get('handle')
         # If a MR handle is passed import MR and finish
         if mr_handle is not None:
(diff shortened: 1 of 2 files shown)


rleon:
Why do we need it in rdma-cm?

TaranovK:
> Why do we need it in rdma-cm?

It was a side effect of implementing is_null in pyverbs. Another solution is simply raise an error in pyverb if a client used rdmacm.
To be honest, I never understood why rdmacm implements rdma_reg_* in the first place, and why somebody reused that useless API in pyverbs. 

If you want I can rewrite the commit and remove changes to rdmacm.

rleon:
> > Why do we need it in rdma-cm?
> 
> It was a side effect of implementing is_null in pyverbs. Another solution is simply raise an error in pyverb if a client used rdmacm.
> To be honest, I never understood why rdmacm implements rdma_reg_* in the first place, and why somebody reused that useless API in pyverbs.
> 
> If you want I can rewrite the commit and remove changes to rdmacm.

I'll leave to @EdwardSro to comment about pyverbs, but regarding rdma-cm, please remove.
We don't want to add API that not in use.

Thanks

pyverbs/mr.pyx
```diff
@@ -95,15 +100,19 @@ cdef class MR(PyverbsCM):
             memset(self.buf, 0, length)
         if isinstance(creator, PD):
             pd = <PD>creator
-            if implicit:
+            if null_mr:
+                self.mr = v.ibv_alloc_null_mr(pd.pd)
```
EdwardSro, line 104: why not to add this check at the beginning (right after line 60), and after successfully allocating a null mr you can just return. IMO it'd be more clear.

pyverbs/mr.pyx
```diff
@@ -95,15 +100,19 @@ cdef class MR(PyverbsCM):
...
                 self.mr = v.ibv_reg_mr(pd.pd, NULL, SIZE_MAX, access)
             else:
                 self.mr = v.ibv_reg_mr(pd.pd, self.buf, length, access)
             self.pd = pd
             pd.add_ref(self)
         elif isinstance(creator, CMID):
             cmid = <CMID>creator
-            if access == e.IBV_ACCESS_LOCAL_WRITE:
+            if null_mr:
+                self.mr = NULL
```
EdwardSro, line 114: I think it's better to just remove this check, and add an explanation to the docstring, something like "Allocate a null memory region (valid only with PD creator)"

EdwardSro (changes requested):

EdwardSro:
> To be honest, I never understood why rdmacm implements rdma_reg_* in the first place, and why somebody reused that useless API in pyverbs.

They were added to Pyverbs as part of other rdmacm APIs coverage and test flows.
But I agree that they're useless...

rleon:
What is the status of this PR?

Thanks

TaranovK:
Hi,

Our team decided not to continue working on this issue, as we fully switched to C/C++ in our project.  

Best,
Konstantin

PR #966, opened 2021-03-20
Set correct errno in rdma_get_request

dmitrygx:
Fix `rdma_get_request()` function according to `rdma_get_request` man:
```
RETURN VALUE
       Returns 0 on success, or -1 on error.  If an error occurs, errno will be set to indicate the failure reason.”
```
If event is `RDMA_CM_EVENT_REJECTED`, set `ECONNREFUSED` to errno.
If event's status is set to any value (i.e. a non-zero value), set errno to `-event->status`.

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -1861,8 +1861,13 @@ int rdma_get_request(struct rdma_cm_id *listen, struct rdma_cm_id **id)
 	if (ret)
 		return ret;
 
+	if (event->event == RDMA_CM_EVENT_REJECTED) {
+		ret = ERR(ECONNREFUSED);
+		goto err;
+	}
+
 	if (event->status) {
-		ret = ERR(event->status);
+		ret = ERR(-event->status);
 		goto err;
 	}
 


librdmacm/cma.c
```diff
@@ -1861,8 +1861,13 @@ int rdma_get_request(struct rdma_cm_id *listen, struct rdma_cm_id **id)
 	if (ret)
 		return ret;
 
+	if (event->event == RDMA_CM_EVENT_REJECTED) {
+		ret = ERR(ECONNREFUSED);
```
rleon, line 1865: This hunk is ok

librdmacm/cma.c
```diff
@@ -1861,8 +1861,13 @@ int rdma_get_request(struct rdma_cm_id *listen, struct rdma_cm_id **id)
...
+		goto err;
+	}
+
 	if (event->status) {
-		ret = ERR(event->status);
+		ret = ERR(-event->status);
```
rleon, line 1870: This is not.

ERR(positive_number)

dmitrygx: Why it is not correct?
It sets `errno` to negative value
The fix is to convert negative value to positive one

rleon: Because ERR() macro receives positive numbers
For example, ENODEV above is positive too.
/usr/include/asm-generic/errno-base.h:#define	ENODEV		19	/* No such device */

dmitrygx: `ERR()` receives `int` which can be +, -, 0
`event->status` is set to negative errno (i.e. `-errno`):
https://github.com/linux-rdma/rdma-core/blob/46b84953153c1fd7ed8e7c20a6df04054cb1e4f9/librdmacm/man/rdma_get_cm_event.3#L46
So, we should convert negative value to positive one when setting `errno`, because a user doesn't expect to get negative value from `errno`

rleon: Excellent, so write it in commit message

dmitrygx: @rleon updated

PR #967, opened 2021-03-20
Return GAI-specific error codes to fix return value and errno

dmitrygx:
Return GAI-specific error codes to fix return value and errno.
If  `rdma_getaddrinfo`->`ucma_getaddrinfo()`->`getaddrinfo()` flow fails with an error, it returns some GAI-specific error and sets errno only if `EAI_SYSTEM` is a return value.
Align `rdma_getaddrinfo()` with `getaddrinfo()` to return more informative errors which can be handled by users appropriate and `gai_strerror()` can be used to get a human-readable string for a return value. If it fails due to some system error (e.g. memory allocation error), return `EAI_SYSTEM` and set a specific `errno` (e.g. `ENOMEM`) - introduce and use `GAI_ERR()` function.
Fix `rdma_getaddrinfo` man copying `EAI_*` return values description + minor fixes in `rdma_getaddrinfo` man.

### DIFF
--- a/librdmacm/man/rdma_getaddrinfo.3
+++ b/librdmacm/man/rdma_getaddrinfo.3
@@ -26,8 +26,37 @@ Resolves the destination node and service address and returns
 information needed to establish communication.  Provides the
 RDMA functional equivalent to getaddrinfo.
 .SH "RETURN VALUE"
-Returns 0 on success, or -1 on error.  If an error occurs, errno will be
-set to indicate the failure reason.
+Returns 0 on success, or -1 on error (errno will be set to indicate the failure
+reason), or one of the following nonzero error codes:
+.IP "EAI_ADDRFAMILY" 12
+The specified network host does not have any network addresses in the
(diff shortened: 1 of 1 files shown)


rleon:
I afraid that this change will break compatibility with user space applications.

dmitrygx:
> I afraid that this change will break compatibility with user space applications.

Then we should just document GAI return value and don't set errno to GAI return value

rleon:
> > I afraid that this change will break compatibility with user space applications.
> 
> Then we should just document GAI return value and don't set errno to GAI return value

I think so.

dmitrygx:
> > > I afraid that this change will break compatibility with user space applications.
> > 
> > 
> > Then we should just document GAI return value and don't set errno to GAI return value
> 
> I think so.

@rleon thanks! fixed.
no codes updates to fix `errno` - it works fine.
the PR contains only man changes to describe GAI-specific errors.

PR #970, opened 2021-03-21
cmake: Bump minimal cmake version

rleon:
The rdma-core doesn't really need anything specific from 2.8.11, so
simply raise the minimal level to be 2.8.12 and silence the warning
below.

 CMake Deprecation Warning at CMakeLists.txt:7 (cmake_minimum_required):
   Compatibility with CMake < 2.8.12 will be removed from a future version of
   CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.

### DIFF
--- a/infiniband-diags/CMakeLists.txt
+++ b/infiniband-diags/CMakeLists.txt
@@ -16,7 +16,7 @@ add_library(ibdiags_tools STATIC
 function(ibdiag_programs)
   foreach(I ${ARGN})
     rdma_sbin_executable(${I} "${I}.c")
-    target_link_libraries(${I} LINK_PRIVATE ${RT_LIBRARIES} ibumad ibmad ibdiags_tools ibnetdisc)
+    target_link_libraries(${I} LINK_PRIVATE ${RT_LIBRARIES} ibdiags_tools ibumad ibmad ibnetdisc)
   endforeach()
 endfunction()
 
@@ -44,6 +44,6 @@ ibdiag_programs(
   )
 
 rdma_test_executable(ibsendtrap "ibsendtrap.c")
-target_link_libraries(ibsendtrap LINK_PRIVATE ibumad ibmad ibdiags_tools)
+target_link_libraries(ibsendtrap LINK_PRIVATE ibdiags_tools ibumad ibmad)
 rdma_test_executable(mcm_rereg_test "mcm_rereg_test.c")
-target_link_libraries(mcm_rereg_test LINK_PRIVATE ibumad ibmad ibdiags_tools)
+target_link_libraries(mcm_rereg_test LINK_PRIVATE ibdiags_tools ibumad ibmad)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,12 @@
 #      (without man pages) when neither pandoc/rst2man nor the pandoc-prebuilt
 #      directory are available.
 
-cmake_minimum_required(VERSION 2.8.11 FATAL_ERROR)
+if (${CMAKE_VERSION} VERSION_LESS "3.18.1")
+	# Centos 6 support
+	cmake_minimum_required(VERSION 2.8.11 FATAL_ERROR)
+else()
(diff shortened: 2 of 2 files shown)


jgunthorpe:
Woah, IIRC, 2.8.11 was selected because it is what Centos6 ships.

rleon:
This is time to drop centos, most likely FC34 will have incompatible cmake.

rleon:
So, what do we plan to do?

jgunthorpe:
We don't actually need compat with 2.8.11, we don't rely on any special behaviors. Can we just conditionalize it based on the cmake version?

rleon:
> We don't actually need compat with 2.8.11, we don't rely on any special behaviors. Can we just conditionalize it based on the cmake version?

It will look strange, but doable.

rleon:
> > We don't actually need compat with 2.8.11, we don't rely on any special behaviors. Can we just conditionalize it based on the cmake version?
> 
> It will look strange, but doable.

done

jgunthorpe:
Why doesn't it build?

rleon:
I saw the failure but didn't have time to look yet.

rleon:
> It would be nice if we could find a better cmake version that just .12 - check cbuild's matrix and pick the next lowest?

Centos7 uses 2.8.12.2

rleon:
Repushed to check CI

rleon:
Finally passed

libibnetdisc/CMakeLists.txt
```diff
@@ -20,5 +20,6 @@ rdma_pkg_config("ibnetdisc" "libibumad libibmad" "")
 rdma_test_executable(testleaks tests/testleaks.c)
 target_link_libraries(testleaks LINK_PRIVATE
   ibmad
+  ibumad
```
jgunthorpe, line 23: This isn't how things are supposed to work, ibmad shuld be self contained

target_link_libraries(ibmad LINK_PRIVATE
  ibumad
  )

I assume changing the cmake version also changed how some of this stuff works?

infiniband-diags/CMakeLists.txt
```diff
@@ -8,16 +8,13 @@ install(FILES
   etc/ibdiag.conf
   DESTINATION "${IBDIAG_CONFIG_PATH}")
 
-add_library(ibdiags_tools STATIC
```
jgunthorpe, line 11: The library prevents multiple recompilation, why can't we fix the library?

jgunthorpe (changes requested):

libibnetdisc/CMakeLists.txt
rleon: Yes, for some reason SuSE 15.0 didn't build without this change.
See first commit for an error.

infiniband-diags/CMakeLists.txt
rleon: It doesn't worth, I see "fixing" library as optimization of something that no one cares.

libibnetdisc/CMakeLists.txt
rleon: ok, I found the reason to missing -rpath error.
It was caused by CMP0022
https://cmake.org/cmake/help/latest/policy/CMP0022.html
I tried as a workaround the following change
cmake_policy(SET CMP0022 OLD)

and suse built correctly.

jgunthorpe: That makes alot more sense

jgunthorpe: But should we make the update described in that policy document, whatever it is?

rleon: Yes, we need, the cmake_policy() was simplest way to check that we are hitting this specific policy change.

infiniband-diags/CMakeLists.txt
rleon: BTW, I fixed it locall by putting ibdiag_tools as first in the library list.

Just can't push till we fix the suse failure.

PR #973, opened 2021-03-29
ibtroute: Change lft to uint8_t in dump_unicast_tables method

haimbo:
Change lft to unsigned uint8_t in dump_unicast_tables method instead
of char in case we have 128 ports and up.

Signed-off-by: Haim Boozaglo <haimbo@nvidia.com>

### DIFF
--- a/infiniband-diags/ibroute.c
+++ b/infiniband-diags/ibroute.c
@@ -321,7 +321,7 @@ static int dump_lid(char *str, int strlen, int lid, int valid)
 static const char *dump_unicast_tables(ib_portid_t *portid, int startlid,
 				       int endlid)
 {
-	char lft[IB_SMP_DATA_SIZE] = { 0 };
+	uint8_t lft[IB_SMP_DATA_SIZE] = { 0 };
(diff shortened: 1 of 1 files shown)


infiniband-diags/ibroute.c
```diff
@@ -321,7 +321,7 @@ static int dump_lid(char *str, int strlen, int lid, int valid)
 static const char *dump_unicast_tables(ib_portid_t *portid, int startlid,
 				       int endlid)
 {
-	char lft[IB_SMP_DATA_SIZE] = { 0 };
+	unsigned char lft[IB_SMP_DATA_SIZE] = { 0 };
```
jgunthorpe, line 324: Shouldn't this should be uint8_t because it is data from the network?

jgunthorpe (changes requested):

infiniband-diags/ibroute.c
haimbo: Dobe, was changed to uint8_t .

haimbo: Is it ok now?

PR #974, opened 2021-04-04
buildlib: Continue build on old GCC versions without SSE

rleon:
The gcc version 4.8.5 in Centos 7 fails to compile rdma-core with
the following error:

-- Performing Test HAVE_TARGET_SSE
-- Performing Test HAVE_TARGET_SSE - Failed
-- Performing Test NEED_MSSE_FLAG
-- Performing Test NEED_MSSE_FLAG - Failed
CMake Error at buildlib/RDMA_EnableCStd.cmake:128 (message):
  Can not figure out how to turn on sse instructions for i386
  CMakeLists.txt:398 (RDMA_Check_SSE)

Instead of aborting the compilation, compile the library without that flag.

Fixes: bf343ecc783c ("buildlib: Use -msse if the compiler does not support target(sse)")

### DIFF
--- a/buildlib/RDMA_EnableCStd.cmake
+++ b/buildlib/RDMA_EnableCStd.cmake
@@ -58,9 +58,15 @@ function(RDMA_EnableCStd)
   endif()
 endfunction()
 
+function(RDMA_Check_C_Compiles TO_VAR CHECK_PROGRAM)
+  set(CMAKE_REQUIRED_FLAGS "${ARGV2} -Werror")
+  CHECK_C_SOURCE_COMPILES("${CHECK_PROGRAM}" ${TO_VAR})
+  set(${TO_VAR} ${${TO_VAR}} PARENT_SCOPE)
+endfunction()
+
 function(RDMA_Check_Aliasing TO_VAR)
   SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2")
-  CHECK_C_SOURCE_COMPILES("
+  RDMA_Check_C_Compiles(HAVE_WORKING_STRICT_ALIASING "
 struct in6_addr {unsigned int u6_addr32[4];};
 struct iphdr {unsigned int daddr;};
 union ibv_gid {unsigned char raw[16];};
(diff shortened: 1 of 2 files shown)


jgunthorpe:
What? When did this start happening? C7 has been in the CI forever

alialnu:
> 
> 
> What? When did this start happening? C7 has been in the CI forever

It started reproducing when `ninja-build` was updated from `1.7.2-2.el7.x86_64`  to `1.10.2-3.el7.x86_64` in epel-7 recently.

jgunthorpe:
That makes no sense. Please find the root cause and file a bug on ninja if necessary

rleon:
I didn't success to reproduce it manually, only reproduces with cmake and
ninja-build

jgunthorpe:
It needs more than that, every fail_regex needs fixing, I think -Werror is the better approach.

rleon:
> It needs more than that, every fail_regex needs fixing, I think -Werror is the better approach.

Why do we see this failure for SSE only?
My guess is that all other code snippets are actually fail to compile and this is only one that returns warning.

jgunthorpe:
sse is the only one that explodes, the other ones just silently misfunction.

rleon:
ok, pushed new version with -Werror

rleon:
At least it passes on FC33

buildlib/RDMA_EnableCStd.cmake
```diff
@@ -58,9 +58,16 @@ function(RDMA_EnableCStd)
   endif()
 endfunction()
 
+function(RDMA_Check_C_Compiles TO_VAR CHECK_PROGRAM)
+  set(CMAKE_REQUIRED_FLAGS "${ARGV2} -Werror")
+  CHECK_C_SOURCE_COMPILES("${CHECK_PROGRAM}" ${TO_VAR})
+  set(${TO_VAR} ${${TO_VAR}} PARENT_SCOPE)
+  unset(TO_VAR CACHE)
```
rleon, line 65: This line can be dropped

PR #975, opened 2021-04-05
Report when ibv_fork_init() is not needed

galpress:
Followup from #883, this series identifies kernel with copy-on-fork support and reports that through the recently introduced ibv_is_fork_initialized() verb.

Kernel submission:
v1: https://lore.kernel.org/linux-rdma/20210405114722.98904-1-galpress@amazon.com/
v2: https://lore.kernel.org/linux-rdma/20210407101606.80737-1-galpress@amazon.com/
v3: https://lore.kernel.org/linux-rdma/20210412064150.40064-1-galpress@amazon.com/
v4: https://lore.kernel.org/linux-rdma/20210418121025.66849-1-galpress@amazon.com/

### DIFF
--- a/libibverbs/ibdev_nl.c
+++ b/libibverbs/ibdev_nl.c
@@ -217,3 +217,39 @@ err:
 	nl_socket_free(nl);
 	return EINVAL;
 }
+
+static int get_copy_on_fork_cb(struct nl_msg *msg, void *data)
+{
+	struct nlattr *tb[RDMA_NLDEV_ATTR_MAX];
+	int ret;
+
+	ret = nlmsg_parse(nlmsg_hdr(msg), 0, tb, RDMA_NLDEV_ATTR_MAX - 1,
+			  rdmanl_policy);
+	if (ret < 0)
+		return ret;
+
+	/* Older kernels don't support COF and don't report it through nl */
+	if (!tb[RDMA_NLDEV_SYS_ATTR_COPY_ON_FORK]) {
+		*(uint8_t *)data = 0;
+		return NL_OK;
+	}
+
+	*(uint8_t *)data = nla_get_u8(tb[RDMA_NLDEV_SYS_ATTR_COPY_ON_FORK]);
+	return NL_OK;
+}
+
+bool get_copy_on_fork(void)
+{
+	struct nl_sock *nl;
+	uint8_t cof;
+
+	nl = rdmanl_socket_alloc();
+	if (!nl)
+		return false;
+
+	if (rdmanl_get_copy_on_fork(nl, get_copy_on_fork_cb, &cof))
+		cof = false;
+
+	nl_socket_free(nl);
+	return cof;
+}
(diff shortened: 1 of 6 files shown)


libibverbs/ibdev_nl.c
```diff
@@ -217,3 +217,41 @@ int find_sysfs_devs_nl(struct list_head *tmp_sysfs_dev_list)
...
+	if (!tb[RDMA_NLDEV_SYS_ATTR_COPY_ON_FORK]) {
+		*(uint8_t *)data = 0;
+		return NL_OK;
+	}
+
+	*(uint8_t *)data = nla_get_u8(tb[RDMA_NLDEV_SYS_ATTR_COPY_ON_FORK]);
+	return NL_OK;
+}
+
+bool get_copy_on_fork(void)
```
mark-bloch, line 241: Do you except it to change during the RDMA device's lifetime (kernel patch always reports it as supported)?
Seems kind of wasteful to create a socket -> talk with the kernel -> close socket
on every call to ibv_is_fork_initialized() if the result never changes.

Maybe store it (like we do with ibdev_idx and such) while in find_sysfs_devs_nl_cb()?

libibverbs/ibdev_nl.c
```diff
@@ -217,3 +217,41 @@ int find_sysfs_devs_nl(struct list_head *tmp_sysfs_dev_list)
...
+{
+	struct nl_sock *nl;
+	uint8_t cof;
+
+	nl = rdmanl_socket_alloc();
+	if (!nl)
+		return false;
+
+	if (rdmanl_get_copy_on_fork(nl, get_copy_on_fork_cb, &cof)) {
+		nl_socket_free(nl);
```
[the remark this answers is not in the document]
galpress: Hey Mark 😃 
Sure.

libibverbs/ibdev_nl.c
galpress: No, it's not expected to change.
On the other hand, I don't expect this function to get called more than once/twice on init time.

galpress:
I'm trying to think whether we should change the behavior of ibv_fork_init() in case cof is supported (do nothing and return success? return an error?).
Curious to hear other people thoughts about it.

galpress:
Gentle reminder.

rleon (approved):

galpress:
Thanks Leon!

> I'm trying to think whether we should change the behavior of ibv_fork_init() in case cof is supported (do nothing and return success?).
> Curious to hear other people thoughts about it.

This questions remains open.

rleon:
> Thanks Leon!
> 
> > I'm trying to think whether we should change the behavior of ibv_fork_init() in case cof is supported (do nothing and return success?).
> > Curious to hear other people thoughts about it.
> 
> This question remains open.

I don't know what such change will give us.

Thanks

galpress:
This could allow existing apps that don't use ibv_is_fork_initialized() to benefit from the kernel's cof, even if they called ibv_fork_init().

rleon:
This answer is too vague.

Do you expect any performance improvement? Will it improve any runtime/initialization parameter significantly?

galpress:
It should improve performance for cases where registration is done in data path such as MPI, as it will skip the tree traversal and kernel calls.

I had some numbers about the fork overhead with huge pages back at: https://github.com/linux-rdma/rdma-core/pull/767.

krehm:
In my case it's not a performance question, it's a way to prevent segfaults when users specify environment variables that enable ibv_fork_init().

PR #977, opened 2021-04-11
mlx5: Add a DV query port API

yishaih:
Introduce a mlx5dv query port API to supply the user with device specific attributes.
Those attributes are to be used with DEVX resources and direct rules.

The series includes also some pyverbs stuff on top of the new added API.


kernel-headers/rdma/mlx5_user_ioctl_cmds.h
```diff
@@ -41,6 +41,25 @@ enum mlx5_ib_create_flow_action_attrs {
 	MLX5_IB_ATTR_CREATE_FLOW_ACTION_FLAGS = (1U << UVERBS_ID_NS_SHIFT),
 };
 
+enum mlx5_ib_dm_methods {
```
jgunthorpe, line 44: Make sure this is regenerated, this shouldn't have the dm stuff..

yishaih: Right, done

PR #979, opened 2021-04-12
pyverbs: Support mlx5 steering and add tests

EdwardSro:
This series introduces mlx5 dv and direct rules (dr) steering to Pyverbs.
It includes the ability to create mlx5dv flow, flow matchers, and actions such as "got to QP" and reformat.
In addition, it supports creating dr domains, tables, matchers, and rules.
Tests that cover different traffic flows are also added.


rleon:
It needs to be rebased

EdwardSro:
@rleon is there a reason for not merging this PR yet?

PR #984, opened 2021-04-20
verbs: Misc fixes

yishaih:
This series includes few small fixes in both verbs and mlx drivers.

### DIFF
--- a/providers/mlx5/dr_domain.c
+++ b/providers/mlx5/dr_domain.c
@@ -343,17 +343,19 @@ mlx5dv_dr_domain_create(struct ibv_context *ctx,
 	list_head_init(&dmn->tbl_list);
 
 	ret = pthread_spin_init(&dmn->info.rx.lock, PTHREAD_PROCESS_PRIVATE);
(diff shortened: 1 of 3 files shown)


providers/mlx5/dr_domain.c
```diff
@@ -344,7 +344,7 @@ mlx5dv_dr_domain_create(struct ibv_context *ctx,
 
 	ret = pthread_spin_init(&dmn->info.rx.lock, PTHREAD_PROCESS_PRIVATE);
 	if (!ret)
-		pthread_spin_init(&dmn->info.tx.lock, PTHREAD_PROCESS_PRIVATE);
+		ret = pthread_spin_init(&dmn->info.tx.lock, PTHREAD_PROCESS_PRIVATE);
```
mark-bloch, line 347: Shouldn't pthread_spin_destroy() be called here? (on both locks tx and rx).

yishaih: Thanks Mark, the patch was updated to handle that.

PR #986, opened 2021-04-20
cbuild: Avoid permission denied error due to SElinux

rleon:
When using SELinux for controlling processes within a container,
we need to make  sure any content that gets volume mounted into
the container is readable.

Use :Z option for volume mounts, so docker will set correct labels.

This change fixes the error below:
➜  rdma-core git:(master) buildlib/cbuild make centos7
CMake Error: The source directory "/home/leonro/src/rdma-core" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
....

### DIFF
--- a/buildlib/cbuild
+++ b/buildlib/cbuild
@@ -977,7 +977,7 @@ def cmd_make(args):
 
         home = os.path.join(os.path.sep,"home",os.getenv("LOGNAME"));
 
-        dirs = [os.getcwd(),"/tmp"];
+        dirs = [os.getcwd()];
         # Import the symlink target too if BUILD is a symlink
         BUILD_r = os.path.realpath(BUILD);
         if not BUILD_r.startswith(os.path.realpath(SRC)):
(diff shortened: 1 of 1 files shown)


jgunthorpe:
Why doesn't /tmp need a label? I'm not sure about this, doesn't re-labeling mean the filesystem is changed?

rleon:
I tried with /tmp and it didn't accept ":Z", my explanation is that it is not part of SElinux.

And it is not relabeling but actually call to the following command inside docker:
 chcon -Rt svirt_sandbox_file_t -l s0:c1,c2 /path/to/dir
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/container_security_guide/docker_selinux_security_policy

jgunthorpe:
Doesn't that command change the filesystem? Does this work on non-selinux systems? I wonder if we should just give up and make this stuff use podman

rleon:
> Doesn't that command change the filesystem? Does this work on non-selinux systems? I wonder if we should just give up and make this stuff use podman

I don't know about other OSes, but on my plain FC 33 cbuild doesn't work.

jgunthorpe:
I'm not really sure about this, it touches the filesystem to write the labels.. This is why people turn selinux off

rleon:
ok, I disabled SElinux and it fixed permission error. 

Let's keep this PR maybe more people will complain, because disabling SElinux is not a solution but a workaround.

PR #985, opened 2021-04-20
redhat: Add missing rdma-ndd binary to initramfs

pjgeorg:
The dracut modules-setup.sh for Red Hat installs rdma-ndd.service, but misses to install the required rdma-ndd binary. This causes the service to fail:
```
systemd[1]: Starting RDMA Node Description Daemon...
systemd[2512]: rdma-ndd.service: Failed to execute command: No such file or directory
systemd[2512]: rdma-ndd.service: Failed at step EXEC spawning /usr/sbin/rdma-ndd: No such file or directory
systemd[1]: rdma-ndd.service: Main process exited, code=exited, status=203/EXEC
systemd[1]: rdma-ndd.service: Failed with result 'exit-code'.
systemd[1]: Failed to start RDMA Node Description Daemon.
```

Fix by also installing the missing rdma-ndd binary.

Note: The suse/module-setup.sh is not affected by this issue as it does not install the rdma-ndd.service

PR #987, opened 2021-04-20
redhat: Do not add rdma-ndd to initramfs

pjgeorg:
See #985 for further details.

### DIFF
--- a/redhat/rdma.modules-setup.sh
+++ b/redhat/rdma.modules-setup.sh
@@ -21,11 +21,10 @@ install() {
 	inst /usr/lib/modprobe.d/libmlx4.conf
 	inst_multiple lspci setpci awk sleep
 	inst_multiple -o /etc/modprobe.d/mlx4.conf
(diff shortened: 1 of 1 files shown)


Honggang-LI:
Looks good for me. Thanks

rleon:
Please add Fixes line and all the needed details into the commit message. The change should be explained sufficient enough to understand without looking on github webpage.

pjgeorg:
Added further information to the commit message as requested.

rleon:
Thanks

nmorey:
The commit does not have the proper Fixes: tag 

@pjgeorg  What Leon and I mean when talking about Fixes line/tag is something like this:
`Fixes: 8bb38f6cb1b2 (redhat: update dracut setting)`
See 3166f3f3d557c6053a826830066993d50564e929 for example.

This tag can be used by tools to detect when and where patches need to be backported among other things.
With such a tag, my tool would have picked up your fix to be applied almost automatically.
In your case, I did see the github chat and will do it manually.

pjgeorg:
@nmorey Sorry, I didn't know about this particular form of Fixes line/tag. Seems like either none of the open source projects I contributed to so far (just a few) used this format or no one ever bothered to tell me.
Thanks for informing me about this workflow and sorry for any inconveniences caused.

rleon:
It's ok, we are not so picky for one time contributors :)

PR #988, opened 2021-04-21
tests: Avoid running the test_efa* tests over non EFA devices

Kamalheib:
To avoid the failures that could happen when running EFA tests over non
EFA devices, Change the tests to skip if they run over non EFA devices.


galpress:
Did you encounter such issues or is it a theoretical? I thought we return -EOPNOTSUPP where needed so the tests would skip anyway. The same way the mlx5 specific tests get skipped.

tests/efa_base.py
```diff
@@ -13,6 +13,19 @@
...
 
+AMAZON_VENDOR_ID = 0x1d0f
+
+
+def is_efa_dev(ctx):
+    dev_attrs = ctx.query_device()
+    return dev_attrs.vendor_id == AMAZON_VENDOR_ID
+
+
+def skip_if_not_efa_dev(ctx):
```
galpress, line 24: Maybe we can make that into a decorator somehow and mark the EFA tests with it?

tests/test_efadv.py
```diff
@@ -27,6 +28,7 @@ def test_efadv_query(self):
         """
         Verify that it's possible to read EFA direct-verbs.
         """
+        efa_base.skip_if_not_efa_dev(self.ctx)
```
galpress, line 31: I prefer we add a setUp function that verifies the device (like you did in test_efa_srd) instead of coding this in every test case.

galpress (changes requested):

rleon:
General note, I asked from Edward to come with global solution of running only tests that specific to the device on which they run.

For example, my machine with cx6 shouldn't execute EFA tests at all, even print their existence.

[leonro@vm ~]$ lspci |grep Ethernet
00:01.0 Ethernet controller: Red Hat, Inc. Virtio network device
00:09.0 Ethernet controller: Mellanox Technologies MT28908 Family [ConnectX-6]
[leonro@vm ~]$ ./src/rdma-core/build/bin/run_tests.py -v -k efa
test_qp_ex_srd_old_send (tests.test_efa_srd.QPSRDTestCase) ... skipped 'Extended SRD QP is not supported on this device'
test_qp_ex_srd_old_send_imm (tests.test_efa_srd.QPSRDTestCase) ... skipped 'Extended SRD QP is not supported on this device'
test_qp_ex_srd_rdma_read (tests.test_efa_srd.QPSRDTestCase) ... skipped 'Extended SRD QP is not supported on this device'
test_qp_ex_srd_send (tests.test_efa_srd.QPSRDTestCase) ... skipped 'Extended SRD QP is not supported on this device'
test_qp_ex_srd_send_imm (tests.test_efa_srd.QPSRDTestCase) ... skipped 'Extended SRD QP is not supported on this device'
test_efadv_query_ah (tests.test_efadv.EfaAHTest)
Test efadv_query_ah() ... skipped 'Not supported on non EFA devices'
test_efadv_create_qp_ex (tests.test_efadv.EfaQPExTest)
Test efadv_create_qp_ex() ... skipped 'Create SRD QPEx is not supported'
test_efadv_create_driver_qp (tests.test_efadv.EfaQPTest)
Test efadv_create_driver_qp() ... skipped 'Create SRD QP is not supported'
test_efadv_query (tests.test_efadv.EfaQueryDeviceTest)
Verify that it's possible to read EFA direct-verbs. ... skipped 'Not supported on non EFA devices'
test_default_allocators_rc_traffic (tests.test_parent_domain.ParentDomainTrafficTest) ... ok

Ran 10 tests in 1.024s

OK (skipped=9)

Kamalheib:
> Did you encounter such issues or is it theoretical? I thought we return -EOPNOTSUPP where needed so the tests would skip anyway. The same way the mlx5 specific tests get skipped.

As you can see below, I found this issue while trying to run the pyverbs tests over ib_qib.

I don't see any reasonable reason to run efa/mlx5 specific tests over other providers, that is why I'm pushing this MR and I plan to do the same for the test_mlx5* tests, especially that you guys don't have all providers hardware to verify that the efa/mlx5 specific tests pass overall them.

[root@rdma-qe-08 rdma-core]$ ./build/bin/run_tests.py -v --dev ibp5s0 -k test_efadv
test_efadv_query_ah (tests.test_efadv.EfaAHTest)
Test efadv_query_ah() ... ERROR
test_efadv_create_qp_ex (tests.test_efadv.EfaQPExTest)
Test efadv_create_qp_ex() ... skipped 'Create SRD QPEx is not supported'
test_efadv_create_driver_qp (tests.test_efadv.EfaQPTest)
Test efadv_create_driver_qp() ... skipped 'Create SRD QP is not supported'
test_efadv_query (tests.test_efadv.EfaQueryDeviceTest)
Verify that it's possible to read EFA direct-verbs. ... skipped 'Not supported on non EFA devices'

ERROR: test_efadv_query_ah (tests.test_efadv.EfaAHTest)
Test efadv_query_ah()
  File "/home/rdma-core/tests/test_efadv.py", line 60, in test_efadv_query_ah
    raise ex
  File "/home/rdma-core/tests/test_efadv.py", line 53, in test_efadv_query_ah
    ah = efa.EfaAH(pd, attr=ah_attr)
  File "addr.pyx", line 410, in pyverbs.addr.AH.__init__
pyverbs.pyverbs_error.PyverbsRDMAError: Failed to create AH. Errno: 22, Invalid argument

Ran 4 tests in 0.027s

FAILED (errors=1, skipped=3)
-bash: __git_ps1: command not found
[root@rdma-qe-08 rdma-core]$ ibv_devinfo -d ibp5s0
hca_id:	ibp5s0
	transport:			InfiniBand (0)
	fw_ver:				0.0.0
	node_guid:			0011:7500:006f:33c6
	sys_image_guid:			0011:7500:006f:33c6
	vendor_id:			0x1175
	vendor_part_id:			29474
	hw_ver:				0x2
	board_id:			InfiniPath_QLE7342
	phys_port_cnt:			2
		port:	1
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		4096 (5)
			sm_lid:			13
			port_lid:		45
			port_lmc:		0x00
			link_layer:		InfiniBand

		port:	2
			state:			PORT_ACTIVE (4)
			max_mtu:		4096 (5)
			active_mtu:		4096 (5)
			sm_lid:			1
			port_lid:		9
			port_lmc:		0x00
			link_layer:		InfiniBand

galpress:
I'm fine with not running EFA tests on other devices, although I think this failure might indicate a real issue where we don't return the proper return value.

tests/test_efadv.py
Kamalheib: Do you mean add setUp functions to all test classes under test_efadv.py? 

Or, I think it will be better to create two new class "EfaAPITestCase" and "EfaRDMATestCase" under efa_base.py and change all the efa tests to inhirent from them + make sure to implement the setUp function will the check of device/vendor id for skipping the test if the device is not EFA, thoughts?

galpress: I can't see the old version now, but I think you added the EFA device check to the setUp() function in QPSRDTestCase, right?

Why not add a setUp function to EfaQueryDeviceTest as well and perform the check there, instead of in each test case.

Kamalheib: @galpress Please review the latest changes, I think that this approach is much cleaner.

galpress (approved):

EdwardSro:
I'm fine with this patch.
But maybe that specific failure that Kamal has mentioned (test_efadv_query_ah) is indicating another issue?
I see that efadv_query_ah does check if the device is an efa_dev, and if not it returns EOPNOTSUPP, so I would expect that the test would be skipped... 

Not loading device-specific tests on other devices is still planned, I'll try to get to it and provide a clean solution. (but yes, we will lose catching the "not returning EOPNOTSUPP" bugs)

PR #993, opened 2021-04-29
verbs: Fix data races in multithreaded calling ibv_reg_mr

ewanglong:
When calling ibv_reg_mr in valgrind --tool=helgrind or --tool=drd under multihreaded test, there were  data race or conflicting store reported. By changing the variable too_late writting into atomic_store, it can pass the valgrind test.

### DIFF
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -713,7 +713,7 @@ int ibv_dontfork_range(void *base, size_t size)
 	if (mm_root)
 		return ibv_madvise_range(base, size, MADV_DONTFORK);
 	else {
-		too_late = 1;
+		atomic_store(&too_late, 1);
 		return 0;
 	}
 }
@@ -723,7 +723,7 @@ int ibv_dofork_range(void *base, size_t size)
 	if (mm_root)
 		return ibv_madvise_range(base, size, MADV_DOFORK);
 	else {
-		too_late = 1;
+		atomic_store(&too_late, 1);
 		return 0;
 	}
 }


libibverbs/memory.c
```diff
@@ -713,7 +713,7 @@ int ibv_dontfork_range(void *base, size_t size)
 	if (mm_root)
 		return ibv_madvise_range(base, size, MADV_DONTFORK);
 	else {
-		too_late = 1;
+		atomic_store(&too_late, 1);
```
rleon, line 716: too_late is declared as "int", but atomic_store receives _Atomic type.
https://en.cppreference.com/w/c/atomic/atomic_store

jgunthorpe:
Huh? This needs a huge commit message explaining the problem and generally we can't solve races by using atomics like this

rleon:
> Huh? This needs a huge commit message explaining the problem and generally we can't solve races by using atomics like this

It doesn't solve race, but silence valgrind test.

jgunthorpe:
I'm not excited about silencing valgrind by wrongly using atomcis. valgrind is correct to say there is wrong locking here.

janekmi:
This change does not fix the race itself. After this change, it is still an option to miss a memory range that should appear in the `mm_root` tree. Too bad if it fixes the drd/helgrind error.

Some documentations to `ibv_fork_init(3)` states explicitly it should be called before any other libibverbs function. In my opinion, it is the only rational use of `ibv_fork_init(3)` which maybe should be suggested by the manpage. Otherwise, even ignoring the possible race, the application may behave differently from provider-to-provider, because there is no documentation that may explain when the provider will use either `ibv_dontfork_range()` or `ibv_dofork_range()`.

Ref: https://www.rdmamojo.com/2012/05/24/ibv_fork_init/

PR #995, opened 2021-05-03
Broadcom's rdma provider lib update

dsharma283:
This patch series is a part of bigger feature submission which
changes the wqe posting logic. The current series adds the
ground work in the direction to change the wqe posting algorithm.


nmorey:
@dsharma283 Please do not put Fixes: in every commit if they don't actually fix something.
It makes my job on stable branches a whole lot harder.

Unless I'm mistaken commit [234]  are not fixes but just optimizations.

PR #996, opened 2021-05-04
efa: Fix use of uninitialized query device response

galpress:
The query device response struct must be cleared before passing it to
the kernel ioctl call. Otherwise, old drivers that do not fill the
entire response would cause the userspace to read uninitialized data.

Fixes: c95038eb598f ("efa: Move the context initialization out of efa_query_device_ex()")


galpress:
Thanks Leon!
@nmorey just to be sure, this will get backported to v33, 34 stable versions due to the fixes tag, right?

nmorey:
@galpress Yes. My script picked it up. Should be in the stable branches soon

PR #998, opened 2021-05-05
Upgrade to fc34

rleon:


rleon:
It is a false failure due to the missing FC34 image in azure.

rleon:
@jgunthorpe can you please push such image?

Thanks

libibmad/register.c
```diff
@@ -117,15 +117,15 @@ int mad_register_client_via(int mgmt, uint8_t rmpp_version,
 }
 
 int mad_register_server(int mgmt, uint8_t rmpp_version,
-			long method_mask[], uint32_t class_oui)
+			long method_mask[16 / sizeof(long)], uint32_t class_oui)
```
jgunthorpe, line 120: Why not just use "long method_mask *" ?

buildlib/azure-pipelines.yml
```diff
@@ -34,7 +34,7 @@ resources:
       image: ucfconsort.azurecr.io/rdma-core/centos8:25.0
       endpoint: ucfconsort_registry
     - container: fedora
-      image: ucfconsort.azurecr.io/rdma-core/fc33:33.0
+      image: ucfconsort.azurecr.io/rdma-core/fc34:34.0
```
jgunthorpe, line 37: Change the descriptive string below as well

jgunthorpe:
Hum those new warnings are interesting, they are bogus becuase 'hdr' is enclosed in a larger object which is the thing we are actually trying to write. I wonder if a void * cast would silence it?

libibmad/register.c
rleon: It still produces same warning of differences between .c and .h. We can't change .h because it is ABI.

rleon:
Interesting why can't I reproduce CI failures locally?

jgunthorpe:
We don't run the fc packaging build with -Werror outside AZP. I think the rpm scripts are injecting some addtional compile options compared to our normal make. Usually what I try to do is pull the extra options into the cmake so everything turns them on

rleon:
> We don't run the fc packaging build with -Werror outside AZP. I think the rpm scripts are injecting some addtional compile options compared to our normal make. Usually what I try to do is pull the extra options into the cmake so everything turns them on

I tried, but maybe not enough.

rleon:
> > We don't run the fc packaging build with -Werror outside AZP. I think the rpm scripts are injecting some addtional compile options compared to our normal make. Usually what I try to do is pull the extra options into the cmake so everything turns them on
> 
> I tried, but maybe not enough.

ok, the flag that reproduces it is "-flto=auto"

PR #1002, opened 2021-05-05
Avoid running mlx5 tests over non mlx5 devices

Kamalheib:
There is no point in trying to run the mlx5 tests over non-mlx5 devices,
change the tests to skip if the tested device is a non-mlx5 device.


rleon:
It needs to be rebased.

Thanks

Kamalheib:
> It needs to be rebased.
> 
> Thanks

Done, Anything else blocking this PR now?

Thanks,
Kamal

PR #1009, opened 2021-05-19
verbs: Introduce ibv_query_qp_data_in_order() API

yishaih:
This series Introduces ibv_query_qp_data_in_order() API.

This enables an application to check whether WQE data is guaranteed to be written in-order on the given QP and thus reader may poll for data instead of poll for completion.

The API was implemented by mlx5 driver over DEVX.
A detailed man page and some pyverbs stuff were added as well.

### DIFF
--- a/libibverbs/dummy_ops.c
+++ b/libibverbs/dummy_ops.c
@@ -397,6 +397,12 @@ static int query_ece(struct ibv_qp *qp, struct ibv_ece *ece)
 	return EOPNOTSUPP;
 }
 
+static int query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
+				  uint32_t flags)
+{
+	return 0;
+}
+
 static int query_port(struct ibv_context *context, uint8_t port_num,
 		      struct ibv_port_attr *port_attr)
 {
@@ -559,6 +565,7 @@ const struct verbs_context_ops verbs_dummy_ops = {
 	query_ece,
 	query_port,
 	query_qp,
+	query_qp_data_in_order,
 	query_rt_values,
 	query_srq,
 	read_counters,
@@ -683,6 +690,7 @@ void verbs_set_ops(struct verbs_context *vctx,
 	SET_PRIV_OP_IC(vctx, query_ece);
 	SET_PRIV_OP_IC(ctx, query_port);
 	SET_PRIV_OP(ctx, query_qp);
+	SET_PRIV_OP_IC(ctx, query_qp_data_in_order);
 	SET_OP(vctx, query_rt_values);
 	SET_OP(vctx, read_counters);
 	SET_PRIV_OP(ctx, query_srq);
--- a/debian/libibverbs1.symbols
+++ b/debian/libibverbs1.symbols
@@ -11,6 +11,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  IBVERBS_1.11@IBVERBS_1.11 32
  IBVERBS_1.12@IBVERBS_1.12 34
  IBVERBS_1.13@IBVERBS_1.13 35
+ IBVERBS_1.14@IBVERBS_1.14 36
  (symver)IBVERBS_PRIVATE_34 34
  _ibv_query_gid_ex@IBVERBS_1.11 32
  _ibv_query_gid_table@IBVERBS_1.11 32
@@ -98,6 +99,7 @@ libibverbs.so.1 libibverbs1 #MINVER#
  ibv_query_port@IBVERBS_1.1 1.1.6
  ibv_query_qp@IBVERBS_1.0 1.1.6
  ibv_query_qp@IBVERBS_1.1 1.1.6
+ ibv_query_qp_data_in_order@IBVERBS_1.14 36
  ibv_query_srq@IBVERBS_1.0 1.1.6
  ibv_query_srq@IBVERBS_1.1 1.1.6
  ibv_rate_to_mbps@IBVERBS_1.1 1.1.8
--- a/libibverbs/CMakeLists.txt
+++ b/libibverbs/CMakeLists.txt
@@ -21,7 +21,7 @@ configure_file("libibverbs.map.in"
 
 rdma_library(ibverbs "${CMAKE_CURRENT_BINARY_DIR}/libibverbs.map"
   # See Documentation/versioning.md
-  1 1.13.${PACKAGE_VERSION}
+  1 1.14.${PACKAGE_VERSION}
   all_providers.c
   cmd.c
   cmd_ah.c
--- a/libibverbs/driver.h
+++ b/libibverbs/driver.h
@@ -370,6 +370,8 @@ struct verbs_context_ops {
 			  struct ibv_port_attr *port_attr);
 	int (*query_qp)(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 			int attr_mask, struct ibv_qp_init_attr *init_attr);
+	int (*query_qp_data_in_order)(struct ibv_qp *qp, enum ibv_wr_opcode op,
+				      uint32_t flags);
 	int (*query_rt_values)(struct ibv_context *context,
 			       struct ibv_values_ex *values);
 	int (*query_srq)(struct ibv_srq *srq, struct ibv_srq_attr *srq_attr);
--- a/libibverbs/libibverbs.map.in
+++ b/libibverbs/libibverbs.map.in
@@ -160,6 +160,11 @@ IBVERBS_1.13 {
 		ibv_unimport_dm;
 } IBVERBS_1.12;
 
+IBVERBS_1.14 {
+	global:
+		ibv_query_qp_data_in_order;
+} IBVERBS_1.13;
+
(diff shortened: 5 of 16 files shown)


jgunthorpe:
Did this get posted to the mailing list?

libibverbs/dummy_ops.c
```diff
@@ -397,6 +397,11 @@ static int query_ece(struct ibv_qp *qp, struct ibv_ece *ece)
 	return EOPNOTSUPP;
 }
 
+static int query_qp_data_in_order(struct ibv_qp *qp, uint8_t *data_in_order)
+{
+	return EOPNOTSUPP;
```
jgunthorpe, line 402: No, return "out of order" and success. Why is this API allowed to fail anyhow?

jgunthorpe (changes requested):

galpress:
Can't ibv_query_qp() be extended to include this information?
Or maybe add ibv_query_qp_ex()?

yochaic:
> Can't ibv_query_qp() be extended to include this information?
> Or maybe add ibv_query_qp_ex()?

We preferred to use a new verb (as opposed to adding an attribute to ibv_query_qp/_ex or so) because:

It reduces complexity of struct ibv_qp_attr.
It is easier to understand e.g. gets a separate man page.
A new dedicated API could have a lightweight implementation in some cases.

galpress:
OK.
Shouldn't this also take the operation type into account? For example, maybe sends are in order, but RDMA operations aren't?

galpress:
I'm not talking about the sender side, you're assuming this property is consistent for the receiving QP regardless of how the received data got there (send vs RDMA), this is not necessarily true for all hardware vendors.

yochaic:
Note this verb indicates data is written in-order **within** each WQE. It says nothing about ordering **between** separate WQEs. Do you have hardware that does ordering differently for different operation types on the same WQE?

rajachan:
I've always wondered about this. As much as applications do this (NCCL and FaRM come to mind) and despite the perf gains to be had, isn't polling for the last byte rather than waiting for a completion in violation of the IB spec at least?

jgunthorpe:
Yes it always violates the spec to do last data polling, but since it worked anyhow for a long time applications were written to it. This API is intended to provide some order - if an app is doing last data polling then it should call this API and disable the polling, going back to processing CQEs, if the rdma-core indicates data ordering is not guarenteed.

galpress:
> Note this verb indicates data is written in-order **within** each WQE. It says nothing about ordering **between** separate WQEs. 

Yes, this is within each WQE submitted by the remote QP. But the QP type and the ordering property refers to the receiving QP.

> Do you have hardware that does ordering differently for different operation types on the same WQE?

I have to verify about existing EFA hardware, but it is likely that with future hardware the ordering will depend on the remote WQE operation type.

Also, doesn't it depend on where the memory is? I assume the ordering on GPU memory is not necessarily the same as host memory?

jgunthorpe:
At the first level it reflects what the device does when it injects DMA into the memory fabric. Ie the PCI relaxed ordering bits or the order of TLPs from a single WQE.

The CPU platform behavior got mixed into here too, but I wouldn't extend that any further. If GPU's add another dimension then it should be some GPU API that apps have to query to know if GPUs can have ordering fidelity.

I'm not sure I belive that differing by operations is a really useful thing, there is only SEND and WRITE as possible inputs. You really think real HW is going to put TLPs out of order for SEND and in order for WRITE or something? Seems like a very strange thing to do.

galpress:
It is relevant for READ as well.

And for EFA hardware, yes, I can see it being different according to the operation.

jgunthorpe:
Well, next you are going to ask to report if the behavior is different for > MTU or < MTU - it seems like some rat hole. Is there actual applications that want to use beyond-spec last data polling and can make use of all these distinctions?

galpress:
Are you saying you can't make this verb generic enough? Expose it as an mlx5 direct verb.

Also, this API deserves better documentation. The TLPs ordering you explained should be part of the man page.
What's the expected app behavior when encountering errors? How are they detected if the completion isn't polled? Are you assuming that if the data is written then it is necessarily completed successfully?

jgunthorpe:
This is intended to support long existing applications that already work in a certain narrow and non-standard way. It is not to define a whole new way of writing verbs app and certainly not to encourage the use of non-standard last data polling beyond its historical usage.

I'm generally not keen on being excessively general, especially if there is no reasonable implementation or user of the generality.

galpress:
I'm not the one to decide whether the apps usage justifies adding a new verb, but apparently the author of this PR thinks it does.
But this PR adds a new verb which is only implemented in a single provider. If it's only relevant for mlx5 devices then it's implemented in the wrong place.
If it's relevant for all RDMA devices and intended for more implementations in the future, it is not good enough - at least for EFA (can't speak for other providers).

Exposing this with lack of documentation won't result in it being used less, but used wrongly..
If apps are doing something wrong it's their own fault, if libibverbs wants to take responsibility about such apps, surely we can do better than this?

jgunthorpe:
The apps that need to call this are RC apps and aren't going to be using EFA anyhow. The idea is not to expose some new facet of verbs, but to have some very minimal support so that the few apps that have been doing this for decades can continue to work when the HW is no longer strongly in order. For this reason I'd rather it is an app driven API, not an API that is just exposing some wild range of HW ability. I don't think there are real apps that can make use of the distinction you are proposing. EFA should just delcare it doesn't support strong ordering - and if it wants to make use of this optimization then it needs to have strong ordering provided in the way todays apps expect it to be provided.

If I remember correct the only real usage is multi-packet RDMA WRITE operations where an app can avoid the CQE traffic if it can  last data poll. In theory can use this in other places, but it rapidly becomes much less useful. Observing a SEND buffer update requires knowing which buffer and RDMA READ is already slow.

jgunthorpe:
And thinking about it, I'd be happy enough if we documented it as the above, being specific to an inbound RDMA WRITE.. Unless somone knows of existing cases that actually are doing something different?

galpress:
> The apps that need to call this are RC apps and aren't going to be using EFA anyhow. The idea is not to expose some new facet of verbs, but to have some very minimal support so that the few apps that have been doing this for decades can continue to work when the HW is no longer strongly in order.

Doesn't nccl do last byte polling? nccl plugs in to EFA over SRD, and AFAIK doesn't exist for decades?
And you don't control how apps are going to use this API, we target it for existing apps but can't know which apps are going to use it in the future.

About the which apps can make use of such distinction, the truth is that I'm not familiar enough with all EFA apps, I'll have to ask the team about it.
But I don't need to know all use cases and worry about how future apps are going to use this API, I'd much rather not go through this one-way door in the first place.

> EFA should just delcare it doesn't support strong ordering - and if it wants to make use of this optimization then it needs to have strong ordering provided in the way todays apps expect it to be provided.

You're leaving us in a sport where if in the future we need a similar functionality provided by this PR for EFA, we would have to expose it as an EFA DV, and this doesn't make any sense.

> Observing a SEND buffer update requires knowing which buffer and RDMA READ is already slow.

RDMA READ is already slow? On which hardware?

> I'd be happy enough if we documented it as the above, being specific to an inbound RDMA WRITE..

OK, do you want to change the verb to ibv_query_qp_data_in_order_write()?

shefty:
Any generic query mechanism likely needs to take into account the MR properties.  I doubt any assumptions can be made about non-system memory.

jgunthorpe:
NCCL uses relaxed ordering MRs which are incompatible with last data polling, not sure what you mean? It also uses EFA through libfabric last I checked, so I'm not sure how a verbs API is relavent to EFA at all???

Do you have an alternative proposal? I really don't want this so complicated that it covers the entire matrix of every possible packet an HCA can generate, that is not the point.

Sean, I covered that above, this API is only about the behavior up until verbs delivers PCIe TLPs out of the HCA, it doesn't try very hard to cover platform behavior. MR specific configuration including relaxed ordering and the behavior of the target memory is out of scope because we can't get that kind of information from verbs.

galpress:
I'm talking about nccl's LL/LL128 protocol. Last I checked libfabric's EFA provider was running over verbs, ofcourse it's relevant.

I wouldn't say that adding an additional parameter to the function call is complicated, anyway, I'm fine with your suggestion of exposing this for RDMA writes only, but please name it and document it as such.

maorgottlieb:
Hi Gal,

Is the below will satisfy:

/**
 * ibv_query_qp_data_in_order - Checks whether the data is guaranteed to be
 *   written in-order.
 * @qp: The QP to query.
 * @op: Operation type.
 * @flags: Extra field for future input. For now must be 0.
 *
 * Return Value
 * ibv_query_qp_data_in_order() returns 1 if the data is guaranteed to be
 *   written in-order for the operation type, 0 otherwise.
 */
int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
                               uint32_t flags);

galpress:
Hmm for the RDMA READ case it is both incoming and outgoing, but you're right, it's not necessarily posted on a remote QP.

galpress (approved):

PR #1013, opened 2021-05-20
Libirdma 052021

tatyana-en:
Hi,

This patch series replaces the current i40iw library with a unified Intel Ethernet RDMA library that supports a new network device E810 (iWARP and RoCEv2 capable) and the existing X722 iWARP device.

The new library is fully backward compatible with the i40iw driver and is designed to be used with multiple future HW generations.

Thank you,
Tatyana

PR #1014, opened 2021-05-20
Libirdma 052021 test

tatyana-en:
Same as Libirdma 052021 PR, but the file irdma-abi.h is added so that AZP tool can run without compilation issues.


jgunthorpe:
Do not make new pull requests, just force push your updates to the existing one.

shirazsaleem:
Hi Jason - I read your comment on the list and from what I understand, we need to update the original PR #1013 with an extra commit "update kernel headers" that adds the irdma-abi.h. Unfortunately, Tatyana is out till June 7. Shall I just construct a new PR?

jgunthorpe:
Well, it won't be needed until the kernel driver gets sorted out

jgunthorpe:
Replaced with Shiraz's PR

PR #1015, opened 2021-05-21
kernel-boot: Exclude loading rdma-core modules for s390x

Prashanth684:
When booting a system with RoCE cards on s390x, the rdma load modules was failing along with infiniband (https://bugzilla.redhat.com/show_bug.cgi?id=1782876). These modules are not needed for Melanox RoCE cards and infiniband is not supported by s390x at all.

### DIFF
--- a/kernel-boot/rdma-load-modules@.service.in
+++ b/kernel-boot/rdma-load-modules@.service.in
@@ -16,6 +16,8 @@ Before=network-pre.target
 # Orders all kernel module startup before rdma-hw.target can become ready
 Before=rdma-hw.target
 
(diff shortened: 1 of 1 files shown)


jgunthorpe:
I don't think this should be in the general packaging, this is a RHEL issue that they compiled only part of the kernel on s390. Other distros won't make this same choice

Prashanth684:
> I don't think this should be in the general packaging, this is a RHEL issue that they compiled only part of the kernel on s390. Other distros won't make this same choice

that makes sense. in fact, on some more digging, the libpcap library pulls in rdma-core. this needs to be changed in the packaging for RedHat to not build with rdma-core if s390x. closing this.

PR #1017, opened 2021-05-27
fixed small typo

Min4sT1r1th:
I assume "Ethernet" is meant here.. !?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,7 +457,7 @@ FIND_PACKAGE(PkgConfig REQUIRED)
 
 # libnl
 if (NOT DEFINED ENABLE_RESOLVE_NEIGH)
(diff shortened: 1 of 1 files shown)


Min4sT1r1th:
yep, I'm still working on that (beside many other things .. =) ) - problem is, when I try to use the 'gh' program to perform the required procedure according to github's own manual .. it simply doesn't work - gh always fail with some error message .. very strange

rleon:
> Signed-off-by: Min4sT1r1th ( https://github.com/Min4sT1r1th )

Sorry, but Signed-off-by shouldn't be with nicknames.
    Signed-off-by: Random J Developer <random@developer.example.org>
https://github.com/linux-rdma/rdma-core/blob/master/Documentation/contributing.md

PR #1022, opened 2021-06-10
Pyverbs fork tests

galpress:
This PR adds basic fork unit testing by calling ibv_fork_init() and ibv_is_fork_initialized().
The fork tests are impacted by the order of tests execution, for example, if a previous test has registered an MR then ibv_fork_init() is expected to fail. If only the fork tests are run, it should succeed.

### DIFF
--- a/pyverbs/libibverbs.pxd
+++ b/pyverbs/libibverbs.pxd
@@ -5,7 +5,6 @@ include 'libibverbs_enums.pxd'
 from libc.stdint cimport uint8_t, uint16_t, uint32_t, uint64_t
 from posix.time cimport timespec
 
-
 cdef extern from 'infiniband/verbs.h':
 
     cdef struct anon:
@@ -724,6 +723,8 @@ cdef extern from 'infiniband/verbs.h':
     int ibv_get_async_event(ibv_context *context, ibv_async_event *event)
     void ibv_ack_async_event(ibv_async_event *event)
     int ibv_query_qp_data_in_order(ibv_qp *qp, ibv_wr_opcode op, uint32_t flags)
+    int ibv_fork_init()
+    ibv_fork_status ibv_is_fork_initialized()
 
 
 cdef extern from 'infiniband/driver.h':
--- a/tests/test_fork.py
+++ b/tests/test_fork.py
@@ -1,0 +1,30 @@
+# SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All rights reserved.
+
+import errno
+import pyverbs.enums as e
+from pyverbs.fork import fork_init, is_fork_initialized
+from pyverbs.pyverbs_error import PyverbsRDMAError
+
+from tests.base import PyverbsAPITestCase
+
+
+class ForkAPITest(PyverbsAPITestCase):
+    """
+    Test the API of the fork functions.
+    """
+
+    def test_is_fork_initialized(self):
+        try:
+            fork_init()
+            expected_ret = [e.IBV_FORK_ENABLED, e.IBV_FORK_UNNEEDED]
+        except PyverbsRDMAError as ex:
+            # Depends on the order of the tests EINVAL could be returned if
+            # fork_init() is called after an MR has already been registered.
+            self.assertEqual(ex.error_code, errno.EINVAL)
+            expected_ret = [e.IBV_FORK_DISABLED, e.IBV_FORK_UNNEEDED]
+
+        ret = is_fork_initialized()
+        if self.config['verbosity']:
+            print(f'is_fork_initialized() = {ret}')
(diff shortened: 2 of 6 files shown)


rleon:
> The fork tests are impacted by the order of tests execution,

This doesn't sound good.

Thanks

galpress:
To clarify, they obviously always pass regardless of the order of execution.
But it means that the test has to take into account that ibv_fork_init() could fail for example.

I'm open for suggestions if anyone has an elegant solution.

galpress:
@EdwardSro thoughts?

tests/test_fork.py
```diff
@@ -0,0 +1,41 @@
...
+            fork_init()
+        except PyverbsRDMAError as ex:
+            # Depends on the order of the tests EINVAL could be returned if
+            # fork_init() is called after an MR has already been registered.
+            self.assertEqual(ex.error_code, errno.EINVAL)
+
+    def test_is_fork_initialized(self):
+        try:
+            fork_init()
+            expected_ret = [e.IBV_FORK_ENABLED, e.IBV_FORK_UNNEEDED]
```
EdwardSro, line 31: you can call is_fork_initialized() before calling fork_init and check its returned value to make a more "precise" check after fork_init.

tests/test_fork.py
```diff
@@ -0,0 +1,41 @@
...
+from tests.base import PyverbsAPITestCase
+
+
+class ForkAPITest(PyverbsAPITestCase):
+    """
+    Test the API of the fork functions.
+    """
+
+    def setUp(self):
+        super().setUp()
```
EdwardSro, line 18: no need, the parent's setup would be called automatically...

pyverbs/libibverbs.pxd
```diff
@@ -4,7 +4,7 @@
 include 'libibverbs_enums.pxd'
 from libc.stdint cimport uint8_t, uint16_t, uint32_t, uint64_t
 from posix.time cimport timespec
-
+cimport pyverbs.libibverbs_enums as e
```
EdwardSro, line 7: no need, there's an "include 'libibverbs_enums.pxd'" line, so you can directly use the enums

pyverbs/libibverbs.pxd
```diff
@@ -723,7 +723,8 @@ cdef extern from 'infiniband/verbs.h':
     int ibv_query_rt_values_ex(ibv_context *context, ibv_values_ex *values)
     int ibv_get_async_event(ibv_context *context, ibv_async_event *event)
     void ibv_ack_async_event(ibv_async_event *event)
-
+    int ibv_fork_init()
+    e.ibv_fork_status ibv_is_fork_initialized()
```
EdwardSro, line 727: remove the "e." after removing the enums cimport

EdwardSro (changes requested):
If memory registration was done, fork_init would fail? regardless whether the memory was freed by the user or not?

galpress:
Yes, if ibv_reg_mr() is called prior to ibv_fork_init() then the call fails:
https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/man/ibv_fork_init.3.md#return-value

tests/test_fork.py
galpress: I didn't want to get into corner cases with ordering of the different test cases.
If a prior test case already called fork_init() then is_fork_initialized() could return IBV_FORK_ENABLED, even if fork_init() is not called yet in this test case.

tests/test_fork.py
galpress: Done

pyverbs/libibverbs.pxd
galpress: Thanks, didn't know that's how it works.

pyverbs/libibverbs.pxd
galpress: Done

tests/test_fork.py
EdwardSro: I know, what I meant, is to call is_fork_initialized(), according to the returned value you would know the exact expected value of the second is_fork_initialized call after fork_init().
But yes, then you're required to handle multiple cases.
You can keep it as is... it wouldn't have much added value.

EdwardSro (commented):
Are you planning to use these verbs in future pyverbs tests?
Since the tests don't have much-added value (still they're needed to cover newly added verb(s) to pyverbs) IMO, I would drop the first test (test_fork_init) and keep the second one. What do you think?

galpress:
> Are you planning to use these verbs in future pyverbs tests?

I have some thoughts about using it in other tests, but I don't think that would happen soon.

> IMO, I would drop the first test (test_fork_init) and keep the second one. What do you think?

Sure, will do.

Thanks for the review!

EdwardSro (approved):
Thanks.

PR #1029, opened 2021-06-27
mlx5: Introduce mlx5dv_wr_raw_wqe builder

yishaih:
Introduce mlx5dv_wr_raw_wqe builder to be used for building a raw work request.

The user needs to set the raw WQE according to mlx5 device specification, the driver internally may override some fields which are state related. 

The work completion opcode for a raw WQE will be IBV_WC_DRIVER2 that was added for a specific driver operation. 

Man pages were updated with the expected usage accordingly.

The series includes also some pyverbs stuff to test this functionality.

### DIFF
--- a/libibverbs/man/ibv_poll_cq.3
+++ b/libibverbs/man/ibv_poll_cq.3
@@ -83,7 +83,8 @@ overrun from occurrence.  In case of a CQ overrun, the async event
 .B IBV_EVENT_CQ_ERR
 will be triggered, and the CQ cannot be used.
 .PP
-IBV_WC_DRIVER1 will be reported as a response to IBV_WR_DRIVER1 opcode.
+IBV_WC_DRIVER1 will be reported as a response to IBV_WR_DRIVER1 opcode;
+IBV_WC_DRIVER2 will be reported on a specific driver operation.
 .SH "SEE ALSO"
 .BR ibv_post_send (3),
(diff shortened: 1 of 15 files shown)


libibverbs/man/ibv_poll_cq.3
```diff
@@ -83,7 +83,8 @@ overrun from occurrence.  In case of a CQ overrun, the async event
 .B IBV_EVENT_CQ_ERR
 will be triggered, and the CQ cannot be used.
 .PP
-IBV_WC_DRIVER1 will be reported as a response to IBV_WR_DRIVER1 opcode.
+IBV_WC_DRIVER1 will be reported as a response to IBV_WR_DRIVER1 opcode;
+IBV_WC_DRIVER2 will be reported on a specific driver operation.
```
galpress, line 87: What does that mean? Why isn't IBV_WC_DRIVER2 used with IBV_WR_DRIVER2?

yishaih: It points on a driver specific completion opcode with that value (i.e. IBV_WR_DRIVER2).

For mlx5 it will be used with raw WQE builder, see its man page, any driver can use it along with its specific usage.

There was no need / usage for IBV_WR_DRIVER2 and as such it wasn't added as part of this series.

PR #1030, opened 2021-06-29
Verbs logging API

galpress:
This PR adds a generic logging API to libibverbs and its providers controlled by a new environment variable VERBS_LOG_LEVEL.

This PR demonstrates the API usage by the EFA provider, other providers (mlx5, hns for ex.) can be converted to make use of it as well.

### DIFF
--- a/Documentation/libibverbs.md
+++ b/Documentation/libibverbs.md
@@ -56,3 +56,21 @@ need to increase this limit.  This is usually done for ordinary users
 via the file /etc/security/limits.conf.  More configuration may be
 necessary if you are logging in via OpenSSH and your sshd is
 configured to use privilege separation.
+
+# Debugging
+
+### Enabling debug prints
+
+Library and providers debug prints can be enabled using the `VERBS_LOG_LEVEL`
+environment variable, the output shall be written to the file provided in the
+`VERBS_LOG_FILE` environment variable. When the library is compiled in debug
+mode and no file is provided the output will be written to stderr.
+
+Note: some of the debug prints are only available when the library is compiled
+in debug mode.
+
+The following table describes the expected behavior when VERBS_LOG_LEVEL is set:
+|                 | Release                         | Debug                                          |
+|-----------------|---------------------------------|------------------------------------------------|
+| Regular prints  | Output to VERBS_LOG_FILE if set | Output to VERBS_LOG_FILE, or stderr if not set |
+| Datapath prints | Compiled out, no output         | Output to VERBS_LOG_FILE, or stderr if not set |
--- a/buildlib/RDMA_BuildType.cmake
+++ b/buildlib/RDMA_BuildType.cmake
@@ -39,4 +39,8 @@ function(RDMA_BuildType)
 	    CACHE STRING "Default flags for RelWithDebInfo configuration" FORCE)
     endif()
   endforeach()
+
+  if (CMAKE_BUILD_TYPE STREQUAL Debug OR CMAKE_BUILD_TYPE STREQUAL RelWithDebInfo)
+    add_definitions("-DVERBS_DEBUG")
+  endif()
 endfunction()
--- a/libibverbs/driver.h
+++ b/libibverbs/driver.h
@@ -49,6 +49,56 @@
 
 struct verbs_device;
 
+enum {
+	VERBS_LOG_LEVEL_NONE,
+	VERBS_LOG_ERR,
+	VERBS_LOG_WARN,
+	VERBS_LOG_INFO,
+	VERBS_LOG_DEBUG,
+};
+
+void __verbs_log(struct verbs_context *ctx, uint32_t level,
+		 const char *fmt, ...);
+
+#define verbs_log(ctx, level, format, arg...)                                  \
+do {                                                                           \
+	int tmp = errno;                                                       \
+	__verbs_log(ctx, level, "%s: %s:%d: " format,                          \
+		    (ctx)->context.device->name, __func__, __LINE__, ##arg);   \
+	errno = tmp;                                                           \
+} while (0)
+
+#define verbs_debug(ctx, format, arg...) \
+	verbs_log(ctx, VERBS_LOG_DEBUG, format, ##arg)
+
+#define verbs_info(ctx, format, arg...) \
+	verbs_log(ctx, VERBS_LOG_INFO, format, ##arg)
+
+#define verbs_warn(ctx, format, arg...) \
+	verbs_log(ctx, VERBS_LOG_WARN, format, ##arg)
+
+#define verbs_err(ctx, format, arg...) \
+	verbs_log(ctx, VERBS_LOG_ERR, format, ##arg)
+
+#ifdef VERBS_DEBUG
+#define verbs_log_datapath(ctx, level, format, arg...) \
+	verbs_log(ctx, level, format, ##arg)
+#else
+#define verbs_log_datapath(ctx, level, format, arg...) {}
+#endif
+
+#define verbs_debug_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_DEBUG, format, ##arg)
+
+#define verbs_info_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_INFO, format, ##arg)
+
+#define verbs_warn_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_WARN, format, ##arg)
+
+#define verbs_err_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_ERR, format, ##arg)
+
 enum verbs_xrcd_mask {
 	VERBS_XRCD_HANDLE	= 1 << 0,
 	VERBS_XRCD_RESERVED	= 1 << 1
--- a/libibverbs/init.c
+++ b/libibverbs/init.c
@@ -36,6 +36,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <glob.h>
+#include <stdarg.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <sys/stat.h>
@@ -52,10 +53,29 @@
 #include <rdma/rdma_netlink.h>
 
 #include <util/util.h>
+#include "driver.h"
 #include "ibverbs.h"
 #include <infiniband/cmd_write.h>
 
 int abi_ver;
+
+static uint32_t verbs_log_level;
+static FILE *verbs_log_fp;
+
+__attribute__((format(printf, 3, 4)))
+void __verbs_log(struct verbs_context *ctx, uint32_t level,
+		 const char *fmt, ...)
+{
+	va_list args;
+
+	if (level <= verbs_log_level) {
+		int tmp = errno;
+		va_start(args, fmt);
+		vfprintf(verbs_log_fp, fmt, args);
+		va_end(args);
+		errno = tmp;
+	}
+}
 
 struct ibv_driver {
 	struct list_node	entry;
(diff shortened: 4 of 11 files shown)


galpress:
Fixed a clang compilation error.

rleon:
I'm ok with the idea of this PR, however I would like to see provides conversion patch too.

Also, I saw and access of verbs_get_ctx() inside provides, maybe it is ok, but this is first user outside of libibverbs,

galpress:
I thought of converting mlx5 but it is probably better if someone from mlx5 side would do it as it requires removal of the debug mask (QP/CQ/CQE/etc) and the optional fp.
I'll add hns and siw as additional consumers.

rleon:
> I thought of converting mlx5 but it is probably better if someone from mlx5 side would do it as it requires removal of the debug mask (QP/CQ/CQE/etc) and the optional fp.
> I'll add hns and siw as additional consumers.

Let's first see that this PR is acceptable, and I'll provide mlx5 patch, once we ready.

yishaih:
Few notes on this PR:

We can't drop the mlx5 internal debug granulation (e.g. QP/CQ/CQE/etc.) which is some kind of API, in addition any change may need to preserve the logic of the target of the output. (e.g. log file, etc.).

In addition,
It seems that introducing an extra environment variable as appears in this patch is by contrast to the way we go in the last years. The patch also lacks control on the target output and use stderr, we should prevent printing to stderr from the library as was discussed in the past.

IMO, adding an extra logging mechanism should be considered carefully and need to be more robust and controllable API.
Let's get more people's feedback here before taking next steps.

galpress:
If mlx5 wishes to preserve backwards compatibility with the provider specific MLX5_DEBUG_MASK environment variable then it probably shouldn't be changed and should not use this API at all.

I agree that a library shouldn't print to stderr, it shouldn't print anything to anywhere, but we're talking about the app being run in debug mode with a specific verbs flag. Adding a target output fp is doable, I don't find it very useful though.

What is wrong with an environment variable? Setting this through some kind of verbs programatic API makes no sense as it would require recompiling the app. Do you have other suggestions?

rleon:
> Few notes on this PR:
> 
> We can't drop the mlx5 internal debug granulation (e.g. QP/CQ/CQE/etc.) which is some kind of API, in addition any change may need to preserve the logic of the target of the output. (e.g. log file, etc.).

There are no plans to drop it, but provide wrapper from MLX5_* variables to general one.

> 
> In addition,
> It seems that introducing an extra environment variable as appears in this patch is by contrast to the way we go in the last years. The patch also lacks control on the target output and use stderr, we should prevent printing to stderr from the library as was discussed in the past.

We are against adding new environment variables that change library behavior to run its in normal mode. This is different because it is needed for the debug and not going to change library flow at all.

rleon:
> If mlx5 wishes to preserve backwards compatibility with the provider specific MLX5_DEBUG_MASK environment variable then it probably shouldn't be changed and should not use this API at all.

I'll provide wrapper.

> 
> I agree that a library shouldn't print to stderr, it shouldn't print anything to anywhere, but we're talking about the app being run in debug mode with a specific verbs flag. Adding a target output fp is doable, I don't find it very useful though.

We found fp usable when you want to print despite not running in debug mode. For the mlx5, it is CQE error.

galpress:
Added an additional environment variable VERBS_LOG_FILE and converted hns and siw.

libibverbs/device.c
```diff
@@ -222,6 +222,44 @@ static bool has_ioctl_write(struct ibv_context *ctx)
...
+}
+
+static void verbs_set_log_file(struct verbs_context *context_ex)
+{
+	char *env;
+
+	if (context_ex->log_level == VERBS_LOG_LEVEL_NONE)
+		return;
+
+	env = getenv("VERBS_LOG_FILE");
```
rleon, line 241: The mlx5 implementation is slightly different and IMHO better.

The mlx5 prints are visible only if library is compiled with MLX5_DEBUG flag, without such special compilation flag, nothing is printed to the screen. This allows us to use logging both in data and control paths. Yes, "if ((level) <= (ctx)->log_level)" costs us enough to do not put it in hot paths.

The FD is needed where you anyway want to print something, but the library is not compiled with MLX5_DEBUG.

I think that you can rely CMAKE_BUILD_TYPE to provide same behavior.

galpress: mlx5 wraps some of the prints with the MLX5_DEBUG define, you can still do that for prints in data path (or use the CMAKE_BUILD_TYPE).
I think it's better for control path prints to not be dependent on such define, this way they will always be available.

rleon: You need two level macros/defines anyway. One for control path, another for data path. Let's add data path macros too.

galpress: What do you mean?
I meant that in data path you can wrap the verbs_* calls with #ifdefs, do you want a version of verbs_* that does that internally?

rleon: Something like that, but I still didn't convince myself that debug prints for library compiled as "Release" is right thing to do.

galpress: Taking mlx5 as an example, the mlx5_err() prints do not depend on how the library is compiled, the mlx5_dbg() does.
I can do something similar where verbs_debug() only prints in debug mode and other stuff are always available.

rleon: It is not always available, but only if you provided DEBUG_FILE

galpress: You're right, it always checks the if statement but doesn't print if there's no DEBUG_FILE.
So do you have any preferences how to do it here?

rleon: I would expect two modes:
1. library compiled in debug mode - print whatever you want without any need in DEBUG_FILE
2. library compiled in release mode - request from the user to supply extra variable VERBS_DEBUG_FILE. it will be super nice if we will be civil and provide a way to use syslog in addition/instead of debug file. Or at least will leave a room to implement it later.

galpress: I thought the motivation was to not run the additional if statement when compiled in release mode, I'm not sure we can do that with what you described?

rleon: > I thought the motivation was to not run the additional if statement when compiled in release mode, I'm not sure we can do that with what you described?

Yes, this is the motivation, this "if" will be for verbs_err() macro only. All other will be compiled out. Maybe even special macro: verbs_err_datapath()

galpress:
Updated the PR, let me know what you think.

Documentation/libibverbs.md
```diff
@@ -56,3 +56,15 @@ need to increase this limit.  This is usually done for ordinary users
...
 configured to use privilege separation.
+
+# Debugging
+
+### Enabling debug prints
+
+Library and providers debug prints can be enabled using the `VERBS_LOG_LEVEL`
+environment variable, the output shall be written to the file provided in the
+`VERBS_LOG_FILE` environment variable. When the library is compiled in debug
+mode and no file is provided the output shall be written to stderr.
```
rleon, line 67: shall -> will

libibverbs/verbs.h
```diff
@@ -84,6 +85,54 @@ struct ibv_gid_entry {
...
 
+enum {
+	VERBS_LOG_LEVEL_NONE,
+	VERBS_LOG_ERR,
+	VERBS_LOG_WARN,
+	VERBS_LOG_INFO,
+	VERBS_LOG_DEBUG,
+};
+
+#define verbs_log(ctx, level, format, arg...)                                         \
```
rleon, line 96: Don't we need to add __printf(3, 4) attribute to silence some GCC 10 warnings?

rleon:
> Updated the PR, let me know what you think.

I liked it and think that it is ready to be merged.

libibverbs/verbs.h
galpress: I guess we would if it was a function instead of a macro.

Documentation/libibverbs.md
galpress: Done

galpress:
Thanks a lot for the review Leon, updated the documentation.

rleon:
Ahh, I forgot to mention that other solutions than prints exist in modern software.
One of them is https://lttng.org/

galpress:
Thanks, I was not aware of this framework.

Few questions:
- Tracing isn't necessarily the same as logging, shouldn't it be an orthogonal feature (as is in the kernel)?
- Do we want to pull in lttng as a dependency?
- Do we want the extra complexity involved in enabling the tracepoints through the daemon?
- Are you interested in using the tracepoints API or tracef()/tracelog()? Creating a tracepoint for every place you want a simple debug print might be a bit of an overkill.

rleon:
> Thanks, I was not aware of this framework.
> 
> Few questions:
> 
>     * Tracing isn't necessarily the same as logging, shouldn't it be an orthogonal feature (as is in the kernel)?
> 
>     * Do we want to pull in lttng as a dependency?
> 
>     * Do we want the extra complexity involved in enabling the tracepoints through the daemon?
> 
>     * Are you interested in using the tracepoints API or tracef()/tracelog()? Creating a tracepoint for every place you want a simple debug print might be a bit of an overkill.

No, to everything :)
I'm looking for dynamic debug prints in the user-space, like we have in the kernel.
It will allow us to enable per-type/file/vendor prints that Yishai mentioned.

Thanks

yishaih:
The options in mlx5 gives more granularity and supports vendor specific options, this may be needed down the road by other vendors as well.

Current suggestion doesn't allow a clean transition for mlx5, so we won't use it unless the API will give a matching alternative.

I believe that If we come with a new API/mechanism need to consider such options in the general case.

galpress:
Vendor specific options can be easily implemented by wrapping the generic calls, there's no reason to implement a "generic" vendor specific API.

You want to enable DR prints only, implement a dr_* logging api that calls the verbs one and checks the mlx5 specific environment variables.

rleon:
I refuse to treat and debug/logging/environment variables as API.

yishaih:
It means that any driver may still need to have its own env variable to control its private/verbs options (e.g. DR, QP, CQ, etc.) and the verbs layer will just give the path to print the data based on debug level and may control the target destination.

This is doable, however, we could consider come with more robust mechanism to have all of this in one shared place.
This may include for example a bit mask options which some range can be verbs and others may be driver specific and maintain only one env variable for all drivers for this.

Still need to consider whether we want to go with extra env variables or use some other API for all of that despite that this may cause an application change to use.

galpress:
The way I see it, we can obviously add more masks for more fine grained control, but at the end of the day we're talking about a debug interface which shouldn't care about performance too much, and a simple grep on the entire log file will likely be be easier to use than a bunch of extra environment variables.

A generic mask for QP/CQ/etc might make sense (have to admit that I'm not very convinced), but there's no reason to reserve a range for provider specific ranges. You can't really make use of this vendor specific range in a generic way, so might as well have an extra variable for vendor specific masks.

Having a kernel-like dynamic debug interface is surely more convenient, but I don't think we're gonna implement the whole thing in rdma-core..

rleon:
From the bug reports and debug session that I saw, users don't really use per-object mask.

libibverbs/verbs.h
```diff
@@ -84,6 +85,54 @@ struct ibv_gid_entry {
...
+
+#define verbs_info_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_INFO, format, ##arg)
+
+#define verbs_warn_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_WARN, format, ##arg)
+
+#define verbs_err_datapath(ctx, format, arg...) \
+	verbs_log_datapath(ctx, VERBS_LOG_ERR, format, ##arg)
+
```
jgunthorpe, line 135: Provider stuff belongs in driver.h, not verbs.h

libibverbs/verbs.h
```diff
@@ -2078,6 +2127,8 @@ struct ibv_values_ex {
 
 struct verbs_context {
 	/*  "grows up" - new fields go here */
+	uint32_t log_level;
+	FILE *log_fp;
```
jgunthorpe, line 2131: Given that every context opens the same log_fp, this should not be in the struct verbs_context or in verbs.h..

jgunthorpe (changes requested):

libibverbs/verbs.h
galpress: I don't see a reason why libibverbs core function won't use this api as well, should it still move in such case?

libibverbs/verbs.h
galpress: Where else? struct verbs_device?

jgunthorpe: verbs_device is Ok, a global is good too.  Also keep in mind that opening the same file twice can be problematic, so better to have a global for the FILE *

libibverbs/verbs.h
jgunthorpe: Yes. It is not public API toward the end user so it does not belong in this header

libibverbs/verbs.h
galpress: I'm thinking of changing both to global variables and initialize them in ibverbs_init(), though that means the file will only get closed implicitly on process cleanup, what do you think?

libibverbs/verbs.h
galpress: Done.

galpress:
I moved stuff around and changed to global variables, updated the PR.

Documentation/libibverbs.md
```diff
@@ -56,3 +56,15 @@ need to increase this limit.  This is usually done for ordinary users
...
 configured to use privilege separation.
+
+# Debugging
+
+### Enabling debug prints
+
+Library and providers debug prints can be enabled using the `VERBS_LOG_LEVEL`
+environment variable, the output shall be written to the file provided in the
+`VERBS_LOG_FILE` environment variable. When the library is compiled in debug
+mode and no file is provided the output will be written to stderr.
```
rleon, line 67: I'm not sure that this is fully correct.

Can you please write it as a table for every build mode?
What is expected if variables set/not set and regular prints vs. datapath.

It will be clearer for everyone.

BTW, we will remove our MLX5 debug variables without any need in legacy wrapper.

Thanks

galpress: Thanks, added a table.

rleon:
Thanks

PR #1034, opened 2021-07-20
mlx5: Introduce user space driver over VFIO

yishaih:
This series introduces mlx5 user space driver over VFIO.

This enables an application to take full ownership on the opened device and run any firmware command (e.g. port up/down) without any concern to hurt someone else.

The application look and feel is like regular RDMA application over DEVX, it uses verbs API to open/close a device and then mostly uses DEVX APIs to interact with the device.

To achieve it, few mlx5 DV APIs were introduced, it includes:
- An API to get ibv_device for a given mlx5 PCI name.
- APIs to manage device specific events.

Detailed man pages were added to describe the expected usage of those APIs.

The mlx5 VFIO driver implemented the basic verbs APIs as of managing MR/PD and the DEVX APIs which are required to write an RDMA application.

The VFIO uAPIs were used to setup mlx5 vfio context by reading the device initialization segment, setup DMA and enables the device command interface.

In addition, the series includes pyverbs stuff which runs DEVX like testing over RDMA and VFIO mlx5 devices.

Some extra documentation of the benefits and the motivation to use VFIO can be found here [1].
[1] https://www.kernel.org/doc/html/latest/driver-api/vfio.html


providers/mlx5/mlx5.h
```diff
@@ -90,10 +91,12 @@ enum {
 	MLX5_DBG_CQ_CQE		= 1 << 4,
 	MLX5_DBG_CONTIG		= 1 << 5,
 	MLX5_DBG_DR		= 1 << 6,
+	MLX5_DBG_VFIO		= 1 << 7,
```
rleon, line 94: As I said, I don't want to see any change here and I agree with Gal, this mask is better to be deleted.

IMHO, this mask represents misunderstanding on how libraries are distributed/installed/used in the real world. And if someone wants to debug library, he/she needs to dump everything to the file and use editor later.

Thanks

yishaih: OK, this extra bit mask / usage was dropped, a rebased version was uploaded.

PR #1035, opened 2021-07-20
util: Add barriers support for LoongArch

cnmushiba:
Add barriers support for LoongArch architecture(a new RISC architecture).

### DIFF
--- a/util/udma_barrier.h
+++ b/util/udma_barrier.h
@@ -96,6 +96,8 @@
 #define udma_to_device_barrier() asm volatile("dsb st" ::: "memory");
 #elif defined(__sparc__) || defined(__s390x__)
 #define udma_to_device_barrier() asm volatile("" ::: "memory")
+#elif defined(__loongarch__)
+#define udma_to_device_barrier() asm volatile("dbar 0" ::: "memory")
 #else
 #error No architecture specific memory barrier defines found!
 #endif
@@ -128,6 +130,8 @@
 #define udma_from_device_barrier() asm volatile("dsb ld" ::: "memory");
 #elif defined(__sparc__) || defined(__s390x__)
 #define udma_from_device_barrier() asm volatile("" ::: "memory")
+#elif defined(__loongarch__)
+#define udma_from_device_barrier() asm volatile("dbar 0" ::: "memory")
 #else
 #error No architecture specific memory barrier defines found!
 #endif
@@ -192,6 +196,8 @@
 #define mmio_flush_writes() asm volatile("dsb st" ::: "memory");
 #elif defined(__sparc__) || defined(__s390x__)
 #define mmio_flush_writes() asm volatile("" ::: "memory")
+#elif defined(__loongarch__)
+#define mmio_flush_writes() asm volatile("dbar 0" ::: "memory")
 #else
 #error No architecture specific memory barrier defines found!
 #endif


util/udma_barrier.h
```diff
@@ -96,6 +96,8 @@
 #define udma_to_device_barrier() asm volatile("dsb st" ::: "memory");
 #elif defined(__sparc__) || defined(__s390x__)
 #define udma_to_device_barrier() asm volatile("" ::: "memory")
+#elif defined(__loongarch__)
+#define udma_to_device_barrier() asm volatile("dbar 0" ::: "memory");
```
xen0n, line 100: Uh-oh. The macros should not have trailing semicolon in its expansion; this was neglected in previous reviews but got caught by upstream CI. You should remove the semicolons like in the previous block.

rleon:
1. Please don't use empty commit messages.
2. Does rdma-core even work on MIPS?

cnmushiba:
> 1. Please don't use empty commit messages.
thanks.
> 2. Does rdma-core even work on MIPS?
May not work.

util/udma_barrier.h
cnmushiba: removed, thanks.

rleon:
> 
> > 2. Does rdma-core even work on MIPS?
> 
> May not work.

It is far from "may not work". I didn't hear from anyone that he/she used RDMA stack on MIPS arch.
RDMA HW is supported on x86, ARM and PowerPC machines, while rdma-core never compiled on other platforms.

So the question is what HW that part of rdma-core and upstream kernel really works on MIPS?

cnmushiba:
> It is far from "may not work". I didn't hear from anyone that he/she used RDMA stack on MIPS arch.

LoongArch is not MIPS, it is a new RISC architecture(such as aarch64, riscv).

[Here](https://github.com/loongson/LoongArch-Documentation/releases/download/2021.07.05/LoongArch-Vol1-v1.00-EN.html) is some information about LoongArch.

> RDMA HW is supported on x86, ARM and PowerPC machines, while rdma-core never compiled on other platforms.

We have the Mellanox ib cards. We need its some benchmark data. Users will also use it in the future. The card info as follow:

```
04:00.0 InfiniBand: Mellanox Technologies MT26428 [ConnectX VPI PCIe 2.0 5GT/s - IB QDR / 10GigE] (rev b0)
```

Our kernel(development version) already supported mlx4_ib.ko. We also need some other userspaces software, such as ib_write_bw, ib_send_bw  and so on. These software need libmlx4.so, librdmacm.so, libibumad.so.3, libibverbs.so, etc. Compile libmlx4.so needs that adding barriers support for LoongArch architecture. Because of

```
CHECK_C_SOURCE_COMPILES("                                                                                                                                                                                            
#include \"${CMAKE_CURRENT_SOURCE_DIR}/util/udma_barrier.h\"                                                                                                                                                         
 int main(int argc,const char *argv[]) {return 0;}"                                                                                                                                                                  
  HAVE_COHERENT_DMA)
```

Mlx4 does also need barrier support in LoongArch architecture.

We've also got the benchmark data on Mellanox ib cards.

rleon:
Ohh, great.

I read about it https://www.tomshardware.com/news/loongson-technology-develops-its-own-cpu-instruction-set-architecture and they talked about MIPS

PR #1042, opened 2021-08-05
Link util static library to pyverbs

EdwardSro:
Some internal util functions are needed by pyverbs, thus the static
rdma_util library should be linked in order to expose the required
internal headers and symbols.

### DIFF
--- a/buildlib/pyverbs_functions.cmake
+++ b/buildlib/pyverbs_functions.cmake
@@ -2,6 +2,8 @@
 # Copyright (c) 2018, Mellanox Technologies. All rights reserved.  See COPYING file
 # Copyright (c) 2020, Intel Corporation. All rights reserved.  See COPYING file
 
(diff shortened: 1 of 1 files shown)


galpress:
What is this fixing? Is there a specific failure?

rleon:
> What is this fixing? Is there a specific failure?

i386 builds
https://lore.kernel.org/linux-rdma/162810725329.8972.6442681920407682185.launchpad@alphecca.canonical.com/T/#u

rleon:
@EdwardSro  What should we update/change in the AZP checker to catch such failures?

EdwardSro:
> @EdwardSro What should we update/change in the AZP checker to catch such failures?

I think what should be added is Cython and Python packages for 32bit - otherwise, pyverbs won't be built on i386.
I'll try to do it locally first and make sure that this is what's missing.

PR #1047, opened 2021-08-15
mlx5: Fix compilation on 32 bit systems with gcc 7.5

yishaih:
Fix some compilation issues as of below on 32 bit systems with gcc 7.5.
error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]

### DIFF
--- a/providers/mlx5/mlx5_vfio.c
+++ b/providers/mlx5/mlx5_vfio.c
@@ -1297,7 +1297,7 @@ static void init_eq_buf(struct mlx5_eq *eq)
 
 static uint64_t uar2iova(struct mlx5_vfio_context *ctx, uint32_t index)
 {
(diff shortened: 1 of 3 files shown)


providers/mlx5/mlx5_vfio.c
```diff
@@ -1297,7 +1297,7 @@ static void init_eq_buf(struct mlx5_eq *eq)
 
 static uint64_t uar2iova(struct mlx5_vfio_context *ctx, uint32_t index)
 {
-	return (uint64_t)((void *)ctx->bar_map + (index * MLX5_ADAPTER_PAGE_SIZE));
+	return (uint64_t)(uintptr_t)((void *)ctx->bar_map + (index * MLX5_ADAPTER_PAGE_SIZE));
```
jgunthorpe, line 1300: All of these (uint64_t)(uintptr_t) things are redundant, just cas tto uinptr_t and allow the compiler to widen the type as necessary.

PR #1060, opened 2021-09-13
librdmacm: Do not fail init completely if one device is failing

sydidelot:
Hi, I have a node equipped with 2x dual-port Mellanox HDR adapters (4 ports in total) and running Ubuntu 20.04.
That node is hosting a Docker container where only 2 ports (out of 4) have a uverbs device exposed to the Docker container.

Here is how the system looks like inside the container:
```
# ls /sys/class/infiniband
mlx5_0  mlx5_1  mlx5_2  mlx5_3

# ls /dev/infiniband/
rdma_cm  ucm0  ucm2  uverbs0  uverbs2
```
(uverbs1 and uverbs3 are not listed as you can see above).

With that setup, any application based on the RDMA CM library would fail in rdma_get_devices() with an ENODEV error. Here, I would expect rdma_get_devices() to return an array of ibv_context * for the 2 verbs devices that are available in the container (uverbs0 and uverbs2) even if not all the 4 devices are configured.

I looked at the librdmacm source code and I found that the call to ibv_get_device_list() in sync_devices_list() would detect all the Infiniband ports listed in /sys/class/infiniband, but ucma_init_all() would return an error and would cause rdma_get_devices() to fail if at least one device cannot be opened (ibv_open_device() would return NULL for any unexposed uverbs device).
See https://github.com/linux-rdma/rdma-core/blob/master/librdmacm/cma.c#L486

Will it be possible to have a more "relaxed" version of ucma_init_all(), which would skip any verbs device that is missing and let rdma_get_devices() to return the devices that could effectively be opened?

As a temporary solution, I wrote the following patch for lirdmacm.
Let me know if you need more information to understand or reproduce the issue.

Regards,
Sylvain Didelot

Signed-off-by: Sylvain Didelot <sdidelot@ddn.com>

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -483,12 +483,13 @@ static int ucma_init_all(void)
 		if (dev->is_device_dead)
 			continue;
 
-		ret = ucma_init_device(dev);
-		if (ret)
-			break;
+		if (ucma_init_device(dev)) {
+			/* Couldn't initialize the device: mark it dead and continue */
+			dev->is_device_dead = true;
+		}
 	}
 	pthread_mutex_unlock(&mut);
-	return ret;
+	return 0;
 }
 
 struct ibv_context **rdma_get_devices(int *num_devices)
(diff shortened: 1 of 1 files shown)


rleon (approved):

sydidelot:
Thanks for approving the patch, @rleon 
Is there anything blocking the merge?

rleon:
I waited to anyone share his thoughts, but it looks like that we waited enough.

Thanks

PR #1061, opened 2021-09-17
change max cpus

swimlessbird:
In now,some cpu can have 96 or more cores, 8 is not enough


swimlessbird:
image

rleon:
> Signed-off-by: swimlessbird <swimlessbird@gmail.com>
Please use name and nickname for SOB signature, so we can merge it.

Thanks

rleon:
I fixed it locally and pushed the patch.

Applied, thanks.

PR #1063, opened 2021-09-23
ibdiags:remove the error printing in the ibchecknet command

swimlessbird:
Use ibchecknet command will print " this command is deprecated; Please use \"ibqueryerrors -f\" ".
But use "ibqueryerrors -f" will print "invalid option -- 'f'".
That means the printing in the ibchecknet command is error.

Signed-off-by: Suwan Sun <swimlessbird@gmail.com>

image


rleon:
Unfortunately github UI doesn't allow to comment on commit messages, so I fixed everything manually and merged it.

Thanks

PR #1064, opened 2021-09-26
ibdiags:make escape character tack effect

swimlessbird:
For most operating systems, bash is used to be sh.
In bash, the escape characters in the output of echo
are often output directly without participating in
the escape.
image
image
image


jgunthorpe:
Is it necessary to change to /bin/bash or is the echo -e enough?

kircherlike:
> > Is it necessary to change to /bin/bash or is the echo -e enough?
> 
> I think should change to bash. In Ubuntu,dash is used to be sh.
> image
> 
> In dash , echo has no -e parameter.
> image

lgtm

rleon:
I would prefer to see change that keeps /bin/sh.

nmorey:
> In dash , echo has no -e parameter.

/usr/bin/echo does support `-e` though

swimlessbird:
> > In dash , echo has no -e parameter.
> 
> /usr/bin/echo does support `-e` though

Try it soon.

nmorey:
Please add a Fixes tag:
`Fixes: c670c59cecbd (infiniband-diags: mark duplicate/lower performing scripts as deprecated.)`

swimlessbird:
> Please add a Fixes tag: `Fixes: c670c59cecbd (infiniband-diags: mark duplicate/lower performing scripts as deprecated.)`

Thanks. I added it as quickly as I could.

rleon:
You need to update commit message.

Thanks

PR #1081, opened 2021-11-10
debian: Exclude compressing python files while building the rdma-core

rleon:
Over Debian, the large packages are compressed while building the
rdma-core, due this the rdma-core tests are compressed and this cause it
to fail.

So, we need to exclude compressing python files over debian.

Signed-off-by: Lena Asfour <lenaa@nvidia.com>
Reviewed-by: Tzafrir Cohen <tzafrirc@mellanox.com>

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -90,6 +90,9 @@ override_dh_installsystemd:
 	dh_installsystemd -pibacm ibacm.socket
 	dh_installsystemd --remaining-packages
 
+override_dh_compress:
+	dh_compress -X .py
+
 # Provider plugin libaries are not shared libraries and do not belong in the
 # shlibs file.
 # librspreload is a LD_PRELOAD library and does not belong in the shlib files


bdrung:
You are talking about the tests that python3-pyverbs installs in `/usr/share/doc/rdma-core/tests/`? In that case, I am in favor of not installing the tests in `/usr/share/doc`. They should be installed in `./usr/lib/python3/dist-packages/` (if you want to install them).

rleon:
> You are talking about the tests that python3-pyverbs installs in `/usr/share/doc/rdma-core/tests/`? In that case, I am in favor of not installing the tests in `/usr/share/doc`. They should be installed in `./usr/lib/python3/dist-packages/` (if you want to install them).

It will cause to difference between distros, other distros install tests in "doc" folder.

bdrung:
I would expect to change it for all distros. Since the tests are Python code as well, they should be installed in the location for Python code, shouldn't they?

rleon:
I don't think so. We separate between pyverbs and tests. Pyverbs is real python package that users will import into their scripts. It is indeed installed in dist-packages, which by the way Debian location that doesn't exist in other distros.

Tests are treated as documentation/example/nice-to-have thing and not supposed to be used as part of any application. This is why the decision was to put them as part of documentation. 

Thanks

bdrung:
Do we want to ship the test cases in the binary package in that case? I tend to say that having the test cases in the source package is enough.

rleon:
> Do we want to ship the test cases in the binary package in that case? I tend to say that having the test cases in the source package is enough.

rleon:
Sorry, closed by mistake.

What is the best practice in distro world?

@Kamalheib , @nmorey

Kamalheib:
> Sorry, closed by mistake.
> 
> What is the best practice in distro world?
> 
> @Kamalheib , @nmorey

Adding @Honggang-LI  the maintainer of the rdma-core.

nmorey:
Looking at my installed distro, it seems the usual way is to put the test under the python package dir.
Some packages do create a separate `pythonN-XXX-test` so that install of the tests is not required.

rleon:
We are shipping tests together with pyverbs, so if I understand from Nicolas answer that it is pretty common way to distribute tests.

The second question was there to put them.

nmorey:
Yes, seems like most of the python packages do ship there tests in the same RPM.
As said, most install their tests in their python site-package dir.

For example, samba installs there:
`/usr/lib64/python3.6/site-packages/samba/tests`

bdrung:
So from the responses here, I propose to put the tests in the Python package dir (probably in the pyverbs directory). Then you can import them using `import pyverbs.tests`. The tests could be split into a separate `python3-pyverbs-test` package then.

Honggang-LI:
```bash
#!/bin/bash

for pkg in $(dnf whatprovides '*/tests/' | grep -i ^py | awk '{print $1}'); do
	echo ${pkg}
	dnf repoquery -l ${pkg} | grep 'tests$'
	echo
done
```

Checked 480 python packages for Fedora 35, install `tests` in `site-packages` directory works for Fedora. But we don't need a separate like `python3-pyverbs-test` for Fedora base distros.

Thanks

nmorey:
The separated package is not needed for SUSE either.
My point was that rdma-core and libs + python being a core package for most images including minimalist ones, it might make sense to split the test package as it is pretty unnecessary in most of the cases.
But the test are tiny and I doubt it'll be required. ANyway, it's easy enough for any distro to tune its own deb/spec file to get it split

jgunthorpe:
If the tests are going to be installed in site-packages then we should also still have a way to run them, eg run-test 

The tests should go in the same package that has include/infiniband/verbs.h

Honggang-LI:
> If the tests are going to be installed in site-packages then we should also still have a way to run them, eg run-test

Do you mean package up `build/bin/run_tests.py`?

rleon:
> > If the tests are going to be installed in site-packages then we should also still have a way to run them, eg run-test
> 
> Do you mean package up `build/bin/run_tests.py`?

Yep

rleon:
Thanks, let's close this PR for now.

PR #1089, opened 2021-11-22
systemd hardening

nmorey:
SUSE is launching a system wide effort to harden systemd service files:
See https://en.opensuse.org/openSUSE:Security_Features#Systemd_hardening_effort

### DIFF
--- a/ibacm/ibacm.service.in
+++ b/ibacm/ibacm.service.in
@@ -17,6 +17,10 @@ After=rdma-hw.target
 [Service]
 Type=notify
 ExecStart=@CMAKE_INSTALL_FULL_SBINDIR@/ibacm --systemd
+ProtectSystem=full
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
 
 [Install]
 Also=ibacm.socket
--- a/iwpmd/iwpmd.service.in
+++ b/iwpmd/iwpmd.service.in
@@ -24,5 +24,9 @@ Before=rdma-hw.target
 Type=notify
 ExecStart=@CMAKE_INSTALL_FULL_SBINDIR@/iwpmd --systemd
 LimitNOFILE=102400
+ProtectSystem=full
+ProtectHome=true
+ProtectHostname=true
+ProtectKernelLogs=true
 
 # iwpmd is automatically wanted by udev when an iWarp RDMA device is present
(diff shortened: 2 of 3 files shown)


nmorey:
@Honggang-LI  @bdrung @Kamalheib Is this OK with your distros ?
I'd rather avoid having SUSE specific service files if possible (or maintain in-house patches on top)

rleon:
> @Honggang-LI @bdrung @Kamalheib Is this OK with your distros ? I'd rather avoid having SUSE specific service files if possible (or maintain in-house patches on top)

if it isn't possible to use for all distros, we can save these files in suse/* folder.

Thanks

nmorey:
> if it isn't possible to use for all distros, we can save these files in suse/* folder.
> 
> Thanks

Sure. That was my second plan. But if it works for other distros, let's make them all more secure :)

Honggang-LI:
>  Is this OK with your distros ? 

```
ProtectSystem=full
ProtectHome=true
ProtectHostname=true
ProtectKernelLogs=true
```
systemd version required for those?

nmorey:
`ProtectSystem` / `ProtectHome`: 214
`ProtectHostname`: 242
`ProtectKernelLogs`: 244

bdrung:
So Debian 11 (bullseye) and Ubuntu 20.04 (focal) have recent enough systemd versions. Will older systemd version just ignore these setting?

I am totally for sharing these hardening improvements for the general systemd service files. So go for it. I haven't tested the changes yet, but the diff looks good to me.

jgunthorpe:
Centos7 is 219 and is still something in the CI. I think the thing to do is to sed away the unsupported logic when we install the systemd files. Should be easy enough to do in cmake?

nmorey:
> So Debian 11 (bullseye) and Ubuntu 20.04 (focal) have recent enough systemd versions. Will older systemd version just ignore these setting?
> 

Older versions ignore these and outputs a warning somewhere in journalct
`Unknown lvalue 'ProtectKernelLogs' in section 'Service'`

So for CI, it shouldn't be an issue either

Honggang-LI:
It is ok for me to share these hardening improvements for the general systemd service files.

> Older versions ignore these and outputs a warning somewhere in journalct `Unknown lvalue 'ProtectKernelLogs' in section 'Service'`

We had fixed such issue for rdma stack package as required by RHEL user. I think jgunthorpe's suggestion looks good for me.

https://github.com/linux-rdma/rdma-core/pull/1089#issuecomment-975993813

> So for CI, it shouldn't be an issue either

PR #1096, opened 2021-12-02
rsocket: fix raccept can't handle disconnected cm event 

raygift:
Application requires that nonblocking accept calls progress without
application intervention, similar to the connect calls.

So reuse existing connect_svc, add new rsocket returned by rs_accept()
to the connect_svc. Because connect service monitors the rsocket continuously,
so cm event such like RDMA_CM_EVENT_DISCONNECTED can be handled during
rsocket being polled in rpoll. This prevent server's acceptor be waiting
for new msg on rsocket after client exist without executing rclose/rshutdown.

For more informations about connect_svc addition and cleanup, see: 659c4f9

Signed-off-by: raygift <bgtickt@gmail>.com

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -1345,6 +1345,10 @@ int raccept(int socket, struct sockaddr *addr, socklen_t *addrlen)
 
 	if (addr && addrlen)
 		rgetpeername(new_rs->index, addr, addrlen);
+	/* The app can still drive the CM state on failure */
+	int save_errno = errno;
+	rs_notify_svc(&connect_svc, new_rs, RS_SVC_ADD_CM);
+	errno = save_errno;
 	return new_rs->index;
 }
 


librdmacm/rsocket.c
```diff
@@ -1345,6 +1349,10 @@ int raccept(int socket, struct sockaddr *addr, socklen_t *addrlen)
 
 	if (addr && addrlen)
 		rgetpeername(new_rs->index, addr, addrlen);
+	/* The app can still drive the CM state on failure */
+	int save_errno = errno;
+	rs_notify_svc(&accept_svc, new_rs, RS_SVC_ADD_CM);
+	errno = save_errno;
```
shefty, line 1355: The commit message needs more detail of the problem and how the change helps.  For example, see:

https://github.com/linux-rdma/rdma-core/commit/659c4f903a2aec8c60adfbb5a315b107e84acdd5

for when the connect_svc was added. I suspect the existing connect_svc can be re-used here, which avoids the possibility of starting up another thread, and is already coded for cleanup, which is missing from this patch.

raygift: The example is very useful for me to unstand connect svc!! thank very much. I think it can be reused to handle cm event for acceptor new_rs. And while reusing connect_svc, i think  no more cleanup codes is needed, right?

And the commit message is modified, please take a look

shefty: Correct - cleanup is already there for the connect_svc.  Commit message is better.  It may be worth referencing the original commit for the cleanup svc as part of the message, but someone interested could use git blame to find it fairly quickly.

raygift: Agreed, added refer for the original commit 659c4f9 in commit message.

Wondering whether the service name "connect_svc" would make someone else confused or not? As it's not only used by connect api calls, but also raccept reusing it now. Is there any suggestion or just keep this at present?

shefty: It's managing the connection state, so I wouldn't bother changing the name.

raygift: got it

rleon:
@shefty, Sean are you ok with the commit now?

Thanks

shefty:
@rleon - Yes, thanks.  Sorry for the delay, I was on vacation.

rleon:
> @rleon - Yes, thanks. Sorry for the delay, I was on vacation.

Thanks

PR #1107, opened 2021-12-09
providers/irdma: Validate input before memory window bind

Sindhu-Devale:
Check for a non null MR, address and length
before allowing bind operation.

Also, add check to make sure the MR and MW are in
same PD before posting a bind MW op.

Fixes: 14a0fc824f16 ("rdma-core/irdma: Implement device supported verb APIs")
Signed-off-by: Sindhu-Devale <sindhu.devale@intel.com>

### DIFF
--- a/providers/irdma/uverbs.c
+++ b/providers/irdma/uverbs.c
@@ -213,7 +213,7 @@ int irdma_ubind_mw(struct ibv_qp *qp, struct ibv_mw *mw,
 	struct ibv_send_wr *bad_wr;
 	int err;
 
-	if (vmr->mr_type != IBV_MR_TYPE_MR || mw->type != IBV_MW_TYPE_1)
+	if (vmr->mr_type != IBV_MR_TYPE_MR)
 		return ENOTSUP;
 
 	if (umr->acc_flags & IBV_ACCESS_ZERO_BASED)
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -2634,8 +2634,16 @@ static inline uint32_t ibv_inc_rkey(uint32_t rkey)
 static inline int ibv_bind_mw(struct ibv_qp *qp, struct ibv_mw *mw,
 			      struct ibv_mw_bind *mw_bind)
 {
+	struct ibv_mw_bind_info *bind_info = &mw_bind->bind_info;
+
 	if (mw->type != IBV_MW_TYPE_1)
 		return EINVAL;
+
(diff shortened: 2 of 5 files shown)


providers/irdma/uverbs.c
```diff
@@ -206,19 +206,28 @@ int irdma_ubind_mw(struct ibv_qp *qp, struct ibv_mw *mw,
...
 	int err;
 
-	if (vmr->mr_type != IBV_MR_TYPE_MR || mw->type != IBV_MW_TYPE_1)
-		return ENOTSUP;
-
-	if (umr->acc_flags & IBV_ACCESS_ZERO_BASED)
+	if (!bind_info->mr && (bind_info->addr || bind_info->length))
 		return EINVAL;
 
+	if (bind_info->mr) {
```
rleon, line 219: 1. Why do we need these checks in user-space and not in the driver?
It is preferable to have one place for the checks and drivers are supposed to have these type of checks anyway.
2. I imagine almost all drivers need these checks. Why irdma is special here?
3. You have double SOB for Tatyana in the commit message.
Signed-off-by: Signed-off-by: Tatyana Nikolova <tatyana.e.nikolova@intel.com>

shirazsaleem: > 1. Why do we need these checks in user-space and not in the driver?
>    It is preferable to have one place for the checks and drivers are supposed to have these type of checks anyway.
> 2. I imagine almost all drivers need these checks. Why irdma is special here?
> 3. You have double SOB for Tatyana in the commit message.
>    Signed-off-by: Signed-off-by: Tatyana Nikolova [tatyana.e.nikolova@intel.com](mailto:tatyana.e.nikolova@intel.com)

Hi Leon - Thanks for the review. Sindhu is out for the rest of the year. And will work on these review comments when she is back on Jan 4th.

Shiraz

rleon: Thanks for the update.

jgunthorpe: Bind isn't supposed to call to the kernel, but perhaps the checkes should be done in the core rdma-core code..

jgunthorpe (changes requested):
Address Leon's remarks

providers/irdma/uverbs.c
Sindhu-Devale: Will send out a patch. Thanks Jason/Leon.

rleon:
Looks good, if no one complains, I will merge it tomorrow, before rdma-core release.

Thanks

PR #1111, opened 2021-12-20
libhns: Remove support for HIP06

Wenpeng-Liang:
HIP06 is no longer supported, so the code of HIP06 is removed.


rleon:
Regarding removing HIP06, I don't know.

We usually drop support from the kernel, wait enough time and only after that removing user-space component.
You are doing it in opposite way,

Wenpeng-Liang:
Thank you, I will resubmit this PR after waiting for enough time. 

Related kernel mode patch:
[v2,for-next] RDMA/hns: Remove support for HIP06
https://patchwork.kernel.org/project/linux-rdma/patch/20211220130558.61585-1-liangwenpeng@huawei.com/

rleon:
So please remove the support drop patch from this PR.
I don't have an option in github UI to merge part of PR.

PR #1121, opened 2022-01-04
libhns: Fix the calculation of QP/SRQ table size

Wenpeng-Liang:
The table_size means the maximum number of QP/SRQ. This value may not be
a power of two. The old algorithm will lead to a result that allocates a
mismathed table.

Fixes: 887b78c ("libhns: Add initial main frame")
Fixes: 9e3df75 ("libhns: Support ibv_create_srq_ex")
Signed-off-by: Wenpeng Liang <liangwenpeng@huawei.com>

### DIFF
--- a/providers/hns/hns_roce_u.c
+++ b/providers/hns/hns_roce_u.c
@@ -92,6 +92,13 @@ static const struct verbs_context_ops hns_common_ops = {
 	.get_srq_num = hns_roce_u_get_srq_num,
 };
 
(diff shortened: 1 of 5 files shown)


rleon:
Please spell check the commit message.
----------------------------
The 'table_size' response from the kernel means the QP/SRQ max number. This value may not be a power of two and will lead the user to allocate a mismatched table.

rleon:
I fixed typo in commit message: "mismathed -> mismatched" and pushed manually.

PR #1122, opened 2022-01-10
Exploit s390's new and enhanced PCI memory I/O (MIO) instructions

niklas88:
Hi All,

In the existing MMIO implementation s390 relies on special syscalls to
access PCI memory. This was necessary as s390 originally only had
special privileged instructions for accessing PCI memory. With z15
however comes support for 4 new PCI memory I/O (MIO) instructions which
operate on virtually mapped PCI memory spaces.

While these are still special PCI access instructions instead of real
MMIO they behave much more like standard MMIO access. There is a load
like instruction _pcilgi_, a store like instruction _pcistgi_ a block store
variant for efficient memcpy _pcistbi_ and a write barrier instruction
_pciwb_. The load and store variants always operate on a 64 bit register
but only load/store the right most  bytes of the register controlled by
a length value in a paired register (even numbered register rN + odd
numbered register r(N+1)).

As the previous PCI instructions did not operate on virtual memory
mappings at all a kernel using them does not setup virtual memory
mappings and thus can't support user-space using the new instructions.

Also as use of PCI MIO instructions can be disabled via the
`pci=nomio` kernel parameter we can't solely rely on hardware support and
kernel version. Instead Linux exposes whether PCI MIO instructions are
enabled via an ELF hardware capability. With this patch we check for
this capability and if enabled use the newly introduced PCI MIO
instructions for MMIO access and barriers.

**Note to reviewers:** One thing I'm not sure about is the handling of the
ELF hardware capability in udma_barrier.h. Because that is used without
mmio.h/mmio.c at least in the compile test for DMA coherency support.
So I didn't want to add a dependency to the `extern s390_mio_supported`
variable. Instead I added a separate `getauxval()` call but that doesn't
seem ideal either, I couldn't measure any performance impact even on
a 100 Gbit/s card but still it's kind of unnecessary overhead.

Thanks,
Niklas

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -502,8 +502,11 @@ check_type_size("long" SIZEOF_LONG BUILTIN_TYPES_ONLY LANGUAGE C)
 # Determine if this arch supports cache coherent DMA. This isn't really an
 # arch specific property, but for our purposes arches that do not support it
 # also do not define wmb/etc which breaks our compile.
+# As a special case s390x always has coherent DMA but needs linking for its wmb
 CHECK_C_SOURCE_COMPILES("
+#if !defined(__s390x__)
 #include \"${CMAKE_CURRENT_SOURCE_DIR}/util/udma_barrier.h\"
+#endif
  int main(int argc,const char *argv[]) {return 0;}"
   HAVE_COHERENT_DMA)
 
@@ -608,6 +611,21 @@ RDMA_Check_C_Compiles(HAVE_GLIBC_FXSTAT "
  int main(int argc,const char *argv[]) {
  struct stat stat = {};  __fxstat(0, 0, &stat);  return 0;}")
 RDMA_DoFixup("${HAVE_GLIBC_FXSTAT}" "sys/stat.h")
+
+
+# glibc before 2.35 does not necesarily define the HWCAP_S390_PCI_MIO hardware
+# capability bit constant. Check for it and if necessary shim it in such that
+# kernel support for PCI MIO instructions can always be checked.
+RDMA_Check_C_Compiles(HAVE_GLIBC_HWCAP_S390_PCI_MIO "
+ #if defined(__s390x__)
+ #include <sys/auxv.h>
+ int main(int argc, const char *argv[]) {
+ return !!(getauxval(AT_HWCAP) & HWCAP_S390_PCI_MIO);}
+ #else
+ int main(int argc, const char *argv[]) {return 0;}
+ #endif
+ ")
+RDMA_DoFixup("${HAVE_GLIBC_HWCAP_S390_PCI_MIO}" "sys/auxv.h")
 
 #-------------------------
 # Build Prep
@@ -805,4 +823,7 @@ if (NOT HAVE_TARGET_SSE)
 endif()
 if (NOT DRM_INCLUDE_DIRS)
   message(STATUS " DMABUF NOT supported (disabling some tests)")
+endif()
+if (NOT HAVE_GLIBC_HWCAP_S390_PCI_MIO )
+  message(STATUS " Glibc version does not contain the HWCAP_S390_PCI_MIO bit, using shim version")
 endif()
--- a/util/CMakeLists.txt
+++ b/util/CMakeLists.txt
@@ -20,6 +20,7 @@ set(C_FILES
 if (HAVE_COHERENT_DMA)
   publish_internal_headers(util
     mmio.h
+    s390_mmio_insn.h
     udma_barrier.h
     )
 
--- a/util/mmio.c
+++ b/util/mmio.c
@@ -6,6 +6,73 @@
 #include <pthread.h>
 #include <stdbool.h>
 
+#ifdef __s390x__
+#include <sys/auxv.h>
+#include <ccan/minmax.h>
+
+bool s390_is_mio_supported;
+
+static __attribute__((constructor)) void check_mio_supported(void)
+{
+	s390_is_mio_supported = !!(getauxval(AT_HWCAP) & HWCAP_S390_PCI_MIO);
+}
+
+typedef void (*mmio_memcpy_x64_fn_t)(void *, const void *, size_t);
+/* This uses the STT_GNU_IFUNC extension to have the dynamic linker select the
+   best above implementations at runtime. */
+#if HAVE_FUNC_ATTRIBUTE_IFUNC
+void mmio_memcpy_x64(void *, const void *, size_t)
+	__attribute__((ifunc("resolve_mmio_memcpy_x64")));
+static mmio_memcpy_x64_fn_t resolve_mmio_memcpy_x64(uint64_t);
+#else
+__asm__(".type mmio_memcpy_x64, %gnu_indirect_function");
+write64_fn_t resolve_mmio_memcpy_64(uint64_t)
+	__asm__("mmio_memcpy_x64");
+#endif
+
+#define S390_MAX_WRITE_SIZE 128
+#define S390_BOUNDARY_SIZE (1 << 12)
+#define S390_BOUNDARY_MASK (S390_BOUNDARY_SIZE - 1)
+
+static uint8_t get_max_write_size(void *dst, size_t len)
+{
+	size_t offset = ((uint64_t __force)dst) & S390_BOUNDARY_MASK;
+	size_t size = min_t(int, len, S390_MAX_WRITE_SIZE);
+
+	if (likely(offset + size <= S390_BOUNDARY_SIZE))
+		return size;
+
+	return S390_BOUNDARY_SIZE - offset;
+}
+
+static void mmio_memcpy_x64_mio(void *dst, const void *src, size_t bytecnt)
+{
+	size_t size;
+
+	/* Input is 8 byte aligned 64 byte chunks. The alignment matches the
+	 * requirements of pcistbi but we must not cross a 4K byte boundary.
+	 */
+	while (bytecnt > 0) {
+		size = get_max_write_size(dst, bytecnt);
+		if (size > 8)
+			s390_pcistbi(dst, src, size);
+		else
+			s390_pcistgi(dst, *(uint64_t *)src, 8);
+		src += size;
+		dst += size;
+		bytecnt -= size;
+	}
+}
+
+mmio_memcpy_x64_fn_t resolve_mmio_memcpy_x64(uint64_t hwcap)
+{
+	if (hwcap & HWCAP_S390_PCI_MIO)
+		return &mmio_memcpy_x64_mio;
+	else
+		return &s390_mmio_write_syscall;
+}
+#endif /* __s390x__ */
+
 #if SIZEOF_LONG != 8
 
 static pthread_spinlock_t mmio_spinlock;
--- a/util/mmio.h
+++ b/util/mmio.h
@@ -67,58 +67,60 @@
    code is always identical.
 */
 #ifdef __s390x__
-#include <unistd.h>
-#include <sys/syscall.h>
-
-/* s390 requires a privileged instruction to access IO memory, these syscalls
-   perform that instruction using a memory buffer copy semantic.
-*/
-static inline void s390_mmio_write(void *mmio_addr, const void *val,
-				   size_t length)
-{
-	// FIXME: Check for error and call abort?
-	syscall(__NR_s390_pci_mmio_write, mmio_addr, val, length);
-}
-
-static inline void s390_mmio_read(const void *mmio_addr, void *val,
-				  size_t length)
-{
-	// FIXME: Check for error and call abort?
-	syscall(__NR_s390_pci_mmio_read, mmio_addr, val, length);
-}
+#include <util/s390_mmio_insn.h>
 
 #define MAKE_WRITE(_NAME_, _SZ_)                                               \
 	static inline void _NAME_##_be(void *addr, __be##_SZ_ value)           \
 	{                                                                      \
-		s390_mmio_write(addr, &value, sizeof(value));                  \
+		if (s390_is_mio_supported)                                     \
+			s390_pcistgi(addr, value, sizeof(value));              \
+		else                                                           \
+			s390_mmio_write_syscall(addr, &value, sizeof(value));  \
 	}                                                                      \
 	static inline void _NAME_##_le(void *addr, __le##_SZ_ value)           \
 	{                                                                      \
-		s390_mmio_write(addr, &value, sizeof(value));                  \
+		if (s390_is_mio_supported)                                     \
+			s390_pcistgi(addr, value, sizeof(value));              \
+		else                                                           \
+			s390_mmio_write_syscall(addr, &value, sizeof(value));  \
 	}
 #define MAKE_READ(_NAME_, _SZ_)                                                \
 	static inline __be##_SZ_ _NAME_##_be(const void *addr)                 \
 	{                                                                      \
 		__be##_SZ_ res;                                                \
-		s390_mmio_read(addr, &res, sizeof(res));                       \
+		if (s390_is_mio_supported)                                     \
+			res = s390_pcilgi(addr, sizeof(res));                  \
+		else                                                           \
+			s390_mmio_read_syscall(addr, &res, sizeof(res));       \
 		return res;                                                    \
 	}                                                                      \
 	static inline __le##_SZ_ _NAME_##_le(const void *addr)                 \
 	{                                                                      \
 		__le##_SZ_ res;                                                \
-		s390_mmio_read(addr, &res, sizeof(res));                       \
+		if (s390_is_mio_supported)                                     \
+			res = s390_pcilgi(addr, sizeof(res));                  \
+		else                                                           \
+			s390_mmio_read_syscall(addr, &res, sizeof(res));       \
 		return res;                                                    \
 	}
 
 static inline void mmio_write8(void *addr, uint8_t value)
 {
-	s390_mmio_write(addr, &value, sizeof(value));
+	if (s390_is_mio_supported)
+		s390_pcistgi(addr, value, sizeof(value));
+	else
+		s390_mmio_write_syscall(addr, &value, sizeof(value));
 }
 
 static inline uint8_t mmio_read8(const void *addr)
 {
 	uint8_t res;
-	s390_mmio_read(addr, &res, sizeof(res));
+
+	if (s390_is_mio_supported)
+		res = s390_pcilgi(addr, sizeof(res));
+	else
+		s390_mmio_read_syscall(addr, &res, sizeof(res));
+
 	return res;
 }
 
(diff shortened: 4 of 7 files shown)


rleon:
I honestly don't know what to do this PR.
It is all black magic for me :(.

niklas88:
> I honestly don't know what to do this PR. It is all black magic for me :(.

It is quite special so that's understandable. Sadly I can't currently refer you to a specification but I'm here to answer
questions. Also, while I tried to give a good summary in the commit message I do realize I might
have missed giving some context being very deep in the topic myself.

First some general s390 PCI and I/O background:
* s390 historically doesn't have MMIO but does have DMA. Instead of using MMIO, I/O is done with special I/O CPU
  instructions and so called [Channel I/O](https://en.wikipedia.org/wiki/Channel_I/O) devices of which some are backwards compatible to the original channel I/O devices which predate PCI by about 35 years. So a lot of history there.
* Today's channel devices use PCIe on the physical layer but this is entirely hidden from the OS.
* As there already was physical PCI, support for the OS talking to normal PCI devices was introduced. Because there
  was no concept of MMIO this was done with special PCI load/store instructions that refer to PCI BARs with a handle plus a 
  BAR number plus an offset not via a virtual address.
* In Linux the above old style PCI instructions are used with an indirect to adapt them for `ioread()`/`iowrite()`
  which of course refer to PCI BARs with an `__iomem` address. I.e. the PCI support fakes MMIO for the rest of the kernel.
* Because the above limits performance and is kind of an ugly hack in Linux new instructions were added
  that refer to PCI memory with a virtual address. These new style instructions behaves more like MMIO except that one
  can only access the mapped I/O memory with these select instructions. Any other access like a normal load
  from this mapping will result in an addressing exception. This also added support for Write Combining and
  an explicit barrier. Also unlike the old instructions these may be executed in user-space.
* The new PCI instructions are currently not available in KVM or z/VM guests so co-exist with the old style
  and we need to switch between them at boot time to support the same kernel image on both.

I can also give some more pointers to related code in the Linux Kernel:
* `arch/s390/pci/pci_insn.c`: This is where the PCI instructions are defined. Both the legacy ones (`_fh` suffix) and the new ones (`_mio` suffix).   As in this PR this uses the `.insn` ASM thing because the assembler doesn't know the mnemonic. The wrapping functions are then used to implement `ioread()`/`iowrite()` in the kernel via the following define/call chain:
```
iowrite64 → writeq → __raw_writeq (in include/asm-generic/io.h)
__raw_writeq → zpci_write_u64 (in arch/s390/include/io.h)
zpci_write_u64 → zpci_store  (in arch/s390/include/pci_io.h)
zpci_store → zpci_store_fh (old style) or __pcistg_mio (new style) (in arch/s390/pci/pci_insn.c)
```
* `arch/s390/pci/pci_mmio.c`: This is the implementation for the syscall that the current rdma-core code calls for s390. Note that this already had to be updated to work at all when the kernel uses the newer instructions, this was commit f058599e22d ("s390/pci: Fix s390_mmio_read/write with MIO"). Sadly that one required even darker magic to basically do what this patch does but from kernel space and as if it was executed by user-space. You can see the same `pcilgi`, `pcistgi` and `pcistbi` again encoded via the raw `.insn`  thing. They are however surrounded with extra incantations using an instruction called `sacf` (Set Address Control Fast) to execute the PCI loads/stores with the syscall invoking user-space address space. This of course adds overhead compared with the user-space implementation in this PR though our syscalls are quite fast so it's less than one might think.

util/mmio.c
```diff
@@ -6,6 +6,21 @@
 #include <pthread.h>
 #include <stdbool.h>
 
+#ifdef __s390x__
+#include <sys/auxv.h>
+
+int s390_mio_supported;
```
jgunthorpe, line 12: bool

util/mmio.h
```diff
@@ -69,10 +69,64 @@
 #ifdef __s390x__
 #include <unistd.h>
 #include <sys/syscall.h>
-
-/* s390 requires a privileged instruction to access IO memory, these syscalls
-   perform that instruction using a memory buffer copy semantic.
+#include <sys/auxv.h>
+
+/* s390 requires special instructions to access IO memory. Originally there
```
jgunthorpe, line 74: This stuff should not be compiled except on s390, and probably be in its own header too

util/mmio.h
```diff
@@ -69,10 +69,64 @@
...
+/* The following pcilgi and pcistgi instructions allow IO memory access from
+   user-space but are only available on z15 and newer.
 */
+static inline uint64_t s390_pcilgi(const void *ioaddr, size_t len)
+{
+	union register_pair ioaddr_len = {.even = (uint64_t)ioaddr, .odd = len};
+	uint64_t val;
+
+	asm volatile (
+		"       .insn   rre,0xb9d60000,%[val],%[ioaddr_len]\n"
```
jgunthorpe, line 100: Please put the actualy mmenoic form of the instruction in a comment...

util/mmio.h
```diff
@@ -90,38 +144,56 @@ static inline void s390_mmio_read(const void *mmio_addr, void *val,
 #define MAKE_WRITE(_NAME_, _SZ_)                                               \
 	static inline void _NAME_##_be(void *addr, __be##_SZ_ value)           \
 	{                                                                      \
-		s390_mmio_write(addr, &value, sizeof(value));                  \
+		if (likely(s390_mio_supported))                                \
```
jgunthorpe, line 147: Do not annotate with likely, you can't know at compile time which leg the running system will prefer but the branch predictor can..

util/mmio.h
```diff
@@ -206,11 +278,47 @@ __le64 mmio_read64_le(const void *addr);
...
+	int size;
+
+	if (unlikely(!s390_mio_supported))
+		s390_mmio_write(dst, src, bytecnt);
+
+	while (bytecnt > 0) {
+		size = s390_get_max_write_size((uint64_t __force) dst,
+					       (uint64_t) src, bytecnt,
+					       S390_PCI_MAX_WRITE_SIZE);
+		if (size > 8) /* main path */
```
jgunthorpe, line 305: x64 is always chunks of exactly 64 bytes, all this logic seems impossible

util/udma_barrier.h
```diff
@@ -194,10 +194,29 @@
...
+
+static inline void s390_pciwb(void)
+{
+#ifdef HWCAP_S390_PCI_MIO
+	/* As udma_barrier.h is stand alone we can't access
+	   the s390_mio_enabled variable used in mmio.h. Instead just check
+	   the ELF HWCAP directly which in testing seems to have very low
+	   overhead.
+	 */
+	if (!(getauxval(AT_HWCAP) & HWCAP_S390_PCI_MIO))
```
jgunthorpe, line 212: This should be preloaded too

jgunthorpe: This is also only meaningful with WC memory, does s390 have that?

util/mmio.h
niklas88: Ah, thanks, didn't realize this restriction. So you're saying the the bytecnt is always a multiple of 8, in that case yes we can greatly simplify this and get rid of the whole `s390_get_max_write_size()` thing.  I somehow read it as a general `mmio_memcpy()` and thought the x64 referred to the fact that it's only available on architectures with a 64 bit atomic.

jgunthorpe: Not 8 bytes, 64 bytes. This is designed to be optimal for entire cache lines

util/udma_barrier.h
niklas88: Yes we do. Do you have a suggestion for where to put the preloading?

util/mmio.h
niklas88: Makes sense, will do.

util/udma_barrier.h
jgunthorpe: You might just do all of this like we did sse on x86 with HAVE_FUNC_ATTRIBUTE_IFUNC

util/mmio.h
niklas88: Will do

util/udma_barrier.h
niklas88: I'll look into it. I'd assume since this is a linker thing it can't be used in a header inline function though because they get inline before the code is been by the linker so I would still need a `.c` file, no?

jgunthorpe: Yes, like we did for sse

jgunthorpe: You'd do it for s390_mmio_Write and inline the pcistgi stuff in the C file

util/mmio.h
niklas88: @jgunthorpe if this is designed for entire cache lines can I also assume that `dst` and `src` are 64 byte aligned?

util/udma_barrier.h
niklas88: Ok sure that's doable but what would I do here for the `mmio_flush_write()`?

jgunthorpe: Same thing, just one implementation is a NOP

util/mmio.h
jgunthorpe: I can't remember,  think it is only 8 bytes required.

util/udma_barrier.h
niklas88: I meant more which C file to put it in. The `udma_barrier.h` is also directly included in a compile test snippet for `HAVE_COHERENT_DMA` so that would then also have to compile that in.

jgunthorpe: Just make a new one for this s390 stuff. The compile test may need to be changed to avoid linking.

util/mmio.h
niklas88: Thanks

niklas88: Sorry for the confusion yesterday, I realized later that we still need another restriction currently ensured by `s390_get_max_write_size()` the `pcistbi` instruction can't normally write accross a 4096 byte boundary in the MMIO destination. So we will need some logic there.

util/udma_barrier.h
niklas88: @jgunthorpe I'm still wondering about IFUNC attribute vs a flag like in this version. With the flag I understand the disadvantage is the global variable I would however guess that it has less overhead because all the `readNN_be/le()` functions are inlined with a known length at compile-time. Also we don't have the function call overhead with the flag boiling down to a single load that as you correctly pointed out is likely well predicted by the branch predictor. How would you feel about the following setup.
* S390 PCI instruction wrappers in `util/s390_mmio.h` and the `MAKE_READ/WRITE` macros. This would be included in `util/mmio.h` and `util/udma_barrier.h` and would use a flag. As you said we need to avoid linking in the compile test either way.
* S390 `mmio_memcpy_x64()` in `util/mmio.c` either with IFUNC or the flag, here the function call overhead is spread over a larger data size and we usually don't know the length at compile time so inlining is less important. Personally I find the IFUNC harder to read than a flag, at least when it's not available as an attribute. Either way I don't think we need to have `mmio_memcpy_x64()` in a header.

niklas88:
@jgunthorpe I have pushed an updated version with the following changes:
* Moved PCI instruction wrapper functions to `util/s390_pci_insn.h`
* Dropped `likely()` around flag check
* Moved `mmio_memcpy_x64()` to `util/mmio.c`. Simplified `get_s390_max_write_size()`. The 8 byte alignment matches the requirements of `pcistbi` but we need to make sure not to cross a 4K boundary.
* Added flag check for `mmio_flush_writes()` and include `util/s390_pci_insn.h`, this needed a s390 special check for the compile test
* Added loop for retrying `pcistgi` in case hardware indicates being busy, this matches the kernel implementation but should be very rarely needed and isn't currently handled by the syscall implementation.

I did not use the IFUNC mechanism in this version because I think a flag is simpler and allows us save the function call overhead for the simple load/store case where I'd also expect more gain from inlining especially since the length parameter
will always be known at compile time for these.

util/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ set(C_FILES
 if (HAVE_COHERENT_DMA)
   publish_internal_headers(util
     mmio.h
+    s390_mmio_insn.h
```
jgunthorpe, line 23: Keep sorted please

util/mmio.c
```diff
@@ -6,6 +6,59 @@
...
+		return size;
+
+	return S390_BOUNDARY_SIZE - offset;
+}
+
+void mmio_memcpy_x64(void *dst, const void *src, size_t bytecnt)
+{
+	size_t size;
+
+	if (!s390_is_mio_supported)
```
jgunthorpe, line 43: See, this is a good place to use ifunc, you already too the function call, may as well land on the right function directly..

jgunthorpe (changes requested):

util/CMakeLists.txt
niklas88: @jgunthorpe I'm confused, this is alphabetically sorted. First letters are "m", "s", "u"?

niklas88: Sorry I'm a little confused. the `mmio.h`, `s390_mmio_insn.h`, and `udma_barrier.h` are alphabetically sorted and `util` is just the destination so needs to be the first argument to `publish_internal_headers()`. What am I missing?

jgunthorpe: Opps my mistake, it is right

niklas88:
I just pushed a new version with the following changes:

* Use `__attribute__((ifunc))` for `mmio_memcpy_x64()`. Talking to our toolchain team it turns out on s390 we actually get the ELF hwcaps as argument to the resolver function so that turns out nicely. The toolchain folks mentioned that on s390 there shouldn't really be a window where the attribute doesn't work but the binutils support does but I still included the fallback to be consistent with the x86_64 ifunc use.
* I signed the commit with GPG. Mostly as an exercise as I recently created this [key](https://keyserver.ubuntu.com/pks/lookup?search=schnelle%40linux.ibm.com&fingerprint=on&op=index). It does have signatures from the s390 kernel maintainers and some other IBM folks as well as my ancient University times keys though.
* I did some basic testing of the `mmio_memcpy_x64()` function as it doesn't seem to be called by a simple `qperf -cm1 rc_bw` test. Also tested the ifunc mechanism

util/mmio.c
niklas88: @jgunthorpe should be resolved in the newly pushed v3

util/mmio.c
```diff
@@ -6,6 +6,77 @@
...
+			s390_pcistgi(dst, *(uint64_t *)src, 8);
+		src += size;
+		dst += size;
+		bytecnt -= size;
+	}
+}
+
+mmio_memcpy_x64_fn_t resolve_mmio_memcpy_x64(uint64_t hwcap)
+{
+	if (hwcap & HWCAP_S390_PCI_MIO)
```
niklas88, line 73: Sorry, just noticed I completely forgot the check if `HWCAP_S390_PCI_MIO` is defined here. Will push a fixed version shortly

niklas88: Should be resolved in the latest version.

niklas88:
I just pushed a v5. It is rebased on current master and removes again the busy wait for `pcistgi`. It's pretty embarrassing but while such a busy wait is needed for the old PCI store it isn't for the new one and that was correct in my initial version. When I thought I had missed it I was looking at the wrong specification...

util/mmio.c
```diff
@@ -6,6 +6,81 @@
...
 
+#ifdef __s390x__
+#include <sys/auxv.h>
+#include <ccan/minmax.h>
+
+bool s390_is_mio_supported;
+
+static __attribute__((constructor)) void check_mio_supported(void)
+{
+#ifdef HWCAP_S390_PCI_MIO
```
jgunthorpe, line 17: I'd rather you use the usual shim mechanism for this constant to ensure it is always declared regardless of the libc version. ie have cmake test it and replace sys/auxv.h like we do for other things

util/udma_barrier.h
```diff
@@ -198,12 +198,21 @@
...
 #define mmio_flush_writes() asm volatile("dsb st" ::: "memory");
-#elif defined(__sparc__) || defined(__s390x__)
+#elif defined(__sparc__)
 #define mmio_flush_writes() asm volatile("" ::: "memory")
 #elif defined(__loongarch__)
 #define mmio_flush_writes() asm volatile("dbar 0" ::: "memory")
 #elif defined(__riscv)
 #define mmio_flush_writes() asm volatile("fence ow,ow" ::: "memory")
+#elif defined(__s390x__)
+#include "s390_mmio_insn.h"
```
jgunthorpe, line 208: Why is read and write being defined through udma_barrier.h ?

jgunthorpe (changes requested):
It seems fine, just a few nitpicky things

util/mmio.c
niklas88: Ok, thanks, will look into that.

util/udma_barrier.h
niklas88: I include the `s390_mmio_insn.h` for the `s390_is_mio_supported` flag. One way not to include the read/write would be to move the `MAKE_READ/MAKE_WRITE` to `mmio.h` and keep `s390_mmio_insn.h` just for the flag and the instruction wrappers. Would that be acceptable @jgunthorpe?

jgunthorpe: You can just add an extern for the flag, it doesn't have be in exactly one header. Just be sure to include all the headers that define the extern in the C file that instantiates it.

niklas88: @jgunthorpe I tried that, it causes a redundant redecleration warning in all files which include both.

niklas88: I tried that, sadly it causes a redundant redecleration warning in all files that include both `udma_barrier.h` and `mmio.h`.

niklas88: Now clearly we could work around this with an extra define much like a header guard such that `util/udma_barrier.h` or `util/s390_mmio_insn.h` depending on which comes first. But that's kind of ugly. No?

Personally I think, moving the MAKE_READ/MAKE_WRITE as well as the 8 bit read/write to `util/mmio.h` while including `util/s390_mmio_insn.h` is a clean solution. It even improves the match with the file name, keeping `s390_mmio_insn.h` strictly about the instructions. I also noticed I had the flag check in both the `s390_pciwb()` itself and again in `mmio_flush_writes()` with that removed I find it pretty clean. Basically in `util/udma_barrier.h` we're just left with `#define mmio_flush_writes() s390_pciwb()`

niklas88: I've pushed a v6 with the above idea for you to take a look.

util/mmio.c
niklas88: In v6 that I just pushed, I added a `buildlib/fixup-include/sys-auxv.h` with the `HWCAP_S390_PCI_MIO`. We'll likely have backports for glibc 2.34 for this flag in the major distributions with s390 support anyway but in the code the shim method is definitely nicer than the inline `#ifdef`.

util/udma_barrier.h
jgunthorpe: Oh blah, yes we have to deal with that extra warning in these headers.

jgunthorpe (commented):
It seems OK overall, just do the cmake thing please

CMakeLists.txt
```diff
@@ -609,6 +612,21 @@ RDMA_Check_C_Compiles(HAVE_GLIBC_FXSTAT "
  struct stat stat = {};  __fxstat(0, 0, &stat);  return 0;}")
 RDMA_DoFixup("${HAVE_GLIBC_FXSTAT}" "sys/stat.h")
 
+
+# glibc before 2.35 does not necesarily define the HWCAP_S390_PCI_MIO hardware
+# capability bit constant. Check for it and if necessary shim it in such that
+# kernel support for PCI MIO instructions can always be checked.
+RDMA_Check_C_Compiles(HAVE_GLIBC_HWCAP_S390_PCI_MIO "
```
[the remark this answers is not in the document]
niklas88: Added one at the bottom, wasn't sure before because there's already a Success/Failed message for the check itself.

PR #1123, opened 2022-01-11
rsocket: fix read_all/write_all to handle EINTR

typedefstruct32:
Rsocket uses socket pair to communicate with threads. Sometimes EINTR is encountered when performing blocking calls on fd.

So add a loop to check the return value and error to handle EINTR and artificially restart the interrupted calls

Signed-off-by: Yilun-Li <420463729@qq.com>

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -430,15 +430,15 @@ struct ds_udp_header {
 
 static void write_all(int fd, const void *msg, size_t len)
 {
-	// FIXME: if fd is a socket this really needs to handle EINTR and other conditions.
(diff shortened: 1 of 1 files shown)


rleon:
Please use real name in Signed-off-by line and write the commit message that explains why retry is needed and "failure" is not good.

The commit message shou,d be in the kernel style format:
-------------------
Title

Commit message body

Signed-off-by: ....
------------------------------

Thanks

jgunthorpe (changes requested):
Use a proper signed-off-by, use a loop not a goto

jgunthorpe:
Real names are needed in the signed-off by line, it is a legal statement.

PR #1124, opened 2022-01-13
rsocket: fix heap use after free in svc run

typedefstruct32:
When rs_svc_grow_sets() is executed, a new heap space is always allocated, and svc->contexts points to this new space.

However, when tcp_svc_run()/udp_svc_run()/cm_svc_run() 's do-while is executed, svc->contexts is not updated to fds/timeouts in time.This causes memory access error and affects the behavior of svc.

Signed-off-by: Yilun-Li <420463729@qq.com>


jgunthorpe (changes requested):
Please use a proper signed-off by, do not include unmerged commits in new PRs

PR #1125, opened 2022-01-16
bnxt_re/lib: Check pointer validity while freeing queue pointers

selvintxavier:
qp->jrqq can be NULL in SRQ case or when accessed from error path.
Avoid segfault by adding check before accessing qp->jrqq and qp->jsqq.

Fixes: f92837e29fd4 ("bnxt_re/lib: consolidate hwque and swque in common structure")
Signed-off-by: Selvin Xavier <selvin.xavier@broadcom.com>

### DIFF
--- a/providers/bnxt_re/verbs.c
+++ b/providers/bnxt_re/verbs.c
@@ -820,10 +820,14 @@ static int bnxt_re_check_qp_limits(struct bnxt_re_context *cntx,
 
 static void bnxt_re_free_queue_ptr(struct bnxt_re_qp *qp)
 {
-	free(qp->jrqq->hwque);
-	free(qp->jrqq);
-	free(qp->jsqq->hwque);
-	free(qp->jsqq);
+	if (qp->jrqq) {
+		free(qp->jrqq->hwque);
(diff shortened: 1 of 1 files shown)


rleon:
This free(qp->jrqq); and free(qp->jsqq); can be outside of if check.

selvintxavier:
Thanks Leon. 
@nmorey can you please pick this fix for stable/v36, stable/v37, stable/v38 branch also.
Thanks

rleon:
@selvintxavier, as a reminder https://github.com/linux-rdma/rdma-core/blob/master/Documentation/stable.md#option-1

selvintxavier:
> @selvintxavier, as a reminder https://github.com/linux-rdma/rdma-core/blob/master/Documentation/stable.md#option-1

ok. thanks @rleon . Will follow from next submission on-wards.

nmorey:
@selvintxavier As it has the proper fixes tag (and it applies cleanly on these branches), you're guaratneed that I'll pick them too.
v36/v37 updated. Had a branch glitch on v38. Will be pushed today

PR #1129, opened 2022-01-20
rsocket: fix forksafety bug

en4bz:
This fixes a bug where valid memory is unmapped in the child process due
to MADV_DONTFORK only working with page granularity. Valid memory may be
allocated by the allocator before/after rsocket's registered allocations
but in the same page causing such allocations to be unmapped after fork
and causing a segfault in the child.

The solution is simply to align allocations on page boundaries and round
the allocation size up to the next page boundary.

Signed-off-by: Ian Forbes <ian@forbes.io>

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -442,6 +442,25 @@ static void read_all(int fd, void *msg, size_t len)
 	assert(rc == len);
 }
 
+/**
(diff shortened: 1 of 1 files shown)


rleon:
Fixed title and coding style and pushed manually.

Thanks

PR #1132, opened 2022-01-27
libibverbs: consistently parse environment variables in ibverbs_init

nzmsv:
Prior to this change, setting some of the variables to "0" would be confusingly interpreted as enabling the feature.

### DIFF
--- a/libibverbs/init.c
+++ b/libibverbs/init.c
@@ -665,20 +665,14 @@ static void verbs_set_log_file(void)
 
 int ibverbs_init(void)
 {
-	char *env_value;
-
-	if (getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE"))
+	if (check_env("RDMAV_FORK_SAFE") || check_env("IBV_FORK_SAFE"))
 		if (ibv_fork_init())
 			fprintf(stderr, PFX "Warning: fork()-safety requested "
 				"but init failed\n");
 
-	/* Backward compatibility for the mlx4 driver env */
-	env_value = getenv("MLX4_DEVICE_FATAL_CLEANUP");
-	if (env_value)
-		verbs_allow_disassociate_destroy = strcmp(env_value, "0") != 0;
-
-	if (getenv("RDMAV_ALLOW_DISASSOC_DESTROY"))
-		verbs_allow_disassociate_destroy = true;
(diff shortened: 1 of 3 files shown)


rashikakheria (approved):

libibverbs/init.c
```diff
@@ -667,18 +667,27 @@ int ibverbs_init(void)
...
+		want_ibv_fork_init = env_value && (strcmp(env_value, "0") != 0);
+	}
+	if (want_ibv_fork_init)
 		if (ibv_fork_init())
 			fprintf(stderr, PFX "Warning: fork()-safety requested "
 				"but init failed\n");
 
-	/* Backward compatibility for the mlx4 driver env */
-	env_value = getenv("MLX4_DEVICE_FATAL_CLEANUP");
+	env_value = getenv("RDMAV_ALLOW_DISASSOC_DESTROY");
```
wzamazon, line 682: I think this part is from another commit?

nzmsv: No, I am trying to make all the variables behave the same way.

libibverbs/init.c
```diff
@@ -667,18 +667,27 @@ int ibverbs_init(void)
 {
 	char *env_value;
 
-	if (getenv("RDMAV_FORK_SAFE") || getenv("IBV_FORK_SAFE"))
+	bool want_ibv_fork_init = false;
+	env_value = getenv("RDMAV_FORK_SAFE");
+	want_ibv_fork_init = env_value && (strcmp(env_value, "0") != 0);
```
jgunthorpe, line 672: Please just make a function to query an environment variable and return a bool if it is to be true or false. Then it is just

`
want_ibv_work_init = ibv_check_env("RDMAV_FORK_SAFE") || ibv_check_env("IBV_FORK_SAFE");
`

jgunthorpe (changes requested):

libibverbs/init.c
nzmsv: Great suggestion!  Done.

util/util.c
```diff
@@ -45,3 +46,9 @@ unsigned int get_random(void)
 
 	return rand_r(&seed);
 }
+
+bool check_env(const char *var)
+{
+	const char *env_value = getenv(var);
```
rleon, line 52: Please add extra line here to follow kernel coding style.

Besides that, patch is ok.

Thanks

PR #1134, opened 2022-01-29
Do not use `__THROW` in prototypes

SoapGentoo:
`__THROW` is a glibc-internal macro, as indicated by the
preceding double underscore. The absence of this macro breaks
builds on musl.

Bug: https://bugs.gentoo.org/828894


rleon:
Yeah, __THROW makes sense only in header files which are used for C++.

PR #1136, opened 2022-02-07
cmake: Use PROJECT_SOURCE_DIR in cmake

Pusnow:
When using rdma-core via cmake's add_subdirectory, CMAKE_SOURCE_DIR
points project's root directory, and build fails. This patch replaces
CMAKE_SOURCE_DIR with PROJECT_SOURCE_DIR. It enables builing rdma-core
via add_subdirectory.

Signed-off-by: Wonsup Yoon <pusnow@kaist.ac.kr>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,26 +89,26 @@ set(IBVERBS_PROVIDER_SUFFIX "-rdmav${IBVERBS_PABI_VERSION}.so")
 
 # Override the CMAKE_INSTALL_ dirs to be under the build/ directory
 if (IN_PLACE)
-  set(CMAKE_INSTALL_SYSCONFDIR "${CMAKE_BINARY_DIR}/etc")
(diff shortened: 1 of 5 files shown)


rleon:
Don't you need to change CMAKE_BINARY_DIR to PROJECT_BINARY_DIR too?

Won't other CMAKE_* variables collide with your top level CmakeFile which included rdma-core?

Thanks

Pusnow:
You are right.  CMAKE_BINARY_DIR and other CMAKE variables should be changed to PROJECT versions. I'll fix them and ping you again.

Pusnow:
@rleon I've also changed CMAKE_BINARY_DIR to PROJECT_BINARY_DIR.
The other CMAKE_XXX_DIRs are  CURRENT directories (CMAKE_CURRENT_SOURCE_DIR and CMAKE_CURRENT_BINARY_DIR) , and they should not be changed (pointing currently processing directories).

rleon:
Thanks

PR #1138, opened 2022-02-16
tests: Modify eswitch mode checking

EdwardSro:
The devlink output behavior would be changed by having two outputs for
Eswitch modes, which are legacy and switchdev, instead of having three
outputs which are None, legacy, and switchdev. So, the legacy mode will be
the default mode.

So, the function that checks the Eswitch mode should be adjusted, since
it only checks the legacy mode in case the command returns a non zero
value, but with the newer devlinks, the command will return zero value for
both legacy and switchdev, so there is important to check the value of
the output.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1148,11 +1148,13 @@ def eswitch_mode_check(dev_name):
         raise unittest.SkipTest(f'Could not find the PCI device of {dev_name}')
     pci_name = pci_name[0].split('/')[5]
     try:
-        subprocess.check_output(['devlink', 'dev', 'eswitch', 'show', f'pci/{pci_name}'],
-                                stderr=subprocess.DEVNULL)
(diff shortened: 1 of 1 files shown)


rleon:
Do you have kernel commit that supports the claim of "going to be changed"?

Thanks

EdwardSro:
> Do you have kernel commit that supports the claim of "going to be changed"?
> 
> Thanks

The patch is going to be sent soon to upstream, once it's done I'll point to it.
This patch though supports both the current implementation and the new candidate and was sent as a precautionary measurement.
But we can wait for the kernel patch, no rush.

EdwardSro:
I see that the kernel patch was accepted: [https://patchwork.kernel.org/project/netdevbpf/patch/20220702190213.80858-7-saeed@kernel.org/]

EdwardSro:
@rleon can we merge this PR?

EdwardSro:
A similar change was already merged as part of a different PR.
Will close this.

PR #1139, opened 2022-02-17
Providers/rxe: Fix bugs in rxe_post_recv

rpearsonhpe:
Recent test case additions in tests/test_qpex.py surfaced three
API errors in rxe.c for post_recv. Two errors were returned with
the wrong sign and an immediate error was not returned if the qp
was in the reset state. This patch fixes these errors.

Signed-off-by: Bob Pearson <rpearsonhpe@gmail.com>


rpearsonhpe:
fixes errors in rxe.c exposed by test_qpex.py

PR #1140, opened 2022-02-18
srp_daemon: Detect proper path to systemctl

nmorey:
While debian uses /bin/systemctl, SUSE only supports /usr/bin/systemctl,
causing srp_daemon to fail silently starting on all ports.
Detect at built which path is the right one and use it.

Fixes: b03beb142e0d ("srp_daemon: Call systemctl properly from udev")

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,8 @@ else()
   set(CYTHON_EXECUTABLE "")
 endif()
 
+find_program(SYSTEMCTL_BIN systemctl HINTS "/usr/bin" "/bin")
+
 RDMA_CheckSparse()
 
 # Require GNU99 mode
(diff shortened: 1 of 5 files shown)


nmorey:
@Honggang-LI @bdrung 
Hopefully this patch should get the proper path for all distros. Could you check it does work well on yours ?

Honggang-LI:
> @Honggang-LI @bdrung Hopefully this patch should get the proper path for all distros. Could you check it does work well on yours ?

Tested it with fedora-35. It works for me.

Thanks

bdrung:
I tested building the master branch on Debian unstable, and `60-srp_daemon.rules` got evaluated to:
```
SUBSYSTEM=="infiniband_mad", KERNEL=="*umad*", PROGRAM=="SYSTEMCTL_BIN-NOTFOUND show srp_daemon -p ActiveState", RESULT=="ActiveState=active", ENV{SYSTEMD_WANTS}+="srp_daemon_port@$attr{ibdev}:$attr{port}.service"
```

This is caused by not having systemd installed in the minimal chroot environment. Please fall back to use `/bin/systemctl` if the binary is not found or provide a way to configure it during build.

rleon:
@nmorey, are you fixing it?

PR #1143, opened 2022-02-25
libibverbs: Check for valid completion channel before arming CQ

Sindhu-Devale:
Check for valid completion channel before arming the CQ.

Fixes: 91fc39561d04 ("Commit libibverbs code from roland-uverbs branch back onto trunk")

Signed-off-by: Sindhu Devale <sindhu.devale@intel.com>


jgunthorpe:
Why? This is a performance path, it should not have needless checks

PR #1144, opened 2022-02-28
Add new bitmap API

yishaih:
This series adds new bitmap implementation to util directory which replaces the ccan equivalent.

The ccan/bitmap stuff was published under the LGPLv2+ license which is not fitting in rdma-core.

As of the above, the applicable bitmap code in rdma-core was adapted to use the util new API.


providers/mlx5/buf.c
```diff
@@ -38,155 +38,38 @@
...
-{
-	int i;
+/* Only ia64 requires this */
+#ifdef __ia64__
+#define MLX5_SHM_ADDR ((void *)0x8000000000000000UL)
+#define MLX5_SHMAT_FLAGS (SHM_RND)
+#else
+#define MLX5_SHM_ADDR NULL
+#define MLX5_SHMAT_FLAGS 0
+#endif
```
jgunthorpe, line 53: What is this stuff?

yishaih: This comes from the deleted file providers/mlx5/bitmap.h to control shmat(), it's orthogonal to the bitmap changes.

PR #1154, opened 2022-03-15
srp_daemon: Prevent starting when the link layer is not InfiniBand

SergeyGorenko:
The srp_daemon uses the InfiniBand specific features and cannot work
with other networks. So let's check the link layer and prevent running
the srp_dameon with unsupported networks. It is needed to avoid running
redundant srp_daemon instances and noise in the logs. The change is
primarily helpful for hosts that mix InfiniBand and RoCE devices.

Signed-off-by: Sergey Gorenko <sergeygo@nvidia.com>

### DIFF
--- a/srp_daemon/srp_daemon.c
+++ b/srp_daemon/srp_daemon.c
@@ -1902,6 +1902,27 @@ static void umad_resources_destroy(struct umad_resources *umad_res)
 	umad_done();
 }
 
+static int check_link_layer(const char *port_sysfs_path)
+{
+	const char expected_link_layer[] = "InfiniBand";
+	char link_layer[sizeof(expected_link_layer)];
+	int ret;
+
+	ret = srpd_sys_read_string(port_sysfs_path, "link_layer", link_layer,
+				   sizeof(link_layer));
+	if (ret < 0) {
+		pr_err("Couldn't read link layer\n");
+		return ret;
+	}
+
+	if (strcmp(link_layer, expected_link_layer)) {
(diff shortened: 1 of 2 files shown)


rleon:
How will it work for mlx4 which can change link type in runtime?
As a scenario, mlx4 started as ethernet, but later changed to IB. Will user need to start srp_daemon manually?

Thanks

SergeyGorenko:
You are right. Manual srp_daemon restart will be required when the link type is changed in runtime. It doesn't look like an issue to me because the srp_daemon restart does not trigger ib_srp module reload and does not affect existing SRP connections. So, it is safe to restart srp_daemon. Moreover, the SRP configuration is not changed often. Once the SRP storage is configured, it is usually not touched for months or years.

Do you see any issue with srp_daemon restart?

rleon:
> You are right. Manual srp_daemon restart will be required when the link type is changed in runtime. It doesn't look like an issue to me because the srp_daemon restart does not trigger ib_srp module reload and does not affect existing SRP connections. So, it is safe to restart srp_daemon. Moreover, the SRP configuration is not changed often. Once the SRP storage is configured, it is usually not touched for months or years.
> 
> Do you see any issue with srp_daemon restart?

The manual part of "srp_daemon restart" worries me. I would expect that srp_daemon will use udev rules to trigger itself.

Thanks

rleon:
> The udev rule behavior looks like a bug to me. The bug is not related to my patch and should be fixed in a separate commit.

Can you please send a fix for that?

PR #1156, opened 2022-03-19
libhns: Extended QP supports the new post send mechanism

Wenpeng-Liang:
The ofed provides a new set of post send APIs for extended QP. With the new
APIs, users can post send WR more efficiently. The hns driver provides
support for the new APIs.

Signed-off-by: Xinhao Liu <liuxinhao5@hisilicon.com>
Signed-off-by: Yixing Liu <liuyixing1@huawei.com>
Signed-off-by: Wenpeng Liang <liangwenpeng@huawei.com>

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -338,6 +338,12 @@ struct hns_roce_qp {
 	unsigned long			flags;
 	int				refcnt; /* specially used for XRC */
 	void				*dwqe_page;
+
+	/* specific fields for the new post send APIs */
+	int				err;
+	void				*cur_wqe;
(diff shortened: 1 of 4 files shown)


providers/hns/hns_roce_u.h
```diff
@@ -362,6 +368,11 @@ struct hns_roce_u_hw {
 	struct verbs_context_ops hw_ops;
 };
 
+static inline unsigned long ROUND_UP(unsigned long n, unsigned long d)
```
rleon, line 371: I don't see any use of this function. Where do you use it?

Wenpeng-Liang: Thanks, this function is redundant, I will remove it.

PR #1157, opened 2022-03-19
cmake: Add -DSYSTEMCTL_BIN option to set path to systemctl

nmorey:
Make cmake fails if systemctl is not found automatically, unless
its path is manually set using -DSYSTEMCTL_BIN
This fixes build issues on debian systems where the minimal chroot does
not contain systemd.

Fixes: 05a19e050278 ("srp_daemon: Detect proper path to systemctl")
Reported-by: Benjamin Drung <benjamin.drung@ionos.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,9 @@ else()
 endif()
 
 find_program(SYSTEMCTL_BIN systemctl HINTS "/usr/bin" "/bin")
+if (NOT SYSTEMCTL_BIN)
+  set (SYSTEMCTL_BIN "/bin/systemctl")
+endif()
 
 RDMA_CheckSparse()
 


nmorey:
@bdrung Could you make sure this solves your issues?
I tried building for debian-11/debian-experimental using cbuild but both failed for other reasons:
```
$ ./buildlib/cbuild pkg debian-experimental
HEAD is now at b5b5359cbe15 cmake: Add -DSYSTEMCTL_BIN option to set path to systemctl
make: 'debian/rules' is up to date.
dh build --with python3,systemd --builddirectory=build-deb
dh: warning: Compatibility levels before 10 are deprecated (level 9 in use)
dh: error: unable to load addon python3: Can't locate Debian/Debhelper/Sequence/python3.pm in @INC (you may need to install the Debian::Debhelper::Sequence::python3 module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.34.0 /usr/local/share/perl/5.34.0 /usr/lib/x86_64-linux-gnu/perl5/5.34 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.34 /usr/share/perl/5.34 /usr/local/lib/site_perl) at (eval 10) line 1.
BEGIN failed--compilation aborted at (eval 10) line 1.

```

@rleon @jgunthorpe I added systemd to the AZP image to avoid specifying -DSYSTEMCTL_BIN for every single test.
Meaning that CI should fail, unless the image gets updated.
I locally ran build-images/pkg azp and it's all OK on my side

bdrung:
@nmorey That error message is caused by not having dh-python installed in your environment. Check Build-Depends from debian/control for the required dependencies.

nmorey:
> @nmorey That error message is caused by not having dh-python installed in your environment. Check Build-Depends from debian/control for the required dependencies.

@bdrung Shouldn't cbuild take care of it?

CMakeLists.txt
```diff
@@ -219,7 +219,12 @@ else()
   set(CYTHON_EXECUTABLE "")
 endif()
 
-find_program(SYSTEMCTL_BIN systemctl HINTS "/usr/bin" "/bin")
+find_program(SYSTEMCTL_BIN_DEFAULT systemctl HINTS "/usr/bin" "/bin")
```
rleon, line 222: I'm missing something important.

The issue is that debian doesn't have systemctl, but this code makes cmake to fail if systemctl doesn't exist. So what is this commit fixing?

jgunthorpe:
It is a big pain to make CI breaking changes like this, are you sure this is the best thing?

nmorey:
I feel like it's the cleanest way to handle this.
I see two way around this:
- Have a fallback to /bin if systemctl binary is not found. It should keep every one working, as long as distro with different path (like SUSE), use either the -D flag, or have systemctl installed
- Change all CI tests that break to use the -DSYSTEMCTL_BIN=/bin/systemctl
I'd rather have this one, as it will be easier to detect a distro built "badly" ( systemctl is not in /bin and not present at compile time)

bdrung:
Can you add `-DSYSTEMCTL_BIN=/bin/systemctl` to DH_AUTO_CONFIGURE in debian/rules?

nmorey:
@bdrung Done.

CMakeLists.txt
nmorey: As discussed in the PR comments, the goal of this is to not be able to built a broken package.
Either systemctl is present and we autodetect it, or the distro/build can specify the path to it.

jgunthorpe: We still support centos6, so breaking the build without systemd doesn't seem like an option? Confused by this remark

nmorey: My bad. I thought we only had systemd based distro left.

nmorey:
In the end, I just dropped the option and add a fallback to /bin/systemctl for system where it's not installed.
Hopefully fixes everything, and it works out of the box in CI

PR #1166, opened 2022-04-26
tests: Fixes and improvements

EdwardSro:
The series includes multiple fixes for some specific tests or flows and general improvements in the tests infrastructure.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -62,6 +62,11 @@ def set_rnr_attributes(qp_attr):
     qp_attr.rnr_retry = RNR_RETRY
 
 
+def is_gid_available(gid_index):
+    if gid_index is None:
+        raise unittest.SkipTest(f'No relevant GID found')
+
+
 class PyverbsAPITestCase(unittest.TestCase):
     def __init__(self, methodName='runTest'):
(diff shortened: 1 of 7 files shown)


tests/base.py
```diff
@@ -62,6 +62,11 @@ def set_rnr_attributes(qp_attr):
     qp_attr.rnr_retry = RNR_RETRY
 
 
+def is_gid_valid(gid_index):
```
rleon, line 65: I'm adding my comment here, just because github has no option to add a comment to commit messages.

This change needs more explanation. How is it possible that GID not valid? Isn't this situation indicate a problem and test should fail and not be skipped?

Thanks

EdwardSro: Sorry for the delayed reply...

It needs to be better explained, I agree.
Previously, if there were no GIDs (if RoCE is disabled for example), then the majority of the tests were getting skipped.
This is not the right behavior in all cases. rdmacm tests for instance can run regardless of the availability of the GIDs. This commit adjusts the infra, so the tests that get skipped if there is no GID are only the ones that MUST be skipped.
I'll fix the commit message and rename the "is_gid_valid"

PR #1167, opened 2022-04-27
mlx5: Fix check for SQ overflow in bind_mw

AlekseyMarchuk:
There is a typo in the check for end of SQ

Signed-off-by: Alexey Marchuk <alexeymar@nvidia.com>


AlekseyMarchuk:
There is a conflict between email address in singned-off and github account. I updated my email in github and pushed another version but the problem is still here. It seems that I need to create another PR, so closing this one

PR #1179, opened 2022-05-30
Add RDMA Atomic Write operation

yangx-jy:
1) Provide new Atomic Write API and make rxe provider support Atomic Write.
2) Add an example using the Atomic Write.
3) Provide python API for Atomic Write and add a python test using the python API.

Note: this PR depends on the corresponding kernel patch set.
https://lore.kernel.org/linux-rdma/617d7d26-e9a1-8919-1553-badf776a7f88@gmail.com/T/#t


jgunthorpe:
I'm not keen on more installed examples, is there a reason why we need the rdmacm based example and pyverbs testing is not enough?

yangx-jy:
> I'm not keen on more installed examples, is there a reason why we need the rdmacm based example and pyverbs testing is not enough?

Hi Jason,

Sorry for the late reply.  I just want to show two different usages (C and python APIs).  It's fine for me to only provide the pyverbs test.

tests/test_qpex.py
```diff
@@ -105,6 +105,12 @@ def create_qps(self):
             self.qps_num.append((recv_qp.qp_num, send_qp.qp_num))
             self.psns.append(random.getrandbits(24))
```
EdwardSro, line 107: leave two empty lines before class definition (PEP-8 convention)

tests/test_qpex.py
```diff
@@ -105,6 +105,12 @@ def create_qps(self):
...
 
+class QpExRCAtomicWrite(RCResources):
+    def create_qps(self):
+        create_qp_ex(self, e.IBV_QPT_RC, e.IBV_QP_EX_WITH_ATOMIC_WRITE)
+
+    def create_mr(self):
+        self.mr = u.create_custom_mr(self, e.IBV_ACCESS_REMOTE_WRITE)
```
EdwardSro, line 114: same

EdwardSro (changes requested):
Minor PEP-8 convention related comments.

jgunthorpe:
This needs to be rebased after addressing Edward's remarks

libibverbs/verbs.h
```diff
@@ -951,6 +951,7 @@ enum ibv_qp_create_send_ops_flags {
 	IBV_QP_EX_WITH_BIND_MW			= 1 << 8,
 	IBV_QP_EX_WITH_SEND_WITH_INV		= 1 << 9,
 	IBV_QP_EX_WITH_TSO			= 1 << 10,
+	IBV_QP_EX_WITH_ATOMIC_WRITE		= 1 << 12,
```
jgunthorpe, line 954: This whole change to verbs.h would be better in the previous patch.

pyverbs/libibverbs_enums.pxd
```diff
@@ -507,4 +510,4 @@ cdef extern from "<infiniband/tm_types.h>":
         IBV_TMH_NO_TAG
         IBV_TMH_RNDV
         IBV_TMH_FIN
-        IBV_TMH_EAGER
\ No newline at end of file
+        IBV_TMH_EAGER
```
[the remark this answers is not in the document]
yangx-jy: @EdwardSro 

I wonder why we need to add an empty line at EoF?  I think there is no empty line at EOF in other files.

yangx-jy:
> This needs to be rebased after addressing Edward's remarks

Done.

jgunthorpe:
Lets drop the rdamcm examples, or at least not install them

yangx-jy:
> Lets drop the rdamcm examples, or at least not install them

I prefer to keep the rdmacm example so I have removed the patch about installing it on different distributions.

Thank you very much for reviewing my atomic write patch set.

kernel-headers/rdma/ib_user_verbs.h
```diff
@@ -466,6 +466,7 @@ enum ib_uverbs_wc_opcode {
 	IB_UVERBS_WC_BIND_MW = 5,
```
jgunthorpe, line 466: This commit should be created by using the command

./kernel-headers/update ../linuxc2d939002934

Not by hand

jgunthorpe (changes requested):

PR #1180, opened 2022-06-01
Dev/stable v20/master

Hatlori9:
-- Could NOT find Systemd (missing: SYSTEMD_LIBRARIES LIBSYSTEMD_INCLUDE_DIRS)


rleon:
This is not how pull request should look

It needs to be based on branch to which it is pushed and with clear description in commit messages.

PR #1185, opened 2022-06-17
fix mem leak for port in cma dev

pim-pesochek:
cma_dev->port is allocated, but never freed

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -304,6 +304,8 @@ static void remove_cma_dev(struct cma_device *cma_dev)
 		ibv_dealloc_pd(cma_dev->pd);
 	if (cma_dev->verbs)
 		ibv_close_device(cma_dev->verbs);
+	if (cma_dev->port)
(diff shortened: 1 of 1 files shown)


librdmacm/cma.c
```diff
@@ -304,6 +304,8 @@ static void remove_cma_dev(struct cma_device *cma_dev)
 		ibv_dealloc_pd(cma_dev->pd);
 	if (cma_dev->verbs)
 		ibv_close_device(cma_dev->verbs);
+	if (cma_dev->port)
```
rleon, line 307: Hi,

Thanks for the patch. It is right one, but needs some small changes before merge.

1. We follow kernel coding style rules. It means that patch should have proper commit message, title, Fixes line and real Signed-off-by tag.
2. There is no need to if (not null) check before free.

Title: cma: Release allocated port array
,,,,
Fixes: 0e3e5c527008 ("librdmacm: Make CMA device list to be dynamic")
Signed-off-by: ....

Thanks

PR #1189, opened 2022-06-26
pyverbs: add property to get cmid's dev name and port num

changchengx:
- After ```rdma bind/resolve addr``` operation, it can get the dev name and port num.
Add API to get these information.

There's below scenario to use rdmacm:
1. create rdma_cm_id
2. resolve addr and route to server
3. get which device is used, create extra ibv_context and ibv_pd to create QP
4. send QP info in the private data space when calling rdma_connect.
5. rdma_connect failed. e.g. get rejected event such as that server doesn't start running the service.
6. repeat 1 & 2 & 4 until get the response event from server. There's no need to go through the 3 step again. 

- fix error when creating Context with cmid without setting device name

### DIFF
--- a/pyverbs/cmid.pyx
+++ b/pyverbs/cmid.pyx
@@ -359,6 +359,10 @@ cdef class CMID(PyverbsCM):
             raise PyverbsError('Unrecognized object type')
 
     @property
+    def dev_name(self):
+        return str(v.ibv_get_device_name(self.id.verbs.device).decode('utf-8'))
+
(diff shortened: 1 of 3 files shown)


changchengx:
Whatever in python2 or python3, it uses below method to always return ```str``` type for the device name
```str(v.ibv_get_device_name(self.id.verbs.device).decode('utf-8'))```

EdwardSro (approved):

PR #1190, opened 2022-06-26
pyverbs: add set rdmacm private_data API

changchengx:
support set rdamcm private data for rdma_connect/accept

Signed-off-by: Liu, Changcheng <changcheng.liu@aliyun.com>

### DIFF
--- a/pyverbs/cmid.pyx
+++ b/pyverbs/cmid.pyx
@@ -821,12 +821,18 @@ cdef class CMID(PyverbsCM):
         if ret != 0:
             raise PyverbsRDMAErrno('Failed to set option')
 
-    def reject(self, private_data=None, private_data_len=0):
+    def reject(self, private_data=None):
         """
         Reject a connection or datagram service lookup request.
         :param private_data: Optional private data to send with the reject message.
         :param private_data_len: Size (in bytes) of the private data being sent.
         """
-        ret = cm.rdma_reject(self.id, <const void*>private_data, private_data_len)
+        data_len = len(private_data) if private_data else 0
+        buffer = ctypes.create_string_buffer(data_len)
+        if (private_data):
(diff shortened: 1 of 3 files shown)


pyverbs/cmid.pyx
```diff
@@ -65,6 +65,10 @@ cdef class ConnParam(PyverbsObject):
     def private_data(self):
         return <object>self.conn_param.private_data
 
+    def set_private_data(self, data, data_len):
```
EdwardSro, line 68: I suggest that you add the option to set the private data in the __init__

changchengx: @EdwardSro Add parameter data_len in the **init**

EdwardSro:
There's something wrong with the ctype data / set_private_data implementation.
Looks like it crashes the cython code in run-time (even though set_private_data is not directly called).

changchengx:
@EdwardSro could you tell me which case trigger the crash?
>
>it crashes the cython code in run-time (even though set_private_data is not directly called).
>

pyverbs/cmid.pyx
```diff
@@ -20,7 +21,7 @@ from pyverbs.cq cimport WC
 cdef class ConnParam(PyverbsObject):
 
     def __init__(self, resources=1, depth=1, flow_control=0, retry=5,
-                 rnr_retry=5, srq=0, qp_num=0):
+                 rnr_retry=5, srq=0, qp_num=0, data_len = 0):
```
EdwardSro, line 24: data_len=0
(according to PEP-8 convention - no spaces around '=' here)

changchengx: @EdwardSro Done.

EdwardSro (approved):

PR #1206, opened 2022-08-18
Changes from Debian upload 42.0-1

bdrung:
* Update overrides for lintian 2.115.2
  * Bump Standards-Version to 4.6.1

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,7 +2,26 @@ rdma-core (43.0-1) unstable; urgency=medium
 
   * New upstream release.
 
- -- Jason Gunthorpe <jgg@obsidianresearch.com>  Thu, 27 Jan 2022 20:08:33 +0100
+ -- Jason Gunthorpe <jgg@obsidianresearch.com>  Thu, 18 Aug 2022 16:02:54 +0200
+
+rdma-core (42.0-1) unstable; urgency=medium
+
+  * New upstream release.
+  * Update overrides for lintian 2.115.2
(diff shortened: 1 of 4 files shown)


bdrung:
The Azure pipelines need to update their lintian version.

rleon:
> 



> The Azure pipelines need to update their lintian version.

It needs to be done in this PR or before it. I can't merge code that fails.

Thanks

bdrung:
I didn't find a way to (easily) configure that in `buildlib/azure-pipelines.yml`. The new lintian version is only in Debian for now. Soon it will be in Ubuntu, but in Ubuntu >= 22.10.

jgunthorpe:
This is really obnoxious that lintian needs upgrading - this has never happened before, can we avoid it somehow?

bdrung:
The lintian output should not change too often and therefore the override won't need adjustments. To align with the Debian package in Debian, lintian from testing or unstable should be used. So can the pipeline run the lintian test inside a Debian testing/unstable container?

jgunthorpe:
I mean, why can't we have an override that is the union of what debian testing and ubuntu can use?

Adding more containers is difficultish, and keeping something rolling like debian testing up to date is not so practical.

We could try to squeeze an updated lintian into the azp container, but that seems hard and maybe not really worth it?

We could also drop the lintian tests from the CI?

I'm not sure?

rleon:
Closed due to missing activity

PR #1207, opened 2022-08-18
Update kernel headers

rleon:


CMakeLists.txt
```diff
@@ -819,6 +829,9 @@ endif()
 if (NOT HAVE_C_WORKING_SHADOW)
   message(STATUS " -Wshadow does NOT work")
 endif()
+if (NOT HAVE_C_VARIABLE_SIZE)
```
jgunthorpe, line 832: This doesn't seem quite right.. I think I would just drop this entire statement. The purpose of the summary block at the end is to highlight defects in the environment that might be problematic. Supressing a warning isn't an issue..

rleon: done

PR #1208, opened 2022-08-19
.gitignore: Add generated files during Debian package build

bdrung:
When building a Debian package in place, several files will be created in `debian/`. Add them to `.gitignore`.


jgunthorpe:
Should this be in debian/.gitignore ?

bdrung:
Okay, I can put it there.

PR #1209, opened 2022-08-23
tests: add error flow tests for qp creation

shuki-zanyovka:
Add coverage for the error conditions of QP
creation limits check.

Signed-off-by: Shuki Zanyovka <shukiz@amazon.com>

### DIFF
--- a/tests/test_qp.py
+++ b/tests/test_qp.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
 # Copyright (c) 2019 Mellanox Technologies, Inc. All rights reserved. See COPYING file
 # Copyright (c) 2020 Kamal Heib <kamalheib1@gmail.com>, All rights reserved.  See COPYING file
+# Copyright 2020-2022 Amazon.com, Inc. or its affiliates. All rights reserved.
 
 """
 Test module for pyverbs' qp module.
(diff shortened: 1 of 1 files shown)


rleon:
These fields are FW controlled ones.
They can or can't be equal to 0xFFFFFFFF

shuki-zanyovka:
> 

This is the maximum value possible.
We are trying to simulate an illegal value error to improve the coverage of this test in each of the tests added in this PR.

rleon:
ok, i will rephrase my question.
Will this test pass if my FW sets 0x1 in all these cap fields?
In such case, create_qp() will fail.

shuki-zanyovka:
> ok, i will rephrase my question. Will this test pass if my FW sets 0x1 in all these cap fields? In such case, create_qp() will fail.

In this case:
1) create_qp() will fail.
2) The tests will still pass (see the exception handling per test).

rleon:
[leonro@vm ~]$ ~/src/rdma-core/build/bin/run_tests.py -k create_raw_qp
ssF....ss
FAIL: test_create_raw_qp_ex_with_illegal_caps_max_inline_data (tests.test_qp.QPTest)
Test Raw Packet QP creation via ibv_create_qp_ex with a QPAttr object with illegal max_inline_data.
  File "/home/leonro/src/rdma-core/tests/test_qp.py", line 228, in test_create_raw_qp_ex_with_illegal_caps_max_inline_data
    with self.assertRaises(PyverbsRDMAError) as ex:
AssertionError: PyverbsRDMAError not raised

Ran 9 tests in 0.113s

I tried it on CX-6 card that in RoCE mode.

shuki-zanyovka:
> 

Hi @rleon ,

I see the issue, looking into a way to improve this.

Thanks,
Shuki

shuki-zanyovka:
Hi @rleon ,

Please see the new commit here, I removed the illegal max_inline_data test completely and added some code to actually query the device for the device capabilities to decide whether or not to run each of the tests.

rleon:
Thanks

PR #1211, opened 2022-08-29
tests: Fix test_create_ud_qp_ex_{no,with}_attr skipped randomly

zhijianli88:
It's noticed that test_create_ud_qp_ex_{no,with}_attr with RXE device will
be skipped randomly:
$ while true; do ./bin/run_tests.py --dev rxe_eth0 --gid 1 tests.test_qp.QPTest.test_create_ud_qp_ex_no_attr tests.test_qp.QPTest.test_create_ud_qp_ex_with_attr ; done 2>&1 | grep OK
OK
OK
OK (skipped=1)
OK
OK (skipped=1)
OK
OK
OK
OK
OK (skipped=2)
OK

That's because the tests will pass a random create_flags(either 0 or 2) to
create_qp_ex() in where RXE doesn't support any create_flags so far.


tests/utils.py
```diff
@@ -309,7 +309,9 @@ def random_qp_init_attr_ex(attr_ex, attr, qpt=None):
         qp_cap.max_send_wr = MAX_RAW_PACKET_SEND_WR
     sig = random.randint(0, 1)
     mask = random_qp_create_mask(qpt, attr_ex)
-    if mask & e.IBV_QP_INIT_ATTR_CREATE_FLAGS:
+    is_rxe = attr.vendor_id == 0xffffff
+    # RXE doesn't support any create_flags
+    if mask & e.IBV_QP_INIT_ATTR_CREATE_FLAGS and not is_rxe:
```
rleon, line 314: I see that EFA doesn't support create_flags. How does it work?
```

   596 static int efa_qp_validate_attr(struct efa_dev *dev,
   597                                 struct ib_qp_init_attr *init_attr)
   598 {
....
   611         if (init_attr->create_flags) {
   612                 ibdev_dbg(&dev->ibdev, "Unsupported create flags\n");
   613                 return -EOPNOTSUPP;
   614         }

```

zhijianli88: Well, I don't have a EFA  hardware,  I only tested mxl5 and RXE
If so, EFA would have the same testing results as the RXE.

zhijianli88: I also checked cxgb4 irdma bnxt_re hns irdma mthca vmw_pvrdma, they all don't support create_flags. 
Currently, AFAIK mlx4 and mlx5 supported create_flags.

how about making it mlx4/mlx5 specific instead?

mrgolin: We also see it skipped on EFA.
Possible solution can be using pytest markers and fixtures to allow granularity in tests execution.

rleon: @EdwardSro ???

zhijianli88:
```diff
[root@iaas-rpma build]# git diff
diff --git a/tests/utils.py b/tests/utils.py
index 9d5b1b1af55d..287a1b438373 100644
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -309,9 +309,9 @@ def random_qp_init_attr_ex(attr_ex, attr, qpt=None):
         qp_cap.max_send_wr = MAX_RAW_PACKET_SEND_WR
     sig = random.randint(0, 1)
     mask = random_qp_create_mask(qpt, attr_ex)
-    is_rxe = attr.vendor_id == 0xffffff
-    # RXE doesn't support any create_flags
-    if mask & e.IBV_QP_INIT_ATTR_CREATE_FLAGS and not is_rxe:
+    is_mlx = attr.vendor_id == 0x02c9
+    # Most HCAs doesn't support any create_flags so far except mlx4/mlx5 so far
+    if mask & e.IBV_QP_INIT_ATTR_CREATE_FLAGS and is_mlx:
         cflags = random_qp_create_flags(qpt, attr_ex)
     else:
         cflags = 0
```

rleon: I don't think that it is correct.
The "_with" test should continue to generate flags and expect error if they are not supported, or expect success if they supported.

zhijianli88: > I don't think that it is correct. The "_with" test should continue to generate flags and expect error if they are not supported, or expect success if they supported.

The create_flags are restricted strongly, only a very few attr + create_flags combinations will be returned by random_qp_create_flags() currently.
Based on this, both "no" and "with" are designed to expect success only, so do you want to change it to expect error as well? If so, Does it mean we have to enumerate all possible valid or invalid combinations per HCA hardwares ?

rleon: I don't think so. I think that random_qp_create_flags should be changed to be random_valid_qp_create_flags() and extended with mlnx vendor_id check. It will make sure that it will always return valid bits.

And please use MLNX_VENDOR_ID instead of open-coded variant.

zhijianli88: @rleon , just updated a new versions as you suggested

PR #1217, opened 2022-09-07
tests: poll_cq: Don't pass wc.status to PyverbsRDMAError

zhijianli88:
When testing fails, we got below confusing error message due to the wrong errno:
    u.rdma_traffic(client, server, self.iters, self.gid_index, self.ib_port,
  File "rdma-core/tests/utils.py", line 1013, in rdma_traffic
    poll_cq(client.cq)
  File "rdma-core/tests/utils.py", line 577, in poll_cq
    raise PyverbsRDMAError('Completion status is {s}'.
pyverbs.pyverbs_error.PyverbsRDMAError: Completion status is local QP operation error. Errno: 2, No such file or directory

That's because PyverbsRDMAError(msg, errno) accept the 2nd parameter as errno.

### DIFF
--- a/tests/mlx5_base.py
+++ b/tests/mlx5_base.py
@@ -234,26 +234,27 @@ class Mlx5DcStreamsRes(Mlx5DcResources):
             except PyverbsRDMAError as ex:
                 self.stream_check = False
 
-    def bad_flow_handler_qp(self, qp_idx, ex, reset=False):
+    def bad_flow_handler_qp(self, qp_idx, status, reset=False):
         str_id = self.get_stream_id(qp_idx)
         bt_stream = (1 << str_id)
-        if isinstance(ex, PyverbsRDMAError):
-            if ex.error_code == e.IBV_WC_LOC_PROT_ERR:
-                self.qp_stream_errors[qp_idx][1] += 1
-                if (self.qp_stream_errors[qp_idx][0] & bt_stream) != 0:
-                    raise PyverbsError(f'Dublicate error from stream id {str_id}')
-                self.qp_stream_errors[qp_idx][0] |= bt_stream
-            if ex.error_code == e.IBV_WC_WR_FLUSH_ERR:
-                qp_attr, _ = self.qps[qp_idx].query(e.IBV_QP_STATE)
-                if qp_attr.cur_qp_state == e.IBV_QPS_ERR and reset:
-                    if self.qp_stream_errors[qp_idx][1] != self.dcis[qp_idx]['errored']:
-                        msg = f'QP {qp_idx} in ERR state with wrong number of counter'
-                        raise PyverbsError(msg)
-                    self.reset_qp(qp_idx)
-                    self.qp_stream_errors[qp_idx][2] = True
+
+        if status == e.IBV_WC_LOC_PROT_ERR:
+            self.qp_stream_errors[qp_idx][1] += 1
+            if (self.qp_stream_errors[qp_idx][0] & bt_stream) != 0:
+                raise PyverbsError(f'Dublicate error from stream id {str_id}')
+            self.qp_stream_errors[qp_idx][0] |= bt_stream
+        if status == e.IBV_WC_WR_FLUSH_ERR:
+            qp_attr, _ = self.qps[qp_idx].query(e.IBV_QP_STATE)
+            if qp_attr.cur_qp_state == e.IBV_QPS_ERR and reset:
+                if self.qp_stream_errors[qp_idx][1] != self.dcis[qp_idx]['errored']:
+                    msg = f'QP {qp_idx} in ERR state with wrong number of counter'
+                    raise PyverbsError(msg)
+                self.reset_qp(qp_idx)
+                self.qp_stream_errors[qp_idx][2] = True
+
         return True
 
-    def bad_flow_handling(self, qp_idx, ex, reset=False):
+    def bad_flow_handling(self, qp_idx, status, reset=False):
         if self.bad_flow == DCI_TEST_GOOD_FLOW:
             return False
         if self.bad_flow == DCI_TEST_BAD_FLOW_WITH_RESET:
@@ -262,7 +263,7 @@ class Mlx5DcStreamsRes(Mlx5DcResources):
                 self.dci_reset_stream_id(qp_idx)
             return True
         if self.bad_flow == DCI_TEST_BAD_FLOW_WITHOUT_RESET:
-            return self.bad_flow_handler_qp(qp_idx, ex, reset)
+            return self.bad_flow_handler_qp(qp_idx, status, reset)
         return False
 
     def set_bad_flow(self, bad_flow):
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -558,8 +558,7 @@ def poll_cq(cq, count=1, data=None):
         for wc in tmp_wcs:
             if wc.status != e.IBV_WC_SUCCESS:
                 raise PyverbsRDMAError('Completion status is {s}'.
-                                       format(s=wc_status_to_str(wc.status)),
-                                       wc.status)
+                                       format(s=wc_status_to_str(wc.status)))
```
EdwardSro, line 561: you're right, that's confusing.
But this change, without further adjustments, will break some tests.
For example, in test_qpex there's the following check:
```
        try:
            u.poll_cq(client.cq)
            if ex.error_code != e.IBV_WC_REM_ACCESS_ERR:
                raise ex
```
You may also find similar cases in some bad flow tests.

EdwardSro: One solution is to generalize the error string by passing "strerror_func" to  PyverbsRDMAError which translates the error code to a string. The os.strerror function would stay the default (converting errno->errno str).

zhijianli88: > But this change, without further adjustments, will break some tests.

Yes, you are right, my mistake.

EdwardSro: > Sorry, I didn't get you. do you mean we need to change the original PyverbsRDMAError ? Actually, I don't want to change the original PyverbsRDMAError class.

I meant to extend it, so it would take strerror_func as a parameter (in the initialization, defaulted to os.strerror).
Create a WC status to str function and pass it when raising that exception.
This is a more general solution that could be also used in other similar cases.

> Instead, I introduced a _poll_cq for tests, _poll_cq will return the error WC or the all successful WCs. So if the tests that want to check the wc.status, they should use _poll_cq instead.

I can stick with this for now. I'll extend the PyverbsRDMAError in the future if needed.

tests/test_qpex.py
```diff
@@ -311,11 +312,10 @@ def test_qp_ex_rc_bind_mw(self):
...
         client.qp.wr_set_sge(client_sge)
         client.qp.wr_complete()
-        try:
-            u.poll_cq(client.cq)
-        except PyverbsRDMAError as ex:
-            if ex.error_code != e.IBV_WC_REM_ACCESS_ERR:
-                raise ex
+        wcs = u._poll_cq(client.cq)
+        if wcs[0].status != e.IBV_WC_REM_ACCESS_ERR:
+            raise PyverbsRDMAError('Completion status is {s}'.
```
EdwardSro, line 317: it's preferable to use f-strings (i.e. f'Completion status is {wc_status_to_str(wc[0].status)}' )

tests/utils.py
```diff
@@ -572,6 +571,25 @@ def poll_cq(cq, count=1, data=None):
...
+    :param cq: CQ to poll from
+    :param count: How many completions to poll
+    :param data: In case of a work request with immediate, the immediate data
+                 to be compared after poll
+    :return: An array of work completions of length <count>, None
+             when events are used
+    """
+    wcs = _poll_cq(cq, count, data)
+    if wcs[0].status != e.IBV_WC_SUCCESS:
+        raise PyverbsRDMAError('Completion status is {s}'.
```
EdwardSro, line 589: it's preferable to use f-strings

tests/mlx5_base.py
```diff
@@ -411,11 +412,15 @@ def traffic_with_bad_flow(client, server, iters, gid_idx, port):
...
                 try:
-                    u.poll_cq(client.cq)
+                    wcs = u._poll_cq(client.cq)
                 except PyverbsError as ex:
-                    if client.bad_flow_handling(qp_idx, ex, True):
+                    if client.bad_flow_handling(qp_idx, e.IBV_WC_SUCCESS, True):
                         continue
                     raise ex
+                if wcs[0].status != e.IBV_WC_SUCCESS:
+                        if client.bad_flow_handling(qp_idx, wcs[0].status, True):
```
EdwardSro, line 421: looks like there's an extra indentation here

tests/mlx5_base.py
```diff
@@ -411,11 +412,15 @@ def traffic_with_bad_flow(client, server, iters, gid_idx, port):
...
                        ah_client, False)
                 try:
-                    u.poll_cq(client.cq)
+                    wcs = u._poll_cq(client.cq)
                 except PyverbsError as ex:
-                    if client.bad_flow_handling(qp_idx, ex, True):
+                    if client.bad_flow_handling(qp_idx, e.IBV_WC_SUCCESS, True):
                         continue
                     raise ex
+                if wcs[0].status != e.IBV_WC_SUCCESS:
```
EdwardSro, line 420: shouldn't his check replace line 417?

EdwardSro (changes requested):

tests/test_qpex.py
zhijianli88: Sounds good

tests/utils.py
zhijianli88: Sounds good

tests/mlx5_base.py
zhijianli88: Good catch.

tests/mlx5_base.py
zhijianli88:
```
414                 try:                                                                                                                                                                                                               
415                     wcs = u._poll_cq(client.cq)                                 
416                 except PyverbsError as ex:                                      
417                     if client.bad_flow_handling(qp_idx, e.IBV_WC_SUCCESS, True):
418                         continue                                                
419                     raise ex                                                    
420                 if wcs[0].status != e.IBV_WC_SUCCESS:                           
421                     if client.bad_flow_handling(qp_idx, wcs[0].status, True):   
422                         continue   
```
In my understanding, line 417 is in a try...except block, in where it cannot get a correct wcs.

EdwardSro: Oh, I see what you did there. you can leave it as is.

tests/utils.py
```diff
@@ -572,6 +571,25 @@ def poll_cq(cq, count=1, data=None):
 
     return wcs
```
EdwardSro, line 573: please add another line here before module-level function definition (PEP convention)

tests/utils.py
```diff
@@ -572,6 +571,25 @@ def poll_cq(cq, count=1, data=None):
...
+                 to be compared after poll
+    :return: An array of work completions of length <count>, None
+             when events are used
+    """
+    wcs = _poll_cq(cq, count, data)
+    if wcs[0].status != e.IBV_WC_SUCCESS:
+        raise PyverbsRDMAError('Completion status is {s}'.
+                               format(s=wc_status_to_str(wcs[0].status)))
+
+    return wcs
```
EdwardSro, line 593: also new line here

EdwardSro (changes requested):

EdwardSro (approved):

PR #1224, opened 2022-09-14
efa: Fix queue creation size uint32_t overflow

shuki-zanyovka:
Fix possible overflow when calculating sq_desc_cnt and rq_desc_cnt and fallback to
QP_DESC_COUNT_MAX=BIT(31) which is the maximum power of 2 in a uint32_t variable.

Signed-off-by: Shuki Zanyovka <shukiz@amazon.com>
Change-Id: If8041a9845d7460d04c7e14a7d738230465d2276

### DIFF
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -49,6 +49,9 @@ static bool is_buf_cleared(void *buf, size_t len)
 
 #define is_reserved_cleared(reserved) is_buf_cleared(reserved, sizeof(reserved))
 
+/* Maximum descriptors per QP (due to sizeof(uint32_t) and the fact that it must be a power of 2) */
+#define QP_DESC_COUNT_MAX BIT(31)
+
 struct efa_wq_init_attr {
 	uint64_t db_mmap_key;
 	uint32_t db_off;
@@ -1158,19 +1161,34 @@ static void efa_setup_qp(struct efa_context *ctx,
 			 struct ibv_qp_cap *cap,
 			 size_t page_size)
 {
-	uint16_t rq_desc_cnt;
+	uint32_t rq_desc_cnt_u32;
+	uint64_t sq_desc_cnt;
+	uint64_t rq_desc_cnt;
 
 	efa_qp_init_indices(qp);
 
-	qp->sq.wq.wqe_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr,
-						     ctx->min_sq_wr));
+	sq_desc_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));
+	if (sq_desc_cnt > QP_DESC_COUNT_MAX) {
+		sq_desc_cnt = QP_DESC_COUNT_MAX;
+		verbs_warn(verbs_get_ctx(qp->verbs_qp.qp.context),
+			   "sq_desc_cnt (%" PRIu64 ") > (%u), setting to 0x%x\n",
+			   sq_desc_cnt, QP_DESC_COUNT_MAX, QP_DESC_COUNT_MAX);
+	}
+	qp->sq.wq.wqe_cnt = (uint32_t)sq_desc_cnt;
 	qp->sq.wq.max_sge = cap->max_send_sge;
 	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;
 
 	qp->rq.wq.max_sge = cap->max_recv_sge;
 	rq_desc_cnt = roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);
-	qp->rq.wq.desc_mask = rq_desc_cnt - 1;
-	qp->rq.wq.wqe_cnt = rq_desc_cnt / qp->rq.wq.max_sge;
+	if (rq_desc_cnt > QP_DESC_COUNT_MAX) {
+		rq_desc_cnt = QP_DESC_COUNT_MAX;
+		verbs_warn(verbs_get_ctx(qp->verbs_qp.qp.context),
+			   "sq_desc_cnt (%" PRIu64 ") > (%u), setting to 0x%x\n",
+			   rq_desc_cnt, QP_DESC_COUNT_MAX, QP_DESC_COUNT_MAX);
+	}
+	rq_desc_cnt_u32 = (uint32_t)rq_desc_cnt;
+	qp->rq.wq.desc_mask = rq_desc_cnt_u32 - 1;
+	qp->rq.wq.wqe_cnt = rq_desc_cnt_u32 / qp->rq.wq.max_sge;
 
 	qp->page_size = page_size;
 }


rleon:
Please remove ChangeID

providers/efa/verbs.c
```diff
@@ -1158,19 +1161,34 @@ static void efa_setup_qp(struct efa_context *ctx,
...
 
 	efa_qp_init_indices(qp);
 
-	qp->sq.wq.wqe_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr,
-						     ctx->min_sq_wr));
+	sq_desc_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));
+	if (sq_desc_cnt > QP_DESC_COUNT_MAX) {
+		sq_desc_cnt = QP_DESC_COUNT_MAX;
+		verbs_warn(verbs_get_ctx(qp->verbs_qp.qp.context),
+			   "sq_desc_cnt (%u) > (%u), setting to 0x%x\n",
```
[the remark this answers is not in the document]
shuki-zanyovka: Fixed.

providers/efa/verbs.c
```diff
@@ -1158,19 +1161,34 @@ static void efa_setup_qp(struct efa_context *ctx,
...
 	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;
 
 	qp->rq.wq.max_sge = cap->max_recv_sge;
 	rq_desc_cnt = roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);
-	qp->rq.wq.desc_mask = rq_desc_cnt - 1;
-	qp->rq.wq.wqe_cnt = rq_desc_cnt / qp->rq.wq.max_sge;
+	if (rq_desc_cnt > QP_DESC_COUNT_MAX) {
+		rq_desc_cnt = QP_DESC_COUNT_MAX;
+		verbs_warn(verbs_get_ctx(qp->verbs_qp.qp.context),
+			   "sq_desc_cnt (%lu) > (%u), setting to 0x%x\n",
```
rleon, line 1186: It is still not correct, in 32bits system, uint64_t is %llu.
You should use %PRid64.

And please remove changeid

shuki-zanyovka: Fixed.

rleon:
It is my third time that I'm asking you to remove ChangeID.

shuki-zanyovka:
> It is my third time that I'm asking you to remove ChangeID.

Fixed.

providers/efa/verbs.c
```diff
@@ -1158,19 +1161,34 @@ static void efa_setup_qp(struct efa_context *ctx,
...
+	uint32_t rq_desc_cnt_u32;
+	uint64_t sq_desc_cnt;
+	uint64_t rq_desc_cnt;
 
 	efa_qp_init_indices(qp);
 
-	qp->sq.wq.wqe_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr,
-						     ctx->min_sq_wr));
+	sq_desc_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));
+	if (sq_desc_cnt > QP_DESC_COUNT_MAX) {
```
rleon, line 1171: How is it possible?

You call to efa_setup_qp() after successful efa_check_qp_limits().
In the latter function you have something like that:

  1217         if (attr->cap.max_send_wr > ctx->max_sq_wr) {
  1218                 verbs_err(&ctx->ibvctx,
  1219                           "Max send WR %u > %u\n", attr->cap.max_send_wr,
  1220                           ctx->max_sq_wr);
  1221                 return EINVAL;
  1222         }
  1223 
  1224         if (attr->cap.max_recv_wr > ctx->max_rq_wr) {
  1225                 verbs_err(&ctx->ibvctx,
  1226                           "Max receive WR %u > %u\n", attr->cap.max_recv_wr,
  1227                           ctx->max_rq_wr);
  1228                 return EINVAL;
  1229         }

I expect that your FW provides proper max_sq_wr and min_sq_wr values from the beginning.
If it is due to FW limitation, it needs to be fixed in the kernel. If there is no such limitation, this patch will be redundant,

shuki-zanyovka: @rleon ,
Normally, as you say, it wouldn't be a problem ==> This is a fix for an issue found in our provider/efa code when bypassing the efa_check_qp_limits() with invalid values.

rleon: How did you bypass efa_check_qp_limits()?
Do you have such flow in the driver?
Why don't you "fix" your efa_check_qp_limits?

shuki-zanyovka: Bypassed efa_check_qp_limits() as part of a test.

Simply:
static int efa_check_qp_limits(struct efa_context *ctx,
			       struct ibv_qp_init_attr_ex *attr)
{
	**if (bypass_enable)
		return 0;**

	if (attr->cap.max_send_sge > ctx->max_sq_sge) {
		verbs_err(&ctx->ibvctx,
			  "Max send SGE %u > %u\n", attr->cap.max_send_sge,
			  ctx->max_sq_sge);
		return EINVAL;
	}

rleon: You "deleted" very important part of QP initialization flow and now expect that other providers/efa functions will work correctly. I don't understand, why do you expect from us to merge it?

shuki-zanyovka: @rleon , This code is a safer initialization sequence of the sq and rx descriptors count. Why not merging it?

rleon: > @rleon , This code is a safer initialization sequence of the sq and rq descriptors count. Why not merging it?

Because this function doesn't exist as standalone and part of proper QP initialization flow. In that flow, QP validation is an important part that you can't skip.

Writing code for something that can't be is bad as it hurts readability later.

shuki-zanyovka: roundup_pow_of_two() returns a uint64_t:
static inline uint64_t roundup_pow_of_two(uint64_t n)

The 64-bit result is assigned below to a 16-bit variable in the original code:
	uint16_t rq_desc_cnt;

rq_desc_cnt = roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);

This code will yield a compilation warning if we enable a higher level of compilation warning.

Note also that (cap->max_recv_sge * cap->max_recv_wr) could be larger than MAX_UINT16 (== 0xFFFF) in case cap->max_recv_sge and cap->max_recv_wr are updated ==> This is unrelated to the test that revealed the issue.

How is this code less readable when it has more comments than the previous code? :)

rleon: > roundup_pow_of_two() returns a uint64_t: static inline uint64_t roundup_pow_of_two(uint64_t n)
> 
> The 64-bit result is assigned below to a 16-bit variable in the original code: uint16_t rq_desc_cnt;
> 
> rq_desc_cnt = roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);
> 
> and in this line, to a 32-bit variable (qp->sq.wq.wqe_cnt): qp->sq.wq.wqe_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));

Everything above should be fixed by returning proper value from FW and having right qp_vaildate_limit function.

> 
> Both lines will yield a compilation warning if we enable a higher level of compiler warning.

But we don't enable.

> 
> Note also that (cap->max_recv_sge * cap->max_recv_wr) could be larger than MAX_UINT16 (== 0xFFFF) in case cap->max_recv_sge and cap->max_recv_wr are updated ==> This is unrelated to the test that revealed the issue.
> 
> How is this code less readable when it has more comments than the previous code? :)

It gives false assumption that such failure is real and needs to be treated specially.

shuki-zanyovka: So how about adding another function in utils.h and use it in provider/efa/verbs.c:

#define MAX_LEADING_ZEROS (sizeof(uint32_t) * 8)
static inline uint32_t roundup_pow_of_two_32bit(uint32_t n)
{
        uint32_t shifts = (MAX_LEADING_ZEROS  - __builtin_clz(n));
        if (shifts == MAX_LEADING_ZEROS)
               shifts = MAX_LEADING_ZEROS - 1;
        return 1UL << shifts;
}

shuki-zanyovka: OK, so how about the following code (to handle only the uint32_t issue):
static void efa_setup_qp(struct efa_context *ctx,
			 struct efa_qp *qp,
			 struct ibv_qp_cap *cap,
			 size_t page_size)
{
	uint32_t rq_desc_cnt_u32;

	efa_qp_init_indices(qp);

	**qp->sq.wq.wqe_cnt = (uint32_t)roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));**
	qp->sq.wq.max_sge = cap->max_send_sge;
	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;

	qp->rq.wq.max_sge = cap->max_recv_sge;
	**rq_desc_cnt_u32 = (uint32_t)roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);**
	qp->rq.wq.desc_mask = rq_desc_cnt_u32 - 1;
	qp->rq.wq.wqe_cnt = rq_desc_cnt_u32 / qp->rq.wq.max_sge;

	qp->page_size = page_size;
}

shuki-zanyovka: > 

1. We can add a check for 0.
2. The function returns uint32_t.

shuki-zanyovka: > I don't think so.
> 
>     1. The result of __builtin_clz is undefined if n == 0.
> 
>     2. You are returning unsigned long, when uint32_t is unsigned int.

Also, regarding the previous proposed scheme,
1.  We can add a check for 0.
2. The function returns a uint32_t.

rleon: > The question is, if this scheme looks like something you can accept...?

I'm missing an answer on "why should I merge it?".

Thanks

mrgolin:
@rleon Our initial intention was making efa_setup_qp() more robust as its current implementation has internal calculations that, although very unlikely, can be tackled by external (seemingly valid) inputs.
Agree with the approach of keeping the code lean as possible but we should try to avoid inexplicit failure behavior as well.
@shuki-zanyovka suggestion was adding a "roundup_and_limit" function to improve robustness with minimal compromise on simplicity.

providers/efa/verbs.c
shuki-zanyovka: On the receive queue: 
qp->rq.wq.desc_mask defined as a uint16.
uint16_t efa_wq::desc_mask

On our NIC, we have:
ctx->max_rq_wr 32768
ctx->max_rq_sge 3

==> The following condition will pass with no error if attr->cap.max_recv_wr == 32767 and max_recv_sge == 3
32767 * 3 = 98301 > MAX(uint16) ==> **It will yield an overflow error**.
```
	if (attr->cap.max_recv_sge > ctx->max_rq_sge) {
		verbs_err(&ctx->ibvctx,
			  "Max receive SGE %u > %u\n", attr->cap.max_recv_sge,
			  ctx->max_rq_sge);
		return EINVAL;
	}

	if (attr->cap.max_recv_wr > ctx->max_rq_wr) {
		verbs_err(&ctx->ibvctx,
			  "Max receive WR %u > %u\n", attr->cap.max_recv_wr,
			  ctx->max_rq_wr);
		return EINVAL;
	}
```

On the send queue, we have:
On the send queue:
We have uint32_t efa_wq::wqe_cnt and 
ctx->max_sq_wr 4096

Indeed, in this scenario there wouldn't be an issue for most values.

rleon:
> @rleon Our initial intention was making efa_setup_qp() more robust as its current implementation has internal calculations that, although very unlikely, can be tackled by external (seemingly valid) inputs. Agree with the approach of keeping the code lean as possible but we should try to avoid inexplicit failure behavior as well. @shuki-zanyovka suggestion was adding a "roundup_and_limit" function to improve robustness with minimal compromise on simplicity.

And what will it give? rdma-core runs as part of user application, mistake there will cause to segfault in application. Why is it different from any other user mistakes that one can do?

rleon:
> > > @rleon Our initial intention was making efa_setup_qp() more robust as its current implementation has internal calculations that, although very unlikely, can be tackled by external (seemingly valid) inputs. Agree with the approach of keeping the code lean as possible but we should try to avoid inexplicit failure behavior as well. @shuki-zanyovka suggestion was adding a "roundup_and_limit" function to improve robustness with minimal compromise on simplicity.
> > 
> > 
> > And what will it give? rdma-core runs as part of user application, mistake there will cause to segfault in application. Why is it different from any other user mistakes that one can do?
> 
> @rleon , The mistake in this case will result in a wrong value sent to the EFA firmware and not a segfault / error or whatever in the application. Please see the last comment above.

I saw the comment and just gave more severe outcome.

The wrong values that were sent to FW can happen anytime, as I can use my own version of rdma-core (without any checks) and my own kernel. Your FW must be protected from such flows.

shuki-zanyovka:
> > > > @rleon Our initial intention was making efa_setup_qp() more robust as its current implementation has internal calculations that, although very unlikely, can be tackled by external (seemingly valid) inputs. Agree with the approach of keeping the code lean as possible but we should try to avoid inexplicit failure behavior as well. @shuki-zanyovka suggestion was adding a "roundup_and_limit" function to improve robustness with minimal compromise on simplicity.
> > > 
> > > 
> > > And what will it give? rdma-core runs as part of user application, mistake there will cause to segfault in application. Why is it different from any other user mistakes that one can do?
> > 
> > 
> > @rleon , The mistake in this case will result in a wrong value sent to the EFA firmware and not a segfault / error or whatever in the application. Please see the last comment above.
> 
> I saw the comment and just gave more severe outcome.
> 
> The wrong values that were sent to FW can happen anytime, as I can use my own version of rdma-core (without any checks) and my own kernel. Your FW must be protected from such flows.

@rleon,
The firmware *supports* these values and these flows but they are not sent correctly to the firmware.

mrgolin:
@rleon How we can move further with this?

shuki-zanyovka:
> @mrgolin @shuki-zanyovka can you please summarize what you're trying to fix here? You started with a custom provider that bypasses qp limits check and it got more and more complicated over time..

@galpress ,


On the receive queue:
qp->rq.wq.desc_mask defined as a uint16.
uint16_t efa_wq::desc_mask

On the EFA, we have:
ctx->max_rq_wr 32768
ctx->max_rq_sge 3

==> The following condition will pass with no error if attr->cap.max_recv_wr == 32767 and max_recv_sge == 3
32767 * 3 = 98301 > MAX(uint16) ==> It will yield an overflow error.

	if (attr->cap.max_recv_sge > ctx->max_rq_sge) {
		verbs_err(&ctx->ibvctx,
			  "Max receive SGE %u > %u\n", attr->cap.max_recv_sge,
			  ctx->max_rq_sge);
		return EINVAL;
	}

	if (attr->cap.max_recv_wr > ctx->max_rq_wr) {
		verbs_err(&ctx->ibvctx,
			  "Max receive WR %u > %u\n", attr->cap.max_recv_wr,
			  ctx->max_rq_wr);
		return EINVAL;
	}

On the send queue, we have:
On the send queue:
We have uint32_t efa_wq::wqe_cnt and
ctx->max_sq_wr 4096

Indeed, in this scenario there wouldn't be an issue for the values in the valid range.

galpress:
```
if (attr->cap.max_recv_wr > ctx->max_rq_wr) {
	verbs_err(&ctx->ibvctx,
		  "Max receive WR %u > %u\n", attr->cap.max_recv_wr,
		  ctx->max_rq_wr);
	return EINVAL;
}
```

For RQs max_rq_wr is misleading..
The driver reports the max RQ depth in units of descriptors, so the comparison of max_recv_wr and max_rq_wr is broken as they're in different units.

I believe the fix here should be changing efa_check_qp_limits() to test whether max_recv_wr * max_rq_sge is greater than max_rq_wr? I'm pretty sure I even had such patch laying around somewhere, maybe you can find it.

shuki-zanyovka:
> max_rq_wr

Does it mean we need to compare it to: roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr) ?

I.e.

max_recv_descs = roundup_pow_of_two(attr->cap.max_recv_sge * attr->cap.max_recv_wr);
if (max_recv_descs  > ctx->max_rq_wr) {
	verbs_err(&ctx->ibvctx,
		  "Max receive WR %u > %u\n", attr->cap.max_recv_wr,
		  ctx->max_rq_wr);
	return EINVAL;
}

This might be a solution for this issue - but still, I think it makes sense to change the variables from uint16_t to uint32_t just to be on the safe side.

We might not need the code to fix the result of roundup_pow_of_two() if it overflows.

galpress:
> > max_rq_wr
> 
> Does it mean we need to compare it to: roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr) ?
> 
> I.e.
> 
> max_recv_descs = roundup_pow_of_two(attr->cap.max_recv_sge * attr->cap.max_recv_wr); if (max_recv_descs > ctx->max_rq_wr) { verbs_err(&ctx->ibvctx, "Max receive WR %u > %u\n", attr->cap.max_recv_wr, ctx->max_rq_wr); return EINVAL; }
> 

Yes, something along these lines looks correct.

> This might be a solution for this issue - but still, I think it makes sense to change the variables from uint16_t to uint32_t just to be on the safe side.

I would start by pushing what fixes the issue.

shuki-zanyovka:
> > > max_rq_wr
> > 
> > 
> > Does it mean we need to compare it to: roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr) ?
> > I.e.
> > max_recv_descs = roundup_pow_of_two(attr->cap.max_recv_sge * attr->cap.max_recv_wr); if (max_recv_descs > ctx->max_rq_wr) { verbs_err(&ctx->ibvctx, "Max receive WR %u > %u\n", attr->cap.max_recv_wr, ctx->max_rq_wr); return EINVAL; }
> 
> Yes, something along these lines looks correct.
> 
> > This might be a solution for this issue - but still, I think it makes sense to change the variables from uint16_t to uint32_t just to be on the safe side.
> 
> I would start by pushing what fixes the issue.

@gal-pressman , We're looking at the solution you propose here, but still - even if we take your solution, can we also apply the following patch here to be on the safe side (to handle only the uint32_t issue)?

static void efa_setup_qp(struct efa_context *ctx,
			 struct efa_qp *qp,
			 struct ibv_qp_cap *cap,
			 size_t page_size)
{
	uint32_t rq_desc_cnt_u32;

	efa_qp_init_indices(qp);

	**qp->sq.wq.wqe_cnt = (uint32_t)roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));**
	qp->sq.wq.max_sge = cap->max_send_sge;
	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;

	qp->rq.wq.max_sge = cap->max_recv_sge;
	**rq_desc_cnt_u32 = (uint32_t)roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);**
	qp->rq.wq.desc_mask = rq_desc_cnt_u32 - 1;
	qp->rq.wq.wqe_cnt = rq_desc_cnt_u32 / qp->rq.wq.max_sge;

	qp->page_size = page_size;
}

gal-pressman:
> > > > max_rq_wr
> > > 
> > > 
> > > Does it mean we need to compare it to: roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr) ?
> > > I.e.
> > > max_recv_descs = roundup_pow_of_two(attr->cap.max_recv_sge * attr->cap.max_recv_wr); if (max_recv_descs > ctx->max_rq_wr) { verbs_err(&ctx->ibvctx, "Max receive WR %u > %u\n", attr->cap.max_recv_wr, ctx->max_rq_wr); return EINVAL; }
> > 
> > 
> > Yes, something along these lines looks correct.
> > > This might be a solution for this issue - but still, I think it makes sense to change the variables from uint16_t to uint32_t just to be on the safe side.
> > 
> > 
> > I would start by pushing what fixes the issue.
> 
> @gal-pressman , We're looking at the solution you propose here, but still - even if we take your solution, can we also apply the following patch here to be on the safe side (to handle only the uint32_t issue)?
> 
> ```
> static void efa_setup_qp(struct efa_context *ctx,
> 			 struct efa_qp *qp,
> 			 struct ibv_qp_cap *cap,
> 			 size_t page_size)
> {
> 	uint32_t rq_desc_cnt_u32;
> 
> 	efa_qp_init_indices(qp);
> 
> 	**qp->sq.wq.wqe_cnt = (uint32_t)roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));**
> 	qp->sq.wq.max_sge = cap->max_send_sge;
> 	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;
> 
> 	qp->rq.wq.max_sge = cap->max_recv_sge;
> 	**rq_desc_cnt_u32 = (uint32_t)roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);**
> 	qp->rq.wq.desc_mask = rq_desc_cnt_u32 - 1;
> 	qp->rq.wq.wqe_cnt = rq_desc_cnt_u32 / qp->rq.wq.max_sge;
> 
> 	qp->page_size = page_size;
> }
> ```

Please explain the issue?

shuki-zanyovka:
> > > > > max_rq_wr
> > > > 
> > > > 
> > > > Does it mean we need to compare it to: roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr) ?
> > > > I.e.
> > > > max_recv_descs = roundup_pow_of_two(attr->cap.max_recv_sge * attr->cap.max_recv_wr); if (max_recv_descs > ctx->max_rq_wr) { verbs_err(&ctx->ibvctx, "Max receive WR %u > %u\n", attr->cap.max_recv_wr, ctx->max_rq_wr); return EINVAL; }
> > > 
> > > 
> > > Yes, something along these lines looks correct.
> > > > This might be a solution for this issue - but still, I think it makes sense to change the variables from uint16_t to uint32_t just to be on the safe side.
> > > 
> > > 
> > > I would start by pushing what fixes the issue.
> > 
> > 
> > @gal-pressman , We're looking at the solution you propose here, but still - even if we take your solution, can we also apply the following patch here to be on the safe side (to handle only the uint32_t issue)?
> > ```
> > static void efa_setup_qp(struct efa_context *ctx,
> > 			 struct efa_qp *qp,
> > 			 struct ibv_qp_cap *cap,
> > 			 size_t page_size)
> > {
> > 	uint32_t rq_desc_cnt_u32;
> > 
> > 	efa_qp_init_indices(qp);
> > 
> > 	**qp->sq.wq.wqe_cnt = (uint32_t)roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));**
> > 	qp->sq.wq.max_sge = cap->max_send_sge;
> > 	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;
> > 
> > 	qp->rq.wq.max_sge = cap->max_recv_sge;
> > 	**rq_desc_cnt_u32 = (uint32_t)roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);**
> > 	qp->rq.wq.desc_mask = rq_desc_cnt_u32 - 1;
> > 	qp->rq.wq.wqe_cnt = rq_desc_cnt_u32 / qp->rq.wq.max_sge;
> > 
> > 	qp->page_size = page_size;
> > }
> > ```
> 
> Please explain the issue?



> > > > > max_rq_wr
> > > > 
> > > > 
> > > > Does it mean we need to compare it to: roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr) ?
> > > > I.e.
> > > > max_recv_descs = roundup_pow_of_two(attr->cap.max_recv_sge * attr->cap.max_recv_wr); if (max_recv_descs > ctx->max_rq_wr) { verbs_err(&ctx->ibvctx, "Max receive WR %u > %u\n", attr->cap.max_recv_wr, ctx->max_rq_wr); return EINVAL; }
> > > 
> > > 
> > > Yes, something along these lines looks correct.
> > > > This might be a solution for this issue - but still, I think it makes sense to change the variables from uint16_t to uint32_t just to be on the safe side.
> > > 
> > > 
> > > I would start by pushing what fixes the issue.
> > 
> > 
> > @gal-pressman , We're looking at the solution you propose here, but still - even if we take your solution, can we also apply the following patch here to be on the safe side (to handle only the uint32_t issue)?
> > ```
> > static void efa_setup_qp(struct efa_context *ctx,
> > 			 struct efa_qp *qp,
> > 			 struct ibv_qp_cap *cap,
> > 			 size_t page_size)
> > {
> > 	uint32_t rq_desc_cnt_u32;
> > 
> > 	efa_qp_init_indices(qp);
> > 
> > 	**qp->sq.wq.wqe_cnt = (uint32_t)roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr, ctx->min_sq_wr));**
> > 	qp->sq.wq.max_sge = cap->max_send_sge;
> > 	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;
> > 
> > 	qp->rq.wq.max_sge = cap->max_recv_sge;
> > 	**rq_desc_cnt_u32 = (uint32_t)roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);**
> > 	qp->rq.wq.desc_mask = rq_desc_cnt_u32 - 1;
> > 	qp->rq.wq.wqe_cnt = rq_desc_cnt_u32 / qp->rq.wq.max_sge;
> > 
> > 	qp->page_size = page_size;
> > }
> > ```
> 
> Please explain the issue?

A few points:

1) The current code is:
```

static void efa_setup_qp(struct efa_context *ctx,
			 struct efa_qp *qp,
			 struct ibv_qp_cap *cap,
			 size_t page_size)
{
	uint16_t rq_desc_cnt;

	efa_qp_init_indices(qp);

	qp->sq.wq.wqe_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr,
						     ctx->min_sq_wr));
	qp->sq.wq.max_sge = cap->max_send_sge;
	qp->sq.wq.desc_mask = qp->sq.wq.wqe_cnt - 1;

	qp->rq.wq.max_sge = cap->max_recv_sge;
	rq_desc_cnt = roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);
	qp->rq.wq.desc_mask = rq_desc_cnt - 1;
	qp->rq.wq.wqe_cnt = rq_desc_cnt / qp->rq.wq.max_sge;

	qp->page_size = page_size;
}
```
1) rq_desc_cnt is declared as a uint16:
	uint16_t rq_desc_cnt;
    Then, we assign a uint64 value to a uint16:
    rq_desc_cnt = **(uint64)**roundup_pow_of_two(cap->max_recv_sge * cap->max_recv_wr);
    At this point, it's possible that rq_desc_cnt is truncated and thus is assigned 0 ==> which goes all the way to the driver and 
    causes a failure to initialize.

    If we change 

2) A minor but still a possible issue if we update the maximum value for cap->max_send_wr in our device:
qp->sq.wq.wqe_cnt is declared as a uint32_t while roundup_pow_of_two() returns a uint64_t ==> With high values we can expect a truncation here.
qp->sq.wq.wqe_cnt = roundup_pow_of_two(max_t(uint32_t, cap->max_send_wr,
						     ctx->min_sq_wr));

gal-pressman:
I don't see how your suggested code solves anything.
There are some inconsistencies in the efa_wq area, but I believe the root cause for all is that the size of the doorbell itself is u16, not u32 (at least that what it looks like from efa_rq_ring_doorbell()/efa_sq_ring_doorbell()).

If the size of doorbell's producer counter is 16 bits, obviously there's no need to change desc_mask to u32.

PR #1225, opened 2022-09-16
util: Add barriers support for MIPS

matoro:
MIPS has the capability to issue memory barriers.  Implement this using a full (heavy) barrier.  This is the only variant that is guaranteed to exist on all hardware.  Other barriers are only available on mipsr6 or some special earlier implementations.  However even on the former the kernel uses a full barrier for safety due to not yet trusting their reliability.

See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881731
See: https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00605-2B-CMPCOHERE-AFP-01.01.pdf
See: https://github.com/torvalds/linux/blob/master/arch/mips/include/asm/sync.h

### DIFF
--- a/util/udma_barrier.h
+++ b/util/udma_barrier.h
@@ -100,6 +100,8 @@
 #define udma_to_device_barrier() asm volatile("dbar 0" ::: "memory")
 #elif defined(__riscv)
 #define udma_to_device_barrier() asm volatile("fence ow,ow" ::: "memory")
+#elif defined(__mips__)
+#define udma_to_device_barrier() asm volatile("sync 0" ::: "memory")
 #else
 #error No architecture specific memory barrier defines found!
 #endif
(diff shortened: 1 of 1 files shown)


rleon:
@jgunthorpe ???

jgunthorpe:
mips may have memory barriers, but do any DMA cache coherent MIPS implementations exist? Which ones are they?

matoro:
> mips may have memory barriers, but do any DMA cache coherent MIPS implementations exist? Which ones are they?

Yes, seems to be everything that doesn't `select DMA_NONCOHERENT`.  A lot of boards select this in `arch/mips/Kconfig` but not all.  My Octeon II is not one of them:
```
$ zgrep -i "coherent" /proc/config.gz
CONFIG_DMA_DECLARE_COHERENT=y
```

Is this correct, or is there a better way to determine this?

PR #1226, opened 2022-09-21
Install 70-persistent-ipoib.rules into docs instead of /etc

bdrung:
This rdma-core Debian package ships the udev rule `70-persistent-ipoib.rules` and installs it under `/etc/udev/rules.d`, which is reserved for user-installed files.

Install `70-persistent-ipoib.rules` into `/usr/share/doc/rdma-core` instead of `/etc/udev/rules.d` and document to copy this file into `/etc/udev/rules.d` before editing.

Bug-Debian: https://bugs.debian.org/958385

### DIFF
--- a/buildlib/centos6.spec
+++ b/buildlib/centos6.spec
@@ -103,7 +103,6 @@ rm -rf %{buildroot}/%{my_unitdir}/
 %config %{_sysconfdir}/libibverbs.d/*
 %config %{_sysconfdir}/rdma/modules/*
 %{perl_vendorlib}/IBswcountlimits.pm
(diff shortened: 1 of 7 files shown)


jgunthorpe:
There is no reason to install a commented out file in /lib/udev, if it should be installed somplace it is /usr/doc/examples

bdrung:
Okay, I change the merge request to install the udev rule into `/usr/share/doc/rdma-core`.

jgunthorpe:
@Honggang-LI  are you Ok with this for Fedora?

nmorey:
OK with the change for SUSE

PR #1229, opened 2022-09-22
tests: ODP testcases for RDMA Write/Read and Atomic operations

daimatsuda:
These patches add new testcases for ODP (test_odp.py).  Currently, only Send/Recv operations are checked, and other operations are not tested with ODP at all. This patchset cleans-up the ODP test code and add four new testcases. I confirmed these new testcases succeeded with RFC implementation of Explicit ODP on rxe[1].

The new testcases succeeded also with ConnectX-6, but currently the ODP Atomic tests are skipped with my device. This is because the capability for RC (rc_odp_caps) is not filled with IB_ODP_SUPPORT_ATOMIC in internal_fill_odp_caps(). However, it seems the device can actually handle Atomic operations. I am not sure if this is a bug of the device or not. If anybody from NVIDIA knows why rc_odp_caps is unfilled, please let me know
cf. https://github.com/torvalds/linux/blame/master/drivers/infiniband/hw/mlx5/odp.c#L321

The first two patches are for clean-up:
 tests: Import test.utils entirely for ODP test cases
 tests: Make ODP test consistent with other traffic tests

The latter three add new testcases:
  tests: Add tests for Atomic operations with ODP
  tests: Add a test for RDMA Read with ODP
  tests: Add a test for RDMA Write with ODP

[1] [RFC PATCH 0/7] RDMA/rxe: On-Demand Paging on SoftRoCE
https://lore.kernel.org/lkml/cover.1662461897.git.matsuda-daisuke@fujitsu.com/t/

### DIFF
--- a/tests/test_odp.py
+++ b/tests/test_odp.py
@@ -85,13 +85,14 @@ class OdpTestCase(RDMATestCase):
                          BaseResources.
         :param resource_arg: Dict of args that specify the resource specific
                              attributes.
-        :return: The (client, server) resources.
         """
-        client = resource(**self.dev_info, **resource_arg)
-        server = resource(**self.dev_info, **resource_arg)
-        client.pre_run(server.psns, server.qps_num)
-        server.pre_run(client.psns, client.qps_num)
-        return client, server
+        self.client = resource(**self.dev_info, **resource_arg)
+        self.server = resource(**self.dev_info, **resource_arg)
+        self.client.pre_run(self.server.psns, self.server.qps_num)
+        self.server.pre_run(self.client.psns, self.client.qps_num)
+        self.traffic_args = {'client': self.client, 'server': self.server,
+                             'iters': self.iters, 'gid_idx': self.gid_index,
+                             'port': self.ib_port}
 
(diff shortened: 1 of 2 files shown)


daimatsuda:
The new testcases are skipped unless all of IB_ODP_SUPPORT_WRITE, IB_ODP_SUPPORT_READ, and IB_ODP_SUPPORT_ATOMIC are set to rc_odp_caps in the driver layer. This is because I did not come up with any way to pass the respective flag to the @requires_odp decorator for each testcase. If you know any solution to do this, please let me know.

EdwardSro (approved):

EdwardSro:
Regarding ODP Atomic tests, I agree that the ODP CAPs are not reporting the Atomic support, even though it's supported.
If you ignore the CAPs check, you'll see that actually the ODP Atomic tests pass.
I'll try to look at it, but anyway, it should not block this PR.

daimatsuda:
I believe this work is ready to be merged.
Could anybody take a look and get this merged?

EdwardSro:
@rleon

PR #1232, opened 2022-09-29
irdma: Verify the provided max_send_wr and max_recv_wr

Kamalheib:
Make sure to verify that the provided max_send_wr and max_recv_wr are in the supported range.

### DIFF
--- a/providers/irdma/uverbs.c
+++ b/providers/irdma/uverbs.c
@@ -1291,6 +1291,12 @@ struct ibv_qp *irdma_ucreate_qp(struct ibv_pd *pd,
 		return NULL;
 	}
 
+	if (attr->cap.max_send_wr > uk_attrs->max_hw_wq_quanta ||
+	    attr->cap.max_recv_wr > uk_attrs->max_hw_rq_quanta) {
+		errno = EINVAL;
+		return NULL;
+	}
+
 	irdma_get_wqe_shift(uk_attrs,
 			    uk_attrs->hw_rev > IRDMA_GEN_1 ? attr->cap.max_send_sge + 1 :
 				attr->cap.max_send_sge,


providers/irdma/uverbs.c
```diff
@@ -1291,6 +1291,12 @@ struct ibv_qp *irdma_ucreate_qp(struct ibv_pd *pd,
 		return NULL;
 	}
 
+	if (attr->cap.max_send_wr > uk_attrs->max_hw_wq_quanta ||
```
rleon, line 1294: In the kernel, the max_hw_wq_quanta indirectly related to max_send_wr. First, why do you need this check in userspace? and second why is it correct thing to claim that max_send_wr > max_hw_wq_quanta check is correct one?

Kamalheib: > In the kernel, the max_hw_wq_quanta indirectly related to max_send_wr. First, why do you need this check in userspace? 

I've added this check in userspace, because I saw that there are other checks for max_send_sge, max_recv_sge, and max_inline_data two lines above it.

Basically, this patch was added to avoid the failure of test_create_raw_qp_ex_with_illegal_caps_max_send_wr and test_create_raw_qp_ex_with_illegal_caps_max_recv_wr pyverbs tests.

> and second why is it correct thing to claim that max_send_wr > max_hw_wq_quanta check is correct one?

Please take a look at irdma_get_sqdepth().

Thanks,
Kamal

rleon: > > In the kernel, the max_hw_wq_quanta indirectly related to max_send_wr. First, why do you need this check in userspace?
> 
> I've added this check in userspace, because I saw that there are other checks for max_send_sge, max_recv_sge, and max_inline_data two lines above it.
> 
> Basically, this patch was added to avoid the failure of test_create_raw_qp_ex_with_illegal_caps_max_send_wr and test_create_raw_qp_ex_with_illegal_caps_max_recv_wr pyverbs tests.

But kernel will fail too, why do we need an extra check in userspace?

> 
> > and second why is it correct thing to claim that max_send_wr > max_hw_wq_quanta check is correct one?
> 
> Please take a look at irdma_get_sqdepth().

This is why I said "indirectly related".

> 
> Thanks, Kamal

PR #1233, opened 2022-10-02
Fix max length of node description

gregoryl-mlnx:
Fix max length of node description (ibnetdiscover and smpquery)

Both utilities truncate node description up to 63 bytes instead 64 bytes.
Node description fields declared as char array of IB_SMP_DATA_SIZE bytes.
Last character for compatibility with c-string set to zero.
In this case node description will be truncated from 64 bytes to 63 bytes.
Size of node description fields was enlarged to the IB_SMP_DATA_SIZE + 1 bytes for using last character for zero char without truncated node description

Signed-off-by: Gregory Linschitz <gregoryl@nvidia.com>

### DIFF
--- a/infiniband-diags/smpquery.c
+++ b/infiniband-diags/smpquery.c
@@ -75,7 +75,7 @@ static const char *node_desc(ib_portid_t *dest, char **argv, int argc)
 {
 	int node_type, l;
 	uint64_t node_guid;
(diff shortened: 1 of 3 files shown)


rleon:
Please prepare one commit that includes all changes.
It should compile, be descriptive and describe why change in this place is enough.
While I grepped rdma-core, I saw many places for nodeinfo that has IB_SMP_DATA_SIZE size.

PR #1234, opened 2022-10-06
Revert "verbs: Allow aligned address & size only for fork init"

rleon:
Unfortunately, this change broke HW steering in DPDK. The reason to it that length can be unaligned on newer kernel.

This reverts commit 561cbce90cf4ef097d80efdcd68d619fa0ddb266.

### DIFF
--- a/libibverbs/man/ibv_reg_mr.3
+++ b/libibverbs/man/ibv_reg_mr.3
@@ -113,9 +113,6 @@ as the rkey field of struct ibv_send_wr passed to the ibv_post_send function.
 .B ibv_dereg_mr()
 returns 0 on success, or the value of errno on failure (which indicates the failure reason).
 .SH "NOTES"
-.B ibv_reg_mr() / ibv_reg_mr_iova()
-fails if the address (\fBaddr\fR) or the length (\fBlength\fR) is not aligned to page size once ibv_fork_init() is called.
-.PP
 .B ibv_dereg_mr()
 fails if any memory window is still bound to this MR.
 .SH "SEE ALSO"
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -635,18 +635,6 @@ static int ibv_madvise_range(void *base, size_t size, int advice)
 	else
 		range_page_size = page_size;
 
-	if (mm_root) {
-		/* 0x7f2764ac5000      -      0x7f2764ac7000
-		 * |ptr0 128| ... |ptr1 4096| ... |ptr2 512|
-		 * After ibv_reg_mr(pd, ptr1, 4096, access), the full 8K range
-		 * becomes DONTFORK. And the child process will hit a segment
-		 * fault during access ptr0/ptr2.
-		 */
-		unsigned long page_mask = range_page_size - 1;
-		if ((unsigned long)base & page_mask || (unsigned long)size & page_mask)
-			return -1;
-	}
-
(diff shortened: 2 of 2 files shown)


rleon:
@pizhenwei

pizhenwei:
> 

Hi @rleon 
Could you please give me more hint about this change?

rleon:
> > 
> 
> Hi @rleon Could you please give me more hint about this change?

From internal bug report:
DPDK tries to initialize the infrastructure by calling rte_eal_init(), which calls mlx5dv_devx_umem_reg()
The function call has been succeded(errno=0) but the pointer mlx5dv_devx_umem *umem_obj is NULL.

Thanks

pizhenwei:
> 

Hi,
I think I followed you about this issue: if ibv_dontfork_range() fails, __mlx5dv_devx_umem_reg_ex returns NULL only without errno raising:
```
rdma-core/providers/mlx5/verbs.c

static struct mlx5dv_devx_umem *
__mlx5dv_devx_umem_reg_ex(struct ibv_context *context,
                         struct mlx5dv_devx_umem_in *in,
                         bool legacy)
{
	...
        if (ibv_dontfork_range(in->addr, in->size))
                goto err; 
	...
}
```

But I don't understand `The reason to it that length can be unaligned on newer kernel.`, I check the madvise.c from linux-6.0, madvise always uses aligned address and size...

rleon:
> But I don't understand `The reason to it that length can be unaligned on newer kernel.`, I check the madvise.c from linux-6.0, madvise always uses aligned address and size...

Are you sure? I see that madvise.c checks that size >= page_size

pizhenwei:
> > But I don't understand `The reason to it that length can be unaligned on newer kernel.`, I check the madvise.c from linux-6.0, madvise always uses aligned address and size...
> 
> Are you sure? I see that madvise.c checks that size >= page_size

You are right!
```
int do_madvise(struct mm_struct *mm, unsigned long start, size_t len_in, int behavior)
{
        unsigned long end;
        int error;
        int write;
        size_t len;
        struct blk_plug plug;

        start = untagged_addr(start);

        if (!madvise_behavior_valid(behavior))
                return -EINVAL;

        if (!PAGE_ALIGNED(start))
                return -EINVAL;
        len = PAGE_ALIGN(len_in);

        /* Check to see whether len was rounded up from small -ve to zero */
        if (len_in && !len)
                return -EINVAL;

        end = start + len;
        if (end < start)
```

madivse checks `addr` strictly, auto align `length` to page size, this syscall really affects an aligned addr&length area. As the comment in code:
```
                /* 0x7f2764ac5000      -      0x7f2764ac7000
                 * |ptr0 128| ... |ptr1 4096| ... |ptr2 512|
                 * After ibv_reg_mr(pd, ptr1, 4096, access), the full 8K range
                 * becomes DONTFORK. And the child process will hit a segment
                 * fault during access ptr0/ptr2.
                 */
```

jgunthorpe:
On a newer kernel we don't even call madvise anymore.

pizhenwei:
> On a newer kernel we don't even call madvise anymore.
Hi @jgunthorpe 

Is there a plan to drop function `static int ibv_madvise_range(void *base, size_t size, int advice)`?
I suppose the addr&length alignment check is needed by this function(to avoid random segfault in the child process)...

rleon:
> ibv_madvise_range

I don't think so.

PR #1235, opened 2022-10-13
rdma-ndd: disable systemd ProtectHostName feature

nmorey:
ProtectHostName prevents dynamic name changes to be noticed by the service. This means that on a system with no static hostname, rdma-ndd is started with a hostname 'localhost' and is not aware of new hostname retreived through a DHCP lease.
Fixes: 384b75b5f624 ("rdma-ndd: systemd hardening")


nmorey:
Fixed.

PR #1236, opened 2022-10-18
efa: Add EFA 0xefa2 PCI ID

mrgolin:
Add 0xefa2 devices to the pciid match table.

Signed-off-by: Michael Margolin <mrgolin@amazon.com>


jgunthorpe:
Where is the kernel version of this????

mrgolin:
I'll submit it shortly.
This PR isn't dependent on a kernel change but has no effect without it.

rleon:
> I'll submit it shortly. This PR isn't dependent on a kernel change but has no effect without it.

Did you send the kernel patch? I don't want to release rdma-core without kernel patch.

mrgolin:
> > I'll submit it shortly. This PR isn't dependent on a kernel change but has no effect without it.
> 
> Did you send the kernel patch? I don't want to release rdma-core without kernel patch.

Sent the patch. Thanks.

PR #1238, opened 2022-10-24
debian: Remove /etc/udev/rules.d/70-persistent-ipoib.rules on upgrade

bdrung:
Commit b9ddb7f8013b274e55727ec390960960b95ae7fa installed `70-persistent-ipoib.rules` into docs instead of `/etc`. Remove `/etc/udev/rules.d/70-persistent-ipoib.rules` on upgrade if it was not modified by the user.


rleon:
Can you please update commit message to include the reason?

Thanks

bdrung:
Updated commit message.

PR #1245, opened 2022-10-30
Fix max length of node description (dump_fst/ibroute/ibtracert)

gregoryl-mlnx:
This fix releted to the previous commit d974c4e398d281ce2993c44f865f5161480307f2 / PR(#1237)

dump_fst/ibroute/ibtracert truncate node description up to 63 bytes instead 64 bytes. Node description field declared as char array of IB_SMP_DATA_SIZE bytes. Last character for compatibility with c-string set to zero. In this case node description will be truncated from 64 bytes to 63 bytes. Size of node description fields was enlarged to the IB_SMP_DATA_SIZE + 1 bytes for using last character for zero char without truncated node description

Signed-off-by: Gregory Linschitz <gregoryl@nvidia.com>

### DIFF
--- a/infiniband-diags/dump_fts.c
+++ b/infiniband-diags/dump_fts.c
@@ -109,7 +109,7 @@ static void dump_multicast_tables(ibnd_node_t *node, unsigned startl,
 				  unsigned endl, struct ibmad_port *mad_port)
 {
 	ib_portid_t *portid = &node->path_portid;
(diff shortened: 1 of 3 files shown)


rleon:
Please use following format for the title: "ibdiags: .....", add Fixes line and break long line in commit message.

Thanks

PR #1248, opened 2022-11-02
RFC kernel-boot: Update udev rules and rdma_rename to handle auxiliary devices

tatyana-en:
Update udev rules for irdma driver to set ID_RDMA_IWARP and ID_RDMA_ROCE environment vars based on the auxiliary device name. For the new rules to work correctly, a patch for the irdma driver is necessary to change the ib device parent from pci to auxiliary device.

Change the rdma_rename logic to recognize auxiliary devices and rename ib devices by_onboard or by_pci the same way it did before they were parented by auxiliary devices.


jgunthorpe:
Thought I commented on the mailing list, send the kernel side as a non-rfc, it looks correct

tatyana-en:
Thank you, Jason. We are going to send a kernel patch soon and I am going to update this PR to drop the RFC. I was asking to check if you are okay with the user space rdma_rename changes.

jgunthorpe:
This is quite old now, what happened to this?

PR #1257, opened 2022-11-21
mlx5: DR, Fixes and improvements 

yishaih:
This series includes fixes and improvements in the DR area, mainly in the pattern and arguments area.

Specifically,
- Fixes an issue causing fallback from pattern/args to legacy modify header mode and some following issues in this area.
- Fixes an issue when both svlan and cvlan matches are provided.
- Improve to adds memory statistics for domain object.
- Improve to calculate the threshold of each pool according to its type.
- Improve to enable up to 64 modifications in modify-header action.


jgunthorpe:
Why is azure pipelines complaining about all the fixes lines? Please fix it. Also please look at the other checkpatch notes about line length

yishaih:
The long lines warning was fixed, commit IDs of the Fixes lines are correct.

PR #1259, opened 2022-11-25
Sort objects before finalize static library

hw-lvguangyan:
os.listdir() method get the list of all files and directories is not sorted, sort objects before finalize static library.
Signed-off-by: lvguangyan [lvguangyan2@huawei.com](mailto:lvguangyan2@huawei.com)

### DIFF
--- a/buildlib/sanitize_static_lib.py
+++ b/buildlib/sanitize_static_lib.py
@@ -144,7 +144,9 @@ class Lib(object):
         os.makedirs(self.final_objdir)
 
         subprocess.check_call([args.ar, "x", libfn], cwd=self.objdir)
-        self.objects = [I for I in os.listdir(self.objdir)]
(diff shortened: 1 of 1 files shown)


buildlib/sanitize_static_lib.py
```diff
@@ -144,7 +144,9 @@ def __init__(self, libfn, tmpdir):
         os.makedirs(self.final_objdir)
 
         subprocess.check_call([args.ar, "x", libfn], cwd=self.objdir)
-        self.objects = [I for I in os.listdir(self.objdir)]
```
jgunthorpe, line 147: This is just sorted(os.listdir(self.objdir))

But why do we want to do this? Please explain in the commit message

Also use force push to update your branches not close/reopen.

jgunthorpe (changes requested):

PR #1266, opened 2022-12-01
buildlib: Remove centos6 era distros

jgunthorpe:
These have left all support windows, time to let go.


rleon:
Need to remove Centos 6 and Amazon Linux 1 from README.md.

And what about?

   56 if (${CMAKE_VERSION} VERSION_LESS "3.18.1")
   57         # Centos 6 support
   58         cmake_minimum_required(VERSION 2.8.11 FATAL_ERROR)
   59 else()
   60         cmake_minimum_required(VERSION 3.18.1 FATAL_ERROR)
   61 endif()

PR #1268, opened 2022-12-03
tests: Add RDMA READ zero length scenario

zhijianli88:
Recently, a regression was introduced into RXE in
commit: 686d348476ee ("RDMA/rxe: Remove unnecessary mr testing") which can be triggered by a zero length RDMA read.

Add this scenario to cover it.

Links: https://lore.kernel.org/linux-rdma/20221202110157.3221952-1-matsuda-daisuke@fujitsu.com/


EdwardSro:
Looks good to me

PR #1270, opened 2022-12-05
tests: Let pyverbs tests run without scapy package

daimatsuda:
Since 'rdma cuda dmabuf' tests were added, it is impossible to run any test without installing scapy package. However, not all users need to run the new tests specific to mlx5. This patch fixes the issue.

### DIFF
--- a/tests/test_mlx5_cuda_umem.py
+++ b/tests/test_mlx5_cuda_umem.py
@@ -5,7 +5,6 @@ import resource
 
 from pyverbs.providers.mlx5.mlx5dv import Mlx5DevxObj, WqeDataSeg, Mlx5UMEM
 from tests.mlx5_base import Mlx5DevxRcResources, Mlx5DevxTrafficBase
-from tests.mlx5_prm_structs import SwMkc, CreateMkeyIn, CreateMkeyOut
 import pyverbs.providers.mlx5.mlx5_enums as dve
 import tests.cuda_utils as cu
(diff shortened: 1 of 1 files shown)


tests/test_mlx5_cuda_umem.py
```diff
@@ -5,11 +5,15 @@
 
 from pyverbs.providers.mlx5.mlx5dv import Mlx5DevxObj, WqeDataSeg, Mlx5UMEM
 from tests.mlx5_base import Mlx5DevxRcResources, Mlx5DevxTrafficBase
-from tests.mlx5_prm_structs import SwMkc, CreateMkeyIn, CreateMkeyOut
```
EdwardSro, line 8: or you can just move this line into create_mkey().

EdwardSro (commented):
No need for all this.
You can move the import to the relevant test function, as we do in other tests.
Then, the test will be skipped after mlx5_prm_structs (which uses Scapy) raises the SkipTest exception.
I don't think it's a good idea to generalize it to every ImportError (may cause an unwanted missing coverage for not careful users)

tests/test_mlx5_cuda_umem.py
daimatsuda: Thanks for taking a look.

This seems the most sensible solution.
I have modified the patch.

EdwardSro (approved):

PR #1271, opened 2022-12-06
Update srq_pingpong.c

Wevistar:
libibverbs/example: optimization of the debug log for srq_pingpong.c

The error branch returned once a wc status failed, but we actually need to known more error information, i.e. the total polled wc numbers, and the first one which failed, under the condition of all wc's stauts cannot be printed.

Signed-off-by: yuwei (O) <yuwei139@huawei.com>
Reviewed-by: zhaofei (K) <zhaofei82@huawei.com>


Wevistar:
invalid commit message format, closed manually

PR #1272, opened 2022-12-06
libibverbs/example: add more debug information

Wevistar:
srq_pingpong.c: optimization of the debug log.

Signed-off-by: yuwei (O) yuwei139@huawei.com
Reviewed-by: zhaofei (K) zhaofei82@huawei.com


Wevistar:
invalid signed-off-by format, closed PR

PR #1273, opened 2022-12-06
srq_pingpong.c: optimization of the debug log

Wevistar:

Signed-off-by: Wei Yu <yuwei139@huawei.com>


Wevistar:
DCO not pass, PR closed

PR #1275, opened 2022-12-06
srq_pingpong.c: optimization of the debug log

Wevistar:

Signed-off-by: Wevistar <yuwei139@huawei.com>


rleon:
1. Please break the commit message from being one line to multiple ones. The lines need to be upto 80 chars.
2. "optimization of the debug log" as a title is a misleading. Please explain why this change ever needed for utility
which whole purpose is to exercise APIs.

Thanks

PR #1282, opened 2022-12-07
libibumad/umad.c: minor cleanups in error path

CaiZhongS:
The strdup() function returns a pointer to a new string which is a duplicate of the string s. Memory for the new string is obtained with malloc, and can be freed with free. Therefore, *ca_name should be freed in error path.

Signed-off-by: CaiZhongS <caizhongshun1@huawei.com>


jgunthorpe:
Huh? How can ca_name be anything other than null at this point?

CaiZhongS:
> Huh? How can ca_name be anything other than null at this point?

I ignored the premise that the above two branches will exit in any case. Thanks!

PR #1297, opened 2023-01-13
irdma: Add support for ibv_reg_dmabuf_mr

zhuyj:
Initialize umr. And I will send out the kernel patch very soon.


zhuyj:
DCO error.

zhuyj:
DCO error.

PR #1304, opened 2023-01-23
tests: Skip CUDA tests if there is no CUDA device

EdwardSro:
Skip CUDA tests in case cuda-python package is installed but no CUDA-cappable device is found.


rleon:
Please rebase to latest master branch which includes centos6 removal.

EdwardSro:
Rebased (again)

PR #1305, opened 2023-01-25
providers/rxe: Remove unused "ssn" field from QP and WQE

rolandd:
librxe sets wqe->ssn to an incrementing value kept in qp->ssn for all send work requests, but there is nothing in either the provider library or the kernel rxe driver that consumes this field. Clean things up by removing this unnecessary work.

Keep the ssn field in the ABI file rdma_user_rxe.h for the forseeable future, although a patch to remove the corresponding kernel code that keeps an ssn in the QP structure and copies the value into each WQE will be submitted too.

Signed-off-by: Roland Dreier <roland@enfabrica.net>

### DIFF
--- a/providers/rxe/rxe.c
+++ b/providers/rxe/rxe.c
@@ -806,7 +806,6 @@ static void wr_atomic_cmp_swp(struct ibv_qp_ex *ibqp, uint32_t rkey,
 	wqe->wr.wr.atomic.swap = swap;
 	wqe->wr.wr.atomic.rkey = rkey;
 	wqe->iova = remote_addr;
-	wqe->ssn = qp->ssn++;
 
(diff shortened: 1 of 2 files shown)


rolandd:
The build failure looks spurious - a problem in the build system?

jgunthorpe:
Yes, azure pipelines changed something in their infra to cause this, we need the centos6 removal PR to fix it.

jgunthorpe:
@rpears0n didn't you work on something similar to this, am I mis remembering?

rleon:
Please rebase to latest master branch which includes centos6 removal.

rolandd:
OK, rebased on top of latest upstream.

PR #1312, opened 2023-02-19
Add ibv_query_qp_data_in_order flags

mrgolin:
Add definition for ibv_query_qp_data_in_order flags to allow querying for specific capabilities related to data polling support, currently adding 128 bytes in order flag (LL128).
Adopt existing code and add support in EFA provider.

### DIFF
--- a/kernel-headers/rdma/efa-abi.h
+++ b/kernel-headers/rdma/efa-abi.h
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: ((GPL-2.0 WITH Linux-syscall-note) OR BSD-2-Clause) */
 /*
- * Copyright 2018-2022 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright 2018-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
 #ifndef EFA_ABI_USER_H
@@ -120,6 +120,7 @@ enum {
 	EFA_QUERY_DEVICE_CAPS_RNR_RETRY = 1 << 1,
 	EFA_QUERY_DEVICE_CAPS_CQ_NOTIFICATIONS = 1 << 2,
 	EFA_QUERY_DEVICE_CAPS_CQ_WITH_SGID     = 1 << 3,
+	EFA_QUERY_DEVICE_CAPS_DATA_POLLING_128 = 1 << 4,
 };
 
 struct efa_ibv_ex_query_device_resp {
--- a/libibverbs/verbs.c
+++ b/libibverbs/verbs.c
@@ -701,7 +701,16 @@ int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
 	 */
 	return 0;
 #else
-	return get_ops(qp->context)->query_qp_data_in_order(qp, op, flags);
+	int result;
+
+	if (!check_comp_mask(flags, IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS))
+		return 0;
+
+	result = get_ops(qp->context)->query_qp_data_in_order(qp, op, flags);
+	if (result & IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG)
+		result |= IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES;
+
+	return flags ? result : !!(result & IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG);
 #endif
 }
 
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -1036,6 +1036,15 @@ enum ibv_qp_attr_mask {
 	IBV_QP_RATE_LIMIT		= 1 << 25,
 };
 
+enum ibv_query_qp_data_in_order_flags {
+	IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS = 1 << 0,
+};
+
+enum ibv_query_qp_data_in_order_caps {
+	IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG = 1 << 0,
+	IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES = 1 << 1,
+};
+
 enum ibv_qp_state {
 	IBV_QPS_RESET,
 	IBV_QPS_INIT,
@@ -3205,11 +3214,15 @@ ibv_modify_qp_rate_limit(struct ibv_qp *qp,
  *   written in-order.
  * @qp: The QP to query.
  * @op: Operation type.
- * @flags: Extra field for future input. For now must be 0.
+ * @flags: Flags are used to select a query type.
+ * For IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS, the function will return a
+ * capabilities vector. If 0, will query for IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG
+ * support and return 0/1 result.
  *
  * Return Value
- * ibv_query_qp_data_in_order() returns 1 if the data is guaranteed to be
- *   written in-order, 0 otherwise.
+ * ibv_query_qp_data_in_order() return value is determined by flags.
+ * For each capability bit, 1 is returned if the data is guaranteed to be
+ * written in-order for selected operation and type, 0 otherwise.
  */
 int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
 			       uint32_t flags);
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-2.0 OR BSD-2-Clause
 /*
- * Copyright 2019-2022 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright 2019-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
 #include <assert.h>
@@ -1522,6 +1522,18 @@ int efa_query_qp(struct ibv_qp *ibvqp, struct ibv_qp_attr *attr,
 
 	return ibv_cmd_query_qp(ibvqp, attr, attr_mask, init_attr,
 				&cmd, sizeof(cmd));
+}
+
+int efa_query_qp_data_in_order(struct ibv_qp *ibvqp, enum ibv_wr_opcode op,
+			       uint32_t flags)
+{
+	struct efa_context *ctx = to_efa_context(ibvqp->context);
+	int caps = 0;
+
+	if (EFA_DEV_CAP(ctx, DATA_POLLING_128))
+		caps |= IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES;
+
+	return caps;
 }
 
 int efa_destroy_qp(struct ibv_qp *ibvqp)
--- a/kernel-headers/rdma/hns-abi.h
+++ b/kernel-headers/rdma/hns-abi.h
@@ -87,10 +87,14 @@ struct hns_roce_ib_create_qp_resp {
 
 enum {
 	HNS_ROCE_EXSGE_FLAGS = 1 << 0,
+	HNS_ROCE_RQ_INLINE_FLAGS = 1 << 1,
+	HNS_ROCE_CQE_INLINE_FLAGS = 1 << 2,
 };
 
 enum {
 	HNS_ROCE_RSP_EXSGE_FLAGS = 1 << 0,
+	HNS_ROCE_RSP_RQ_INLINE_FLAGS = 1 << 1,
+	HNS_ROCE_RSP_CQE_INLINE_FLAGS = 1 << 2,
 };
 
 struct hns_roce_ib_alloc_ucontext_resp {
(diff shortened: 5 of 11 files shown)


mrgolin:
Related kernel patch:
https://lore.kernel.org/linux-rdma/20230219081328.10419-1-mrgolin@amazon.com/T/#u

mrgolin:
@rleon I would really appreciate if this can be reviewed for the upcoming release.

rleon:
> @rleon I would really appreciate if this can be reviewed for the upcoming release.

You posted kernel patch day before merge window starts. I afraid that it is too late for UAPI changes.

mrgolin:
> > @rleon I would really appreciate if this can be reviewed for the upcoming release.
> 
> You posted kernel patch day before merge window starts. I afraid that it is too late for UAPI changes.

Thanks Leon I understand the timeline and I wish we could submit this earlier. Yet, the kernel patch is really simple one and it changes UAPI only by a definition of another bit in an existing field.

rleon:
> > > @rleon I would really appreciate if this can be reviewed for the upcoming release.
> > 
> > 
> > You posted kernel patch day before merge window starts. I afraid that it is too late for UAPI changes.
> 
> Thanks Leon I understand the timeline and I wish we could submit this earlier. Yet, the kernel patch is really simple one and it changes UAPI only by a definition of another bit in an existing field.

I'm aware. however I don't want to take any chances and get angry about sending PR to Linus with patches not being in linux-next.

Thanks

rleon:
@jgunthorpe , you are sending PRs to Linus, so it is your call.

Thanks

mrgolin:
> This doesn't make alot of sense to me.. "in order" reflects the entire message stream, so what does "within a block" even mean? It seems to be some completely orthogonal concept related to cache tearing not message ordering?

From documentation:
```
**ibv_query_qp_data_in_order()** Checks whether WQE data is guaranteed to be
written in-order, and thus reader may poll for data instead of poll for completion.
This function indicates data is written in-order within each WQE, but cannot be used to determine ordering between separate WQEs.
```
This function is used to determine whether data can be polled, i.e. it's guaranteed that if some byte was written then all previous bytes of this message are ready to be used.
Some devices (e.g. EFA) may not support this for the full message but can guarantee write in order inside each data block of a particular size.

mrgolin:
> > This doesn't make alot of sense to me.. "in order" reflects the entire message stream, so what does "within a block" even mean? It seems to be some completely orthogonal concept related to cache tearing not message ordering?
> 
> From documentation:
> 
> ```
> **ibv_query_qp_data_in_order()** Checks whether WQE data is guaranteed to be
> written in-order, and thus reader may poll for data instead of poll for completion.
> This function indicates data is written in-order within each WQE, but cannot be used to determine ordering between separate WQEs.
> ```
> 
> This function is used to determine whether data can be polled, i.e. it's guaranteed that if some byte was written then all previous bytes of this message are ready to be used. Some devices (e.g. EFA) may not support this for the full message but can guarantee write in order inside each data block of a particular size.

@jgunthorpe Does it make sense now?

YonatanNachum:
We have a kernel patch: [link](https://patchwork.kernel.org/project/linux-rdma/patch/20230219081328.10419-1-mrgolin@amazon.com/) 
Thanks.

libibverbs/driver.h
```diff
@@ -420,8 +420,7 @@ struct verbs_context_ops {
 			  struct ibv_port_attr *port_attr);
 	int (*query_qp)(struct ibv_qp *qp, struct ibv_qp_attr *attr,
 			int attr_mask, struct ibv_qp_init_attr *init_attr);
-	int (*query_qp_data_in_order)(struct ibv_qp *qp, enum ibv_wr_opcode op,
-				      uint32_t flags);
```
gal-pressman, line 424: I would keep the flags here, the provider can choose to ignore them.

libibverbs/verbs.c
```diff
@@ -701,7 +701,16 @@ int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
...
+	uint32_t query_mask;
+	uint32_t comp_mask;
+
+	query_mask = flags ? flags : IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG;
+
+	comp_mask = get_ops(qp->context)->query_qp_data_in_order(qp, op);
+	if (comp_mask & IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG)
+		comp_mask |= IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES;
+
+	return !(query_mask & ~comp_mask);
```
gal-pressman, line 713: You can use check_comp_mask() here

libibverbs/verbs.c
```diff
@@ -701,7 +701,16 @@ int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
 	 */
 	return 0;
 #else
-	return get_ops(qp->context)->query_qp_data_in_order(qp, op, flags);
+	uint32_t query_mask;
+	uint32_t comp_mask;
```
gal-pressman, line 705: Not sure comp_mask is the best name for this variable.

gal-pressman:
IIUC, it means every 128 bytes are written "atomically", it doesn't really replace the completion though, as you have no idea when the entire message is completed (unless you check each 128 chunk separately?).

What is this used for?

libibverbs/driver.h
mrgolin: The idea of this interface change is that providers will now return all their supported capabilities (related to data inorder) without depending on the request. I think flags here may be confusing, when would you use it in provider code?

gal-pressman:
Makes sense.
LL128 is where you have a flags header for each 120 bytes of data, right?

mrgolin:
> Makes sense. LL128 is where you have a flags header for each 120 bytes of data, right?

Right, 120 bytes of data and 8 bytes flag.

libibverbs/driver.h
gal-pressman: I don't know, depends on the flags.
Anyway, it's not a big deal, I guess the flags can be reintroduced if/when they're needed.

libibverbs/verbs.c
mrgolin: Done, thanks.

jgunthorpe (changes requested):

libibverbs/verbs.h
```diff
@@ -3205,11 +3210,12 @@ ibv_modify_qp_rate_limit(struct ibv_qp *qp,
  *   written in-order.
  * @qp: The QP to query.
  * @op: Operation type.
- * @flags: Extra field for future input. For now must be 0.
+ * @flags: A bit-mask used to select specific capabilities to query. If 0,
+ * will query for IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG support.
```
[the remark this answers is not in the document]
mrgolin: Do you suggest that passing IBV_QUERY_QP_DATA_RETURN_FLAGS as flags to that function will make it return supported capabilities instead of 0/1?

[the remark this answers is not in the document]
mrgolin: That actually was one of the options we considered but it didn't feel as a natural API extension but more like combining two into a single one. Is there any future use case for this function you can think of that won't be satisfied by querying for a capability bit?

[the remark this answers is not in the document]
mrgolin: @jgunthorpe Changed according to your request.

kernel-headers/rdma/efa-abi.h
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: ((GPL-2.0 WITH Linux-syscall-note) OR BSD-2-Clause) */
 /*
- * Copyright 2018-2022 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright 2018-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
```
wenduwan, line 3: Same as below, the latest open source policy advises dateless notices, e.g.
```
Copyright Amazon.com, Inc. or its affiliates. All rights reserved.
```

libibverbs/verbs.h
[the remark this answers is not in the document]
mrgolin: If it looks ok now, can we get this merged?

libibverbs/verbs.c
```diff
@@ -701,7 +701,22 @@ int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
...
-	return get_ops(qp->context)->query_qp_data_in_order(qp, op, flags);
+	uint32_t query_flags;
+	int result;
+
+	query_flags = flags ? flags : IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS;
+
+	result = get_ops(qp->context)->query_qp_data_in_order(qp, op, query_flags);
+	if (query_flags == IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS) {
+		if (result & IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG ||
+		    get_ops(qp->context)->query_qp_data_in_order(qp, op, 0))
```
jgunthorpe, line 712: There is only one other implementation just fix it to return the new style always

jgunthorpe (changes requested):
The rest looks OK

libibverbs/verbs.c
mrgolin: Ok, done.

providers/efa/verbs.c
```diff
@@ -1524,6 +1524,21 @@ int efa_query_qp(struct ibv_qp *ibvqp, struct ibv_qp_attr *attr,
 				&cmd, sizeof(cmd));
 }
 
+int efa_query_qp_data_in_order(struct ibv_qp *ibvqp, enum ibv_wr_opcode op,
+			       uint32_t flags)
+{
+	struct efa_context *ctx = to_efa_context(ibvqp->context);
+	int caps = 0;
+
+	if (flags != IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS)
```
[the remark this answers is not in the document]
mrgolin: So to make sure we are aligned, you would remove flags param from provider ops as I did in the first place or just move this condition to common code and ignore flags in providers?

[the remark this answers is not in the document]
jgunthorpe: yes just ignore the flag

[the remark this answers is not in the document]
mrgolin: Done

kernel-headers/rdma/efa-abi.h
mrgolin: @wenduwan Thanks.
This is a kernel header, can't change it here so will leave dated copyrights for now.

mrgolin:
@jgunthorpe any additional comments?

jgunthorpe:
I'm still a bit nervous about this - AFAICT the only way to achieve this in a modern system is if the device promises to generate 128 byte PCIe MemWr TLPs so that the platform can execute single TLPs "in order". But the max TLP size is controlled by PCI config space, so how can verbs know it is 128 bytes at this point?

The existing flag is basically saying 'the device does not use relaxed ordering and writes all bytes in any TLPs in order" and further that "the platform does not re-order non-relaxed ordering TLPs", which doesn't have this problem.

And further, semantically, according to verbs MRs created should be of the non-relaxed ordering type anyhow, so how do you get into a situation where otherwise in-order TLPs are re-ordered? Does this only work with relaxed ordering or is something in EFA wrongly forcing relaxed ordering in the MRs?

mrgolin:
> I'm still a bit nervous about this - AFAICT the only way to achieve this in a modern system is if the device promises to generate 128 byte PCIe MemWr TLPs so that the platform can execute single TLPs "in order". But the max TLP size is controlled by PCI config space, so how can verbs know it is 128 bytes at this point?
> 
> The existing flag is basically saying 'the device does not use relaxed ordering and writes all bytes in any TLPs in order" and further that "the platform does not re-order non-relaxed ordering TLPs", which doesn't have this problem.
 
Whenever 128 bytes in-order is promised it's up to the device to make sure each such block isn't getting splitted and reordered nether on the device/communication level nor by PCIe TLPs. It can achieve the PCIe part as you suggested by not using PCIe relaxed ordering (same way it is done for whole message in-order support) or if it is familiar with the platform, ensure that write TLPs don't split data anywhere in-between 128 bytes boundaries.
Either way it is less strict than whole message in-order so I believe that other providers may also correctly support this.

> And further, semantically, according to verbs MRs created should be of the non-relaxed ordering type anyhow, so how do you get into a situation where otherwise in-order TLPs are re-ordered? Does this only work with relaxed ordering or is something in EFA wrongly forcing relaxed ordering in the MRs?

I think it is mostly valuable for relaxed ordering to gain both from relaxed order performance improvement and from LL128 data polling low latency but if 128 bytes in-order is assured for relaxed ordering I don't see a reason why it won't be for non-relaxed ordering.

jgunthorpe:
Having just learned about this NCCL LL128 stuff, this patch is somewhat wrong :(

The query_qp_data_in_order() is all about the *CPU*'s perception of ordering. This new flag you've added is really about the device's ability to generate 128 byte PCIe TLPs.

mrgolin:
I believe this verb documentation can, hopefully not too late for some users, be improved to more explicitly state where its responsibility ends but looking at the discussion on original [PR](https://github.com/linux-rdma/rdma-core/pull/1009) that introduced this there wasn't any intension to cover destination memory behavior. For example your comment from that discussion:
>Sean, I covered that above, this API is only about the behavior up until verbs delivers PCIe TLPs out of the HCA, it doesn't try very hard to cover platform behavior. MR specific configuration including relaxed ordering and the behavior of the target memory is out of scope because we can't get that kind of information from verbs.

So I think what we did here perfectly aligns with this, extending it to support additional cases.

PR #1313, opened 2023-02-20
providers/erdma: Add non-4K page size support

hz-cheng:
Hello,

This PR adds the non-4K page size support. For some aarch64 distributions, the default page size is 64K, not 4K. ERDMA will not work correctly in these systems. The main reason is that we use the wrong doorbell page and page offset. We fix it in this PR. This PR needs kernel patch.

Thanks,
Cheng Xu


hz-cheng:
This PR is not right, so  I close it.

PR #1314, opened 2023-02-24
librdmacm: fix cma dev leak after init failed

KKRainbow:
If a cma dev is failed to initialize during rdma_get_devices, it'll never be freed or retry to init again. This bug can be triggered when rdma_get_devices is called immediately after mlx_ib driver loaded.

Here is a sample to reproduce the problem.

```
#include <stdio.h>
#include <unistd.h>
#include <rdma/rdma_cma.h>

int main() {
  while (1) {
    int size = 0;
    struct ibv_context** contexts = rdma_get_devices(&size);
    for (int i = 0; i < size; i++) {
      const char* name = ibv_get_device_name(contexts[i]->device);
      printf("context: %p, name: %s, i: %d\n", contexts[i], name, i);
    }
    if (contexts) {
            rdma_free_devices(contexts);
    }
    usleep(10);
  }
  return 0;
}
```

run this program and do a pci remove/rescan
```
echo 1 > /sys/class/net/eno12399np0/device/remove
echo 1 > /sys/bus/pci/rescan
```

the cma device may never show up.

### DIFF
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -96,6 +96,7 @@ struct cma_device {
 	uint8_t		    max_responder_resources;
 	int		    ibv_idx;
 	uint8_t		    is_device_dead : 1;
+	uint8_t		    is_device_fail_to_init : 1;
 };
 
 struct cma_id_private {
@@ -514,10 +515,11 @@ struct ibv_context **rdma_get_devices(int *num_devices)
 
 		/* reinit newly added devices */
 		if (ucma_init_device(dev)) {
-			/* Couldn't initialize the device: mark it dead and continue */
-			dev->is_device_dead = true;
+			/* Couldn't initialize the device: mark it unavailable and continue. */
+			dev->is_device_fail_to_init = true;
 			continue;
 		}
+		dev->is_device_fail_to_init = false;
 		cma_dev_cnt++;
 	}
 
@@ -526,7 +528,7 @@ struct ibv_context **rdma_get_devices(int *num_devices)
 		goto out;
 
 	list_for_each(&cma_dev_list, dev, entry) {
-		if (dev->is_device_dead)
+		if (dev->is_device_dead || dev->is_device_fail_to_init)
 			continue;
 		devs[i++] = dev->verbs;
 		dev->refcnt++;


librdmacm/cma.c
```diff
@@ -514,10 +515,11 @@ struct ibv_context **rdma_get_devices(int *num_devices)
 
 		/* reinit newly added devices */
 		if (ucma_init_device(dev)) {
-			/* Couldn't initialize the device: mark it dead and continue */
-			dev->is_device_dead = true;
+			/* Couldn't initialize the device: mark it unavailable and continue. */
+			dev->is_device_fail_to_init = true;
```
rleon, line 519: This change gives same logic as was before, just with another variable name.

KKRainbow: I think is_device_dead means the verbs is invalid and should be destroyed and no need to retry init further more.

is_device_fail_to_init means we should try reinit the verbs so the user can see this verbs after the init completed.

KKRainbow: Another approach is to reinit the verbs even if it's already marked dead. Though I think it will make the `is_device_dead` flag ambiguous

rleon: > I think is_device_dead means the verbs is invalid and should be destroyed and no need to retry init further more.
> 
> is_device_fail_to_init means we should try reinit the verbs so the user can see this verbs after the init completed.

It is a race and library shouldn't try to avoid them. User triggered device removal/addition and later started application, he should synchronize it.

rleon: There are many races like this and rdma-cm device initialization is maybe most visible.

KKRainbow: During our test, this is the only barrier to do RDMA device hot plugin. We have to try call ibv_open_device to test if the verbs is valid before call rdma_get_devices.

rleon: > Without this patch user cannot support hot-plug or hot-unplug without restarting the application.

Are you sure about it? Users who are interested (and using it) in hotplug/unplug listen to udev events to know when device is added and action accordingly.

KKRainbow: Applications rely on RDMA_CM_EVENT_DEVICE_REMOVAL event to do hot-unplug.
To do hot-plugging, User can listen to the udev events, or polling rdma devices changes by calling rdma_get_devices or ibv_get_device_list.
No matter in which method, if rdma_get_devices is called immediately after the device appears, there is a chance the according verbs would be marked dead and user will never see this device.

KKRainbow: The original implementation assumes that the verbs will be invalid forever if init failed, but it's not true.

rleon: What about this change?

diff --git a/librdmacm/cma.c b/librdmacm/cma.c
index 7b924bd0d..0a0da1cce 100644
--- a/librdmacm/cma.c
+++ b/librdmacm/cma.c
@@ -485,10 +485,7 @@ static int ucma_init_all(void)
                if (dev->is_device_dead)
                        continue;
 
-               if (ucma_init_device(dev)) {
-                       /* Couldn't initialize the device: mark it dead and continue */
-                       dev->is_device_dead = true;
-               }
+               ucma_init_device(dev);
        }
        pthread_mutex_unlock(&mut);
        return 0;

KKRainbow: It works too. ucma_init_device checks cma_dev->verbs to avoid duplicated init.
should I update this PR？

KKRainbow: I think a flag is still required to hide the uninitilized cma_devs.

rleon: > Oh wait. this cma_dev should not return to user before init done

Good point.

rleon: > I think a flag is still required to hide the uninitilized cma_devs.

And I am not convinced yet that this is librdmacm issue.

KKRainbow: Only librdmacm can see the newly created cma_dev and it just leaves them with an uninitialized ib_verbs. If librdmacm does not retry init, no one can use the device.

KKRainbow: This patch gives a chance to the user to call rdma_get_devices to retry init the cma_dev.

KKRainbow: This is a patch I submitted to SPDK to recreate IB listeners after device was hotplugged. I have no alternative but to do it with a workaround.
https://review.spdk.io/gerrit/c/spdk/spdk/+/15616

image

jgunthorpe:
I still don't understand this PR, how does a device get marked dead during hot plug in?

What exactly is the race you are seeing? The netlink identity should not be visible until all the cdevs are available.

jgunthorpe:
I saw your test program, it still doesn't explain what the race is.

In upstream kernel we use netlink to discover the devices and upstream kernel should order the netlink visibilty only after the kernel has setup the related cdevs. I did not check again if this is true, but it was the intention

If this is not happening then it should be fixed by a kernel patch.

If your system is using the old sysfs stuff then we don't expect hotplug to work correctly. netlink stuff was created to make hotplug reliable.

So, again, what *exactly* is the problem, how are you able to discover a device but then fail to open parts of it?

xupeng-mingtu:
We meet the same issue in spdk：

KKRainbow:
It seems the kernel module may expose a device that has not been fully ready and cannot be opened by CMA.

I think it's worth handling this scenario in CMA to make it more user-friendly for CMA users.

PR #1326, opened 2023-05-01
tests: add and improve some tests

EdwardSro:
This series adds a new SW Steering test and improve ODP tests by explicitly invalidating the MRs to make sure a page fault occurs.
In addition, adjusts the way we get the net-interface to be dependent on both the device name and port.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -180,11 +180,18 @@ class RDMATestCase(unittest.TestCase):
             has_roce_hw_bug(vendor_id, vendor_pid)
 
     @staticmethod
-    def get_net_name(dev):
-        out = subprocess.check_output(['ls',
-                                       '/sys/class/infiniband/{}/device/net/'
-                                      .format(dev)])
-        return out.decode().split('\n')[0]
+    def get_net_name(dev, port=None):
+        if port is not None:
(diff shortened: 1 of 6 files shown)


tests/base.py
```diff
@@ -180,11 +180,18 @@ def is_eth_and_has_roce_hw_bug(self):
...
-                                       '/sys/class/infiniband/{}/device/net/'
-                                      .format(dev)])
-        return out.decode().split('\n')[0]
+    def get_net_name(dev, port=None):
+        if port is not None:
+            out = subprocess.check_output(['rdma', 'link',  'show', '-j'])
+            loaded_json = json.loads(out.decode())
+            for row in loaded_json:
+                if row['ifname'] == dev and row['port'] == port:
+                    net_name = row['netdev']
```
mrgolin, line 189: This breaks all pyverbs tests for EFA provider.

rleon: Why? Can you please provide a fix to this get_net_name() before release of rdma-core?

mrgolin: Sure. Already have a fix, will send it shortly.

mrgolin: Please see [this PR](https://github.com/linux-rdma/rdma-core/pull/1328).

PR #1331, opened 2023-05-09
Move to fc38

rleon:


rleon:
@jgunthorpe , can you please push fc38 container to registry?
This is the reason why CI fails.

buildlib/fixup-include/sys-capability.h
```diff
@@ -0,0 +1,58 @@
...
+
+static inline cap_t cap_get_proc(void)
+{
+	char fpath[64], line[1024];
+	pid_t current = getpid();
+	long long int cap = 0;
+	FILE *fp;
+
+	sprintf(fpath, "/proc/%d/status", current);
+	printf("fpath %s\n", fpath);
```
jgunthorpe, line 35: ???

rleon, line 35: need to delete this printf

PR #1332, opened 2023-05-12
redhat: Update source filename in spec file

sunkuamzn:
rdma-core release uses .tar.gz
suse/rdma-core.spec also uses .tar.gz

### DIFF
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -9,7 +9,7 @@ Summary: RDMA core userspace libraries and daemons
 #  providers/hfi1verbs Uses the 3 Clause BSD license
 License: GPLv2 or BSD
 Url: https://github.com/linux-rdma/rdma-core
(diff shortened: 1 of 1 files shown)


sunkuamzn:
At AWS, we package rdma-core in our [EFA installer](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-enable). Currently, we have to rename the rdma-core release tarball from `rdma-core*.tar.gz` to `rdma-core*.tgz` to satisfy the spec file. Changing the spec file would it consistent with the release tarball and we won't have to rename during packaging.

jgunthorpe:
@Kamalheib  ?

michich:
As a Fedora packager, the change is fine with me.  

FWIW, rdma-core.spec in Fedora has:
Source: https://github.com/linux-rdma/rdma-core/releases/download/v%{version}/%{name}-%{version}.tar.gz

PR #1334, opened 2023-05-17
libhns: Change arrangement in SL to adapt MPI APP

hginjgerx:
Adapt hns userspace driver and obtain packet type from CQE for MPI APP.

### DIFF
--- a/providers/hns/hns_roce_u.h
+++ b/providers/hns/hns_roce_u.h
@@ -175,6 +175,12 @@ enum hns_roce_db_type {
 	HNS_ROCE_DB_TYPE_NUM
 };
 
+enum hns_roce_pktype {
+	HNS_ROCE_PKTYPE_ROCE_V1,
+	HNS_ROCE_PKTYPE_ROCE_V2_IPV6,
+	HNS_ROCE_PKTYPE_ROCE_V2_IPV4,
+};
+
 struct hns_roce_db_page {
 	struct hns_roce_db_page	*prev, *next;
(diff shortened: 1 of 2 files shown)


libibverbs/verbs.h
```diff
@@ -602,7 +608,14 @@ struct ibv_wc {
 	unsigned int		wc_flags;
 	uint16_t		pkey_index;
 	uint16_t		slid;
-	uint8_t			sl;
+	/* SL for IB or packet type
+	 * (GRH/IPv4/IPv6) for RoCE in the CQE
```
rleon, line 612: @jgunthorpe ????

hginjgerx: Hi, Leon. Let me explain the purpose of this patch more clearly.

According to Annex17_RoCEv2 (A17.4.5.2), for RoCEv2 UD, a CQE should carry a flag that indicates if the received
frame is an IPv4, IPv6 or RoCE packet. But currently, the values of the flag corresponding to these packet types haven't been defined yet in WC. As these values have already been defined In the UCX patch of https://github.com/openucx/ucx/commit/ed28845b88633e65d64fce8ec880060aa61bd59c, I think they should also be specified in WC to support UCX.

jgunthorpe: You can't change this struct, it is ABI.

hginjgerx:
Hi @rleon @jgunthorpe, the changes in verbs.h have been removed. Please re-review the changes in HNS driver.

Thanks.

PR #1348, opened 2023-06-09
stable: Tune stable branch creation process

nmorey:
Currently Azure generates release changelogs by diffing the new tag and the overall latest github release ( git log new_tag ^latest_release). This means that the changelog usually contains all the commits from the stable branch (vXX.0..vXX.N) and not just the one added from the latest release in this branch (vXX.N-1..vXX.N).

This commits changes the stable branch creation process so that azure-pipelines-release.yml will create the changelog using the latest github release in this stable branch.

### DIFF
--- a/Documentation/stable.md
+++ b/Documentation/stable.md
@@ -80,7 +80,10 @@ To do that, the creator of the branch should run
 mkdir ABI
 touch ABI/.gitignore
 git add ABI/.gitignore
-git commit -m "ABI Files"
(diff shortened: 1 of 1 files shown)


nmorey:
@jgunthorpe @rleon If you're OK with this patch, I'll run the new script bit for all maintained stable branches manually.
This way I won't have to manually prune all the changelogs every time I make a stable release. :)

rleon:
> @jgunthorpe @rleon If you're OK with this patch, I'll run the new script bit for all maintained stable branches manually. This way I won't have to manually prune all the changelogs every time I make a stable release. :)

Sure

PR #1351, opened 2023-06-14
Add general support for LTTng tracing and traces for EFA provider

YonatanNachum:
Add support for LTTng: The LTTng User Space Tracing (LTTng-UST) library allows any C/C++ application to be instrumented for and traced by LTTng.

Add CMake flag to compile in LTTng support for providers who added LTTng traces, without suppling this flags the tracing code will be compiled out without any influence on other code.

Also add initial support for EFA tracing: post receive, send and completion.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,8 @@
 #      Disable man pages. Allows rdma-core to be built and installed
 #      (without man pages) when neither pandoc/rst2man nor the pandoc-prebuilt
 #      directory are available.
+#   -DENABLE_LTTNG (default, no tracing support)
+#      Enable LTTng tracing.
 
 if (${CMAKE_VERSION} VERSION_LESS "3.18.1")
 	# Centos 7 support
@@ -555,6 +557,12 @@ if (DRM_INCLUDE_DIRS)
   else()
     unset(DRM_INCLUDE_DIRS CACHE)
   endif()
+endif()
+
+# LTTng Tracer support
+if (DEFINED ENABLE_LTTNG)
+  include(FindLTTngUST REQUIRED)
+  add_definitions(-DLTTNG_ENABLED)
 endif()
 
 #-------------------------
--- a/providers/efa/CMakeLists.txt
+++ b/providers/efa/CMakeLists.txt
@@ -1,5 +1,10 @@
+if (ENABLE_LTTNG AND LTTNGUST_FOUND)
+  set(TRACE_FILE efa_trace.c)
+endif()
+
 rdma_shared_provider(efa libefa.map
 	1 1.2.${PACKAGE_VERSION}
+	${TRACE_FILE}
 	efa.c
 	verbs.c
 )
@@ -9,3 +14,8 @@ publish_headers(infiniband
 )
 
 rdma_pkg_config("efa" "libibverbs" "${CMAKE_THREAD_LIBS_INIT}")
+
+if (ENABLE_LTTNG AND LTTNGUST_FOUND)
+	target_include_directories(efa PUBLIC ".")
+	target_link_libraries(efa LINK_PRIVATE LTTng::UST)
+endif()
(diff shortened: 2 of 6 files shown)


providers/efa/CMakeLists.txt
```diff
@@ -9,3 +14,8 @@ publish_headers(infiniband
 )
 
 rdma_pkg_config("efa" "libibverbs" "${CMAKE_THREAD_LIBS_INIT}")
+
+if (ENABLE_TRACE AND LTTNGUST_FOUND)
+	target_include_directories(efa PUBLIC ".")
```
jgunthorpe, line 19: Huh?

providers/efa/efa_trace.c
```diff
@@ -0,0 +1,9 @@
+/* SPDX-License-Identifier: GPL-2.0 OR BSD-2-Clause */
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#define LTTNG_UST_TRACEPOINT_CREATE_PROBES
+#define LTTNG_UST_TRACEPOINT_DEFINE
+
+#include "./efa_trace.h"
```
jgunthorpe, line 9: No need for ./

CMakeLists.txt
```diff
@@ -556,6 +558,21 @@ if (DRM_INCLUDE_DIRS)
   endif()
 endif()
 
+# Tracer support
+if (DEFINED ENABLE_TRACE)
+  message(STATUS "ENABLE_TRACE is set: ${ENABLE_TRACE}")
+  if ("${ENABLE_TRACE}" STREQUAL "lttng")
+    include(FindLTTngUST REQUIRED)
+    if (NOT LTTNGUST_FOUND)
```
jgunthorpe, line 566: REQUIRED already does this error handling, doesn't it?

CMakeLists.txt
```diff
@@ -556,6 +558,21 @@ if (DRM_INCLUDE_DIRS)
...
+      message(FATAL_ERROR "lttng-UST libraries not found")
+    endif()
+    add_definitions(-DTRACE_ENABLED)
+    add_definitions(-DTRACE_LTTNG)
+  else()
+    message(FATAL_ERROR "Tracer isn't supported: ${ENABLE_TRACE}")
```
jgunthorpe, line 572: Why would we support more than one kind of tracer?

jgunthorpe (changes requested):

CMakeLists.txt
```diff
@@ -556,6 +558,21 @@ if (DRM_INCLUDE_DIRS)
   endif()
 endif()
 
+# Tracer support
+if (DEFINED ENABLE_TRACE)
+  message(STATUS "ENABLE_TRACE is set: ${ENABLE_TRACE}")
```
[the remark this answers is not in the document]
YonatanNachum: Done

providers/efa/efa_trace.c
YonatanNachum: Done

providers/efa/CMakeLists.txt
YonatanNachum: This is needed for the tracepoint provider header file: efa_trace.h, LTTng needs each header file to supply the path to the header itself:
```
#define LTTNG_UST_TRACEPOINT_INCLUDE "efa_trace.h"
```
without It, I get the error:
```
fatal error: efa_trace.h: No such file or directory
 #define LTTNG_UST_TRACEPOINT_INCLUDE "efa_trace.h"
```

I tried to change to "efa/efa_trace.h" and "providers/efa/efa_trace.h" and got the same error.

If you look at LTTng-UST github example for CMake: [link](https://github.com/lttng/lttng-ust/blob/master/doc/examples/cmake-multiple-shared-libraries/CMakeLists.txt), you can see that they also used:
```
include_directories(.)
```

YonatanNachum:
Hey @jgunthorpe, kind reminder for this patch.

Thanks.

jgunthorpe:
I still think we should not try to support multiple tracers? Is there some reason? It isn't code transparent either, right?

mrgolin:
> I still think we should not try to support multiple tracers? Is there some reason? It isn't code transparent either, right?

It should be but I'm ok with supporting only LTTng and removing some of the configuration logic, at least until someone will really need this.

YonatanNachum:
Hey, @jgunthorpe, Kind reminder.

Thanks.

PR #1352, opened 2023-06-21
tests: Make the client and server members of the TestCase

ygoldamzn:
Issue is that tests fail to deregister the MR because it is busy with pending receive requests that never get released.

This change enforces destruction order of the client+server prior to base class, which makes sure the pending receives are aborted before trying to free the MR.  Previous solution was at the mercy of garbage collection and therefore flaky.

This makes the presented changes in line with other test suites.

### DIFF
--- a/tests/test_cq_events.py
+++ b/tests/test_cq_events.py
@@ -30,25 +30,26 @@ class CqEventsTestCase(RDMATestCase):
         super().setUp()
         self.iters = 100
         self.qp_dict = {'ud': CqEventsUD, 'rc': CqEventsRC}
+        self.client = None
+        self.server = None
 
     def create_players(self, qp_type):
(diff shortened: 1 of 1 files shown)


gal-pressman:
I don't think that MR deregistration should fail in such case?

EdwardSro:
But why it fails in such a case? should it?

ygoldamzn:
> But why it fails in such a case? should it? It could fail due to another issue, but that would indicate a problem with pyverbs' GC mechanism (which I doubt at this stage). Anyway, there's already a PR here: #1349 that includes this change

[#1349](https://github.com/linux-rdma/rdma-core/pull/1349) would solve this problem while streamlining test structure, so I will just abandon this PR.

PR #1355, opened 2023-06-26
tests: Add general support for HW accelerator DMA buffer testing and support for HabanaLabs testing

YonatanNachum:
Current Pyverbs HW accelerator testing supports CUDA for mlx provider and DRM if the accelerator supports it. Add a general implementation for HW accelerators for DMA buffer testing.
Also add HabanaLabs accelerators as the first accelerator to support this new implementation and more HW accelerator types will be pushed to review in the near future.


CMakeLists.txt
```diff
@@ -557,6 +558,20 @@ if (DRM_INCLUDE_DIRS)
   endif()
 endif()
 
+# Habana headers
```
rleon, line 561: All this habana support is not related to rdma-core tests which are intended to test code and devices supported by rdma-core.

YonatanNachum: Hey, @rleon.
Pyverbs already have some HW accelerator support like Cuda or DRM for HW accelerators that support it (AFAIK from Linux kernel non relevant currently).
What about testing other HW accelerators that some providers might want to test using pyverbs ? perftest does support adding more HW accelerators to their testing interface.
This can be limiting because there are HW accelerators that we have support for except for CUDA like Amazon's Trainium or Habana's Gaudi.

Thanks.

YonatanNachum: It's important to note that we don't test the accelerator itself but rdma-core - kernel - device integration with different accelerators.

rleon: Existing tests don't test accelerators, but standard dmabuf interface.

YonatanNachum: @rleon, yes exactly. Maybe I miss explained. I didn't add new tests for Habana only make the existing one work with other HW accelerator types.

gal-pressman: How do you test dmabuf without an accelerator?

YonatanNachum: Hey @gal-pressman, we currently doesn't support dma buffers without HW accelerator and this patch is for supporting dma buffers on non-Cuda accelerators.

mrgolin: @rleon How can we move forward with this? We would really like to enable dmabuf flow testing on platforms that don't have Nvidia accelerations.

jgunthorpe:
Before this could even be considered for merging we'd have to see the CI build test this stuff. I don't even have any idea where to find the habana libraries in an open source form that could be used for this. The cuda version which uses an externally provided cuda wrapper for python is in many ways cleaner for how rdma-core is setup.

CMakeLists.txt
sequoiar: @YonatanNachum Do you mean DMA ring-buffer to fetch, or DMA data buffer to transmit between GPU?

mrgolin: This PR is for testing of MR registration using `ibv_reg_dmabuf_mr` with various dmabuf exporters.

YonatanNachum:
We will think of another solution for testing DMA buffers with other HW accelerators.
Closing this for now.

PR #1360, opened 2023-06-29
bnxt_re/lib: Doorbell Drop Prevention

selvintxavier:
The idea behind this series is to prevent Doorbell drops
on some of the Broadcom adapters that require Doorbell
moderation. This is achieved by pacing the doorbell writes
into the hardware FIFO. The rate at which individual doorbells
are written needs to be dynamically adjusted, because
it depends on the ability of the hardware to drain the
FIFO and on the number and behavior of individual
doorbell writers. When congestion is detected by the user
library, it notifies the driver and driver adjust the
pacing parameters dynamically in a shared page, which will
be used for pacing the Doorbells.

### DIFF
--- a/providers/bnxt_re/db.c
+++ b/providers/bnxt_re/db.c
@@ -37,8 +37,62 @@
  */
 
 #include <util/mmio.h>
+#include <ccan/minmax.h>
 #include <stdlib.h>
 #include "main.h"
+
+static uint16_t rnd(struct xorshift32_state *state, uint16_t range)
+{
+	/* range must be a power of 2 - 1 */
+	return (xorshift32(state) & range);
+}
+
+static int calculate_fifo_occupancy(struct bnxt_re_context *cntx)
+{
+	struct bnxt_re_pacing_data *pacing_data =
+		(struct bnxt_re_pacing_data *)cntx->dbr_page;
+	struct bnxt_re_dev *rdev = cntx->rdev;
+	uint32_t read_val, fifo_occup;
+	uint64_t fifo_reg_off;
+	uint64_t *dbr_map;
+
+	fifo_reg_off =  pacing_data->grc_reg_offset & ~(BNXT_RE_PAGE_MASK(rdev->pg_size));
+	dbr_map = cntx->bar_map + fifo_reg_off;
+
+	read_val = *dbr_map;
+	fifo_occup = pacing_data->fifo_max_depth -
+		((read_val & pacing_data->fifo_room_mask) >>
+		 pacing_data->fifo_room_shift);
+
+	return fifo_occup;
+}
+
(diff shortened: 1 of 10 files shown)


providers/bnxt_re/db.c
```diff
@@ -40,6 +40,75 @@
...
+
+	read_val = *dbr_map;
+	fifo_occup = pacing_data->fifo_max_depth -
+		((read_val & pacing_data->fifo_room_mask) >>
+		 pacing_data->fifo_room_shift);
+
+	return fifo_occup;
+}
+
+static inline uint32_t find_min(uint32_t x, uint32_t y)
```
rleon, line 80: Please reuse existing helpers - ccan/minmax.h

providers/bnxt_re/db.c
```diff
@@ -40,6 +40,75 @@
 #include <stdlib.h>
 #include "main.h"
 
+static uint32_t xorshift32(struct xorshift32_state *state)
```
rleon, line 43: This function is called only once.

providers/bnxt_re/verbs.c
```diff
@@ -102,6 +102,20 @@ static int bnxt_re_map_db_page(struct ibv_context *ibvctx,
...
 
+int bnxt_re_notify_drv(struct ibv_context *ibvctx)
+{
+	DECLARE_COMMAND_BUFFER(cmd,
+			       BNXT_RE_OBJECT_NOTIFY_DRV,
+			       BNXT_RE_METHOD_NOTIFY_DRV,
+			       0);
+	int ret;
+
+	ret = execute_ioctl(ibvctx, cmd);
```
rleon, line 113: return execute_ioctl(); directly without need in ret.

providers/bnxt_re/db.c
selvintxavier: > This function is called only once.

isn't it okay to keept it as it is? It is implementing the xorshift algo and I feel its better to keep it as a separate function.  Is it sufficient to mark it as an inline function?

providers/bnxt_re/db.c
selvintxavier: > Please reuse existing helpers - ccan/minmax.h

Will do

providers/bnxt_re/verbs.c
selvintxavier: > return execute_ioctl(); directly without need in ret.

will do. thanks

providers/bnxt_re/db.c
rleon: no inline ,please.

maybe add that function to util/ folder.

selvintxavier: > no inline ,please.
> 
> maybe add that function to util/ folder.

Sure.. will do.

PR #1365, opened 2023-07-04
providers/irdma: Verify max_send_wr and max_recv_wr

Kamalheib:
Make sure to verify that the provided max_send_wr and max_recv_wr are in the supported range.

Fixes: 14a0fc824f16 ("rdma-core/irdma: Implement device supported verb APIs")

### DIFF
--- a/providers/irdma/uverbs.c
+++ b/providers/irdma/uverbs.c
@@ -1328,6 +1328,8 @@ struct ibv_qp *irdma_ucreate_qp(struct ibv_pd *pd,
 
 	if (attr->cap.max_send_sge > uk_attrs->max_hw_wq_frags ||
 	    attr->cap.max_recv_sge > uk_attrs->max_hw_wq_frags ||
+	    attr->cap.max_send_wr > uk_attrs->max_hw_wq_quanta ||
+	    attr->cap.max_recv_wr > uk_attrs->max_hw_rq_quanta ||
 	    attr->cap.max_inline_data > uk_attrs->max_hw_inline) {
 		errno = EINVAL;
 		return NULL;


Kamalheib:
Hi @shirazsaleem, Could you please review this PR?

providers/irdma/uverbs.c
```diff
@@ -1325,10 +1326,13 @@ struct ibv_qp *irdma_ucreate_qp(struct ibv_pd *pd,
...
 			      ibv_ctx.context);
 	uk_attrs = &iwvctx->uk_attrs;
+	dev_attrs = &iwvctx->dev_attrs;
 
 	if (attr->cap.max_send_sge > uk_attrs->max_hw_wq_frags ||
 	    attr->cap.max_recv_sge > uk_attrs->max_hw_wq_frags ||
-	    attr->cap.max_inline_data > uk_attrs->max_hw_inline) {
+	    attr->cap.max_inline_data > uk_attrs->max_hw_inline ||
+	    attr->cap.max_send_wr > dev_attrs->max_qp_wr ||
+	    attr->cap.max_recv_wr > dev_attrs->max_qp_wr) {
```
shirazsaleem, line 1335: Ummm...this will probably make some existing customers/app unhappy.

The max send/recv WRs are really contingent on the what the max SGE and max inline data used in this QP as the WQE's are sized based on this. See irdma_get_wqe_shift. 

So we need a combination of the incoming max send WR's and the max WQE size for this QP [shift] to determine if we will go past the device supported WQ size. This is done irdma_get_sqdepth, irdma_get_rqdepth

max_qp_wr "assumes" the largest WQE size [8 quantas or 256Bytes] and therefore is too conservative an upper bound to check the max_send/recv WRs against for this QP. We could do better if this QP used lower values of max_send_sge/max_inline [for example, it yielded a WQE size of 1 quanta or 32 bytes]. We would end up incorrectly return -EINVAL in this case.

Changing max_qp_wr to be based on the smallest WQE size also has drawbacks. i.e we would end up overflowing the WQ if the incoming  
max_send_sge/max_inline happened to make the WQE size > smallest value.

shirazsaleem: The max values you will get out send_wr and recv_wr are 32,510 and 32767 respectively.

Kamalheib: @shirazsaleem : Fixed, Could you please review?

shirazsaleem: I misunderstood your request, I thought you were trying to update an application with what to pass in max_send/recv_wr.

providers/irdma/uverbs.c
```diff
@@ -1328,7 +1328,9 @@ struct ibv_qp *irdma_ucreate_qp(struct ibv_pd *pd,
 
 	if (attr->cap.max_send_sge > uk_attrs->max_hw_wq_frags ||
 	    attr->cap.max_recv_sge > uk_attrs->max_hw_wq_frags ||
-	    attr->cap.max_inline_data > uk_attrs->max_hw_inline) {
+	    attr->cap.max_inline_data > uk_attrs->max_hw_inline ||
+	    attr->cap.max_send_wr > IRDMA_MAX_SEND_WR ||
+	    attr->cap.max_recv_wr > IRDMA_MAX_RECV_WR) {
```
shirazsaleem, line 1333: I am not seeing a need for this.

irdma_get_sqdepth guards against unsupported values of passed in max_send_wr?

Kamalheib: I was trying to avoid the failure of the following pyverbs tests, The test tries to pass the following values and expects the driver to return EINVAL...

qp_init_attr.max_recv_wr = 0xffffffff # max_uint32
qp_init_attr.max_send_wr = 0xffffffff # max_uint32

```
 ./build/bin/run_tests.py --dev irdma1 test_qp 
ss.F.Fss....ssssss..s.sss.
FAIL: test_create_raw_qp_ex_with_illegal_caps_max_recv_wr (tests.test_qp.QPTest)
Test Raw Packet QP creation via ibv_create_qp_ex with a QPAttr object with illegal max_recv_wr.
  File "/home/rdma-core/tests/test_qp.py", line 234, in test_create_raw_qp_ex_with_illegal_caps_max_recv_wr
    self.create_qp_common_test(e.IBV_QPT_UD, e.IBV_QPS_RTS, False, True,
AssertionError: PyverbsRDMAError not raised

FAIL: test_create_raw_qp_ex_with_illegal_caps_max_send_wr (tests.test_qp.QPTest)
Test Raw Packet QP creation via ibv_create_qp_ex with a QPAttr object with illegal max_send_wr.
  File "/home/rdma-core/tests/test_qp.py", line 201, in test_create_raw_qp_ex_with_illegal_caps_max_send_wr
    self.create_qp_common_test(e.IBV_QPT_UD, e.IBV_QPS_RTS, False, True,
AssertionError: PyverbsRDMAError not raised

Ran 26 tests in 0.020s
```

shirazsaleem: Ah ok. Thanks! Seems like an overflow in irdma_get_sqdepth in the roundup function. Let me review it once and see what is the best path forward.

shirazsaleem: Sorry for the delay!

The limits are stored in uk_attrs struct already and no need for the defines. So something like this should do it.

@@ -2129,6 +2130,8 @@ struct ibv_qp *irdma_ucreate_qp(struct ibv_pd *pd,
 #endif
        if (attr->cap.max_send_sge > uk_attrs->max_hw_wq_frags ||
            attr->cap.max_recv_sge > uk_attrs->max_hw_wq_frags ||
+           attr->cap.max_send_wr > uk_attrs->max_hw_wq_quanta ||
+           attr->cap.max_recv_wr > uk_attrs->max_hw_rq_quanta ||
            attr->cap.max_inline_data > uk_attrs->max_hw_inline) {
                errno = EINVAL;
                return NULL;

Kamalheib: Fixed

shirazsaleem (approved):

PR #1366, opened 2023-07-04
tests: Skip CQ shrinking for irdma devices

Kamalheib:
The test_resize_cq assumes that when shrinking an active CQ with unpolled entries it fail with an EINVAL error, but the irdma hardware is implementing the resize CQ differently and allowing to shrink a CQ without losing the unpolled entries.

To avoid the test_resize_cq failure for irdma devices, skip the test case.

### DIFF
--- a/tests/irdma_base.py
+++ b/tests/irdma_base.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: (GPL-2.0 OR Linux-OpenIB)
+# Copyright (c) 2023 Red Hat, Inc, All rights reserved.  See COPYING file
+
+import unittest
+
+INTEL_VENDOR_ID = 0x8086
+IRDMA_DEVS = {
+        0x1572, # I40E_DEV_ID_SFP_XL710
+        0x1574, # I40E_DEV_ID_QEMU
+        0x1580, # I40E_DEV_ID_KX_B
+        0x1581, # I40E_DEV_ID_KX_C
+        0x1583, # I40E_DEV_ID_QSFP_A
+        0x1584, # I40E_DEV_ID_QSFP_B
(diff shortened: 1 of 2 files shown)


tests/irdma_base.py
```diff
@@ -0,0 +1,68 @@
...
+        0x1891, # Intel(R) Ethernet Connection C822N for QSFP
+        0x1892, # Intel(R) Ethernet Connection C822N for SFP
+        0x1893, # Intel(R) Ethernet Connection E822-C/X557-AT 10GBASE-T
+        0x1894, # Intel(R) Ethernet Connection E822-C 1GbE
+        0x1897, # Intel(R) Ethernet Connection E822-L for backplane
+        0x1898, # Intel(R) Ethernet Connection E822-L for SFP
+        0x1899, # Intel(R) Ethernet Connection E822-L/X557-AT 10GBASE-T
+        0x189A, # Intel(R) Ethernet Connection E822-L 1GbE
+}
+
```
EdwardSro, line 60: leave two empty lines before module-level functions (PEP-8 convention)

Kamalheib: Will do.

Kamalheib: Fixed

tests/irdma_base.py
```diff
@@ -0,0 +1,69 @@
...
+
+def is_irdma_dev(ctx):
+    dev_attrs = ctx.query_device()
+    return dev_attrs.vendor_id == INTEL_VENDOR_ID and \
+            dev_attrs.vendor_part_id in IRDMA_DEVS
+
```
EdwardSro, line 66: also here

EdwardSro (approved):

PR #1369, opened 2023-07-12
tests: Add event-enabled tests over SRD

ygoldamzn:
Added event enabled cq tests over SRD, with new and old send.

### DIFF
--- a/tests/test_efa_srd.py
+++ b/tests/test_efa_srd.py
@@ -4,6 +4,7 @@
 import unittest
 import errno
 
+from pyverbs.cq import CQ, CompChannel
 from pyverbs.pyverbs_error import PyverbsRDMAError
 import pyverbs.enums as e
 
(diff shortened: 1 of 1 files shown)


EdwardSro (changes requested):
Please move the test to test_efa_srd.py or another EFA-specific test file

ygoldamzn:
> Please move the test to test_efa_srd.py or another EFA-specific test file

Done.
I've decided to duplicate the create_cq_with_comp_channel() method because:
1. sharing between test suites introduces a bad dependency since there's no real sense of hierarchy there
2. it seems too specific to move to a base library

tests/test_efa_srd.py
```diff
@@ -12,6 +13,29 @@
...
+        self.cq = CQ(self.ctx, self.num_msgs, None, self.comp_channel)
+        self.cq.req_notify()
+
+
+class CqEventsSRDTestCase(EfaRDMATestCase):
+    def setUp(self):
+        super().setUp()
+        self.iters = 100
+        self.server = None
+        self.client = None
```
EdwardSro, line 31: BTW, no need to initiate server and client as it's already done in RDMATestCase.__init__()
Except for that, it looks good to me.

EdwardSro (approved):

tests/test_efa_srd.py
ygoldamzn: OK, removed.

EdwardSro (approved):

PR #1374, opened 2023-07-27
Wq size raw attr status

tatyana-en:
"Allow accurate reporting on QP max send/recv WR" and "Use HW specific minimum WQ size" are ABI changes and come with corresponding kernel patches (already submitted).

"Remove enum irdma_status_code" applies after the two ABI changes patches.


kernel-headers/rdma/irdma-abi.h
```diff
@@ -1,6 +1,6 @@
-/* SPDX-License-Identifier: (GPL-2.0 WITH Linux-syscall-note) OR Linux-OpenIB */
+/* SPDX-License-Identifier: (GPL-2.0 WITH Linux-syscall-note) OR Linux-OpenIB) */
 /*
- * Copyright (c) 2006 - 2021 Intel Corporation.  All rights reserved.
+ * Copyright (c) 2006 - 2022 Intel Corporation.  All rights reserved.
```
rleon, line 3: If you decided to change it, we are in 2023 now.

rleon: Please use kernel-header/update script to generate this commit, and this patch should be first in the series.

tatyana-en: Thanks for the feedback.

rleon:
The commit which is used in first patach doesn't exist in kernel tree.
"To commit a9e82b8a663b ("RDMA/irdma: Fix checkpatch in PR workflow")."

rleon:
> > The commit which is used in first patach doesn't exist in kernel tree. "To commit a9e82b8a663b ("RDMA/irdma: Fix checkpatch in PR workflow")."
> 
> Hi Leon - Tatyana is out on vacation. We can fix this when she is back. Unless your ok with me spinning a new PR for this.

Whatever works for you.

rleon:
Change first commit to point to right kernel commit
... To commit 72d422c2465e ("RDMA/irdma: Use HW specific minimum WQ size")....

and pushed manually

PR #1377, opened 2023-08-03
Export symbols

rleon:


rleon:
@nmorey can you please take it to all stables once this PR is accepted?

jgunthorpe (changes requested):

rleon:
For lld 16, this is the same as inline function. The COMPAT_SYMVER_FUNC generates right assembly anyway. nm -D libiibverbs.so is the same as before.

On Thu, Aug 3, 2023, at 18:14, Jason Gunthorpe wrote:
> 
> 
> ***@***.**** requested changes on this pull request.
> 
> 
> In libibverbs/libibverbs.map.in <https://github.com/linux-rdma/rdma-core/pull/1377#discussion_r1283351085>:
> 
> > @@ -94,7 +93,6 @@ IBVERBS_1.1 {
>  		ibv_query_srq;
>  		ibv_rate_to_mbps;
>  		ibv_reg_mr;
> -		ibv_register_driver;
> 
> Why this one?
> 
> `COMPAT_SYMVER_FUNC(ibv_register_driver, 1_1, "IBVERBS_1.1",
>                    void,
>                    const char *name, ibv_driver_init_func_1_1 init_func)
> {
>         /* The driver interface is private as of rdma-core 13. This stub is
>          * left to preserve dynamic-link compatibility with old libfabrics
>          * usnic providers which use this function only to suppress a fprintf
>          * in old versions of libibverbs. */
> }
`

jgunthorpe:
Oh I really don't like that, it seems like a LLD bug that throws an error for explicitly versionsed symbols. I'm worried this will quietly break one of the other linkers.

rleon:
So what do you suggest? Leave it?

jgunthorpe:
I'd leave it yes, and file a bug report on lld about it since it is a recent regression

nmorey:
Pushed to stable branches v28-47

rleon:
Thanks

PR #1381, opened 2023-08-08
providers/mlx5: Fixed bug with mlx5_post_wq_recv not supporting Multi-Packet queues 

AjayBrahmakshatriya:
There is a bug in the current implementation of mlx5/mlx5_post_wq_recv when Multi-Packet queues are enabled via - `MLX5DV_WQ_INIT_ATTR_MASK_STRIDING_RQ`. The wqe buffer is allocated with the appropriate size but is not accessed correctly while posting recvs. 

This results in the recv buffers to posted incorrectly and the application receives a local protection error status in the work completion. This PR fixes this with the following changes - 

1. Add a field `bool is_mprq` to the `struct mlx5_rwq`. 
2. Set this field to true in `create_wq` if `MLX5DV_WQ_INIT_ATTR_MASK_STRIDING_RQ` is set. 
3. If this field is set in `mlx5_post_wq_recv`, use `get_wq_mprq_recv_wqe` to get the address of the `struct mlx5_wqe_data_seg` which uses the `struct mlx5_mprq_wqe` to calculate the address. This struct is already defined under mlx5dv.h

This change has been tested and calling `ibv_post_wq_recv` from the application registers the buffers appropriately and the application is able to receive the data. This change doesn't break the behavior when `MLX5DV_WQ_INIT_ATTR_MASK_STRIDING_RQ` is not enabled and has been tested.

### DIFF
--- a/providers/mlx5/qp.c
+++ b/providers/mlx5/qp.c
@@ -72,8 +72,8 @@ static void *get_wq_recv_wqe(struct mlx5_rwq *rwq, int n)
 }
 
 static void *get_wq_mprq_recv_wqe(struct mlx5_rwq *rwq, int n) {
-	struct mlx5_mprq_wqe* wqe = rwq->pbuff;
-	return &(wqe[n].dseg);
+	return rwq->pbuff + (n << rwq->rq.wqe_shift) 
+		+ sizeof(struct mlx5_wqe_srq_next_seg);
 }
 static int copy_to_scat(struct mlx5_wqe_data_seg *scat, void *buf, int *size,
 			 int max, struct mlx5_context *ctx)
--- a/providers/mlx5/qp.c
+++ b/providers/mlx5/qp.c
@@ -71,6 +71,10 @@ static void *get_wq_recv_wqe(struct mlx5_rwq *rwq, int n)
 	return rwq->pbuff  + (n << rwq->rq.wqe_shift);
 }
 
+static void *get_wq_mprq_recv_wqe(struct mlx5_rwq *rwq, int n) {
+	struct mlx5_mprq_wqe* wqe = rwq->pbuff;
+	return &(wqe[n].dseg);
+}
 static int copy_to_scat(struct mlx5_wqe_data_seg *scat, void *buf, int *size,
 			 int max, struct mlx5_context *ctx)
 {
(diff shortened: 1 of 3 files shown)


providers/mlx5/qp.c
```diff
@@ -3771,8 +3775,10 @@ int mlx5_post_wq_recv(struct ibv_wq *ibwq, struct ibv_recv_wr *wr,
...
 			goto out;
 		}
-
-		scat = get_wq_recv_wqe(rwq, ind);
+		if (rwq->is_mprq)
+			scat = get_wq_mprq_recv_wqe(rwq, ind);
+		else 
+			scat = get_wq_recv_wqe(rwq, ind);
 		sig = (struct mlx5_rwqe_sig *)scat;
 		if (unlikely(rwq->wq_sig)) {
```
AjayBrahmakshatriya, line 3783: While this PR is open, I want to ask someone from the provides/mlx5 team if this change would interfere with the `memset` in this branch. Should I change the implementation advance the pointer `scat` after the memset?

yishaih: The functionality for Multi-Packet RQ was exposed over mlx5 DV API for the control path to be used later on by layers who may implement data path by themselves as of UCX/VMA, etc. and are fully familiar with the device specification. 

The data path flow is another full feature that may touch post_recv(), poll_cq(), etc.

As of that you can't just fix/adapt this small part, this will not work.

Please close this PR.

AjayBrahmakshatriya: I see, in that case would it be a good idea to post the `post_recv` function pointer in the ibv_wq structure to point to a function that explicitly errors to the user with the proper message rather than calling the current implementation that doesn't work. 
My concern is that (AFAIK) this not documented anywhere - that users should use rdma_core only for the control path and spin up their own implementation for the data path.

yishaih: As this is a DV API, we may expect its users to be familiar with the device specification/expected usage to some point.

For example the 'stride' properties from struct mlx5dv_striding_rq_init_attr are device specific ones, which to use them properly needs a close device knowledge.
 
As of that, we didn't set/replace a function pointer just for giving a non expected user an error message for such a DV scenario.

In case you really need a full striding RQ working solution please contact NVIDIA support for getting help/consider other layers as of UCX/VMA who may support that already.

Please close that PR as it can't work.

Thanks.

AjayBrahmakshatriya: Thanks for your response. I will close the PR.

PR #1382, opened 2023-08-23
efa: Change resource deallocation order in QP and CQ

YonatanNachum:
When trying to destroy QP or CQ, we first release rdma-core resources for the relevant object (mapping for address ranges, refcounts, locks etc.) and only then request the driver to destroy the object. If the driver or the device fails, the object isn't fully destroyed so the user can try to destroy the object again which will lead to trying to release already released rdma-core resources for the object. Change deallocation order to first call kernel and driver code and if succeeded release rdma-core resources.

### DIFF
--- a/providers/efa/verbs.c
+++ b/providers/efa/verbs.c
@@ -967,17 +967,17 @@ int efa_destroy_cq(struct ibv_cq *ibvcq)
 	struct efa_cq *cq = to_efa_cq(ibvcq);
 	int err;
 
-	munmap(cq->db, to_efa_dev(cq->verbs_cq.cq.context->device)->pg_sz);
-	munmap(cq->buf, cq->buf_size);
-
-	pthread_spin_destroy(&cq->lock);
-
 	err = ibv_cmd_destroy_cq(ibvcq);
 	if (err) {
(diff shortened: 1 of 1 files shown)


gal-pressman:
The patch looks fine, though I'm quite confused why would a user try to destroy the same object twice? If the device failed to destroy it the first time, what is expected to change the second time?

Doesn't it indicate there's something broken in the user's app?

jgunthorpe:
I think this is going after the crazy "destroy can fail". If destroy fails then the expectation is the user will call destroy again and it might succeed. 'again' might be after it does some other stuff. Most reasons destroy fails are because the object is in use and yes it probably always indicates an application bug.

Regardless, if the provider chooses to fail destroy it must not corrupt the object so badly that another destroy will crash.

YonatanNachum:
Yes this is the case and it does point to some application bug, its an edge case that doesn't suppose to happen at all but we prefer to cover this. It is also seems that other providers also first call the driver and only after release rdma-core resources.
This patch is in combination with a driver patch I also uploaded:  [link](https://www.spinics.net/lists/linux-rdma/msg119512.html)

YonatanNachum:
Hey @jgunthorpe, kind reminder for this patch.

Thanks.

PR #1386, opened 2023-09-06
Do not merge this commit. Testing check of Fixes tag

nmorey:
Fixes: deadbeef ("This is not a real commit")


rleon:
> @rleon Seems like we're ignoring all checkpatch errors as well: https://dev.azure.com/ucfconsort/rdma-core/_build/results?buildId=69066&view=logs&j=143eea66-4597-539b-c2e7-6e457837a569&t=cf8c41fc-a364-5748-ae55-aa685dbb48b3
> 
> Just a warning in Azure but all green here in github
> 
> Is this the intended behaviour?

Hard to say, before azp script, I checked for errors.

PR #1389, opened 2023-09-20
libhns: Support SRQ record doorbell

hginjgerx:
Compared with normal doorbell, using record doorbell can shorten the process of ringing the doorbell and reduce the latency.

During SRQ creation, the kernel driver will allocate doorbell buffer and notify userspace whether the SRQ record doorbell is enabled with the flag HNS_ROCE_RSP_SRQ_CAP_RECORD_DB. The userspace driver will decide whether to use record doorbell or normal doorbell based on this flag in post SRQ recv process.

This patch relies on the corresponding kernel patch: RDMA/hns: Support SRQ record doorbell

### DIFF
--- a/kernel-headers/rdma/hns-abi.h
+++ b/kernel-headers/rdma/hns-abi.h
@@ -52,15 +52,25 @@ struct hns_roce_ib_create_cq_resp {
 	__aligned_u64 cap_flags;
 };
 
+enum hns_roce_srq_cap_flags {
+	HNS_ROCE_SRQ_CAP_RECORD_DB = 1 << 0,
+};
+
+enum hns_roce_srq_cap_flags_resp {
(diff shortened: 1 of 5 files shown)


hginjgerx:
Hi, @jgunthorpe @rleon 
the kernel patch has been applied. Would you please review this PR again?

rleon:
UAPI headers patch should be prepared with https://github.com/linux-rdma/rdma-core/blob/master/kernel-headers/update script

hginjgerx:
> UAPI headers patch should be prepared with https://github.com/linux-rdma/rdma-core/blob/master/kernel-headers/update script
Done.

hginjgerx:
Hi, @rleon @jgunthorpe 
the uapi header issue has been fixed. Would you please review this PR again?

PR #1390, opened 2023-09-20
verbs: Add support for XDR speed in ibv_query_port

yishaih:
This series adds support for XDR speed in ibv_query_port.

Its output structure was extended to include the XDR capabilities (i.e. active_speed_ex and IBV_PORT_LINK_SPEED_XDR_SUP) to let applications get this information.

The man page was updated to describe the expected usage from an application point of view.

The series follows the submission that was done in the kernel for rdma-next.


rleon:
@nmorey probably we need to change azp-checkpatch script to ignore commit SHAs that are not in Fixes line.

nmorey:
@rleon It does already. Unless it's prefixed by `commit `... But not by something like `commit:`
As we download checkpatch.pl on the fly, it's not that easy to change that behavior.
It's always possible to fix the regexp with sed but we run the risk of it breaking when it gets updated.

rleon:
> @rleon It does already. Unless it's prefixed by `commit `... But not by something like `commit:` As we download checkpatch.pl on the fly, it's not that easy to change that behavior. It's always possible to fix the regexp with sed but we run the risk of it breaking when it gets updated.

More simple solution will be not run checkpatch on "Update kernel headers .." patches.

nmorey:
A simple change like that to the update scripts would also work around the issue:
```
diff --git a/kernel-headers/update b/kernel-headers/update
index 48a716d583eb..19a1d340dfb9 100755
--- a/kernel-headers/update
+++ b/kernel-headers/update
@@ -151,7 +151,7 @@ def make_commit(args,ntree,kernel_desc):
 
     # And now create the commit
     msg="Update kernel headers\n\n";
-    p = textwrap.fill("To commit %s."%(kernel_desc.decode()),
+    p = textwrap.fill("To commit: %s."%(kernel_desc.decode()),
                       width=74)
     msg = msg + p;
     msg = msg + "\n\nSigned-off-by: %s\n"%(emaila);
```

Because checkpatch uses this regexp:
`		if ($in_commit_log && $line =~ /(^fixes:|\bcommit)\s+([0-9a-f]{6,40})\b/i) {`

The extra `:` means it won't trigger

nmorey:
@rleon It seems to do the trick.
Thanks @yishaih

PR #1391, opened 2023-09-22
providers/bnxt_re: Fix bnxt_re_alloc_queue_ptr error flow

Kamalheib:
Make sure to free the "qp->jsqq" pointer in the right place and add checks to make sure that the "hwque" pointers are allocated before trying to free them.

Fixes: f92837e29fd4 ("bnxt_re/lib: consolidate hwque and swque in common structure")


providers/bnxt_re/verbs.c
```diff
@@ -1012,11 +1012,13 @@ static int bnxt_re_check_qp_limits(struct bnxt_re_context *cntx,
 static void bnxt_re_free_queue_ptr(struct bnxt_re_qp *qp)
 {
 	if (qp->jrqq) {
-		free(qp->jrqq->hwque);
+		if (qp->jrqq->hwque)
```
rleon, line 1015: Why did you add these checks in all places?
https://linux.die.net/man/3/free
If ptr is NULL, no operation is performed.

Kamalheib: fixed

PR #1393, opened 2023-09-24
/sys/class/infiniband_verbs/uverbs0 is a directory

afazekas:
According to the ABI doc this item supposed to have sub elements (yes they does) [1]

It fixes the issue of /etc/rdma/mlx4.conf  not loaded at initramfs time, in hostonly  (default) mode.

[1] https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-class-infiniband

### DIFF
--- a/redhat/rdma.modules-setup.sh
+++ b/redhat/rdma.modules-setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 check() {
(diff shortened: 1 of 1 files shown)


nmorey:
Your commit lacks a Fixes tag:
`Fixes: 39fa824dd80e ("redhat: add udev/systemd/etc infrastructure bits")`

PR #1402, opened 2023-10-25
tests: Fix test_resize_srq

EdwardSro:
According to the IB Spec, Vol 1, Rel-1.6.2022.07-15b: 10.2.9.3 SHARED RECEIVE QUEUE MODIFICATION:
"If the HCA supports the modification of the maximum number of outstanding SRQ WRs, then Consumer may request to change the maximum number of outstanding WRs through the Modify SRQ verb. The CI returns the actual maximum number of WQEs that can be outstanding on the SRQ, this value must be greater than or equal to the number requested."

Change the max_wr assert to greater or equal to the requested value (instead of equal).
In addition, remove max_sge modification as it's irrelevant and not supported.

### DIFF
--- a/tests/test_srq.py
+++ b/tests/test_srq.py
@@ -20,23 +20,21 @@ class SrqTestCase(RDMATestCase):
 
     def test_resize_srq(self):
         """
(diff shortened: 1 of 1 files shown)


EdwardSro:
@rpears0n

PR #1408, opened 2023-11-22
verbs: reserve unused 24bits in sq_sig_all for extension

changchengx:
1. keep compatibility
2. [ib_uverbs_create_qp::sq_sig_all](https://github.com/linux-rdma/rdma-core/blob/v49.0/kernel-headers/rdma/ib_user_verbs.h#L598) occupy 8 bits.

### DIFF
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -922,7 +922,15 @@ struct ibv_qp_init_attr {
 	struct ibv_srq	       *srq;
 	struct ibv_qp_cap	cap;
 	enum ibv_qp_type	qp_type;
-	int			sq_sig_all;
+	struct {
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+		int sq_sig_all:8;
+		int reserved:24;
+#else
+		int reserved:24;
+		int sq_sig_all:8;
+#endif
+	};
 };
 
 enum ibv_qp_init_attr_mask {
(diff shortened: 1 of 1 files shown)


changchengx:
@rleon Please help review this PR when you have time.

jgunthorpe:
No to this, we don't know what clients are doing with this memory, we can't just assume a bitfiled will magically keep all existing users working correctly.

changchengx:
@jgunthorpe The application should set the existed ```int sq_sig_all``` to be 0 or 1. With this change, it could still keep compatibility with the existed application.

jgunthorpe:
I don't think we can assume apps will set it to only 0 or 1 unfortunately. Probably most do but it was never enforced..

changchengx:
@jgunthorpe 
>
>
I agree with this point.

However, in the [ib_uverbs_create_qp::sg_sig_all](https://github.com/linux-rdma/rdma-core/blob/v49.0/kernel-headers/rdma/ib_user_verbs.h#L598) & [ib_uverbs_ex_create_qp::sg_sig_all](https://github.com/linux-rdma/rdma-core/blob/v49.0/kernel-headers/rdma/ib_user_verbs.h#L624) & [ib_uverbs_query_qp_resp::sg_sig_all](https://github.com/linux-rdma/rdma-core/blob/v49.0/kernel-headers/rdma/ib_user_verbs.h#L718), the member type is ```__u8```.
This PR won't touch the LSB 8 bits. For example, for [ibv_query_qp](https://github.com/linux-rdma/rdma-core/blob/v49.0/libibverbs/cmd.c#L477), actually, it only set the LSB 8 bits.

jgunthorpe:
That is only half the problem, if you expect to re-use the upper bits for something else you can't reliably be sure existing apps actually left them at zero since nothing checked.

changchengx:
@jgunthorpe 
>  you can't reliably be sure existing apps actually left them at zero since nothing checked.
>
Besides we can't confirm that the existing applications didn't use the MSB 24bits, do we have other concern about this PR?

jgunthorpe:
I certainly wouldn't take this without something else to motivate doing this kind of risky thing

PR #1412, opened 2023-12-05
Add query MR EFA direct verb

mrgolin:
This PR is adding a new EFA DV interface to query for MR attributes, with an initial implementation that returns PCIe bus IDs used by the device to reach the MR to write data for receive, RDMA read, and RDMA write receive operations.
Documentation was updated accordingly and pyverbs layer and a test to cover this new functionality are being added.

PR #1420, opened 2024-01-16
Add convenience functions to ioctl mechanism

mrgolin:
From discussions [here](https://github.com/linux-rdma/rdma-core/pull/1412) and [here](https://lore.kernel.org/all/20240108180636.GM50406@nvidia.com/) seems it will be useful to have an easy way to get optional output field validity.


rleon:
I don't see any users of these new functions.

PR #1423, opened 2024-01-29
rstream [RFC]: Set INLINE data for iWARP

tatyana-en:
Set rsocket SQ INLINE to RS_MSG_SIZE for iWARP.

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -917,8 +917,12 @@ static int rs_create_ep(struct rsocket *rs)
 	int i, ret;
 
 	rs_set_qp_size(rs);
-	if (rs->cm_id->verbs->device->transport_type == IBV_TRANSPORT_IWARP)
+	if (rs->cm_id->verbs->device->transport_type == IBV_TRANSPORT_IWARP) {
 		rs->opts |= RS_OPT_MSG_SEND;
+
+		if (rs->sq_inline < RS_MSG_SIZE)
+			rs->sq_inline = RS_MSG_SIZE;
+	}
 	ret = rs_create_cq(rs, rs->cm_id);
 	if (ret)
 		return ret;


rleon:
@tatyana-en , rstream as an example application, why is this change marked as RFC?

tatyana-en:
It is marked as RFC to get comments, since I didn't discuss it with an rstream maintainer.

rleon:
> It is marked as RFC to get comments, since I didn't discuss it with an rstream maintainer.

Are you referring to Sean? @shefty

librdmacm/examples/rstream.c
```diff
@@ -304,7 +304,7 @@ static void set_options(int fd)
 			rs_setsockopt(fd, SOL_RDMA, RDMA_INLINE, &inline_size,
 				      sizeof inline_size);
 		} else if (optimization == opt_bandwidth) {
-			val = 0;
+			val = 4;
```
shefty, line 307: I'm soooo out of touch with this code...  One could argue the problem is actually in rs_create_ep() and not rstream.  The spot in rs_create_ep() where it does this:
```
    if (rs->cm_id->verbs->device->transport_type == IBV_TRANSPORT_IWARP)
        rs->opts |= RS_OPT_MSG_SEND;
```
adjusts the rsocket options, which causes the check below (also in rs_create_ep()) to fail:
```
    if ((rs->opts & RS_OPT_MSG_SEND) && (rs->sq_inline < RS_MSG_SIZE))
        return ERR(ENOTSUP);
```
Technically, there's nothing a use did wrong here.  I think a better fix is to handle this internally.  Something like:
```
    /* See comment for RS_OPT_MSG_SEND */
    if (rs->cm_id->verbs->device->transport_type == IBV_TRANSPORT_IWARP) {
        rs->opts |= RS_OPT_MSG_SEND;
        if (rs->sq_inline < RS_MSG_SIZE)
            rs->sq_inline = RS_MSG_SIZE;
    }
```

tatyana-en: Thanks for the feedback!

shefty:
I meant to the pull request itself.  That is, where is says "rstream [RFC]: Set INLINE data for iWARP".  It doesn't really matter, since it doesn't end up in the git logs, but can be helpful for whoever is managing the pull requests.

tatyana-en:
> I meant to the pull request itself. That is, where is says "rstream [RFC]: Set INLINE data for iWARP". It doesn't really matter, since it doesn't end up in the git logs, but can be helpful for whoever is managing the pull requests.

Unfortunately, I don't know how to do this without closing this PR and opening a new one with a different name.

PR #1424, opened 2024-02-02
bnxt_re/lib: Reduce the Doorbells during SRQ and CQ handling

selvintxavier:
Some optimizations in data path for Broadcom devices. This series implements some Doorbell moderation techniques to avoid some latency in data path and help to reduce the Doorbell HW FIFO congestion.


selvintxavier:
> @selvintxavier when we merge PR, everything that is written there will be visible in commit history, including "Please review and apply Thanks, Selvin"
> 
> Thanks

Thanks Leon. Will fix it today.

PR #1431, opened 2024-02-20
libibverbs: ignore IBV_FORK_SAFE et.al. when kernel has fork support

krehm:
ibv_fork_init will now check if IBV_FORK_UNNEEDED is true, and if so will skip initialization of mm_root.  Environment variables like IBV_FORK_SAFE and UCX_IB_FORK_INIT are silently ignored when the kernel has the necessary fork support.

### DIFF
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -134,6 +134,9 @@ int ibv_fork_init(void)
 	if (mm_root)
 		return 0;
 
+	if (ibv_is_fork_initialized() == IBV_FORK_UNNEEDED)
+		return 0;
+
 	if (too_late)
 		return EINVAL;
 


jgunthorpe:
@gal-pressman  do you remember why we didn't do this from the start?

gal-pressman:
https://github.com/linux-rdma/rdma-core/pull/975#issuecomment-824665681

krehm:
In my case, it's not a question of performance improvement, it's a way to prevent segfaults when the user specifies an environment variable that enables ibv_fork_init.

jgunthorpe:
Then it is probably best to move forward like this

chien-intel:
Any issues moving forward with this PR?

PR #1436, opened 2024-03-08
bnxt_re/lib: Code refactoring in the resource allocation verbs

selvintxavier:
Implements few helper functions for better code maintainability. Allocates the queue memory and the shadow queue memory together and manage the  pointer for accessing the queue memory and shadow queue memory. Also splits some big functions to smaller sub functions.

### DIFF
--- a/providers/bnxt_re/main.h
+++ b/providers/bnxt_re/main.h
@@ -168,6 +168,7 @@ struct bnxt_re_srq {
 	struct bnxt_re_wrid *srwrid;
 	struct bnxt_re_dpi *udpi;
 	struct xorshift32_state rand;
+	struct bnxt_re_mem *mem;
 	uint32_t srqid;
 	int start_idx;
 	int last_idx;
--- a/providers/bnxt_re/main.h
+++ b/providers/bnxt_re/main.h
@@ -93,9 +93,10 @@ struct bnxt_re_cq {
 	struct ibv_cq ibvcq;
 	uint32_t cqid;
 	struct bnxt_re_context *cntx;
-	struct bnxt_re_queue cqq;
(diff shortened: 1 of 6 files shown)


providers/bnxt_re/main.h
```diff
@@ -614,4 +620,14 @@ static inline void bnxt_re_sub_sec_busy_wait(uint32_t nsec)
 }
 
 #define BNXT_RE_HW_RETX(a) ((a)->comp_mask & BNXT_RE_COMP_MASK_UCNTX_HW_RETX_ENABLED)
+
+static inline uint32_t llog2(uint32_t x)
```
rleon, line 626: we already have this function, "git grep ilog32"

selvintxavier: > we already have this function, "git grep ilog32"

Thanks. I missed searching for log32. Will update the PR.

providers/bnxt_re/main.h
```diff
@@ -554,6 +557,11 @@ static inline void bnxt_re_jqq_mod_last(struct bnxt_re_joint_queue *jqq,
 	jqq->last_idx = jqq->swque[idx].next_idx;
 }
 
+static inline uint32_t bnxt_re_get_diff(uint64_t cmask)
```
rleon, line 560: 1. It is not used
2. There is no value in function which simply returns defind

selvintxavier: Agree that it is not useful with the current patch. I will remove it.
I have a follow up series which avoid this diff for newer adapters, that includes kernel driver changes. We will have to check the compatibility mask for adding the diff. I will re-introduce the function for that series.

PR #1442, opened 2024-03-22
Make it possible to override PYTHON_VERSION when invoking cmake

mchukhrii:
Currently cmake script will override PYTHON_EXECUTABLE if specified in command line.
This patch should fix it. Now if PYTHON_EXECUTABLE is not empty it will be used and python version discovery will not run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,26 +215,38 @@ if (NOT HAVE_WORKING_WERROR)
 endif()
 
 # Use Python modules based on CMake version for backward compatibility
-set(CYTHON_EXECUTABLE "")
 if (${CMAKE_VERSION} VERSION_LESS "3.12")
-  # Look for Python. We prefer some variant of python 3 if the system has it
-  FIND_PACKAGE(PythonInterp 3 QUIET)
-  if (PythonInterp_FOUND)
-    # pyverbs can only use python3:
-    if (NOT NO_PYVERBS)
-      FIND_PACKAGE(cython)
-    endif()
-  else()
-    # But we still must have python (be it 2) for the build process
-    FIND_PACKAGE(PythonInterp REQUIRED)
-  endif()
-else()
-  # FindPython looks preferably for Python3. If not found, version 2 is searched
-  FIND_PACKAGE(Python COMPONENTS Interpreter REQUIRED)
-  set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
-  if (NOT NO_PYVERBS AND Python_VERSION_MAJOR EQUAL 3)
+        FIND_PACKAGE(PythonInterp REQUIRED)
+        FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} EXACT)
+elseif (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12")
+        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+        FIND_PACKAGE(Python 3 REQUIRED COMPONENTS Interpreter OPTIONAL_COMPONENTS Development)
+        set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
+        if(Python_Development_FOUND)
+            set(PYTHONLIBS_FOUND ${Python_Development_FOUND})
+            set(PYTHON_LIBRARIES ${Python_LIBRARIES})
+            set(PYTHON_INCLUDE_DIRS ${Python_INCLUDE_DIRS})
+        endif()
+endif()
+
+set(CYTHON_EXECUTABLE "")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -215,26 +215,38 @@ if (NOT HAVE_WORKING_WERROR)
...
-  # FindPython looks preferably for Python3. If not found, version 2 is searched
-  FIND_PACKAGE(Python COMPONENTS Interpreter REQUIRED)
-  set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
-  if (NOT NO_PYVERBS AND Python_VERSION_MAJOR EQUAL 3)
+        FIND_PACKAGE(PythonInterp REQUIRED)
+        FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} EXACT)
+elseif (${CMAKE_VERSION} VERSION_GREATER "3.12")
+        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+        FIND_PACKAGE(Python 3 REQUIRED COMPONENTS Interpreter OPTIONAL_COMPONENTS Development)
+        set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
```
jgunthorpe, line 224: Both legs if this if should be setting this, right?

CMakeLists.txt
```diff
@@ -215,26 +215,38 @@ if (NOT HAVE_WORKING_WERROR)
...
-else()
-  # FindPython looks preferably for Python3. If not found, version 2 is searched
-  FIND_PACKAGE(Python COMPONENTS Interpreter REQUIRED)
-  set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
-  if (NOT NO_PYVERBS AND Python_VERSION_MAJOR EQUAL 3)
+        FIND_PACKAGE(PythonInterp REQUIRED)
+        FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} EXACT)
+elseif (${CMAKE_VERSION} VERSION_GREATER "3.12")
+        set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+        FIND_PACKAGE(Python 3 REQUIRED COMPONENTS Interpreter OPTIONAL_COMPONENTS Development)
```
jgunthorpe, line 223: What fails cmake if cython is enabled and we don't find Development?

CMakeLists.txt
```diff
@@ -215,26 +215,38 @@ if (NOT HAVE_WORKING_WERROR)
...
-    # But we still must have python (be it 2) for the build process
-    FIND_PACKAGE(PythonInterp REQUIRED)
-  endif()
-else()
-  # FindPython looks preferably for Python3. If not found, version 2 is searched
-  FIND_PACKAGE(Python COMPONENTS Interpreter REQUIRED)
-  set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
-  if (NOT NO_PYVERBS AND Python_VERSION_MAJOR EQUAL 3)
+        FIND_PACKAGE(PythonInterp REQUIRED)
+        FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} EXACT)
```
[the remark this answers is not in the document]
mchukhrii: no it does not. find_package will not fail if PythonLibs is not found.
https://cmake.org/cmake/help/v3.11/command/find_package.html

CMakeLists.txt
mchukhrii: PythonInterp sets PYTHON_EXECUTABLE

CMakeLists.txt
mchukhrii: If find_package python find developer libs Python_Development_FOUND variable is set so i check for that

mchukhrii:
Updated commit. Now should always prefer python3

CMakeLists.txt
```diff
@@ -215,26 +215,38 @@ if (NOT HAVE_WORKING_WERROR)
...
-    FIND_PACKAGE(PythonInterp REQUIRED)
-  endif()
-else()
-  # FindPython looks preferably for Python3. If not found, version 2 is searched
-  FIND_PACKAGE(Python COMPONENTS Interpreter REQUIRED)
-  set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
-  if (NOT NO_PYVERBS AND Python_VERSION_MAJOR EQUAL 3)
+        FIND_PACKAGE(PythonInterp 3 REQUIRED)
+        FIND_PACKAGE(PythonLibs ${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR} EXACT)
+elseif (${CMAKE_VERSION} VERSION_GREATER "3.12")
```
EdwardSro, line 221: You're missing to handle 3.12:
VERSION_GREATER_EQUAL

PR #1443, opened 2024-03-25
include libgen.h for basename

kraj:
basename prototype has been removed from string.h from latest musl [1] compilers e.g. clang-18 flags the absense of prototype as error. therefore include libgen.h for providing it.

[1] https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

### DIFF
--- a/kernel-boot/rdma_rename.c
+++ b/kernel-boot/rdma_rename.c
@@ -2,6 +2,7 @@
 /* Copyright (c) 2019, Mellanox Technologies. All rights reserved. See COPYING file */
 
 #define _GNU_SOURCE
+#include <libgen.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/librdmacm/examples/rping.c
+++ b/librdmacm/examples/rping.c
@@ -33,6 +33,7 @@
 #define _GNU_SOURCE
 #include <endian.h>
 #include <getopt.h>
+#include <libgen.h>
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
--- a/providers/mlx5/mlx5_vfio.c
+++ b/providers/mlx5/mlx5_vfio.c
@@ -6,6 +6,7 @@
 #define _GNU_SOURCE
 #include <config.h>
 
+#include <libgen.h>
 #include <stdio.h>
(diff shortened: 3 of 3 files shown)


jgunthorpe:
Ah, no, this is not right. Read the man page:

```
NOTES
       There  are two different versions of basename() - the POSIX version described above, and the GNU version, which one gets after

               #define _GNU_SOURCE         /* See feature_test_macros(7) */
               #include <string.h>

       The GNU version never modifies its argument, and returns the empty string when path has a trailing slash, and in  particular
       also when it is "/".  There is no GNU version of dirname().

       With glibc, one gets the POSIX version of basename() when <libgen.h> is included, and the GNU version otherwise.
```

rdma-core expects to have the "never modifies its argument" kind. If musl doesn't provide this gnuism then it needs a cmake thing to provide the gnu implementation.

jmaselbas:
musl has basename as defined by POSIX not the gnu variant, both share the same symbol and are not compatible.
You need to redefine your own implementation of basename if you want to be portable and have the same behavior as the glibc one.

What is the "cmake thing" you are suggesting ?

jgunthorpe:
You'd need to use cmake's rdma-core/buildlib/fixup-include/ scheme to put in the glibc version under the right header file. Looking at it that looks a bit hard.

Maybe the better option is to carefully audit three users and change them to use the POSIX version.

jmaselbas:
one solution is to not depends on the gnu implementation of basename (and not use the one from libgen), and re-implement the equivalent of gnu basename, it can be replaced with:
```
const char *gnu_basename(const char *path) {
        const char *p = strrchr(path, '/');
        return p ? p + 1 : path;
}
```

alternatively I took a quick look at fixups and i think this could work, but i've not tested this (yet)
```
#ifndef _FIXUP_STRING_H
#define _FIXUP_STRING_H

#ifdef __USE_GNU
#define basename(p) (strrchr(p,'/') ? strrchr(p,'/') + 1 : (p))
#endif

#include_next <string.h>

#endif
```

kraj:
another try imeplenting local basename - https://github.com/linux-rdma/rdma-core/pull/1467

PR #1447, opened 2024-04-07
unify using ETHERNET_LL_SIZE

changchengx:


libibverbs/verbs.h
```diff
@@ -1736,9 +1737,10 @@ enum ibv_flow_spec_type {
 	IBV_FLOW_SPEC_ACTION_COUNT	= 0x1003,
 };
 
+#define ETHERNET_LL_SIZE ETH_ALEN
```
rleon, line 1740: Does this ETH_ALEN define exist in all OS in which rdma-core is supported?
ETH_ALEN is defined in uapi in 2012.

Thanks

changchengx: >
>Does this ETH_ALEN define exist in all OS in which rdma-core is supported?
>
I think so. If it's not, could be there some known reported problems about below code?
https://github.com/linux-rdma/rdma-core/blob/master/kernel-headers/rdma/mlx5-abi.h#L432
https://github.com/linux-rdma/rdma-core/blob/master/kernel-headers/rdma/mlx5-abi.h#L38

rleon: > > Does this ETH_ALEN define exist in all OS in which rdma-core is supported?
> 
> I think so. If it's not, could be there some known reported problems about below code? https://github.com/linux-rdma/rdma-core/blob/master/kernel-headers/rdma/mlx5-abi.h#L432 https://github.com/linux-rdma/rdma-core/blob/master/kernel-headers/rdma/mlx5-abi.h#L38

Good point, thanks

PR #1449, opened 2024-04-15
mana: add PCI vendor/device match for 3.10

paboldin:
This patch adds PCI vendor-device match for the MANA RDMA provider, which allows to use it on older kernels such as 3.10 which lacks support of the CHARDEV NetLink interface.

### DIFF
--- a/providers/mana/mana.c
+++ b/providers/mana/mana.c
@@ -28,8 +28,13 @@ DECLARE_DRV_CMD(mana_alloc_pd, IB_USER_VERBS_CMD_ALLOC_PD, empty, empty);
 DECLARE_DRV_CMD(mana_create_cq, IB_USER_VERBS_CMD_CREATE_CQ, mana_ib_create_cq,
 		empty);
 
+#define MICROSOFT_MANA_VENDOR_ID	0x1414
+#define MICROSOFT_MANA_DEVICE_ID	0x00ba
+
 static const struct verbs_match_ent hca_table[] = {
 	VERBS_DRIVER_ID(RDMA_DRIVER_MANA),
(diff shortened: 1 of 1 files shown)


rleon:
I would say that we don't really support such old systems.
Plus mana_ib was added after kernel was extended to support netlink chardev.

paboldin:
@rleon, thanks for the review. We are currently backporting MANA drivers to 3.10 for Microsoft.

jgunthorpe:
And you intend to use a 2024 version of rdma-core on those ancient systems???

paboldin:
@jgunthorpe thanks for the question!

MANA provider was introduced in the
```
commit 443f196deee00b7a7fb4305c1d6a3be42db4f2f8
Author: Long Li <longli@microsoft.com>
Date:   Tue Nov 2 21:40:46 2021 +0000

    mana: Microsoft Azure Network Adapter (MANA) RDMA provider
    
    Introduce a provider that exposes MANA devices to user applications.
    
    Signed-off-by: Long Li <longli@microsoft.com>
```

LISA uses rdma-core-46 release (https://github.com/microsoft/lisa/blob/main/microsoft/testsuites/dpdk/dpdktestpmd.py#L564) for tests, which was released:
```
commit f2789cbd394c5839e8942d585a494ab72fd97e39 (tag: v46.0)
Merge: e4e087276 d8eec8724
Author: Leon Romanovsky <leon@kernel.org>
Date:   Wed May 10 09:40:26 2023 +0300
```

So, the short answer is "probably". We would be glad to backport this to the v46 as well.

rleon:
Sorry, but I didn't find support to the claim that such old kernel is really supported and usable.

rleon:
Let's close for now.

Thanks

PR #1452, opened 2024-04-29
Fix mem leak: initialize refcount in try_driver to 0; not 1

cyberang3l:
When initializing `verbs_device->refcount` to 1 we never free `verbs_device->sysfs`.

refcount is increased by `ibv_get_device_list` and `verbs_init_context`, and decreased by `verbs_uninit_context` and `ibv_free_device_list`. With this matching logic, the `atomic_fetch_sub(&verbs_device->refcount, 1) == 1` line in the if statement of the `ibverbs_device_put` function is never true to free `verbs_device-sysfs`.


libibverbs/init.c
```diff
@@ -391,7 +391,7 @@ static struct verbs_device *try_driver(const struct verbs_device_ops *ops,
 
 	vdev->ops = ops;
 
-	atomic_init(&vdev->refcount, 1);
```
yishaih, line 394: Please see the below commit ID [1], which introduced that code.

ibv_device list is refreshed each time that ibv_get_device_list is called, the device and its sysfs entry are kept alive while they exist in the system.

 In case, that the device was unbound, a call from __ibv_get_device_list() to ibverbs_device_hold() will free its memory.

Once the library goes down any existing device/sysfs memory will be freed as part of its memory cleanup.

So, no leak exists here, let's keep it.

[1] commit 8125fdeb69bb6535e78335efab6d46dcca2dffef
Author: Maor Gottlieb <maorg@mellanox.com>
Date:   Tue Jun 27 17:19:49 2017 +0300

    verbs: Avoid ibv_device memory leak

yishaih: Just fixing a typo in the above note.
In case, that the device was unbound, a call from  ibv_get_device_list() to *ibverbs_device_put()* will free its memory.

PR #1453, opened 2024-04-30
libhns: Support flexible WQE buffer page size

hginjgerx:
Currently, driver fixedly allocates 4K pages for userspace WQE buffer
and results in HW reading WQE with a granularity of 4K even in a 64K
system. HW has to switch pages every 4K, leading to a loss of performance.

In order to improve performance, add support for userspace to allocate
flexible WQE buffer page size between 4K to system PAGESIZE. During
ibv_create_qp(), the userspace driver will dynamically select the
appropriate page size based on ibv_qp_cap, thus improving performance
without consuming too much memory.

For old-version userspace driver that does not support this feature,
the kernel driver will use a fixed 4K pagesize.


rleon:
not relevant for now

PR #1455, opened 2024-05-01
providers/mana_ib: create RNIC CQs

TaranovK:
Create RNIC CQs when no external allocator is used. The external allocator is used by DPDK and it expects ethernet CQ. RNIC CQ can not be used by DPDK.

The code is tested with old kernel (no support of RNIC CQs) and new kernels (with support of RNIC CQs, at the moment is sent to rdma-next).

### DIFF
--- a/kernel-headers/rdma/mana-abi.h
+++ b/kernel-headers/rdma/mana-abi.h
@@ -16,8 +16,20 @@
 
 #define MANA_IB_UVERBS_ABI_VERSION 1
 
+enum mana_ib_create_cq_flags {
+	MANA_IB_CREATE_RNIC_CQ	= 1 << 0,
+};
+
 struct mana_ib_create_cq {
 	__aligned_u64 buf_addr;
+	__u16 flags;
+	__u16 reserved0;
+	__u32 reserved1;
+};
+
+struct mana_ib_create_cq_resp {
+	__u32 cqid;
+	__u32 reserved;
 };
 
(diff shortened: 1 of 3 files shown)


providers/mana/mana.c
```diff
@@ -253,23 +271,18 @@ int mana_destroy_cq(struct ibv_cq *ibcq)
...
 			  "Failed to Destroy CQ\n");
 		return ret;
 	}
 
+	if (cq->buf_external)
+		ctx->extern_alloc.free(cq->buf, ctx->extern_alloc.data);
+	else
+		munmap(cq->buf, cq->cqe * COMP_ENTRY_SIZE);
+
 	ctx->extern_alloc.free(cq->buf, ctx->extern_alloc.data);
```
longlimsft, line 286: double free on extern_alloc.free

kernel-headers/rdma/mana-abi.h
```diff
@@ -16,8 +16,20 @@
 
 #define MANA_IB_UVERBS_ABI_VERSION 1
 
+enum mana_ib_create_cq_flags {
```
rleon, line 19: Changes to this file should be created with kernel-headers/update script and be in separate commit.

Thanks

TaranovK: Thanks! I did not know that.

I will do in the next PR. I abandon this PR as I will make one PR with all RC QP related changes, which will include this one.

PR #1457, opened 2024-05-04
Add MSN table capability check

selvintxavier:
Read MSN capability from driver and allocate queue memory accordingly.


providers/bnxt_re/verbs.c
```diff
@@ -2030,7 +2030,7 @@ int bnxt_re_post_send(struct ibv_qp *ibvqp, struct ibv_send_wr *wr,
 		bnxt_re_fill_wrid(wrid, wr->wr_id, bytes,
 				  sig, sq->tail, slots);
 		wrid->wc_opcd = bnxt_re_ibv_wr_to_wc_opcd(wr->opcode);
-		if (BNXT_RE_HW_RETX(qp->cntx))
+		if (BNXT_RE_MSN_TBL_EN(qp->cntx))
```
rleon, line 2033: Does it mean that new rdma-core will "downgrade" all old devices/kernels to behave as they don't have PSN capability?

Because you will take "else" leaf for such flows now.

selvintxavier: No. All old adapters/rdma-core/driver will work with PSN table. New adapters that export this capability will work with MSN table. 

If you see the first patch in my kernel driver posting, which queries this value from FW. 
       #define CREQ_QUERY_FUNC_RESP_SB_REQ_RETRANSMISSION_SUPPORT_HOST_PSN_TABLE  (0x0UL << 4)
       #define CREQ_QUERY_FUNC_RESP_SB_REQ_RETRANSMISSION_SUPPORT_HOST_MSN_TABLE  (0x1UL << 4)

All older adapters report value as 0 and driver/lib default to PSN table. The new adapter only will report 1 and that indicates the usage of MSN table. If the adapter is new and rdma-core is old, we will default to PSN mode.

The current change is because - earlier the MSN table selection was based on HW re-transmit capability. If HW re-transmit is not enabled, we go to PSN table creation mode.   We have some configuration where the HW re-transmit is disabled by FW, but the adapter still need the MSN table. So we are using the new field to identify that. 
If  BNXT_RE_MSN_TBL_EN is null, we follow PSN table flow. Else MSN table flow.

rleon:
@selvintxavier please update kernel header patch.

Thanks

selvintxavier:
> @selvintxavier please update kernel header patch.
> 
> Thanks

Sorry for the delay. i missed this. Updated.

PR #1459, opened 2024-05-06
efa: Add unsolicited RDMA write with immediate receive support

mrgolin:
This PR extends the EFA direct verbs to enable creation of QPs which RQs do not require explicit receive work requests to anticipate incoming RDMA write with immediate. A new work completion query function allows checking whether a work completion is unsolicited.

Related kernel patch: https://lore.kernel.org/all/20240506104859.9225-1-mrgolin@amazon.com/

### DIFF
--- a/kernel-headers/rdma/efa-abi.h
+++ b/kernel-headers/rdma/efa-abi.h
@@ -85,11 +85,17 @@ enum {
 	EFA_QP_DRIVER_TYPE_SRD = 0,
 };
 
+enum {
+	EFA_CREATE_QP_WITH_UNSOLICITED_WRITE_RECV = 1 << 0,
+};
+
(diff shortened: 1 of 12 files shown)


kernel-headers/rdma/efa-abi.h
```diff
@@ -85,11 +85,17 @@ enum {
 	EFA_QP_DRIVER_TYPE_SRD = 0,
 };
 
+enum {
```
rleon, line 88: Please use real SHA and not ?? when creating this patch.
The kernel headers are pushed to wip/leon-for-next

mrgolin: Thanks, updated the commit.

PR #1462, opened 2024-05-17
libhns: One optimization and one bugfix

hginjgerx:
The first patch optimizes cqe inline in srq scenario.
The second one is a bugfix that add checks for input param.


nmorey:
Is there an actual fix in  88d6ac1 ?
if not, please avoid using Fixes in commits that do not fix issues :)

hginjgerx:
> Is there an actual fix in [88d6ac1](https://github.com/linux-rdma/rdma-core/commit/88d6ac1f8884de03413f621f8a973ee3ce13e5aa) ? if not, please avoid using Fixes in commits that do not fix issues :)

Indeed, it's more of an optimization than a fix. Thanks for pointing it out!

PR #1465, opened 2024-06-02
fix error message in mlx5 verbs

DanCh2020:


rleon:
Hi,

in rdma-core, we follow kernel coding style and require Signed-off-by: ... tag.
Please add it together with more elaborated commit message.

Thanks

nmorey:
Please also add a 'Fixes' tag to your commit message mark which commit introduced the bug.
In this case:
`Fixes: 0c7ac1083831 ("Add XRC support")`

DanCh2020:
Hi,
thank you for you suggestions. 
Fixed and pushed

PR #1466, opened 2024-06-03
debian: Correct libhns library name in lintian override

bdrung:
The libhns library is named `libhns1` instead of `libhns-1`.

Fixes: cf6d9149f8f5 ("libhns: Introduce hns direct verbs")


rleon:
Please add Fixes line.

Thanks

bdrung:
Added Fixes line.

PR #1467, opened 2024-06-03
Provide own basename API implementation

kraj:
basename prototype has been removed from string.h from latest musl [1] compilers e.g. clang-18 flags the absense of prototype as error, it turns out the basename has two different implementations and musl only provides POSIX implementation, whereas glibc has GNU version and posix version both depending upon if string.h is included or not it will pick GNU version,

to avoid this dependency, lets implement own basename function which reflects GNU basename behavior, it will work across all libcs

[1] https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -547,11 +547,7 @@ if (NOT DRM_INCLUDE_DIRS)
 endif()
 
 if (DRM_INCLUDE_DIRS)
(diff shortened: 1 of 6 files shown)


jgunthorpe:
Again, why don't we try to convert the callers over to the POSIX behavior? That seems a bit better doesn't it?

kraj:
> Again, why don't we try to convert the callers over to the POSIX behavior? That seems a bit better doesn't it?

that needs testing, and I am not that familiar with rdma-core code to do it.

rleon:
It was fixed in @nmorey PR.

Thanks

PR #1470, opened 2024-06-03
fix potential CQ deadlock in mlx5 provider

zengyijing:
We saw deadlock in mlx5_destroy_qp() if ibv_start_poll() returns EBUSY failure.
According to reference (https://man7.org/linux/man-pages/man3/ibv_create_cq_ex.3.html), if ibv_start_poll() returns error, ibv_end_poll() shouldn't be called.
Therefore, we must release the CQ lock in mlx5_start_poll() if mlx5dv_get_clock_info() returns error e.g. EBUSY.

### DIFF
--- a/providers/mlx5/cq.c
+++ b/providers/mlx5/cq.c
@@ -1163,8 +1163,11 @@ static inline int mlx5_start_poll(struct ibv_cq_ex *ibcq, struct ibv_poll_cq_att
 		goto out;
 	}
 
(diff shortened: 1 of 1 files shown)


nmorey:
Please add a Fixes tag to your commit

zengyijing:
@nmorey I've updated the commit msg with the Fixes tag. Hope it works this time.

PR #1471, opened 2024-06-04
Switch to POSIX implementation of basename

nmorey:
- Drop the useless basename calls in rping
- Include libgen for the POSIX version if kernel-boot and mlx5

Both call basename on non-const string generated by readlink of a sysfs path. The original string never re-used afterward.
So in these cases, both flavours of basename behave exactly the same from the code PoV

This should solve the issue raised in #1467 and #1443


jgunthorpe (approved):

nmorey:
I'm an idiot. Thought it got merged and deleted the branch :/

PR #1475, opened 2024-06-22
 librdmacm: Use overloadable function attribute with clang

kraj:
This is to fix build warnings seen with upcoming clang19 and
 glibc 2.40, since glibc 2.40 has improved fortyfying this
 function with clang, it ends up with build errors like below

librdmacm/preload.c:796:9: error: at most one overload for a given name may lack the 'overloadable' attribute
|   796 | ssize_t recvfrom(int socket, void *buf, size_t len, int flags,
|       |         ^
| /mnt/b/yoe/master/build/tmp/work/core2-64-yoe-linux/rdma-core/51.0/recipe-sysroot/usr/include/sys/socket.h:163:16: note: previous unmarked overload of function is here
|   163 | extern ssize_t recvfrom (int __fd, void *__restrict __buf, size_t __n,
|       |                ^


kraj:
It fails with musl as well, I will have to look into fixing it differently.

PR #1477, opened 2024-06-25
Added suffix libdrm to CMakeLists.txt for drm

hseara:
Canonical `libdrm` package installation results in headers installed at `\usr\include\libdrm.` The current `CMakeLists.txt` installation script searches the headers only in the `drm` suffix, where headers are found in RedHat-based distros.  This is not the canonical location. I propose adding the canonical path `libdrm`, also used in distros like archlinux. Cuurently such distros must patch the main `CMakeLists.txt`  file before installation.

https://gitlab.archlinux.org/archlinux/packaging/packages/rdma-core


rleon:
@nmorey , can you please pick this patch for stable@???

Thanks

nmorey:
Done for branches v34..v52

PR #1478, opened 2024-06-27
debian: exclude hns provider on archs without coherent DMA

bdrung:
The provider `hns` requires coherent DMA and will not build on architectures that lack coherent DMA. So exclude trying to install them on those architectures.

Bug-Debian: https://bugs.debian.org/1073050


nmorey:
Do you need this to be back ported in stable branches?

bdrung:
Not for the official Debian and Ubuntu packages.

PR #1480, opened 2024-07-01
efa: Add EFA 0xefa3 PCI ID

mrgolin:
Add 0xefa3 devices to the pciid match table.


mrgolin:
The related kernel patch is [here](https://lore.kernel.org/all/20240701095752.20246-1-mrgolin@amazon.com/).

PR #1481, opened 2024-07-02
Ensure RDMA service loads modules in initrd

cjubran:
Fixed an issue where the RDMA service was killed during switch to root. Added a wants symlink for initrd.target in the dracut arrangement and Before=initrd.target to the systemd service to ensure it runs and completes during initrd.

### DIFF
--- a/kernel-boot/rdma-load-modules@.service.in
+++ b/kernel-boot/rdma-load-modules@.service.in
@@ -5,6 +5,9 @@ Documentation=file:@CMAKE_INSTALL_FULL_DOCDIR@/udev.md
 # systemd-modules-load.service
 DefaultDependencies=no
 Before=sysinit.target
+# Kernel modules must load in initrd before initrd.target to avoid being killed
+# when initrd-cleanup.service isolates to initrd-switch-root.target.
+Before=initrd.target
 # Do not execute concurrently with an ongoing shutdown
 Conflicts=shutdown.target
 Before=shutdown.target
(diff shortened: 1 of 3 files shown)


nmorey:
Can i ask why are you running this service in initrd?
Shouldn't any service/target/FS that needs RDMA be the one to have a Before=initrd.target ?

cjubran:
> Can i ask why are you running this service in initrd? Shouldn't any service/target/FS that needs RDMA be the one to have a Before=initrd.target ?

We run this service in initrd to load RDMA modules early in the boot process. It prevents issues like failed NFS mounts over RDMA due to missing modules.

jgunthorpe:
I talked to Chuck and he told me that NFS module autoloading works fine, even checked it out for me.

So why is it failing in the initrd?

cjubran:
> I talked to Chuck and he told me that NFS module autoloading works fine, even checked it out for me.
> 
> So why is it failing in the initrd?

The rdma service is terminated when initrd-cleanup.service isolates for initrd-switch-root.target because initrd.target lacks the dependency on the rdma service.

nmorey:
Does this actually work out of the box? I fill like SUSE/RH spec file (at least) are missing a call to dracut somewhere?

rleon:
> Does this actually work out of the box? I fill like SUSE/RH spec file (at least) are missing a call to dracut somewhere?

I don't think so, we saw this failure in our regressions where we run dracut anyway.
Carolina asked me if we need to add call to dracut in spec file too, but I was afraid that people will be unhappy if installation of rdma-core will require reboot after this spec file change.

PR #1484, opened 2024-07-19
Fix for cm_svc_run

linxiaochou:
When the cm_svc_run thread starts, the four element size rss will be allocated to svc->contexts. When the rss reallocates memory due to an increase in the number of elements in the cm_svc_process_sock function, the rss will point to a new address, but the cm_svc_run function does not get the new address again, leading to an exception in the future

### DIFF
--- a/librdmacm/rsocket.c
+++ b/librdmacm/rsocket.c
@@ -4678,8 +4678,11 @@ static void *cm_svc_run(void *arg)
 			fds[i].revents = 0;
 
 		poll(fds, svc->cnt + 1, -1);
-		if (fds[0].revents)
+		if (fds[0].revents) {
 			cm_svc_process_sock(svc);
+			/* svc->contexts may have been reallocated, so need to assign again */
+			fds = svc->contexts;
+		}
 
 		for (i = 1; i <= svc->cnt; i++) {
 			if (!fds[i].revents)


librdmacm/rsocket.c
```diff
@@ -4678,8 +4678,11 @@ static void *cm_svc_run(void *arg)
 			fds[i].revents = 0;
 
 		poll(fds, svc->cnt + 1, -1);
-		if (fds[0].revents)
+		if (fds[0].revents) {
 			cm_svc_process_sock(svc);
+			/* svc->contexts will be changed, so need to assign again */
+			fds = svc->contexts;
```
rleon, line 4684: Please write an explanation to the change in commit message, together with exception.

We are following kernel coding style, so Fixes line should be added too.
Fixes: b60c79daaacc ("rsockets: Use service thread to accept connections")

@shefty

linxiaochou:
@rleon I'm glad you responded and pointed out my mistake. I've updated the commit msg. Hope it works this time.

rleon:
> @rleon I'm glad you responded and pointed out my problem. I've updated the commit msg. Hope it works this time.

Sure, let's wait for @shefty approval.

librdmacm/rsocket.c
shefty: nit: I would say 'svc->contexts may have been reallocated' rather than 'will be changed'  Fix looks good.

linxiaochou:
@shefty Thank you for your reply and suggestions. I have already revised the corresponding notes. I hope it can work this time.

PR #1490, opened 2024-08-12
verbs: Use provider name when generating provider constructor

jakemoroni:
Previously, the token pasting operator was not being applied to any of the PROVIDER_DRIVER macro aguments, so every provider's constructor had the same name of "drv__register_driver".

Now, these constructors will be named using the provider name as the prefix, resulting in names like "irdma_register_driver".


rleon:
Please add this Fixes line to commit message.

Fixes: 50b2aebc2114 ("verbs: Expose the struct verbs_device_ops for each provider to the linker")

The change LGTM, thanks.

jakemoroni:
Done, thanks.

PR #1492, opened 2024-08-28
bnxt_re/lib: SRQ toggle support for GenP7 adapters

selvintxavier:
Implements the UAPI call to get the SRQ toggle page.
 Also, get the toggle bit from the shared page and
 use it while arming the SRQ.

### DIFF
--- a/kernel-headers/rdma/bnxt_re-abi.h
+++ b/kernel-headers/rdma/bnxt_re-abi.h
@@ -134,8 +134,14 @@ struct bnxt_re_srq_req {
 	__aligned_u64 srq_handle;
 };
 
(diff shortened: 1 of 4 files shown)


rleon:
@selvintxavier please don't put fake SHA-1 in first commit and leave it as ??. It is misleading to see some SHA and remember to double check it every time when I'm merging the PR.

selvintxavier:
> @selvintxavier please don't put fake SHA-1 in first commit and leave it as ??. It is misleading to see some SHA and remember to double check it every time when I'm merging the PR.

Oops. I am sorry. I might have missed using the --not-final option. Will post a revision now

PR #1505, opened 2024-10-16
Add QP service level in EFA DV

mrgolin:
Add SL parameter to EFA create QP direct verb and pass it to the kernel.
Update Pyverbs and documentation accordingly.

The related kernel patches are [here](https://lore.kernel.org/all/20241015174242.3490-1-mrgolin@amazon.com/).

### DIFF
--- a/kernel-headers/rdma/efa-abi.h
+++ b/kernel-headers/rdma/efa-abi.h
@@ -95,7 +95,8 @@ struct efa_ibv_create_qp {
 	__u32 sq_ring_size; /* bytes */
 	__u32 driver_qp_type;
 	__u16 flags;
-	__u8 reserved_90[6];
+	__u8 sl;
+	__u8 reserved_98[5];
 };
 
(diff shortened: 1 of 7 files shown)


jgunthorpe:
Is it safe to rely on efa_attr->sl being zero? It seems like it would be easy for the caller to have not zero'd the reserved field. There aren't many callers out there, did you check libfabric?

mrgolin:
> Is it safe to rely on efa_attr->sl being zero? It seems like it would be easy for the caller to have not zero'd the reserved field. There aren't many callers out there, did you check libfabric?

Yes. The checks [here](https://github.com/linux-rdma/rdma-core/blob/master/providers/efa/verbs.c#L1644) ensure that anyone calling ```efadv_create_qp_ex``` with non zero extra bytes in ```efa_attr``` would have been failing.

mrgolin:
Updated kernel headers patch with real commit SHA.

PR #1509, opened 2024-10-30
tests: Skip CQ shrinking for bnxt_re devices.

mohammadheib:
bnxt_re driver implement the CQ shrinking operation in a way that allows users to shrink CQ with a size smaller than the existing CQE and still have access to old CQEs please see:

 - https://github.com/linux-rdma/rdma-core/pull/1315

The test_resize_cq assumes that when shrinking an active CQ with unpolled entries it fail with an EINVAL error, which is not true for bnxt_re devices.

To avoid this failure for bnxt we skip part of the test that expect EINVAL.

### DIFF
--- a/providers/bnxt_re/verbs.c
+++ b/providers/bnxt_re/verbs.c
@@ -328,8 +328,10 @@ struct ibv_cq *bnxt_re_create_cq(struct ibv_context *ibvctx, int ncqe,
 	struct bnxt_re_context *cntx = to_bnxt_re_context(ibvctx);
 	struct bnxt_re_dev *dev = to_bnxt_re_dev(ibvctx->device);
 
-	if (ncqe > dev->max_cq_depth)
+	if (ncqe > dev->max_cq_depth) {
+		errno = EINVAL;
 		return NULL;
+	}
 
 	cq = calloc(1, (sizeof(*cq) + sizeof(struct bnxt_re_queue)));
 	if (!cq)
--- a/tests/test_cq.py
+++ b/tests/test_cq.py
@@ -112,20 +112,24 @@ class CQTest(RDMATestCase):
 
         # Increase the CQ size.
         new_cq_size = 7
+        post_send_num = new_cq_size - 1
         self.client.cq.resize(new_cq_size)
         self.assertTrue(self.client.cq.cqe >= new_cq_size,
                         f'The actual CQ size ({self.client.cq.cqe}) is less '
                         'than guaranteed ({new_cq_size})')
 
-        irdma.skip_if_irdma_dev(d.Context(name=self.dev_name))
         # Fill the CQ entries except one for avoid cq_overrun warnings.
         send_wr, _ = u.get_send_elements(self.client, False)
         ah_client = u.get_global_ah(self.client, self.gid_index, self.ib_port)
-        for i in range(self.client.cq.cqe - 1):
+        for i in range(post_send_num):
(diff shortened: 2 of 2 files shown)


jgunthorpe:
I'm not keen on a list of devices in the test suite, can we just detect success and skip the test? Or better, validate the CQE are all preserved?

mohammadheib:
Hi @jgunthorpe ,
Thank you so much for the quick review :).
Agree with you that the device list is not the best solution, especially for newly added devices in the future. 
But i had some issues trying to fix this test for bnxt_re if you see below i had two failers with the test :
first failed was in the following loop:
`        
for i in range(self.client.cq.cqe - 1): 
u.send(self.client, send_wr, ah=ah_client)
`
This loop assumes that the CQE in the cq after the resize to 7 is 7 CQE and sends cq->cqe SR, which is a wrong assumption because the CQEs can be more than 7 and that lead us to posting more SR than the qp->max_send_wr.
 I managed to fix it by running the loop only 7 times.   

The second issue which i had to add the device list for is these three lines:
`        # Decrease the CQ size to less than the CQ unpolled entries.
        new_cq_size = 1
        with self.assertRaises(PyverbsRDMAError) as ex:
            self.client.cq.resize(new_cq_size)
        self.assertEqual(ex.exception.error_code, errno.EINVAL)
`

these lines return different values from one provider to another so i can't expect it to return EINVAL for bnxt_re and can't expect it to return errno = 0 for mlx4

So, I decided to skip it, but I really don't like the device list idea. If you have a better way to solve it, I would love to update the PR.

one idea in my mind was to use only the vendor ID to skip it but maybe future devices will have a different way of handling resize in BNXT.

jgunthorpe:
@EdwardSro  ?

EdwardSro:
@mohammadheib Regarding the your first point I agree, the post send loop should do 7 SR and not the actual self.client.cq.cqe times.
Regarding the expected errno on the last decrease: Honestly, I'm starting to suspect the correctness and accuracy of this test.
Although the man page states that "cqe must be at least the number of unpolled entries in the CQ" and "More outstanding entries on CQ than capacity specified" is mentioned as a possible verbs result by the IB Spec, neither one of them states explicitly that in this case the verb would fail. BNXT is the second driver (after iRDMA) that succeeds in such a case.

What is common in this case and all resize cases is that completions must not be lost. So, my suggestion is to adjust the test in this try-except manner:

```python
# Increase the CQ size.
new_cq_size = 7
post_send_num = new_cq_size - 1
self.client.cq.resize(new_cq_size)
self.assertTrue(self.client.cq.cqe >= new_cq_size,
                f'The actual CQ size ({self.client.cq.cqe}) is less '
                'than guaranteed ({new_cq_size})')

# Fill the CQ entries except one for avoid cq_overrun warnings.
send_wr, _ = u.get_send_elements(self.client, False)
ah_client = u.get_global_ah(self.client, self.gid_index, self.ib_port)
for i in range(post_send_num):
    u.send(self.client, send_wr, ah=ah_client)

# Decrease the CQ size to less than the CQ unpolled entries.
new_cq_size = 1
try:
    self.client.cq.resize(new_cq_size)
    self.assertEqual(ex.error_code, errno.EINVAL)
finally:
    for i in range(post_send_num):
        u.poll_cq(self.client.cq)
```
This should cover all drivers.

EdwardSro (approved):

PR #1511, opened 2024-11-05
util/udma_barrier.h: adjust ARMv8 memory barriers

dharmik-arm:
Change the barrier APIs for IO to reflect that Armv8-a is other-multi-copy atomicity memory model.

Armv8-a memory model has been strengthened to require other-multi-copy atomicity. This property requires memory accesses from an observer to become visible to all other observers simultaneously [3]. This means

a) A write arriving at an endpoint shared between multiple CPUs is
   visible to all CPUs
b) A write that is visible to all CPUs is also visible to all other
   observers in the shareability domain

This allows for using cheaper DMB instructions in the place of DSB for devices that are visible to all CPUs

Please refer to [1], [2] and [3] for more information.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=22ec71615d824f4f11d38d0e55a88d8956b7e45f
[2] https://www.youtube.com/watch?v=i6DayghhA8Q
[3] https://www.cl.cam.ac.uk/~pes20/armv8-mca/

### DIFF
--- a/util/udma_barrier.h
+++ b/util/udma_barrier.h
@@ -93,7 +93,7 @@
 #elif defined(__sparc_v9__)
 #define udma_to_device_barrier() asm volatile("membar #StoreStore" ::: "memory")
 #elif defined(__aarch64__)
-#define udma_to_device_barrier() asm volatile("dsb st" ::: "memory");
+#define udma_to_device_barrier() asm volatile("dmb oshst" ::: "memory")
 #elif defined(__sparc__) || defined(__s390x__)
 #define udma_to_device_barrier() asm volatile("" ::: "memory")
 #elif defined(__loongarch__)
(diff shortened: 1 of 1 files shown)


util/udma_barrier.h
```diff
@@ -201,7 +201,7 @@
 #elif defined(__sparc_v9__)
 #define mmio_flush_writes() asm volatile("membar #StoreStore" ::: "memory")
 #elif defined(__aarch64__)
-#define mmio_flush_writes() asm volatile("dsb st" ::: "memory");
+#define mmio_flush_writes() asm volatile("dmb oshst" ::: "memory");
```
hnagarahalli, line 204: We should use DGH here. But, suggest to make this in another commit as it would need a different commit log.

jgunthorpe: I think you should take the mmio_flush_writes change out of this commit and revisit it in a followup. The justification you gave for this change doesn't apply here. DGH alone may also not be correct.

This code needs to ensure that a write combining is entirely contained within a spinlock region. No write combine can be merged across different spinlock critical sections. AFAIK this requirement has nothing to do with other-multi-copy-atomicity.

dharmik-arm: Thank you for the review comments! I have updated the PR to remove the mmio_flush_writes change, will revisit in a follow up.

hnagarahalli (changes requested):
@dharmik-arm Changes look good. CI is throwing errors, can you please resolve them?

dharmik-arm:
Thank you, @hnagarahalli!
CI errors may not be applicable, the macro definition is consistent with other macros.
@jgunthorpe Any suggestion on how can I go about the CI failures?

jgunthorpe:
Just ignore checkpatch errors like that, they don't make sense

PR #1524, opened 2024-11-19
buildlib: set a proper name when creating a github release

nmorey:
By default, AZP creates a release on Github names vXX.YY which does not match the format currently used: rdma-core-XX.YY This changes the release name to match the expected format.

Fixes: 936cf8e17cf3 ("build/azp: Have Azure Pipelines create releases when tags are made")

### DIFF
--- a/buildlib/azure-pipelines-release.yml
+++ b/buildlib/azure-pipelines-release.yml
@@ -28,6 +28,10 @@ stages:
 
           - bash: |
               set -e
+              git_tag=$(git describe --exact-match HEAD)
+              rel_ver=$(echo $git_tag | sed -e 's/^v//')
+              echo "Version is $rel_ver"
+              echo "##vso[task.setvariable variable=rel_ver]$rel_ver"
               mkdir build-pandoc artifacts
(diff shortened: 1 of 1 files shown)


nmorey:
Yes I've renamed all the hundreds of stable releases manually in GH. IIRC we had proper GH names when using Travis but I could never get it to work with AZP for some reason.
The fixes is to be sure that this get picked on the stable branches but I can remove the Fixes tag if you don't like it

rleon:
Maybe these names didn't work for stable, but for main release, I always wrote full rdma-core-XX.YY name and never saw incorrect name.

I don't mind regarding Fixes, just trying to understand if I should take it now or wait after release (in next two weeks).

nmorey:
OK my bad, the commit message is slightly off.
Right now, there is no release title set by AZP (and not vXX.YY).
This commit simply sets it to the proper value automatically. I'll reword that

nmorey:
@relon It should be clearer now.

rleon:
Thanks, will try it in upcoming release.

PR #1526, opened 2024-11-27
mlx5: invalidate cq->cur_rsc when QP is destroyed inside a polling batch

FujiZ:
When using cq_ex inteface, if the user destroys the QP associated with the current work completion, the next ibv_next_poll() call will cause a use-after-free error since it needs to access the QP that has already been destroyed through cq->cur_rsc inside get_req_context().

Fix this error by resetting the cq->cur_rsc in __mlx5_cq_clean if it is associated with the QP to be destroyed.

### DIFF
--- a/providers/mlx5/cq.c
+++ b/providers/mlx5/cq.c
@@ -1823,6 +1823,16 @@ void __mlx5_cq_clean(struct mlx5_cq *cq, uint32_t rsn, struct mlx5_srq *srq)
 		return;
 
 	/*
+	 * For CQ created in single threaded mode serving multiple
+	 * QPs, if the user destroys a QP between ibv_start_poll()
+	 * and ibv_end_poll(), then cq->cur_rsc should be invalidated
+	 * since it may point to the QP that is being destroyed, which
+	 * may cause UAF error in the next ibv_next_poll() call.
+	 */
+	if (unlikely(cq->cur_rsc && rsn == cq->cur_rsc->rsn))
+		cq->cur_rsc = NULL;
+
+	/*
 	 * First we need to find the current producer index, so we
 	 * know where to start cleaning from.  It doesn't matter if HW
 	 * adds new entries after this loop -- the QP we're worried


providers/mlx5/cq.c
```diff
@@ -1070,6 +1070,8 @@ static inline void _mlx5_end_poll(struct ibv_cq_ex *ibcq,
 	if (lock)
 		mlx5_spin_unlock(&cq->lock);
 
+	cq->cur_rsc = NULL;
```
yishaih, line 1073: We already have that code (i.e. cq->cur_rsc = NULL)  as part of mlx5_start_poll(), so it seems redundant.

Note:
In any case, it had to be done before line 1070 to be done under the lock, if the lock was applicable.

providers/mlx5/cq.c
```diff
@@ -1822,6 +1824,14 @@ void __mlx5_cq_clean(struct mlx5_cq *cq, uint32_t rsn, struct mlx5_srq *srq)
 	if (!cq || cq->flags & MLX5_CQ_FLAGS_DV_OWNED)
 		return;
 
+	/*
+	 * Reset the cq->cur_rsc if it is associated with the QP to be
+	 * destroyed in order to prevent use-after-free errors in the
+	 * next ibv_next_poll().
+	 */
```
yishaih, line 1831: What is the scenario that you are referring to here ? a CQ which servers more than a single QP ?

In addition, if the 'lock' mode was used, this code will run only after mlx5_end_poll(), so the next mlx5_start_poll() should do the work by setting the pointer to NULL upon its entrance.

Note:
The code should not protect against incorrect application behavior (e.g., destroying a QP while still polling for its completions), especially in areas that might impact the data path.

yishaih (changes requested):

providers/mlx5/cq.c
FujiZ: > What is the scenario that you are referring to here ? a CQ which servers more than a single QP ?

Yeah in this case multiple QPs are attached to the same CQ.
The code sequence to reproduce this problem would look like this:
```C
ibv_start_poll(); // get a work completion associated with the QP A
ibv_destroy_qp(); // destroy QP A here since we get a completion error
ibv_next_poll(); // try to get the next work completion from other QPs. UAF error is triggered here
```

> destroying a QP while still polling for its completions

Do you mean that destroying a QP between `ibv_start_poll()` and `ibv_end_poll()` is not permitted? However, I haven't find any manual which describes this behaviour.

yishaih: It's permitted but doesn't look like a good practice, a few notes below.

You are referring only to a single-threaded application, correct? In a multi-threaded application, the call to ibv_destroy_qp() will remain blocked until ibv_end_poll() is invoked, ensuring that ibv_next_poll() is safe to use.

Additionally, we are only discussing a scenario where the CQ serves multiple QPs. Otherwise, it would not make sense to destroy a QP and continue polling its CQ, as this would clearly indicate incorrect application behavior.

The commit log should be rephrased to clarify the exact use case that we are talking about.

So, in the specific scenario that you are talking about, it can make sense to set the NULL upon __mlx5_cq_clean() while narrowing the comment near this line to be more specific as was mentioned above.

In any case, no need  for a change in mlx5_end_poll() as I already mentioned.

providers/mlx5/cq.c
FujiZ: Got it. I have just removed this line in the latest patch.

providers/mlx5/cq.c
FujiZ: > You are referring only to a single-threaded application, correct? In a multi-threaded application

Yes, the CQ is configured to be in single threaded mode by specifying `IBV_CREATE_CQ_ATTR_SINGLE_THREADED`.
I have updated the comment to reveal the use case that we have discussed so far.

yishaih (approved):

providers/mlx5/cq.c
YonatanNachum: @FujiZ can you please explain why this is thread safe ? using `IBV_CREATE_CQ_ATTR_SINGLE_THREADED` states the CQ is polled only by one thread but it doesn't state that CQ poll and destroy QP are synchronized. If one thread is polling the CQ and another thread is destroying the QP there is still a chance the CQ polling thread will see a stale value without locks/barriers.

@yishaih what is the expected behavior with  `IBV_CREATE_CQ_ATTR_SINGLE_THREADED` given that also mlx uses CQ lock during QP destroy.

yishaih: IBV_CREATE_CQ_ATTR_SINGLE_THREADED basically means that only a single thread will poll the CQ on a given time, so no lock is needed as part of the polling flow.

However,
It doesn't mean that only a single thread is used in the library for other purposes, in addition a single CQ can serve multiple QPs, as of that we have the CQ locks for cleanup, upon QP destroy.

YonatanNachum: @yishaih thanks for the answer,
> It doesn't mean that only a single thread is used in the library for other purposes

So there is a chance you will poll a single threaded CQ while a its QP is being destroyed. Isn't there a race between the polling functions and `__mlx5_cq_clean`, they touch the same fields in the CQ. Or do you have an assumption on such a case ?

yishaih: We don't plan/need to protect against any un-expected application flow that might hit itself .. 

Application should be careful before taking such an approach while using IBV_CREATE_CQ_ATTR_SINGLE_THREADED.

Need to get into the details here to see what might be the outcome.

jgunthorpe: No verbs objects can be concurrently destroyed while being used by another thread. This is completely illegal as it can trivially result in using an already destroyed object.

YonatanNachum: Thanks for the answers.
This is really important for certain flows.

PR #1535, opened 2024-12-31
mlx5: fix max length of hostname

leoncamel:
The current length of `mlx5_context.hostname` is a 40 bytes long. According to `gethostname`'s man-page, and also the standard of POSIX. The maximum length is `HOST_NAME_MAX`, which is 64 bytes.

And also, the [RFC1034](https://datatracker.ietf.org/doc/html/rfc1034#section-3.1), the maximum length is also about 64 bytes.

So, I made this patch to follow the standards.

### DIFF
--- a/providers/mlx5/mlx5.h
+++ b/providers/mlx5/mlx5.h
@@ -39,6 +39,7 @@
 #include <stdarg.h>
 #include <stdatomic.h>
 #include <util/compiler.h>
+#include <limits.h>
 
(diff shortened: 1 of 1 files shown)


rleon:
We use kernel coding style.

Please add description to commit message and set Fixes line.

Thanks

leoncamel:
@rleon Done. Please review it.

rleon:
I still don't see Fixes line, but ok, it is not important here.

PR #1538, opened 2025-01-08
modify rcopy codes with supporting client to use specific bond ip

zjulcq:
modify rcopy codes with supporting client to use specific bond ip


hxy0307:
请问一下，这个rcopy使用的specific bond ip应该是什么呀

PR #1542, opened 2025-01-15
mlx5: specify unsupport config mkey write relax ordering

changchengx:
The API mlx5dv_wr_set_mkey_access_flags access_flags parameter doesn't support IBV_ACCESS_RELAXED_ORDERING.
https://github.com/linux-rdma/rdma-core/blob/v55.0/providers/mlx5/qp.c#L2951-L2958
```C
	if (unlikely(!check_comp_mask(access_flags,
				      IBV_ACCESS_LOCAL_WRITE |
				      IBV_ACCESS_REMOTE_WRITE |
				      IBV_ACCESS_REMOTE_READ |
				      IBV_ACCESS_REMOTE_ATOMIC))) {
		mqp->err = EINVAL;
		return;
	}
```

### DIFF
--- a/providers/mlx5/man/mlx5dv_wr_mkey_configure.3.md
+++ b/providers/mlx5/man/mlx5dv_wr_mkey_configure.3.md
@@ -161,7 +161,7 @@ struct mlx5dv_mkey_conf_attr {
 
 	:	The desired memory protection attributes; it is either 0 or
 		the bitwise OR of one or more of flags in **enum
-		ibv_access_flags**.
+		ibv_access_flags**. IBV_ACCESS_RELAXED_ORDERING is not supported.
 
 ## Data layout setters
 


providers/mlx5/man/mlx5dv_wr_mkey_configure.3.md
```diff
@@ -161,7 +161,7 @@ struct mlx5dv_mkey_conf_attr {
 
 	:	The desired memory protection attributes; it is either 0 or
 		the bitwise OR of one or more of flags in **enum
-		ibv_access_flags**.
+		ibv_access_flags**. IBV_ACCESS_RELAXED_ORDERING is not supported.
```
yishaih, line 164: This is not the only access flag which is not supported by that DV API (e.g. IBV_ACCESS_ON_DEMAND, IBV_ACCESS_MW_BIND, IBV_ACCESS_ZERO_BASED, etc.).

I would skip/drop this patch, the man page doesn't guarantee that all bits are supported.

yishaih (changes requested):

providers/mlx5/man/mlx5dv_wr_mkey_configure.3.md
changchengx: @yishaih I agree with you. I've closed this PR.
One question: Is there any hardware limitation that ```mlx5dv_wr_mkey_configure``` doesn't support ```IBV_ACCESS_RELAXED_ORDERING``` in ```access_flags``` parameter?

PR #1549, opened 2025-01-21
Changes from Debian upload 55.0 1 plus fixes

bdrung:
* debian: Bump Standards-Version to 4.7.0
* debian: add IBMAD_1.4@IBMAD_1.4 symbol (fixes 57b50e2064f4)
* debian: run wrap-and-sort
* debian: Add Debian uploads up to version 55.0-1
* libibumad: Correct symbols for new API to support SMI/GSI seperation (fixes be54b52e94be)
* libibmad: Correct symbols for new API to support SMI/GSI seperation (fixes fb11e5a40402)
* debian: Switch to automatically created -dbgsym packages

See individual commits for details. Switch to automatically created -dbgsym packages should be safe, because since the stable releases support it.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,7 +2,33 @@ rdma-core (56.0-1) unstable; urgency=medium
 
   * New upstream release.
 
- -- Jason Gunthorpe <jgg@obsidianresearch.com>  Tue, 27 Feb 2024 11:32:22 +0100
+ -- Jason Gunthorpe <jgg@obsidianresearch.com>  Tue, 21 Jan 2025 13:44:36 +0100
+
+rdma-core (55.0-1) unstable; urgency=medium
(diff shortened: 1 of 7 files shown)


nmorey:
@bdrung Can you confirm if you need d122e4996222a79144f40e6fb2e387a813ef1a5f to be backported to all stable branches that contains 57b50e2 ("libibmad: Expose the XDR speed decoding functions")

It applies without any issue but I don't know enough about DEB packaging to make sure if I should do it or not :)

bdrung:
@nmorey Yes, https://github.com/linux-rdma/rdma-core/commit/d122e4996222a79144f40e6fb2e387a813ef1a5f should be backported to the stable releases v54 and v55.

PR #1552, opened 2025-01-28
rping: make pointers translation more robust and descriptive

SiarheiVolkau:
Since size of the `unsigned long` isn't clear for reader and might vary depending on compiler and target platform it has to be replaced by more descriptive `uintptr_t`, also it expresses the intention to bypass pointer "as is".

### DIFF
--- a/librdmacm/examples/rping.c
+++ b/librdmacm/examples/rping.c
@@ -444,13 +444,13 @@ static int rping_disconnect(struct rping_cb *cb, struct rdma_cm_id *id)
 
 static void rping_setup_wr(struct rping_cb *cb)
 {
-	cb->recv_sgl.addr = (uint64_t) (unsigned long) &cb->recv_buf;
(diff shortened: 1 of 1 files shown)


rleon:
Thanks for the contribution, but unfortunately there is nothing to fix here.
Unsigned long type is a common and valid way to represent pointers.

rleon:
Do you run rdma-core in such systems?

SiarheiVolkau:
Me not, but implicitly assuming `sizeof long >= sizeof ptr` is just wrong in a context of portable code.

rleon:
There is no need to fix something that is not broken. The rdma-core library is perfectly portable on any platforms which run Linux/FreeBSD OS flavors.

SiarheiVolkau:
The patch is not about "fix something broken" it's all about semantics.

When I saw that piece of code `info->buf = htobe64((uint64_t) (unsigned long) buf);` the first question was WTF is going on here? Why is it unsigned long? For explicit truncation? What special meaning does it have here?

Nevermind, I understood.

PR #1554, opened 2025-01-29
libibumad: Change find_preffered_port

mazorasaf:
Updated the find_preferred_port function to ensure the same port number is set for SMI/GSI
devices when they belong to the same device.
This prevents selecting different ports for the same device.


rleon:
We user kernel coding style for commits.

Please use Subject in title, add description and Fixes line.
Right now, the patch looks:

Subject: [PATCH] libibumad: Change find_preffered_port to set same portnum for
 smi/gsi device in case that its the same device (otherwise we might choose
 different port for same device)

Signed-off-by: Asaf Mazor <amazor@nvidia.com>
---

rleon:
Same comments as in PR https://github.com/linux-rdma/rdma-core/pull/1555

PR #1555, opened 2025-01-29
infiniband-diags: use correct port info to get cap_mask for ibnd_get_agg_linkspeedextsup

mazorasaf:
Added get_port_info_for_cap_mask to retrieve the correct port information for determining the cap_mask -
in the case of a Switch we should be using port 0, In other cases it should be the current port.
This function is now utilized with the ibnd_get_agg_linkspeedext methods in ibdiag_common to ensure accurate handling.

### DIFF
--- a/infiniband-diags/ibdiag_common.c
+++ b/infiniband-diags/ibdiag_common.c
@@ -777,6 +777,19 @@ static unsigned int get_max(unsigned int num)
 	return (1 << r);
 }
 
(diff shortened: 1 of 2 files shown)


rleon:
Same comment as in previous PR.
In addition, parent and issue are not relevant here.

rleon:
1. You have extra blank line between Fixes and SOB lines.
2. Fixes should have 12 chars and not 7
3. Please double check your commit message, it lacks punctuation and sentences are not complete.

PR #1557, opened 2025-02-12
providers/mana: Add dma-buf support

TaranovK:
Add support for dma-buf MR verb API in mana.


TaranovK:
Added the "Signed-off-by: Konstantin Taranov" to this patch.

PR #1558, opened 2025-02-12
redhat: Add missing rdma-load-modules systemd scriptlets

Kamalheib:
Add the missing systemd restart bits for the rdma-load-modules.

### DIFF
--- a/redhat/rdma-core.spec
+++ b/redhat/rdma-core.spec
@@ -363,6 +363,18 @@ if [ -x /sbin/udevadm ]; then
 /sbin/udevadm trigger --subsystem-match=net --action=change || true
 /sbin/udevadm trigger --subsystem-match=infiniband_mad --action=change || true
 fi
+%systemd_post rdma-load-modules@rdma.service
+%systemd_post rdma-load-modules@infiniband.service
+%systemd_post rdma-load-modules@roce.service
+
+%preun -n rdma-core
+%systemd_preun rdma-load-modules@rdma.service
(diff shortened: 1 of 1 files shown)


rleon:
@nmorey do you need anything like this in SuSE?

nmorey:
@rleon Need? No. It does work fine like that unless I'm missing something.
As the rdma-load-modules service isn't really a usual system service but just a trickery to load kmod, it seems to be working all right without the usual heavy systemd/RPm macro.
But we can probably have something similar (we do use %service_ macros and not %systemd directly). Not sure if it's worth the hassle. I'll check with my systemd colleagues

Thanks for the heads up

nmorey:
@rleon After checking with my colleague, we won't make a similar patch for SUSE.
It now works as it is because rdma-load-module isn't a usual service but just trickery to load modules.

The drawback of the current SUSE approach is that if a rdma-core update were to change the .conf files, the new modules would not get loaded during the update because rdma-hw.target is already loaded and the udevadm trigger won't do anything. But these files have not changed since 2017 so this seems quite acceptable.

The drawback of the RHEL patch is that every single rdma-core update will load RDMA/Infiniband/RoCE ULP modules into the kernel whether they are required or not.

rleon:
> The drawback of the RHEL patch is that every single rdma-core update will load RDMA/Infiniband/RoCE ULP modules into the kernel whether they are required or not.

Right, I saw it, but my assumption it that this patch comes from RH engineer for RH specific file, so they need it like this.

PR #1562, opened 2025-02-16
srp_daemon: remove check after mutex init

ariel-anieli:
Hello maintainers,

Here is my first ever patch to the project; I would like to understand the internals of RDMA.

The patch addresses a warning found by static analysis. It is about [`pthread_mutex_init`](https://man7.org/linux/man-pages/man3/pthread_mutex_lock.3.html):

> pthread_mutex_init always returns 0. The other mutex functions return 0 on success and a non-zero error code on error.

Your feedback is much appreciated

### DIFF
--- a/srp_daemon/srp_sync.c
+++ b/srp_daemon/srp_sync.c
@@ -105,29 +105,16 @@ bool sync_resources_error(struct sync_resources *res)
 
 int sync_resources_init(struct sync_resources *res)
 {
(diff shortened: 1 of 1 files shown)


rleon:
Thank you for your first contribution.
Unfortunately, this code doesn't fix anything. The pthread_mutex_init() returns value, so we can check it anyway.

PR #1563, opened 2025-02-18
efa: Fix receive SGE length overflow

dkkranz:
EFA supports 16 bit length for receive buffers, but will receive only up to MTU size messages, so the rest of the receive buffer is unused. Posting a 64KB buffer will end up wrapping around to zero. Instead, Truncate it to the maximum size, which is higher than EFA MTU.


dkkranz:
Fixed, thanks

rleon:
> Fixed, thanks

Thanks, and now please remove blank line between new Fixes line and ROB tags.

dkkranz:
Fixed again. Thanks

PR #1566, opened 2025-02-21
rsetsockopt: fix bit manipulation

SiarheiVolkau:
I'm not sure if the upper bits are actually used now, but anyway, it might happen in future.
`rsetsockopt` was unable to set/clear bits properly which position is >= 32.
Moreover, any clear bit operation clears also upper bits (32-63).

Found using static analysis tool PVS-Studio.


rleon:
It is not happening now. We will change it when upper bits will be used.

PR #1572, opened 2025-03-03
preload: fix build issue with clang 19

clan:
librdmacm/preload.c:796:9: error: at most one overload for a given name may lack the 'overloadable' attribute

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -395,11 +395,14 @@ if (NOT HAVE_LONG_LONG_U64)
 endif()
 
 # Check __SOCKADDR_ARG exist and is union
+set(SAFE_CMAKE_REQUIRED_INCLUDES "${CMAKE_REQUIRED_INCLUDES}")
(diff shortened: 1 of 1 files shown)


jgunthorpe:
That does not seem correct since it broke everything, I don't think attribute overload is the right answer here

jgunthorpe:
That seems like a more believable change, but it probably needs a cmake fragment to detect what the libc is using for it's function signature.

jgunthorpe:
I don't know where this is coming from, but the usual way to handle it is to have cmake detect what is correct and then have the C code match it.

jgunthorpe:
I suppose the sparse failure means the cmake test is not quite right

jgunthorpe:
I can't remember what the sparse issue is that required that change, but shouldn't cmake should still be able to detect the difference? Hmm. maybe the order of the header patching prevents that from working.
I suggest adjusting your CMake logic to also check HAVE_SPARSE and make the right choice.

jgunthorpe:
that makes sense and looks right

PR #1573, opened 2025-03-04
libibverbs: Fix cast-align warning with verbs_get_ctx().

yishaih:
Fix cast-align warning with verbs_get_ctx().

Casting using uint8_t * pointer arithmetic may generate a cast-align warning, when using "-Wcast-align=strict" compilation flag.

Fix it by using uintptr_t instead of uint8_t *.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,9 @@ include(RDMA_DoFixup)
 include(publish_headers)
 include(rdma_functions)
 include(pyverbs_functions)
+
+check_c_compiler_flag("-Wcast-align=strict" HAVE_WCAST_ALIGN_STRICT)
+
 if (NO_MAN_PAGES)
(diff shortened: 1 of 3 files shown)


rleon:
I don't understand this sentence "... may generate ...". Does it generate an error or not?
If yes, commit message should be clear about it. If not, so what are you fixing here?
Does all rdma-core compile now with -Wcast-align=strict? If yes, please add this check to global CmakeList.txt, if not
let's fix all other places too.

Thanks

yishaih:
Yes, this is a fix for the public header, allowing outer applications to compile cleanly with the -Wcast-align=strict option.

To ensure verbs.h remains compliant with -Wcast-align=strict in future builds, this flag was added as part of libibverbs/examples/CMakeLists.txt.

The commit log was updated accordingly.

PR #1574, opened 2025-03-04
Some fixes found while playing with clang scan-build

nmorey:


jgunthorpe:
The fixes lines should all be in the trailer block, no blank line before the signed-off-vy

PR #1579, opened 2025-03-13
libhns: Cleanup and Bugfixes

hginjgerx:
This PR contains some recent cleanup and bugfixes for libhns.

### DIFF
--- a/providers/hns/hns_roce_u_verbs.c
+++ b/providers/hns/hns_roce_u_verbs.c
@@ -417,7 +417,7 @@ static int verify_cq_create_attr(struct ibv_cq_init_attr_ex *attr,
 {
 	struct hns_roce_pad *pad = to_hr_pad(attr->parent_domain);
 
-	if (!attr->cqe || attr->cqe > context->max_cqe) {
+	if (!attr->cqe || attr->cqe > (uint32_t)context->max_cqe) {
 		verbs_err(&context->ibv_ctx, "unsupported cq depth %u.\n",
 			  attr->cqe);
 		return EINVAL;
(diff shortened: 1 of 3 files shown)


providers/hns/hns_roce_u_verbs.c
```diff
@@ -1307,9 +1317,7 @@ static void set_ext_sge_param(struct hns_roce_context *ctx,
 							attr->cap.max_send_sge);
 
 	if (ctx->config & HNS_ROCE_RSP_EXSGE_FLAGS) {
-		attr->cap.max_inline_data = min_t(uint32_t, roundup_pow_of_two(
```
rleon, line 1310: Just use min3() instead.

hginjgerx: > Just use min3() instead.

Sorry I don't get it. Only two values ​​are compared here

rleon: min3(attr->cap.max_inline_data, roundup_pow_of_two(attr->cap.max_inline_data), ctx->max_inline_data)

hginjgerx: > min3(attr->cap.max_inline_data, roundup_pow_of_two(attr->cap.max_inline_data), ctx->max_inline_data)

This is wrong when attr->cap.max_inline_data is not power of 2

rleon: try it

hginjgerx: I tried, but the result was really not what we expected, since roundup_pow_of_two(attr->cap.max_inline_data) will never be smaller than attr->cap.max_inline_data. What we want is the roundup_pow_of_two() value.
```
attr->cap.max_inline_data = 3
roundup_pow_of_two(attr->cap.max_inline_data) = 4
ctx->max_inline_data = 1024
min3 result = 3
```

PR #1586, opened 2025-03-20
Revert "mlx4: Fix overrun-buffer-arg"

Kamalheib:
This reverts commit 4d92ff74aa3d4a4b0181ea0db103377c7034dcb4.

A regression was found when testing resize_cq using pyverbs.

Fixes: 4d92ff74aa3d ("mlx4: Fix overrun-buffer-arg")


jgunthorpe:
Yeah,  how many other places have been changed like that? Tthe lower level functions are supposed to accept the full size of the struct including the driver data portion

Kamalheib:
> Yeah, how many other places have been changed like that? Tthe lower level functions are supposed to accept the full size of the struct including the driver data portion

The following PR should remove all of them:

https://github.com/linux-rdma/rdma-core/pull/1588

PR #1587, opened 2025-03-28
libhns: Add support for LTTng tracing

hginjgerx:
Add support for LTTng tracing. For now it is used for post_send, post_recv and poll_cq.

### DIFF
--- a/providers/hns/CMakeLists.txt
+++ b/providers/hns/CMakeLists.txt
@@ -1,5 +1,10 @@
+if (ENABLE_LTTNG AND LTTNGUST_FOUND)
+  set(TRACE_FILE hns_roce_u_trace.c)
+endif()
+
 rdma_shared_provider(hns libhns.map
   1 1.0.${PACKAGE_VERSION}
+  ${TRACE_FILE}
   hns_roce_u.c
   hns_roce_u_buf.c
   hns_roce_u_db.c
@@ -12,3 +17,8 @@ publish_headers(infiniband
 )
 
 rdma_pkg_config("hns" "libibverbs" "${CMAKE_THREAD_LIBS_INIT}")
+
+if (ENABLE_LTTNG AND LTTNGUST_FOUND)
+  target_include_directories(hns PUBLIC ".")
+  target_link_libraries(hns LINK_PRIVATE LTTng::UST)
+endif()
--- a/providers/hns/hns_roce_u_hw_v2.c
+++ b/providers/hns/hns_roce_u_hw_v2.c
@@ -38,6 +38,7 @@
 #include "hns_roce_u.h"
 #include "hns_roce_u_db.h"
 #include "hns_roce_u_hw_v2.h"
+#include "hns_roce_u_trace.h"
 
 #define HR_IBV_OPC_MAP(ib_key, hr_key) \
 		[IBV_WR_ ## ib_key] = HNS_ROCE_WQE_OP_ ## hr_key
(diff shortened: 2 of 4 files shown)


providers/hns/hns_roce_u_trace.c
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: GPL-2.0 OR BSD-2-Clause
+/*
+ * Copyright (c) 2025 Hisilicon Limited.
+ *
+ * This software is available to you under a choice of one of two
```
rleon, line 5: Please don't add license text to new files. SPDX tag should be enough

providers/hns/hns_roce_u_trace.h
```diff
@@ -0,0 +1,63 @@
...
+#undef LTTNG_UST_TRACEPOINT_INCLUDE
+#define LTTNG_UST_TRACEPOINT_INCLUDE "hns_roce_u_trace.h"
+
+#if !defined(__HNS_TRACE_H__) || defined(LTTNG_UST_TRACEPOINT_HEADER_MULTI_READ)
+#define __HNS_TRACE_H__
+
+#include <lttng/tracepoint.h>
+#include <infiniband/verbs.h>
+
+#define rdma_tracepoint(arg...) lttng_ust_tracepoint(arg)
```
rleon, line 146: Just asking, I see same line in many other drivers. Shouldn't it be placed in some common place?

providers/hns/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+if (ENABLE_LTTNG AND LTTNGUST_FOUND)
+  rdma_shared_provider(hns_trace libhns_trace.map
```
rleon, line 2: Regular libhns won't have dependency unless you build with -DENABLE_LTTNG, which is by default "false".

There is no need to create extra library just for tracing.

providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -683,6 +684,65 @@ static int parse_cqe_for_cq(struct hns_roce_context *ctx, struct hns_roce_cq *cq
 	return 0;
 }
 
+static UNNEEDED uint8_t read_wc_sl(struct hns_roce_qp *hr_qp,
```
rleon, line 687: I'm aware that UNNEEDED keyword exist in our compiler.h file, but I still prefer to rely on LTTNG_ENABLED here.

providers/hns/hns_roce_u_trace.c
hginjgerx: Thanks, fixed.

providers/hns/hns_roce_u_hw_v2.c
hginjgerx: Okay, fixed.

providers/hns/hns_roce_u_trace.h
hginjgerx: Yeah, but not many drivers are supporting tracing for now though. We can do this if you want.

providers/hns/CMakeLists.txt
hginjgerx: Separating libhns_trace have some benefits:
1. libhns can still be built with -DENABLE_LTTNG even if lttng is not installed.
2. Applications directly relying on libhns (e.g., perftest) won't inherit the lttng dependency.
3. Consistently separating trace libraries across all providers may allow enabling lttng in the release version of rdma-core, as regular libraries like libhns or libmlx5 won't introduce extra dependencies or performance impact. Users not needing traces can install rdma-core and run applications as usual, while trace users can simply install lttng and preload lib*_trace without rebuilding rdma-core

providers/hns/hns_roce_u_trace.h
rleon: Let's do it as a followup patches, after we merge it.

hginjgerx:
Comments above are fixed @rleon

PR #1592, opened 2025-04-14
mthca: Fix malloc() return value without detecting null pointer

YanLongDai:
mthca: Fix malloc() return value without detecting null pointer


YanLongDai:
image

@rleon Checkpatch failed due to upstream CI error. Can the reviewer ignore this error

PR #1606, opened 2025-05-04
verbs: Allow query only device support for QP data in order

dkkranz:
EFA can support 128 bytes blocks write in-order for some Grace platforms. Move the check on x86 architecture to the mlx5 provider implementation.

### DIFF
--- a/libibverbs/verbs.c
+++ b/libibverbs/verbs.c
@@ -695,23 +695,31 @@ LATEST_SYMVER_FUNC(ibv_query_qp, 1_1, "IBVERBS_1.1",
 int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
 			       uint32_t flags)
 {
+	int result;
+
+	if (!check_comp_mask(flags,
+			     IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS |
+			     IBV_QUERY_QP_DATA_IN_ORDER_DEVICE_ONLY))
+		return 0;
+
 #if !defined(__i386__) && !defined(__x86_64__)
 	/* Currently this API is only supported for x86 architectures since most
 	 * non-x86 platforms are known to be OOO and need to do a per-platform study.
+	 * However, it is possible to override this restriction to allow querying
+	 * the device capability directly, regardless of the host CPU architecture.
 	 */
-	return 0;
-#else
-	int result;
-
-	if (!check_comp_mask(flags, IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS))
+	if (!(flags & IBV_QUERY_QP_DATA_IN_ORDER_DEVICE_ONLY))
 		return 0;
+#endif
 
 	result = get_ops(qp->context)->query_qp_data_in_order(qp, op, flags);
 	if (result & IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG)
 		result |= IBV_QUERY_QP_DATA_IN_ORDER_ALIGNED_128_BYTES;
 
-	return flags ? result : !!(result & IBV_QUERY_QP_DATA_IN_ORDER_WHOLE_MSG);
-#endif
+	if (flags & IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS)
+		return result;
+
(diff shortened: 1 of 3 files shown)


dkkranz:
Applied your suggestion, thanks

dkkranz:
@jgunthorpe Kind reminder

jgunthorpe:
I don't know.. I just feel uneasy about this. ARM is such a wide and varied architecture I don't know if we can really say it is actually reliably in order. Grace might be just because it has 128 byte lines, but that doesn't mean a different ARM chip will be. This could break some of the MPIs running on ARM supercomputers.. Do you think you could limit it just to Grace? Did someone from NVIDIA confirm that this is actually true on Grace?

jgunthorpe:
I didn't mean a flag to userspace, there is nothing userspace can do with this. I ment some kind of flag from EFA to the core code to disable the arch checks in the core code just for EFA.

mrgolin:
>I don't really want to have drivers doing architecture checks..

But this is exactly what you are suggesting now, and I'm not sure why we need to have architecture checks in libibverbs either.

I think there is use for a common way to query whether data is written in order from device perspective only, for instance data polling from GPU isn't necessarily affected by the CPU architecture. Don't you agree?

jgunthorpe:
Maybe you could reorganize things so the driver calls out some kind of core helper to test if the CPU/platform is OK for some properties and then just not call that on EFA.

jgunthorpe:
The ability to do last data polling is largely a property of the receiving unit. Yes the verb is intended to cover CPU memory only.

If you want to expose a property it should be if the RDMA device issues PCIe TLPs strictly in order and strictly with non-relaxed ordering. ie it is a PCIe behavior property only. Then you could cross that with some GPU property if it can do last data polling with such TLPs. That could be a general purpose verb.

mrgolin:
>If you want to expose a property it should be if the RDMA device issues PCIe TLPs strictly in order and strictly with non-relaxed ordering.

Yes, this is something RDMA device can tell and it's what I would like to expose as a general purpose verb (or an extension of existing one).

>Then you could cross that with some GPU property if it can do last data polling with such TLPs.

I think we better leave this part to upper layers that can be aware of GPU behavior or system topology.

jgunthorpe:
If this is what you want to do I think a new general verb flag would be fine - report the PCI device behavior only, and explain in the manual this is only half the problem and any software using this must also consider the behavior of the receiving device.

libibverbs/verbs.c
```diff
@@ -695,23 +695,26 @@ LATEST_SYMVER_FUNC(ibv_query_qp, 1_1, "IBVERBS_1.1",
 int ibv_query_qp_data_in_order(struct ibv_qp *qp, enum ibv_wr_opcode op,
 			       uint32_t flags)
 {
+	int result;
+
+	if (!check_comp_mask(flags, IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS))
```
jgunthorpe: This function sure has become weird, flags is less of a flags and more of a selector now..

Still, flags has to be validated.. It can only be 0, IBV_QUERY_QP_DATA_IN_ORDER_RETURN_CAPS or  IBV_QUERY_QP_DATA_IN_ORDER_DEVICE_ONLY, everything else is a failure and returns 0.

dkkranz:
I fixed the flags check, thanks.
There is no change required in the EFA implementation. The API was only blocked in the common verb check on the host architecture.

jgunthorpe:
Something is wrong here, you can't just add new API with new driver semantics without changing the existing drivers, somehow. What is the story?

mrgolin:
Providers' implementations of ```query_qp_data_in_order``` op are expected to return device level only support (network, device internals and PCIe TLPs ordering), what is exactly what ```ibv_query_qp_data_in_order``` with the new ```IBV_QUERY_QP_DATA_IN_ORDER_DEVICE_ONLY``` flag promises. Am I missing something?

rleon:
> The use of LL128 on GB200 and any other ARM + GPU platform with upstream rdma-core is currently blocked by inability to query for its support. This PR is in the queue for almost 3 month now, @jgunthorpe @rleon I think we now have the the solution we agreed on that extends the verb without changing its existing behavior, if there are no other significant concerns can we please merge this?

Jason was more involved in review than me, he needs to merge this PR.

PR #1612, opened 2025-05-29
efa: Fix work request index double use

mrgolin:
When polling a CQE related to a destroyed QP, we do not update the current work queue reference, what results in "returning" the work request index of the CQE to a wrong pool. This can later result in two different inflight work requests using same index and mistakenly completing with same work request id.

Fix by zeroing the work queue pointer on error and unify conditions with unsolicited completion flow.


mrgolin:
Thanks @nmorey , I think v52 is old enough.

PR #1613, opened 2025-06-05
libhns: Remove MW support

hginjgerx:
MW is no longer supported in hns. Delete relevant codes.


rleon:
We are in merge window now.
This PR and kernel patch will need to wait after v6.15-rc1 is released.

hginjgerx:
@rleon  Hi Leon, can this be merged now?

PR #1616, opened 2025-06-19
efa: Fix CQ poll after QP destroy

mrgolin:
After a QP has been destroyed there might still be pending CQEs. As an optimization for some cases, poll functions ignore CQs that are not in use by any QP, preventing completions from being read after destruction of the last QP that uses this CQ. Remove this optimization.


mrgolin:
@jgunthorpe @rleon Can we please merge this small change?

jgunthorpe:
Unbreaking the CI was a Microsoft adventure of "It does not work" errors when clicking a button until it finally worked after restarting web browsers to make the button happy again. Sigh.

PR #1621, opened 2025-07-02
[RFC] Support dynamically loading LTTng tracing

hginjgerx:
This PR is an RFC to support dynamically loading LTTng tracing. 

As mentioned in our brief discussion previously in https://github.com/linux-rdma/rdma-core/pull/1587, there are some benefits to supporting this:
1. Regular driver libraries can still be built with -DENABLE_LTTNG even if LTTng is not installed.
2. Applications directly relying on driver libraries (e.g., perftest) won't inherit the LTTng dependency.
3. No additional dependencies or performance penalty will be introduced to regular libraries. Users not needing tracing can install rdma-core and run applications as usual, while tracing users can simply install LTTng and preload the tracing libraries without rebuilding rdma-core

We believe that this can greatly improve the usability of LTTng tracing.

BTW, the first patch is included incidentally and isn’t actually part of this feature. It’s meant to fix the static compilation failures when enabling LTTng.


rleon:
And I still think that providing extra, special tracing libraries as part of rdma-core is wrong approach.

PR #1623, opened 2025-07-08
verbs: RDMA support for DMA handle

yishaih:
This patch series introduces a new DMA Handle (DMAH) object, along with its corresponding verbs for allocation and deallocation. (i.e. ibv_alloc_dmah(), ibv_dealloc_dmah())

The DMAH object encapsulates attributes relevant for DMA transactions.

While initially intended to support TLP Processing Hints (TPH) [1], the design is extensible to accommodate future features such as PCI multipath for DMA, PCI UIO configurations, traffic class selection, and more.

This addition enables applications to allocate an opaque DMA handle that can be used to optimize DMA and RDMA operations associated with a registered memory region.

Additionally, we introduce a new verb named ibv_reg_mr_ex() with its corresponding ioctl (i.e. UVERBS_METHOD_REG_MR) on the MR object.

This verb consolidates multiple reg_mr variants under a single user-space ioctl interface, supporting: ibv_reg_mr(), ibv_reg_mr_iova(), ibv_reg_mr_iova2() and ibv_reg_dmabuf_mr().

It also enables passing a DMA handle as part of the registration process to associate it with the registered MR. 

The series includes the implementation of the new functionality in the mlx5 driver.

Detailed man pages exist to describe the expected usage.

The matching kernel series was sent to rdma-next.

[1] Background, from PCIe specification 6.2.
TLP Processing Hints (TPH)
------------------------------
TLP Processing Hints is an optional feature that provides hints in Request TLP headers to facilitate optimized processing of Requests that target Memory Space.  These Processing Hints enable the system hardware (e.g., the Root Complex and/ or Endpoints) to optimize platform resources such as system and memory interconnect on a per TLP basis.

### DIFF
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -631,6 +631,10 @@ struct ibv_mw_bind_info {
 	unsigned int	 mw_access_flags; /* use ibv_access_flags */
 };
 
+struct ibv_dmah {
+	struct ibv_context *context;
+};
+
 struct ibv_pd {
 	struct ibv_context     *context;
 	uint32_t		handle;
@@ -2105,6 +2109,24 @@ struct ibv_parent_domain_init_attr {
 	void *pd_context;
 };
 
+enum ibv_tph_mem_type {
+	IBV_TPH_MEM_TYPE_VM, /* volatile memory */
+	IBV_TPH_MEM_TYPE_PM, /* persistent memory */
+};
+
+enum  ibv_dmah_init_attr_mask {
+	IBV_DMAH_INIT_ATTR_MASK_CPU_ID = 1 << 0,
+	IBV_DMAH_INIT_ATTR_MASK_PH = 1 << 1,
+	IBV_DMAH_INIT_ATTR_MASK_TPH_MEM_TYPE = 1 << 2,
+};
+
+struct ibv_dmah_init_attr {
+	uint32_t comp_mask;
+	uint32_t cpu_id;
+	uint8_t ph;
+	uint8_t tph_mem_type; /* From enum ibv_tph_mem_type */
+};
+
 struct ibv_counters_init_attr {
 	uint32_t	comp_mask;
 };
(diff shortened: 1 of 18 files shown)


libibverbs/verbs.h
```diff
@@ -681,6 +681,25 @@ struct ibv_mr {
...
 
+enum  ibv_reg_mr_in_mask {
+	IBV_REG_MR_MASK_IOVA = 1 << 0,
+	IBV_REG_MR_MASK_ADDR = 1 << 1,
+	IBV_REG_MR_MASK_FD = 1 << 2,
+	IBV_REG_MR_MASK_FD_OFFSET = 1 << 3,
+	IBV_REG_MR_MASK_DMAH = 1 << 4,
+};
+
+struct ibv_reg_mr_in {
```
shefty, line 692: Are we still able to change the name of this structure?  Has there been a new release of libibverbs?

Adding UET support to verbs requires extending this structure, and I would also like to use it to return output data, to avoid needing a reg_mr_ex2 call or other changes.

rleon: I still didn't release yet. We are releasing rdma-core according to Linux kernel schedule on -rc1 and sometimes on -rc2 (if we are asked to delay for some reason).

So yes, we can change it and/or revert this change prior release.

yishaih: It has a comp_mask field for future extension.

What is the motivation to rename it ? any other suggested name ?

Note:
Even if upon success it will hold some output fields in the future, (what is the use case ?) it still can be considered as the input struct for the API which holds both the in/out data.

rleon: Structure that provides in and out fields can't be called struct XXX_in {}.

Thanks

yishaih: It wasn't intended for output.

Let's see the use case and then re-consider whether we really need a rename.

PR #1625, opened 2025-07-10
librdmacm/cmtime: Support multiple clients and/or servers

vladum:
Add support to run and synchronize multiple cmtime instances. All clients connect to all servers.

Key flags which can be used when running cmtime in this distributed topology:
  * `-L` (new): decides which instances are servers
  * `-c`: controls the number of connections created per client/server pair
  * `-P <n>` (new): informs the synchronization logic of the total number of instances that are going to be started
  * `-C <IP>` (new): determines which instance acts as synchronization root (i.e., decides when to advance the benchmark phase)
  * `-B <interface>` (new): binds to named interface instead of IP (`-b`) to conveniently share the flag value across multiple hosts

The same set of flags can be used for all instances, except for `-L` which is only used for servers.

Example with 2 clients connecting to 3 servers:

```
client1$ cmtime -B enp1s0f0np0 -P 5 -C 10.10.0.1 -c 1000
client2$ cmtime -B enp1s0f0np0 -P 5 -C 10.10.0.1 -c 1000

server1$ cmtime -B enp1s0f0np0 -P 5 -C 10.10.0.1 -c 1000 -L
server2$ cmtime -B enp1s0f0np0 -P 5 -C 10.10.0.1 -c 1000 -L
server3$ cmtime -B enp1s0f0np0 -P 5 -C 10.10.0.1 -c 1000 -L
```

Startup order does not matter. The first instance to start on the host which has the `10.10.0.1` IP assigned will act as synchronization root. The total number of connections: 6000 (2 clients * 3 servers * 1000 connections per pair).

### DIFF
--- a/librdmacm/examples/cmtime.c
+++ b/librdmacm/examples/cmtime.c
@@ -52,6 +52,7 @@
 #include <arpa/inet.h>
 #include "common.h"
 
+#define SRV_MAX_DISCONNECT_TIME_US 60000000
 
 static struct rdma_addrinfo hints, *rai;
 static struct addrinfo *ai;
@@ -131,6 +132,7 @@ struct conn {
 
 	uint64_t times[STEP_CNT][2];
 	int retries;
+	bool server_disconnected;
 };
 
 static struct work_queue wq;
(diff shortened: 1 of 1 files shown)


shefty (approved):
This is a marvelous series of enhancements!  Thanks, Vlad!

PR #1626, opened 2025-07-24
efa: Parent domains and thread domains support

YonatanNachum:
In order to optimize our CQ polling functions and QP posting functions,
implement the alloc/dealloc thread domain verbs and alloc parent domain
verb so the ULP will have better control on the synchronization mechanisms
EFA provider uses in our data path functions.
We will use the parent domain to bind QPs and CQs on the same thread domain
so we can remove the WQ lock in data path.

### DIFF
--- a/providers/efa/efa.h
+++ b/providers/efa/efa.h
@@ -99,6 +99,7 @@ struct efa_cq {
 	struct efa_wq *cur_wq;
 	struct efa_io_cdesc_common *cur_cqe;
 	struct ibv_device *dev;
+	struct efa_parent_domain *parent_domain;
 	struct efa_sub_cq sub_cq_arr[];
 };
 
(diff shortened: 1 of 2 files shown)


providers/efa/efa.h
```diff
@@ -51,12 +51,24 @@ struct efa_context {
...
 	uint16_t pdn;
+	/* Pointer to original PD used to create parent domain */
+	struct efa_pd *orig_pd;
+	/* Number of parent domains referencing this PD */
+	atomic_int refcount;
 };
 
 struct efa_td {
 	struct ibv_td ibvtd;
-	/* Number of parent domains referncing this TD */
```
rleon: This line should be in previous patch

YonatanNachum: Hey, what to you mean ? the line is in `efa: Add support to allocate thread domains` which is first in the chain.

rleon: > Hey, what to you mean ? the line is in `efa: Add support to allocate thread domains` which is first in the chain.

In the second patch, you are fixing line which was introduced in first one.
Just write it correctly in first place.

YonatanNachum: Fixed, thanks.

PR #1627, opened 2025-07-29
libibverbs: Fix create CQ max number of attributes

YonatanNachum:
With CQ creation with external memory support the max number of attributes was increased but was left unchanged in the cmd declaration.

Fixes: 8d0ce825e0cc ("libibverbs: Add option to pass CQ buffers to kernel")


rleon:
Please remove empty line between Fixes and SOB.

Thanks

PR #1628, opened 2025-07-30
librdmacm: Allow using SA to resolve IB addresses

shefty:
The following series adds support for using SA ServiceRecords to resolve a service name to a usable IB address.  This is in addition to the current approach which uses DNS.  To add this support, a couple enhancements are added to the librdmacm.

* Address resolution can now be done asynchronously.  A new function, rdma_resolve_addrinfo(), is added to the API.  The behavior is similar to rdma_getaddrinfo(), only the operation is asynchronous.  Once the resolution completes, an event is written to the event channel.  An rdma_cm_id is used to track the operation while it is outstanding.  The new call supports both SA name resolution and DNS.  In the case of DNS support, asynchronous behavior is emulated.
* The application can use flags passed into rdma_resolve_addrinfo() to control whether DNS or SA or both are used for resolution.
* Once asynchronous resolution completes, a new rdma_query_addrinfo() call allows the application to retrieve the results.  The flow of calling rdma_resolve_addrinfo() + reading the event + rdma_query_addrinfo() provides similar behavior as calling rdma_getaddrinfo().
* In order to convert synchronous DNS resolution into an asynchronous operation, a third API call is added, rdma_write_cm_event().  This call allows user space to inject an event into the event channel queue.  This allows rdma_resolve_addrinfo() to provide consistent application behavior, independent of how resolution is implemented under the API.  Because such an API may be useful for applications, it is exposed for direct use.

I'm submitting this PR on behalf of Mark Zhang, who is the author.


debian/librdmacm1.symbols
```diff
@@ -4,6 +4,7 @@ librdmacm.so.1 librdmacm1 #MINVER#
  RDMACM_1.1@RDMACM_1.1 16
  RDMACM_1.2@RDMACM_1.2 23
  RDMACM_1.3@RDMACM_1.3 31
+ RDMACM_1.4@RDMACM_1.4 57
```
rleon: 57 needs to be 60

debian/librdmacm1.symbols
```diff
@@ -52,6 +52,7 @@ librdmacm.so.1 librdmacm1 #MINVER#
  rdma_resolve_route@RDMACM_1.0 1.0.15
  rdma_set_local_ece@RDMACM_1.3 31
  rdma_set_option@RDMACM_1.0 1.0.15
+ rdma_write_cm_event@RDMACM_1.4 57
```
rleon: Same as before

PR #1641, opened 2025-09-04
infiniband-diags: Fix sa_get_handle to use smi/gsi API

mazorasaf:
- Use mad_rpc_open_port_2 inside sa_get_handle
 - Update handle members according to gsi port from mad_rpc_open_port2
 - Replace umad_close_port in mad_rpc_close_port_2 inside sa_free_handle

This fixes behavior where we can take smi handle by mistake. mad_rpc_open_port2 does inside umad_open_port and umad_register, so we just need to assign the relevant values to the handle.

Fixes: aaf9f699c10f ("ibqueryerrors: Use API supporting SMI/GSI seperation")

### DIFF
--- a/infiniband-diags/ibdiag_sa.c
+++ b/infiniband-diags/ibdiag_sa.c
@@ -48,6 +48,8 @@
  * the saquery tool and provides it to other utilities.
  */
 
+static struct ibmad_ports_pair *srcports;
+
 struct sa_handle *sa_get_handle(char *ca_name)
 {
 	struct sa_handle *handle;
@@ -57,7 +59,13 @@ struct sa_handle *sa_get_handle(char *ca_name)
 
 	char *name = ca_name ? ca_name : ibd_ca;
 
-	resolve_sm_portid(name, ibd_ca_port, &handle->dport);
+	int mgmt_classes[1] = { IB_SA_CLASS };
+
+	srcports =  mad_rpc_open_port2(name, ibd_ca_port, mgmt_classes, 1, 0);
+	if (!srcports)
+		IBEXIT("Failed to open '%s' port '%d'", name, ibd_ca_port);
+
+	resolve_sm_portid(srcports->gsi.ca_name, ibd_ca_port, &handle->dport);
 	if (!handle->dport.lid) {
 		IBWARN("No SM/SA found on port %s:%d",
 			name ? "" : name,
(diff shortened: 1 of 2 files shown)


infiniband-diags/ibdiag_sa.c
```diff
@@ -39,6 +39,7 @@
 
 #include <errno.h>
 #include <infiniband/umad.h>
+#include "../libibmad/mad_internal.h"
```
rleon: Something wrong if you need to provide such path. You should handle it as part of CMakeList.txt.

mad_internal.h even in the name says that it shouldn't be used outside of libibmad

infiniband-diags/ibdiag_sa.c
```diff
@@ -38,8 +38,8 @@
 
 
 #include <errno.h>
+#include <infiniband/mad.h>
 #include <infiniband/umad.h>
-#include "../libibmad/mad_internal.h"
```
rleon: You added this line in previous patch.

infiniband-diags/ibdiag_sa.c
```diff
@@ -78,8 +78,8 @@ struct sa_handle *sa_get_handle(char *ca_name)
 	if (!handle->dport.qkey)
 		handle->dport.qkey = IB_DEFAULT_QP1_QKEY;
 
-	handle->fd = srcports->gsi.port->port_id;
-	handle->agent = srcports->gsi.port->class_agents[IB_SA_CLASS];
+	handle->fd = mad_rpc_portid(srcports->gsi.port);
```
rleon, line 80: You added these lines in previous patch.

infiniband-diags/ibdiag_sa.c
mazorasaf: yes, I removed it. It was added by my IDE and missed that

infiniband-diags/ibdiag_sa.c
mazorasaf: ?

infiniband-diags/ibdiag_sa.c
rleon: > yes, I removed it. It was added by my IDE and missed that

So probably you need to squash your two patches into one.

infiniband-diags/ibdiag_sa.c
mazorasaf: oh i guess the change is not related to the commit.
ok, will fix

mazorasaf:
I messed up the whole PR..

infiniband-diags/ibdiag_sa.c
mazorasaf: ok fixed it, now one commit for each file.
Do you still want me to squash?
Or is it ok like that?

rleon: it is ok, let's wait for CI to finish.

PR #1642, opened 2025-09-04
librdmacm/cmtime: Update man page and drop unused option

vladum:
PR #1625 significantly changed `cmtime` usage.

Update man page to match and remove command line argument which is no longer used.


rleon:
Please add context to commit message for first patch.

Thanks

PR #1643, opened 2025-09-09
dracut: unify and improve dracut rdma module

jozzsi:
The intention of this PR is eliminate the need to have a separate module in the redhat and suse directories.

Additional improvements beyond the original goal
 - moves the dracut module ordering to 50, which is the recommended ordering for out of tree dracut modules

 - fixes dracut dependency and shellcheck issues


jgunthorpe:
This seems reasonable, the rpm builds are failing looks like a typo.

Can we put the dracut related files under kernel-boot/ please?

jgunthorpe:
Please rebase it, that will fix the CI failure. Also please enhance the commit message to explain whatever is going on in that bug report link. It seems this is more than just "recommened". Let's document what version of dracut has broken rdma-core.

The changes look fine otherwise.

PR #1644, opened 2025-09-09
kernel-boot: Add rdma_topo tool

jgunthorpe:
For some time now modern multi-NIC servers now have very complex topology. Often with NICs, GPUs and NVMe devices that are topologically co-located. These systems tend to come with specialized ACS requirements for PCI Peer to Peer, for instance ACS disable or ACS setup specially for translated traffic.

NVIDIA's latest systems have a novel PCI multipath system that requires special asymmetric ACS.

Introduce a tool to help users configure the ACS on such systems. The tool will be able to parse the PCI topology and identify the topological features then generate the require ACS settings.

Modern kernels support the config_acs kernel command line parameter to allow fine grained settings so the correct ACS for the topology can be fed into Grub and to the kernel command line to configure it at boot

The tool has four functions:
 topo - Print out the topology from the RDMA perspective. Indicate what
        devices are P2P connected to the NIC.
 write-grub-acs - Emit the config_acs kernel command line parameter for
                  the required ACS configuration
 setpci-acs - Use setpci after booting to set the required ACS
              configuration. This is not recommended but provided to help
              legacy systems without config_acs.
 check - Read the live ACS settings and compare them to the required
         configuration

This initial version supports two NVIDIA platforms. There is an expectation it will grow to more broadly support more common topologies as well.

### DIFF
--- a/debian/rdma-core.install
+++ b/debian/rdma-core.install
@@ -22,6 +22,7 @@ lib/udev/rules.d/90-rdma-ulp-modules.rules
 lib/udev/rules.d/90-rdma-umad.rules
 usr/lib/truescale-serdes.cmds
 usr/sbin/iwpmd
+usr/sbin/rdma_topo
 usr/sbin/rdma-ndd
 usr/share/doc/rdma-core/70-persistent-ipoib.rules
 usr/share/doc/rdma-core/MAINTAINERS
--- a/kernel-boot/CMakeLists.txt
+++ b/kernel-boot/CMakeLists.txt
@@ -41,6 +41,9 @@ rdma_subst_install(FILES "persistent-ipoib.rules.in"
   DESTINATION "${CMAKE_INSTALL_DOCDIR}"
   PERMISSIONS OWNER_WRITE OWNER_READ GROUP_READ WORLD_READ)
 
+install(FILES "rdma_topo"
+  DESTINATION "${CMAKE_INSTALL_SBINDIR}")
+
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
 
 # Create an installed executable (under /usr/lib/udev)
--- a/kernel-boot/rdma_topo
+++ b/kernel-boot/rdma_topo
@@ -1,0 +1,764 @@
+#!/usr/bin/env python3
+# SPDX-License-Identifier: Linux-OpenIB
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES
+# PYTHON_ARGCOMPLETE_OK
+from __future__ import annotations
+import argparse
+import collections
+import importlib
+import inspect
+import itertools
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
+from typing import *
+
+BDF_RE = re.compile(r"^([0-9a-f]+?):([0-9a-f]{2}?):([0-9a-f]{2}?)\.([0-9a-f])$")
+KERNEL_ACS_ISOLATED = "xx111x1"
+pci_vendors = {
+    "MELLANOX": 0x15B3,
+    "NVIDIA": 0x10DE,
+}
+
+
+class CommandError(Exception):
+    pass
+
+
+def sysfs_read_str(sysdir: str, fn: str) -> str:
+    """Read the entire content of a sysfs file to a string"""
+    with open(os.path.join(sysdir, fn)) as F:
+        return F.read().strip()
+
+
+def sysfs_read_link(sysdir: str, fn: str) -> str:
+    """Read a link in sysfs to an absolute path string"""
+    return os.readlink(os.path.join(sysdir, fn))
+
+
+def PCI_VDEVICE(vendor: str, device_id: int) -> re.Pattern:
+    """Match a Vendor and device ID"""
+    vendor_id = pci_vendors[vendor]
+    return re.compile(rf"^pci:v{vendor_id:08X}d{device_id:08X}.*$")
+
+
+def PCI_DEVICE_CLASS(cid: int) -> re.Pattern:
+    """Match by exact programming class using the int coding from the kernel"""
+    class_id = (cid >> 16) & 0xFF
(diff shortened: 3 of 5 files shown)


debian/rdma-core.install
```diff
@@ -22,6 +22,7 @@ lib/udev/rules.d/90-rdma-ulp-modules.rules
 lib/udev/rules.d/90-rdma-umad.rules
 usr/lib/truescale-serdes.cmds
 usr/sbin/iwpmd
+usr/sbin/rdma_topo
```
rleon, line 25: Can we name it rdma-topo and not rdma_topo?

Also why did you put it in kernel-boot directory? I would expect that you will create new directory tools/ and put it there.

kernel-boot/rdma_topo
```diff
@@ -0,0 +1,764 @@
...
+def cmd_write_grub_acs(args):
+    """Generate a grub dropin file to have the kernel commandline set the
+    required ACS flags during system boot. This is the recommended way to
+    configure ACS on systems but requires an NVIDIA Optimized 6.14 kernel (or
+    newer), or otherwise compatible kernel.
+
+    If the system does not have any need of ACS flags the dropin file will be
+    removed. This command is intended for Debian style systems with a
+    /etc/default/grub.d and update-grub command"""
+    # FIXME: What about RHEL and grubby?
```
rleon: ?

kernel-boot/rdma_topo
```diff
@@ -0,0 +1,764 @@
...
+        f"{acs}@{pdev.bdf}"
+        for pdev, acs in sorted(acs.items(), key=lambda x: x[0].bdf)
+        if acs != KERNEL_ACS_ISOLATED
+    ]
+    acs_arg = ";".join(config_acs)
+    grub_conf = [
+        f"# Generated by {sys.argv[0]} do not change. ACS settings for RDMA GPU Direct",
+        f'GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX pci=config_acs=\\"{acs_arg}\\""',
+    ]
+    grub_conf = "\n".join(grub_conf)
```
rleon, line 593: I don't see check for previous runs of this tool, and maybe you already have GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX pci=config_acs=... line.

More broadly speaking, I think that it is wrong to write to grub file directly.
It is perfectly fine to print expected line, but write itself should be done by the user.

kernel-boot/rdma_topo
```diff
@@ -0,0 +1,764 @@
...
+        "-n",
+        "--dry-run",
+        action="store_true",
+        dest="dry_run",
+        help="Output the grub configuration to stdout and make no changes",
+    )
+    parser.add_argument(
+        "--output",
+        action="store",
+        default="/etc/default/grub.d/config-acs.cfg",
```
rleon, line 561: grub.d folder doesn't exist in my Fedora

kernel-boot/rdma_topo
```diff
@@ -0,0 +1,764 @@
...
+            subsys = pdev.get_subsystems()
+            if subsys:
+                jnvcx["subsystems"][str(pdev.bdf)] = {
+                    subsys: list(devs) for subsys, devs in subsys.items()
+                }
+        jtop.append(jnvcx)
+    print(json.dumps(jtop, indent=4))
+
+
+def cmd_topo(args):
```
rleon: Let's not shorten english words.
topo -> topology

debian/rdma-core.install
jgunthorpe: Most of the tools we have in bin use _ not -
I put it there because it is mainly about writing out the kernel command line and didn't think another directory was really needed.

kernel-boot/rdma_topo
jgunthorpe: Each run updates the file with the correct command line. The entire point of the sub command is to write the grub file, if the user doesn't want to do that they shouldn't call the command. Eventually the intention is to call this as part of the post-install packaging scripts on debian.

kernel-boot/rdma_topo
jgunthorpe: Yeah, it doesn't use that mechanism. I couldn't figure out how to work the RHEL/Fedora system, that will come with some followup commit.

kernel-boot/rdma_topo
jgunthorpe: OK

kernel-boot/rdma_topo
jgunthorpe: I dropped the fixme, the doc text above already explains it only works on debian style systems.

rleon: and I think that it is bad thing to add such functionality to main rdma-core repository which suppose to work on all systems

jgunthorpe: I hope to get something for Fedora someday, but right now I couldn't figure out how to safely use their complicated system from a packaging script. This is working for Debian systems, and provides a way to manually do it on RHEL like systems with one of the admin-only methods RHEL documents. I don't want to force Debian systems to have to use manual methods just because Fedora lacks these features for grub integration.

kernel-boot/CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ rdma_subst_install(FILES "persistent-ipoib.rules.in"
   DESTINATION "${CMAKE_INSTALL_DOCDIR}"
   PERMISSIONS OWNER_WRITE OWNER_READ GROUP_READ WORLD_READ)
 
+install(FILES "rdma_topo"
```
vladum:
```suggestion
install(PROGRAMS "rdma_topo"
```

From https://cmake.org/cmake/help/latest/command/install.html#programs:
> The PROGRAMS form is identical to the FILES form except that the default permissions for the installed file also include OWNER_EXECUTE, GROUP_EXECUTE, and WORLD_EXECUTE. This form is intended to install programs that are not targets, such as shell scripts.

PR #1646, opened 2025-09-14
efa: Extend DV query CQ to return doorbell

mrgolin:
Return the CQ reverse doorbell to give external data path implementations (GPU, etc.) an option to notify the device of polled entries.

### DIFF
--- a/providers/efa/efadv.h
+++ b/providers/efa/efadv.h
@@ -127,6 +127,7 @@ struct efadv_cq_attr {
 	uint8_t *buffer;
 	uint32_t entry_size;
 	uint32_t num_entries;
+	uint32_t *doorbell;
 };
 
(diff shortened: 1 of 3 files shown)


providers/efa/efadv.h
```diff
@@ -127,6 +127,7 @@ struct efadv_cq_attr {
 	uint8_t *buffer;
 	uint32_t entry_size;
 	uint32_t num_entries;
+	uint32_t *doorbell;
```
rleon, line 130: If it is address, why is it uint32_t and not uint64_t?

mrgolin: The doorbell is 32 bits and I prefer returning a relevant pointer type rather than an address, in similar to what we do in ```efadv_query_qp_wqs```.
Changed the commit message to better reflect on this.

mrgolin:
@rleon if there are no further comments, can you please merge this change?

PR #1653, opened 2025-10-31
[RFC] [NOT FOR MAIN] libibverbs: Add ultra ethernet support

shefty:
This series targets merging into a staging branch until an underlying implementation is available.

The following set of patches introduce extensions to the libibverbs API to support Ultra Ethernet Transport (UET).  The UE specification is available from the UE website:

https://ultraethernet.org/wp-content/uploads/sites/20/2025/10/UE-Specification-1.0.1.pdf

UEC is working towards kernel and user space implementations for UET devices.  The purpose behind this PR is to share the targeted user space API changes between UEC developers and rdma-core maintainers.  This allows for an open discussion to ensure development continue to moves forward in an acceptable path.  It is anticipated that libibverbs API changes will be reflected to some degree in the kernel ABI and APIs and some of the detailed changes here will result in deeper discussions both within the UEC and with rdma-core.

The first patch in this series describes the verbs semantic model.  The intent is to document how verbs, despite its naming, has evolved beyond its original Infiniband constructs.  For UET, a clear mapping between verbs objects and field structures to transport specific fields will eventually fall out.

Subsequent patches extend the verbs API for UET.  UET has several significant features.  These include:

- Reliable-unconnected communication semantic as viewed by the application.  We introduce a new QP type, updates to QP attributes, and defines UET QP states.

- UET uses job-based communication.  This series specifically targets the UEC AI base profile.  The AI base profile targets support for CCL style jobs, while excluding more complex transport operations such as tag matching.  To support job-based communication, new job-related objects are introduced.  Transport operations that target the AI full and HPC profiles are deferred for future changes.

- UET allows finer control over the exposure of registered memory regions.  The memory registration API and QP operations are extended to support UET allowed memory registration models.  This includes restricting a MR to being used by a specific job.  It also allows exposing a MR to the network separate from the creation of the MR.

- UET is designed for sending and receiving data out-of-order.  UET defines 3 reliable traffic ordering modes: reliable-ordered delivery (ROD), reliable-unordered delivery (RUD), and reliable-unordered delivery for idempotent transports (RUDI).  Traffic between 2 UET QPs may use all delivery mechanism simultaneously.  For example, message headers may use ROD, while message data may use RUD.  This is a significant deviation from the verbs RC QP model.  To allow the application and verbs provider to negotiate when message and/or data ordering may be relaxed, a new QP semantic structure is introduced.

Additional details are documented in each patch.

### DIFF
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -140,6 +140,7 @@ enum ibv_qp_msg_order {
 
 enum ibv_qp_use_flags {
 	IBV_QP_USAGE_IMM_DATA_RQ = (1 << 0),
+	IBV_QP_USAGE_ATTACH_MR = (1 << 1),
 };
 
 struct ibv_qp_semantics {
@@ -2357,6 +2358,8 @@ struct ibv_values_ex {
 
 struct verbs_context {
 	/*  "grows up" - new fields go here */
+	int (*attach_mr)(struct ibv_qp *qp, struct ibv_mr *mr);
+	int (*detach_mr)(struct ibv_qp *qp, struct ibv_mr *mr);
 	int (*query_qp_semantics)(struct ibv_context *context,
 				 enum ibv_qp_type qp_type,
 				 struct ibv_ah_attr *ah_attr,
@@ -2923,6 +2926,22 @@ static inline struct ibv_mw *ibv_alloc_mw(struct ibv_pd *pd,
 static inline int ibv_dealloc_mw(struct ibv_mw *mw)
 {
 	return mw->context->ops.dealloc_mw(mw);
+}
+
+static inline int ibv_attach_mr(struct ibv_qp *qp, struct ibv_mr *mr)
+{
+	struct verbs_context *vctx = verbs_get_ctx_op(qp->context, attach_mr);
+
+	if (!vctx)
+		return EOPNOTSUPP;
+
+	return vctx->attach_mr(qp, mr);
+}
+
+static inline int ibv_detach_mr(struct ibv_qp *qp, struct ibv_mr *mr)
+{
+	struct verbs_context *vctx = verbs_get_ctx_op(qp->context, attach_mr);
+	return vctx->detach_mr(qp, mr);
 }
 
 /**
--- a/libibverbs/verbs.h
+++ b/libibverbs/verbs.h
@@ -1160,7 +1160,8 @@ enum ibv_send_flags {
 	IBV_SEND_SIGNALED	= 1 << 1,
 	IBV_SEND_SOLICITED	= 1 << 2,
 	IBV_SEND_INLINE		= 1 << 3,
-	IBV_SEND_IP_CSUM	= 1 << 4
+	IBV_SEND_IP_CSUM	= 1 << 4,
+	IBV_SEND_DELIVERY_COMPLETE = 1 << 5,
 };
 
 enum ibv_placement_type {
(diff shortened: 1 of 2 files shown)


libibverbs/verbs.h
```diff
@@ -2524,6 +2719,21 @@ int ibv_query_pkey(struct ibv_context *context, uint8_t port_num,
 int ibv_get_pkey_index(struct ibv_context *context, uint8_t port_num,
 		       __be16 pkey);
 
+static inline int ibv_query_qp_semantics(struct ibv_context *context,
+					 enum ibv_qp_type qp_type,
+					 struct ibv_ah_attr *ah_attr,
```
shefty, line 2724: Simplify, replace ah_attr with port_num + sgid index.

libibverbs/verbs.h
```diff
@@ -1993,6 +2136,51 @@ struct ibv_flow_action_esp_attr {
...
+	uint32_t handle;
+};
+
+struct ibv_job_attr {
+	uint32_t comp_mask;
+	unsigned int flags;
+	uint64_t id;
+	uint32_t max_addr_entries;
+	enum ibv_qp_type qp_type;
+	struct ibv_ah_attr ah_attr;
```
shefty, line 2151: Simplify, remove qp_type.  Replace ah_attr with port_num + sgid index.

libibverbs/verbs.h
```diff
@@ -1463,6 +1569,25 @@ static inline void ibv_wr_set_ud_addr(struct ibv_qp_ex *qp, struct ibv_ah *ah,
 	qp->wr_set_ud_addr(qp, ah, remote_qpn, remote_qkey);
 }
 
+static inline void ibv_wr_set_ru_addr(struct ibv_qp_ex *qp, struct ibv_ah *ah,
+				      uint32_t remote_qpn, uint32_t jkey)
```
shefty, line 1573: We need a new object and API that combines the ah + remote_qpn + jkey together, which can be created prior to posting a WR.

Consider an implementation that only has an address table underneath.  This API would force an expensive lookup when posting a wr to determine if the remote qpn is in the table or not and its location.

The new object would be some sort of 'transport address'.

libibverbs/verbs.h
```diff
@@ -679,6 +728,7 @@ struct ibv_mr {
 	uint32_t		handle;
 	uint32_t		lkey;
 	uint32_t		rkey;
+	uint64_t		rkey64;
```
shefty, line 731: Also allow for lkey64 for NICs that use the same key value for lkeys and rkeys.  If an rkey 64 is valid, the lkey 64 must also be valid.  Update post wr APIs accordingly.  Note that a vendor can set the upper bits of the keys to 0 if they are not needed.

libibverbs/verbs.h
```diff
@@ -557,6 +604,8 @@ enum ibv_create_cq_wc_flags {
 	IBV_WC_EX_WITH_FLOW_TAG		= 1 << 9,
 	IBV_WC_EX_WITH_TM_INFO		= 1 << 10,
 	IBV_WC_EX_WITH_COMPLETION_TIMESTAMP_WALLCLOCK	= 1 << 11,
+	IBV_WC_EX_WITH_IMM64		= 1 << 12,
+	IBV_WC_EX_WITH_SRC_ID		= 1 << 13, /* implies job id */
```
shefty, line 608: remove comment

shefty (commented):
Comments based on discussions within UEC which need to be reflected in the proposed API.

shefty:
There have been several changes to this proposal needed to meet UET semantics.  As is, this RFC might be 70-80% of the way there.  Closing to open a new PR at a later date which will be closer aligned with UEC discussions.

PR #1654, opened 2025-11-03
bnxt_re/lib: Support direct verbs

sbasavapatna:
Hi,

This patchset supports Direct Verbs in the bnxt_re library.

This is required by vendor specific applications that need to manage
the HW resources directly and to implement the datapath in the
application.

To support this, the library and the driver are being enhanced to
provide Direct Verbs using which the application can allocate and
manage the HW resources (Queues, Doorbell etc) . The Direct Verbs
enable the application to implement the control path.

Patch#1 Update kernel headers
Patch#2 Add a helper for VA-backed buffer descriptors
Patch#3 Add ibv_buf APIs for user allocated memory
Patch#4 Export HSI file
Patch#5 Support DBR verbs
Patch#6 Update bnxt_re_cq to include struct verbs_cq
Patch#7 Support CQ verbs
Patch#8 Support QP verbs
Patch#9 Add man pages

Thanks,
-Harsha


providers/bnxt_re/bnxt_re_dv.h
```diff
@@ -0,0 +1,78 @@
...
+	int dmabuf_fd;
+};
+
+struct bnxt_re_dv_db_region_attr *
+bnxt_re_dv_alloc_db_region(struct ibv_context *ctx);
+int bnxt_re_dv_free_db_region(struct ibv_context *ctx,
+			      struct bnxt_re_dv_db_region_attr *attr);
+int bnxt_re_dv_get_default_db_region(struct ibv_context *ibvctx,
+				     struct bnxt_re_dv_db_region_attr *out);
+void *bnxt_re_dv_umem_reg(struct ibv_context *ibvctx,
```
jgunthorpe: void *? That shouldn't be. Why do you have these functions when there is no kernel side?

jgunthorpe:
So give it an opaque type, don't just randomly use void *s. 

Again,why is this stuff here when there is no kernel side for umems?

sbasavapatna:
On Thu, Jan 8, 2026 at 1:28 AM Jason Gunthorpe ***@***.***>
wrote:

> *jgunthorpe* left a comment (linux-rdma/rdma-core#1654)
> <https://github.com/linux-rdma/rdma-core/pull/1654#issuecomment-3720537042>
>
> So give it an opaque type, don't just randomly use void *s.
>
Are you suggesting to use a pointer to the forward declaration of that
structure? Is that what you meant by "give it an opaque type"? Please
clarify.

> Again,why is this stuff here when there is no kernel side for umems?
>
The application registers the addr and the total size of the memory area
with the library, initially using this API. Later during each object
creation, the library uses this information to provide the actual va
(base-addr + offset) for mapping/pinning, to the kernel driver. This has
been captured in the kernel driver review thread:
***@***.***/

sbasavapatna:
While looking into changes needed in the bnxt_re kernel driver to process DBR passed through driver specific attributes (during QP creation), I noticed that the corresponding libibverbs support seems to be missing in cmd_qp.c.
That is, ibv_cmd_create_qp_ex() doesn't support "ibv_command_buffer *driver" arg that should be linked to the main command buffer that it creates. This is supported in CQ creation command though (ibv_cmd_create_cq_ex2()).

sbasavapatna:
On Thu, Jan 15, 2026 at 10:19 PM Jason Gunthorpe ***@***.***>
wrote:

> *jgunthorpe* left a comment (linux-rdma/rdma-core#1654)
> <https://github.com/linux-rdma/rdma-core/pull/1654#issuecomment-3755833542>
>
> As the first user you'd need to do a similar thing as CQ.
>
Pushed the next version with a new create_qp function in libibverbs.
Thanks,
-Harsha

libibverbs/driver.h
```diff
@@ -853,4 +853,25 @@ fill_attr_in_buf_umem(struct ibv_command_buffer *cmdb, uint16_t attr_id,
...
+	*storage = (struct ib_uverbs_buffer_desc){
+		.type	= IB_UVERBS_BUFFER_TYPE_VA,
+		.fd	= -1,
+		.addr	= (uintptr_t)(addr ? addr : buf->addr),
+		.length	= length ? length : buf->size,
+	};
+	fill_attr_in_ptr(cmdb, attr_id, storage);
+	cmdb->fallback_ioctl_only = 1;
+}
+
```
jpirko, line 895: Does not make sense to have this as separate helper. Your caller checks "buf->fd ==-1". You can do the check inside fill_attr_in_buf_umem and fill up things (like type) appropriately.

providers/bnxt_re/bnxt_re_dv.h
```diff
@@ -0,0 +1,78 @@
...
+	uint64_t umdbr;
+	uint64_t *dbr;
+};
+
+enum  bnxt_re_dv_umem_in_flags {
+	BNXT_RE_DV_UMEM_FLAGS_DMABUF = 1 << 0,
+};
+
+struct bnxt_re_dv_umem_reg_attr {
+	void *addr;
```
jpirko: This duplicates the layout of struct ibv_buf. Why you can't use it directly?

libibverbs/driver.h
sbasavapatna: Done, will update it in the next version.

providers/bnxt_re/bnxt_re_dv.h
sbasavapatna: This is a user facing structure and we don't want to change it as it would impact applications that are already using it.

jpirko: This puzzles me. How an application can use this struct if you are only introducing it now?

libibverbs/driver.h
jpirko: looks okay to me now.

providers/bnxt_re/bnxt_re_dv.h
sbasavapatna: OOB solution which already exists is being upstreamed.

jpirko: Wait, you say that you form the API of a feature currently being pushed upon some out of tree code and use it as argument the API can't be changed? I have troubles to understand that reasoning, sorry.

sbasavapatna: The structure layouts are slightly different. We have added a comp_mask for vendor specific extensions.
 
struct bnxt_re_dv_umem_reg_attr {
        void *addr;
        size_t size;
        uint64_t comp_mask;
        int dmabuf_fd;
};
struct ibv_buf {
        void *addr;
        size_t size;
        struct ibv_pd *pd;
        int dmabuf_fd;
};

Also,  since this is a Direct Verb to the user application, we want to maintain the same ABI for both our inbox and out-of-tree libraries, so that the applications can work seamlessly on both installations. So unless you see a design level issue here, we would prefer to keep this as-is and avoid any impact to existing customers that are using out-of-tree packages.

jpirko: Why you have comp_mask here is a mystery. Not needed (and if really, really need it, you can embed ibv_buf - so no argument). Basically they are same structures. Don't screw-up the API in a name of out-of-tree code. That is fundamentally wrong. Please fix.

providers/bnxt_re/dv.c
```diff
@@ -0,0 +1,196 @@
...
+	DECLARE_COMMAND_BUFFER(cmd,
+			       BNXT_RE_OBJECT_DEFAULT_DBR,
+			       BNXT_RE_METHOD_GET_DEFAULT_DBR,
+			       1);
+
+	fill_attr_out_ptr(cmd, BNXT_RE_DEFAULT_DBR_ATTR, &attr);
+
+	ret = execute_ioctl(ibvctx, cmd);
+	if (ret) {
+		fprintf(stderr, "%s: execute_ioctl() failed: %d\n", __func__, ret);
```
jgunthorpe: Do not print errors from libraries like this, remove them all

providers/bnxt_re/verbs.c
```diff
@@ -375,10 +377,15 @@ struct ibv_cq *bnxt_re_create_cq(struct ibv_context *ibvctx, int ncqe,
...
-			      &resp.ibv_resp, sizeof(resp)))
+	cq_attr_ex.cqe = ncqe;
+	cq_attr_ex.channel = channel;
+	cq_attr_ex.comp_vector = vec;
+	if (ibv_cmd_create_cq_ex(ibvctx, &cq_attr_ex, NULL,
+				 (struct verbs_cq *)&cq->ibvcq,
+				 (struct ibv_create_cq_ex *)&cmd.ibv_cmd,
+				 sizeof(cmd),
+				 (struct ib_uverbs_ex_create_cq_resp *)&resp.ibv_resp,
+				 sizeof(resp), 0))
```
jgunthorpe, line 388: Why all the casts here? It shouldn't have casts if written properly

providers/bnxt_re/verbs.c
```diff
@@ -2107,7 +2117,7 @@ static struct ibv_qp *__bnxt_re_create_qp(struct ibv_context *ibvctx,
 		req.sq_slots = qattr[BNXT_RE_QATTR_SQ_INDX].slots;
 
 	if (ibv_cmd_create_qp_ex(ibvctx, &qp->vqp, attr,
-				&req.ibv_cmd, sizeof(req), &resp.ibv_resp, sizeof(resp)))
+				 &req.ibv_cmd, sizeof(req), &resp.ibv_resp, sizeof(resp)))
```
jgunthorpe: stray whitespace?

providers/bnxt_re/dv.c
```diff
@@ -194,3 +194,120 @@ int bnxt_re_dv_umem_dereg(struct bnxt_re_dv_umem *umem)
...
+	cq_attr_ex.cqe = cq_attr->ncqe;
+	cq_attr_ex.comp_mask = 0;
+	cq_attr_ex.flags = 0;
+
+	memset(resp, 0, sizeof(*resp));
+	ret = ibv_cmd_create_cq_ex2(ibvctx, &cq_attr_ex, NULL,
+				    (struct verbs_cq *)&cq->ibvcq,
+				    (struct ibv_create_cq_ex *)&cmd.ibv_cmd, sizeof(cmd),
+				    (struct ib_uverbs_ex_create_cq_resp *)&resp->ibv_resp,
+				    sizeof(*resp), 0, driver_attrs);
```
jgunthorpe, line 190: Also why casts?

providers/bnxt_re/bnxt_re_dv.h
jgunthorpe: I agree with Jiri, we have ibv_buf now that is the user facing API surface for a dmabuf or a VA or whatever, you should add some API to allow the user to create an ibv_buf from other things and consistently use that in your here in DV. Get rid of the umem here, it completely duplicates what ibv_buf is trying to do.

You guys choose to do the OOT thing, the broken ABIs are all your problem to deal with.

providers/bnxt_re/dv.c
```diff
@@ -311,3 +311,305 @@ int bnxt_re_dv_destroy_cq(struct ibv_cq *ibvcq)
...
+	dst->ah_attr.dlid = src->ah_attr.dlid;
+	dst->ah_attr.is_global  = src->ah_attr.is_global;
+	memcpy(&dst->ah_attr.grh.dgid, &src->ah_attr.grh.dgid, 16);
+	dst->ah_attr.grh.sgid_index = src->ah_attr.grh.sgid_index;
+	dst->ah_attr.grh.hop_limit = src->ah_attr.grh.hop_limit;
+	dst->ah_attr.grh.traffic_class = src->ah_attr.grh.traffic_class;
+	dst->ah_attr.grh.flow_label = src->ah_attr.grh.flow_label;
+}
+
+int bnxt_re_dv_query_qp(struct ibv_qp *ibvqp, struct ib_uverbs_qp_attr *qp_attr)
```
jgunthorpe: This type ib_uverbs_qp_attr should not be returned from a library interface, it is a kernel facing type

I don't even understand why this function exists in the first place

providers/bnxt_re/dv.c
```diff
@@ -311,3 +311,305 @@ int bnxt_re_dv_destroy_cq(struct ibv_cq *ibvcq)
...
+{
+	struct bnxt_re_qp *qp = to_bnxt_re_qp(ibvqp);
+	struct ibv_qp_init_attr init_attr = {};
+	struct ibv_qp_attr attr = {};
+	struct ibv_query_qp cmd;
+	int rc;
+
+	rc = ibv_cmd_query_qp(ibvqp, &attr, qp_attr->qp_attr_mask, &init_attr,
+			      &cmd, sizeof(cmd));
+	if (!rc) {
```
jgunthorpe: success oriented flow

providers/bnxt_re/man/bnxt_re_dv_create_cq.3.md
```diff
@@ -0,0 +1,57 @@
...
+
+# DESCRIPTION
+
+**bnxt_re_dv_create_cq**() creates a completion queue using memory registered
+with **bnxt_re_dv_umem_reg**(3). The **umem_handle** field must be the opaque
+handle returned by that function; **umem_offset** is the byte offset into that
+registration, aligned to the device page size; **ncqe** is the requested number
+of CQEs.
+
+This path requires **IB_UVERBS_CORE_SUPPORT_ROBUST_UDATA** (robust udata) on the
```
jgunthorpe: Doesn't seem like something the user needs to know. You could write 'this call requires linux kernel version >= 'xyz

providers/bnxt_re/man/bnxt_re_dv_destroy_qp.3.md
```diff
@@ -0,0 +1,46 @@
...
+# NAME
+
+bnxt_re_dv_destroy_qp - destroy a QP created with bnxt_re_dv_create_qp
+
+# SYNOPSIS
+
+```c
+#include <infiniband/bnxt_re_dv.h>
+
+int bnxt_re_dv_destroy_qp(struct ibv_qp *ibvqp);
```
jgunthorpe: Why does this exist? What is wrong with the normal QP destruction?

jgunthorpe (changes requested):

providers/bnxt_re/man/bnxt_re_dv_alloc_db_region.3.md
```diff
@@ -0,0 +1,64 @@
...
+# RETURN VALUE
+
+On success, returns a pointer to a **struct bnxt_re_dv_db_region_attr**
+with all fields filled in:
+
+```c
+struct bnxt_re_dv_db_region_attr {
+        uint32_t  handle;   /* kernel handle for this doorbell region */
+        uint32_t  dpi;      /* doorbell page index */
+        uint64_t  umdbr;    /* unmapped doorbell BAR offset */
```
mplaneta-qc: What is the intended flow for using `umdbr`? Does unmapped mean not mapped into application virtual address space?

providers/bnxt_re/dv.c
sbasavapatna: will remove these error prints.

providers/bnxt_re/verbs.c
sbasavapatna: will fix this.

providers/bnxt_re/verbs.c
sbasavapatna: accept.

providers/bnxt_re/dv.c
sbasavapatna: will be fixed.

providers/bnxt_re/dv.c
sbasavapatna: Thanks for catching this; in the next version this will be restored to standard bnxt_re_query_qp() and the dv function will be removed.

providers/bnxt_re/dv.c
sbasavapatna: This function itself won't be there in the next version.

providers/bnxt_re/man/bnxt_re_dv_create_cq.3.md
sbasavapatna: done.

providers/bnxt_re/man/bnxt_re_dv_destroy_qp.3.md
sbasavapatna: DV variants of QP functions - destroy, query and modify will be removed; they were mostly wrappers.  Similarly dv_destroy_cq will also be removed.

providers/bnxt_re/bnxt_re_dv.h
sbasavapatna: Will add a couple of APIs as suggested to associate/disassociate an ibv_buf with user allocated memory. This ibv_buf will then be passed to bnxt_re_dv_create_cq/qp APIs.

providers/bnxt_re/man/bnxt_re_dv_alloc_db_region.3.md
sbasavapatna: Yes, this is the raw PCI-BAR/physical-address of the doorbell register page.

mplaneta-qc: How user space can use this field?

sbasavapatna: It is not meant for normal user-space applications. It is intended to support future extensions for peer devices like GPUs that may need to directly write to the Doorbell BAR.

mplaneta-qc: Is it supposed to be used somehow with DMA buf then? https://www.kernel.org/doc/html/latest/driver-api/dma-buf.html

jgunthorpe: What? You cannot expose a physical address to userspace. Send a kernel patch to take this away please.

Aswin5005: Can we pass a File Descriptor with help of dma-buf framework to user space instead of returning raw Physical address ?

sbasavapatna: > What? You cannot expose a physical address to userspace. Send a kernel patch to take this away please.

Since the ABI structure is already in released kernels (7.1, 7.2), we can't remove the field right? I have a patch that changes it to a reserved field and avoids setting it.  If that's ok, I can send out that patch, let me know.

insanum:
Are there any comments regarding the new `ibv_buf` allocation APIs defined in eb805d4 (i.e., `ibv_alloc_user_buf` and `ibv_free_user_buf`)?

omor1:
I've noticed that there are several files in the PR with SPDX identifiers indicating that they are licensed under `GPL-2.0 OR Linux-OpenIB`, but the following full copyright header carries the text for [`BSD-2-Clause`](https://spdx.org/licenses/BSD-2-Clause.html) rather than [`Linux-OpenIB`](https://spdx.org/licenses/Linux-OpenIB.html).

Can you please clarify which is correct? Thanks!

rleon:
> I've noticed that there are several files in the PR with SPDX identifiers indicating that they are licensed under `GPL-2.0 OR Linux-OpenIB`, but the following full copyright header carries the text for [`BSD-2-Clause`](https://spdx.org/licenses/BSD-2-Clause.html) rather than [`Linux-OpenIB`](https://spdx.org/licenses/Linux-OpenIB.html).
> 
> Can you please clarify which is correct? Thanks!

You are right, we should have only SPDX tags without copied license text.

sbasavapatna:
> > I've noticed that there are several files in the PR with SPDX identifiers indicating that they are licensed under `GPL-2.0 OR Linux-OpenIB`, but the following full copyright header carries the text for [`BSD-2-Clause`](https://spdx.org/licenses/BSD-2-Clause.html) rather than [`Linux-OpenIB`](https://spdx.org/licenses/Linux-OpenIB.html).
> > Can you please clarify which is correct? Thanks!
> 
> You are right, we should have only SPDX tags without copied license text.

Thanks, pushed the next version with this change.

PR #1671, opened 2025-12-13
Fix enum-enum math for compatibility with C++20

r-barnes:
When compiling code that includes the header `infiniband/verbs.h` with CUDA using C++20 we observe:
```
buck-out/v2/gen/fbsource/third-party/rdma-core/stablev60/__ibverbs__/c01e142197f8fe04/buck-headers/infiniband/verbs.h:573:59: error: bitwise operation between different enumeration types ('(unnamed enum at buck-out/v2/gen/fbsource/third-party/rdma-core/stablev60/__ibverbs__/c01e142197f8fe04/buck-headers/infiniband/verbs.h:562:1)' and 'ibv_create_cq_wc_flags') is deprecated [-Werror,-Wdeprecated-anon-enum-enum-conversion]
  573 | IBV_CREATE_CQ_SUP_WC_FLAGS = ((((((IBV_WC_STANDARD_FLAGS) | (IBV_WC_EX_WITH_COMPLETION_TIMESTAMP)) | (IBV_WC_EX_WITH_CVLAN)) | (IBV_WC_EX_WITH_FLOW_TAG)) | (IBV_WC_EX_WITH_TM_INFO)) | (IBV_WC_EX_WITH_COMPLETION_TIMESTAMP_WALLCLOCK))
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This use of enums seems to be a pre-C++11 way of generating compile-time constants. This PR uses explicit casts to suppress the compile-time warning.

This code, is of course C, but the enum-enum conversion issues leak out into C++ because they're in a header. There are a few options:
* Use explicit casts to suppress the warning (what I've done here)
* Change the code so it uses constants - the enums don't seem semantically meaningful here.
* Upgrade to the C23 standard to gain access to `constexpr` (heavyweight solution).
* Suppress the relevant compiler flags in this header using `#pragma clang diagnostic push` and `#pragma clang diagnostic pop` (it's more complicated than this, but folly has a [macro](https://github.com/facebook/folly/blob/94a8e82cf16a0e229fc4fc89140219434ba78fa2/folly/CPortability.h#L324) that could be copied to cover all scenarios)

The current code's use of enums seems as though it could cause increasing compilation issues over time, so fixing it seems like a good idea.

Let me know what you'd like to do - I'm happy to help.


jgunthorpe:
I think since these are intended to be used by an application that will | them together the only answer is to change them to #define's. Otherwise the app will hit these warnings too.

jgunthorpe:
And if we are going to do this we should do it for all the enum's being used for defining bit fields.

PR #1693, opened 2026-01-21
verbs: Support the PROCMAP_QUERY ioctl

gerd-rausch:
The PROCMAP_QUERY ioctl, available since Linux-6.11, is able to obtain the size of an underlying mapping without the need to parse /proc/self/smaps.

Because kernels that support the PROCMAP_QUERY ioctl also support copy_on_fork, environment variable
RDMAV_USE_MADVISE must be used to bypass the
copy_on_fork feature, and allow PROCMAP_QUERY to be used instead.

If PROCMAP_QUERY isn't available on the running kernel, the "smaps" file will be parsed afterall.

### DIFF
--- a/libibverbs/memory.c
+++ b/libibverbs/memory.c
@@ -43,6 +43,10 @@
 #include <dirent.h>
 #include <limits.h>
 #include <inttypes.h>
+#include <sys/types.h>
+#include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


gerd-rausch:
copy_on_fork is slow with very large mappings.

PR #1696, opened 2026-02-04
providers/mana: expose PD number via manadv 

TaranovK:
Expose PD number to manadv using manadv_init_obj().
Allow users to request short PDN using new manadv_alloc_pd() API.

### DIFF
--- a/debian/ibverbs-providers.symbols
+++ b/debian/ibverbs-providers.symbols
@@ -211,8 +211,10 @@ libhns.so.1 ibverbs-providers #MINVER#
 libmana.so.1 ibverbs-providers #MINVER#
 * Build-Depends-Package: libibverbs-dev
  MANA_1.0@MANA_1.0 41
+ MANA_1.1@MANA_1.1 65
  manadv_init_obj@MANA_1.0 41
  manadv_set_context_attr@MANA_1.0 41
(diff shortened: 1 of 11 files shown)


jgunthorpe:
This needs to be refreshed with a kernel git hash from the accepted patches

TaranovK:
Hi @jgunthorpe,

I have rebased the changes and updated the kernel header. Could you have a look?
Thanks!

jgunthorpe:
aren't the commits in the wrong order? Shouldn't the kernel update be first? The CI doesn't check it but we expect every commit to compile

TaranovK:
@jgunthorpe I have reordered the commits. Thanks

jgunthorpe:
Doesn't mana have to check IB_UVERBS_CORE_SUPPORT_ROBUST_UDATA before attempting to use these flows? Do all old kernels have the necessary checks to be properly compatible?

TaranovK:
We check MANA_IB_UCNTX_ALLOC_PDN_SUPPORT to use new flows. The check of IB_UVERBS_CORE_SUPPORT_ROBUST_UDATA is redundant. 
Old kernels have the necessary checks as they follow the ROBUST UDATA and they check the comp_mask in newer version and in older version expect empty request.

PR #1701, opened 2026-02-16
Introduce Completion Counters verbs

mrgolin:
Extend verbs interface to support Completion Counters that can be seen as a light-weight alternative to polling CQ. A completion counter object separately counts successful and error completions, can be attached to multiple QPs and be configured to count completions of a subset of operation types. This is especially useful for batch or credit based workloads running on accelerators but can serve many other types of applications as well.

Kernel changes are [here](https://lore.kernel.org/all/20260720132221.2551-1-mrgolin@amazon.com/).

Thanks.


jgunthorpe:
I've been feeling a high reluctance lately to introduce new verbs that rely on a single HW implementations.. That has not gone well in the past. Consider using dv, but also consider the questions below if it should be a new verb:

Can any existing HW implement this, or just one?
Are there any prior-art standards that already implement this idea, like UE or something?
What does hardware actually have to do to implement it?

shijin-aws:
@shefty Yes, the proposal is aimed to be aligned with Libfabric semantic as a co-design between @mrgolin and my team

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,245 @@
...
+
+# DESCRIPTION
+
+Completion counters provide a lightweight completion mechanism as an
+alternative or extension to completion queues (CQs). Rather than generating
+individual completion queue entries, a completion counter tracks the aggregate
+number of completed operations. This makes them well suited for applications
+that need to know how many requests have completed without requiring
+per-request details, such as credit based flow control or tracking responses
+from remote peers.
```
shefty, line 52: libfabric and UEC call out 2 separate types of counters.  One counts completed data transfer operations.  The other counts the number of application bytes carried by a data transfer.  For example, the number of bytes written into a RDMA write target buffer.

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,245 @@
...
+Each completion counter maintains two distinct 64-bit values: a completion
+count that is incremented on successful completions, and an error count that
+is incremented when operations complete in error.
+
+**ibv_create_comp_cntr**() allocates a new completion counter for the RDMA
+device context *context*. The properties of the counter are defined by
+*cc_attr*. On success, the returned **ibv_comp_cntr** structure contains
+pointers to the completion and error count values. The maximum number of
+completion counters a device supports is reported by the *max_comp_cntr*
+field of **ibv_device_attr_ex**.
```
shefty: Returning direct pointers to the memory locations where the counters are stored is forcing a specific implementation.  Counters could be local to the NIC and require specific commands to read.  It is more generic and aligns with libfabric to provide function calls to read/write the counters.

From what I can tell looking at the CXI libfabric implementation, this is how counters work on CXI.  Counters are a HW object.  Reading a counter submits a command that copies the value back to host memory.  A counter may also be associated with a threshold value.  Once the counter reaches that value, it will also update host memory with the value.

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,245 @@
...
+**ibv_memory_location** structure which supports two modes: a virtual address
+(**IBV_MEMORY_LOCATION_VA**), where the application supplies a direct pointer, or
+a DMA-BUF reference (**IBV_MEMORY_LOCATION_DMABUF**), where the application
+supplies a file descriptor and offset into an exported DMA-BUF. When using
+DMA-BUF, the *ptr* field may also be set to provide a process-accessible
+mapping of the memory; if provided, the *comp_count* and *err_count* pointers
+in the returned **ibv_comp_cntr** will point to it. Using external memory
+allows the counter values to reside in application-managed buffers or in
+memory exported through DMA-BUF, enabling zero-copy observation of completion
+progress by co-located processes or devices.
```
shefty: This deviates from libfabric and isn't called out by UEC.  Letting the app specify where the counter should reside is forcing a very specific HW implementation.  I agree with Jason here and would definitely want to have multiple NIC implementations available before adding this.

Can we simplify the API to meet more minimal requirements?

AFAICT, CXI can support an option such as this, but the buffer would be a write-back buffer, not the actual counter.  This suggests that a call such as ibv_read_cntr(cntr, &value, flags) might be an option to direct the counter value to a specific memory location, versus the libfabric method of returning the value directly from the function.

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,245 @@
...
+
+## ibv_comp_cntr_init_attr
+
+```c
+struct ibv_comp_cntr_init_attr {
+	uint32_t comp_mask;
+	uint32_t flags;
+	struct ibv_memory_location comp_cntr_ext_mem;
+	struct ibv_memory_location err_cntr_ext_mem;
+};
```
shefty, line 112: libfabric attributes include an enum for events to count (transfers or bytes).  UEC lists support for both as mandatory.  Adding an enum here, even if only transfers are included to start with, would make this more extensible.

libfabric also supports blocking counter read calls.  It's possible this might align with a verbs ibv_comp_channel (or ibv_cntr_channel?) object.

I don't see any statement in UEC regarding blocking counters (or CQs).  However, since libfabric counters include blocking support, they may simply not have thought it important to call that support out separately.  (I don't think MPI uses blocking counters.)

CXI implements blocking reads by setting a triggered threshold.  HW writes the counter value to memory once it hits the trigger.  SW in this case spins until the counter is triggered.

At least based on UEC 1.0, I would defer adding wait or trigger APIs until there's a larger discussion.

libibverbs/man/ibv_qp_attach_comp_cntr.3.md
```diff
@@ -0,0 +1,118 @@
...
+	IBV_COMP_CNTR_ATTACH_OP_REMOTE_RDMA_READ        = 1 << 3,
+	IBV_COMP_CNTR_ATTACH_OP_RDMA_WRITE              = 1 << 4,
+	IBV_COMP_CNTR_ATTACH_OP_REMOTE_RDMA_WRITE       = 1 << 5,
+};
+
+struct ibv_comp_cntr_attach_attr {
+	uint32_t comp_mask;
+	uint32_t op_mask;
+};
+```
```
shefty, line 89: This behavior seems aligned with libfabric, so it should work with UEC.

shefty (commented):
Comments are based on comparing to UEC spec requirements, libfabric API, and looking at the CXI implementation, which is the only other HW-based counter implementation I'm aware of.

libibverbs/man/ibv_create_comp_cntr.3.md
mrgolin: Thanks for your feedback, replied in another comment below.

libibverbs/man/ibv_create_comp_cntr.3.md
mrgolin: One of the use cases we want this API to support is direct polling of the counters from GPUs, what requires having counter memory addresses. In this case I actually expect counters to be created with external memory located in HBM so exposing addresses might be redundant.
I think we can go with your suggestion, add ```ibv_cntr_read``` verb and pave the path to kernel, then each provider can decide to optimize it by simple memory read when possible.

libibverbs/man/ibv_create_comp_cntr.3.md
mrgolin: I added this to common verbs because I believe completion counters in GPU memory will be one of the main use cases for verbs completion counters. Device implementations can vary, but one of the options would be maintaining an internal counter and syncing it back to the user-supplied buffer, as you suggested CXI is doing. That said, I understand your concern that this may be difficult or impossible to implement on some devices that otherwise support the rest of the API.

We can either make ```IBV_COMP_CNTR_INIT_WITH_EXTERNAL_MEM``` support optional, or move the capability to create counters with external memory to direct verbs. @jgunthorpe what's your opinion on this?

libibverbs/man/ibv_create_comp_cntr.3.md
mrgolin: Regarding events vs. bytes I think it's already easy to extend by adding another flag like ```IBV_COMP_CNTR_INIT_COUNT_BYTES```.

For your comment about blocking counters, I would try keep ```comp_cntr``` and ```comp_channel``` separate in rdma-core. Upper layers (like Libfabric) can combine these constructs to implement more advanced behaviors. Anyway I think we would be able to add this support later if we see the need.

libibverbs/man/ibv_create_comp_cntr.3.md
rleon: > One of the use cases we want this API to support is direct polling of the counters from GPUs, what requires having counter memory addresses.

Isn't this handled through dmabuf mechanism?

mrgolin:
@jgunthorpe 
> Can any existing HW implement this, or just one?

It depends on device implementation, EFA is going to support this in existing devices.

> Are there any prior-art standards that already implement this idea, like UE or something?

As @shefty mentioned Libfabric has a similar interface and UE is going to support it as well.

> What does hardware actually have to do to implement it?

I think the bare minimum needed is the ability to map QP + opcode to a counter during completion handling and increment this counter. With @shefty's suggestion to add ibv_cntr_read, read can be implemented using an admin command.

libibverbs/man/ibv_create_comp_cntr.3.md
mrgolin: Yes, right. In the scenario I described the buffer is allocated in GPU memory, exported as dmabuf which is then passed in when creating counters. GPU will directly read counter values from its memory.
For this case exposing counter VA isn't needed and it might not be available.

mrgolin:
Updated according to suggestions above.
Looks there is some issue with the pipeline.

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,246 @@
...
+**ibv_memory_location** structure which supports two modes: a virtual address
+(**IBV_MEMORY_LOCATION_VA**), where the application supplies a direct pointer, or
+a DMA-BUF reference (**IBV_MEMORY_LOCATION_DMABUF**), where the application
+supplies a file descriptor and offset into an exported DMA-BUF. When using
+DMA-BUF, the *ptr* field may also be set to provide a process-accessible
+mapping of the memory, which may enable more efficient counter reads. Using
+external memory allows the counter values to
+reside in application-managed buffers or in memory exported through DMA-BUF,
+enabling zero-copy observation of completion progress by co-located processes
+or devices.
```
shefty, line 99: I still question trying to support external memory for counters as a standard API feature.  This is similar to supporting external memory for a CQ.  I don't see a benefit for supporting this for host memory.  GPU/device memory could have a benefit, but, again, this is forcing a specific NIC implementation.

libfabric took a different approach for NIC to GPU interactions in the 1.x release (removed from 2.x because of a lack of implementation).  That involved a trigger API, where once a counter hit a certain threshold, memory on the GPU would be updated.  That's not the same as keeping the actual counter value on the GPU.

https://github.com/ofiwg/libfabric/blob/v1.22.x/include/rdma/fi_trigger.h

The advantage that approach had is that the trigger could come from the NIC or CPU, based on the vendor implementation.  And even the counters could exist totally in SW.

I would remove external memory support and defer discussion for how to best support completions to a GPU.

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,246 @@
...
+# ARGUMENTS
+
+## ibv_comp_cntr
+
+```c
+struct ibv_comp_cntr {
+	struct ibv_context *context;
+	uint32_t handle;
+	uint64_t comp_count_max_value;
+	uint64_t err_count_max_value;
```
shefty, line 110: I would move these to device attributes, not set per counter.  An app may need to know this prior to created the counters, rather than after the fact.

libibverbs/man/ibv_create_comp_cntr.3.md
shefty: I would still remove external memory support, particularly in the initial API submission.

I would also include the enum to select the counter type (bytes vs. completions).  This isn't something that should be a flag, as only one is usable at a time.  Even if the enum only has 1 option to start with, introducing it now makes compatibility much easier, both from the perspective of verbs, but also the user.

mrgolin:
@shefty I've moved external memory support to EFA direct verbs.

pyverbs/comp_cntr.pyx
```diff
@@ -0,0 +1,99 @@
...
+cdef class CompCntr(PyverbsCM):
+    """Represents a Completion Counter."""
+    def __init__(self, Context ctx not None, CompCntrInitAttr attr not None):
+        super().__init__()
+        self.comp_cntr = v.ibv_create_comp_cntr(ctx.context, &attr.attr)
+        if self.comp_cntr == NULL:
+            raise PyverbsRDMAErrno('Failed to create comp_cntr')
+        self.ctx = ctx
+
+    def close(self):
```
ShacharKagan: CompCntr needs a __dealloc__ that calls close(). PyverbsCM doesn't provide one, so without it the C counter leaks whenever a CompCntr is garbage-collected outside an explicit close() / with block.

pyverbs/comp_cntr.pyx
```diff
@@ -0,0 +1,99 @@
...
+
+
+cdef class CompCntr(PyverbsCM):
+    """Represents a Completion Counter."""
+    def __init__(self, Context ctx not None, CompCntrInitAttr attr not None):
+        super().__init__()
+        self.comp_cntr = v.ibv_create_comp_cntr(ctx.context, &attr.attr)
+        if self.comp_cntr == NULL:
+            raise PyverbsRDMAErrno('Failed to create comp_cntr')
+        self.ctx = ctx
```
ShacharKagan, line 44: Register the new counter with its Context so teardown ordering works:
self.ctx = ctx
ctx.add_ref(self)

Also update pyverbs/device.pyx:
Add self.comp_cntrs = weakref.WeakSet() in Context.__init__.
Add an elif isinstance(obj, CompCntr): self.comp_cntrs.add(obj) branch in Context.add_ref.

pyverbs/comp_cntr.pyx
```diff
@@ -0,0 +1,99 @@
...
+#cython: language_level=3
+
+from pyverbs.base import PyverbsRDMAErrno, PyverbsRDMAError
+from pyverbs.device cimport Context
+cimport pyverbs.libibverbs as v
+
+
+cdef class CompCntrInitAttr(PyverbsObject):
+    """Represents ibv_comp_cntr_init_attr struct."""
+    def __init__(self, comp_mask=0, type=0, flags=0):
```
ShacharKagan: type shadows a Python builtin. Better to rename the parameter. Suggestion: cntr_type

pyverbs/comp_cntr.pyx
```diff
@@ -0,0 +1,99 @@
...
+        self.attr.op_mask = op_mask
+
+    @property
+    def op_mask(self):
+        return self.attr.op_mask
+
+
+cdef class CompCntr(PyverbsCM):
+    """Represents a Completion Counter."""
+    def __init__(self, Context ctx not None, CompCntrInitAttr attr not None):
```
ShacharKagan, line 34: Please add meaningful docstrings for the methods that wrap the verbs: __init__, close, set, set_err, inc, inc_err, read, read_err.

pyverbs/providers/efa/efadv.pyx
```diff
@@ -293,6 +296,50 @@ cdef class EfaCQ(CQEX):
...
+
+    def set_err_ext_mem_va(self, uintptr_t ptr):
+        self.attr.flags |= dve.EFADV_COMP_CNTR_INIT_WITH_ERR_EXTERNAL_MEM
+        self.attr.err_cntr_ext_mem.type = dve.EFADV_MEMORY_LOCATION_VA
+        self.attr.err_cntr_ext_mem.ptr = <uint8_t *>ptr
+
+
+cdef class EfaCompCntr(CompCntr):
+    """
+    Initializes an EFA Completion Counter with EFA-specific options.
```
ShacharKagan: Class docstrings don't take :param: / :return:. Move the parameter block to __init__ and replace the class docstring with a one-line description like "EFA-specific Completion Counter."

pyverbs/providers/efa/efadv.pyx
```diff
@@ -293,6 +296,50 @@ cdef class EfaCQ(CQEX):
...
+    :param ctx: Context object
+    :param attr: CompCntrInitAttr object
+    :param efa_attr: EfaCompCntrInitAttr object
+    :return: An initialized EfaCompCntr
+    """
+    def __init__(self, Context ctx not None, CompCntrInitAttr attr not None,
+                 EfaCompCntrInitAttr efa_attr=None):
+        if efa_attr is None:
```
ShacharKagan: EfaCompCntr.__init__ skips super().__init__() (correct, since calling CompCntr.__init__ would re-create the counter), but as a result PyverbsObject.__init__ never runs and self.logger is never assigned. Any later use of the logger will raise AttributeError.

Suggestion: Initialize the parent chain explicitly, bypassing only CompCntr:
PyverbsCM.__init__(self)

pyverbs/providers/efa/efadv.pyx
```diff
@@ -293,6 +296,50 @@ cdef class EfaCQ(CQEX):
...
+            efa_attr = EfaCompCntrInitAttr()
+        self.comp_cntr = dv.efadv_create_comp_cntr(ctx.context, &attr.attr, &efa_attr.attr, sizeof(efa_attr.attr))
+        if self.comp_cntr == NULL:
+            raise PyverbsRDMAErrno('Failed to create EFA comp_cntr')
+        self.ctx = ctx
```
ShacharKagan: call ctx.add_ref(self) here once CompCntr is registered with Context (see the corresponding comment on comp_cntr.pyx)

pyverbs/providers/efa/efadv.pyx
```diff
@@ -293,6 +296,50 @@ cdef class EfaCQ(CQEX):
...
+    :param ctx: Context object
+    :param attr: CompCntrInitAttr object
+    :param efa_attr: EfaCompCntrInitAttr object
+    :return: An initialized EfaCompCntr
+    """
+    def __init__(self, Context ctx not None, CompCntrInitAttr attr not None,
+                 EfaCompCntrInitAttr efa_attr=None):
+        if efa_attr is None:
+            efa_attr = EfaCompCntrInitAttr()
+        self.comp_cntr = dv.efadv_create_comp_cntr(ctx.context, &attr.attr, &efa_attr.attr, sizeof(efa_attr.attr))
```
ShacharKagan: The line is too long. Break it:
self.comp_cntr = dv.efadv_create_comp_cntr(ctx.context, &attr.attr,
                                                                        &efa_attr.attr,
                                                                         sizeof(efa_attr.attr))

pyverbs/qp.pyx
```diff
@@ -1169,6 +1170,18 @@ cdef class QP(PyverbsCM):
...
+        """
+        Attach a completion counter to this QP.
+        :param comp_cntr: The completion counter to attach
+        :param attr: Attach attributes including op_mask
+        """
+        rc = v.ibv_qp_attach_comp_cntr(self.qp, comp_cntr.comp_cntr,
+                                        &attr.attr)
+        if rc != 0:
+            raise PyverbsRDMAError('Failed to attach comp cntr to QP', rc)
+
```
ShacharKagan, line 1261: Track attached QPs on the counter and close them before destroying it.
A CompCntr cannot be destroyed while a QP is still attached to it — the QP must be destroyed first (as described in the man page). The patch attaches QPs to a counter but never tracks the relationship, so an explicit comp_cntr.close() while QPs are still alive will fail with EBUSY.

ShacharKagan (changes requested):
This review covers only the pyverbs commit (pyverbs: Add Completion Counters support). I have not reviewed the libibverbs/efa changes in this PR.

libibverbs/man/ibv_create_comp_cntr.3.md
shefty: I still hold that the max values should be device attributes, not set per counter.

libibverbs/man/ibv_create_comp_cntr.3.md
```diff
@@ -0,0 +1,189 @@
...
+
+struct ibv_comp_cntr_init_attr {
+	uint32_t comp_mask;
+	enum ibv_comp_cntr_type type;
+	uint32_t flags;
+};
+```
+
+*comp_mask*
+:	Bitmask specifying what fields in the structure are valid.
```
shefty, line 116: Add: Used for extensibility.  Should be set to 0.

libibverbs/verbs.h
```diff
@@ -2112,6 +2113,38 @@ struct ibv_cq_init_attr_ex {
 	struct ibv_pd		*parent_domain;
 };
 
+struct ibv_comp_cntr {
+	struct ibv_context *context;
+	uint32_t handle;
+	uint64_t comp_count_max_value;
+	uint64_t err_count_max_value;
```
shefty: Repeating, but these should be device attributes, similar to the max QP, CQ sizes.

libibverbs/verbs.h
```diff
@@ -2112,6 +2113,38 @@ struct ibv_cq_init_attr_ex {
...
+	uint32_t flags;
+};
+
+enum ibv_qp_attach_comp_cntr_op {
+	IBV_QP_ATTACH_COMP_CNTR_OP_SEND			= 1 << 0,
+	IBV_QP_ATTACH_COMP_CNTR_OP_RECV			= 1 << 1,
+	IBV_QP_ATTACH_COMP_CNTR_OP_RDMA_READ		= 1 << 2,
+	IBV_QP_ATTACH_COMP_CNTR_OP_REMOTE_RDMA_READ	= 1 << 3,
+	IBV_QP_ATTACH_COMP_CNTR_OP_RDMA_WRITE		= 1 << 4,
+	IBV_QP_ATTACH_COMP_CNTR_OP_REMOTE_RDMA_WRITE	= 1 << 5,
```
shefty: libfabric calls out counters for remote RDMA operations as a separate feature.  The primary reason is that local counters can be implemented within SW, but counting remote ops needs some level of HW support.  Should there be a device CAP to indicate that counters support remote RDMA?  max_comp_cntr indicates if counters are supported at all.  Under the current proposal, the app won't know if remote RDMA is supported until it tries to create a counter and that fails.

shefty (commented):
My comments are based on examining the API changes.  The overall approach from that level seems reasonable to me, with 2 open comments:

* IMO, the counter size (max value) should be a device attribute and not change per counter.  Counters differ from CQs in that the user doesn't request what size of counter to use.  It gets what's supported by HW.
* Should support for counting remote RDMA operations be exposed as a separate capability?  I believe it should as that has a larger impact on implementation.

libibverbs/verbs.h
mrgolin: Ok NP, I'll move it.

libibverbs/verbs.h
mrgolin: Agree, I'll add supported ops mask so that applications don't need to try and fail.

pyverbs/comp_cntr.pyx
mrgolin: Fixed, thanks.

pyverbs/comp_cntr.pyx
mrgolin: Added the refcounting as you suggested.

pyverbs/comp_cntr.pyx
mrgolin: Done

pyverbs/comp_cntr.pyx
mrgolin: Done

pyverbs/providers/efa/efadv.pyx
mrgolin: Fixed, thanks.

pyverbs/providers/efa/efadv.pyx
mrgolin: Done

pyverbs/qp.pyx
mrgolin: This is already handled in the kernel and will fail with the relevant error code as you mentioned. I don't think we need to have another mechanism here to handle wrong usage.

libibverbs/man/ibv_create_comp_cntr.3.md
mrgolin: Done

libibverbs/verbs.h
mrgolin: Moved to a device level verb.

libibverbs/verbs.h
mrgolin: Now there is ```supported_qp_attach_ops``` that gives the support mask.

mrgolin:
> My comments are based on examining the API changes. The overall approach from that level seems reasonable to me, with 2 open comments:
> 
> * IMO, the counter size (max value) should be a device attribute and not change per counter.  Counters differ from CQs in that the user doesn't request what size of counter to use.  It gets what's supported by HW.
> * Should support for counting remote RDMA operations be exposed as a separate capability?  I believe it should as that has a larger impact on implementation.

@shefty Thanks for your feedback, I've resolved all the comments.

pyverbs/providers/efa/efadv.pyx
mrgolin: Done.

pyverbs/providers/efa/efadv.pyx
mrgolin: Fixed.

mrgolin:
Rebased and updated kernel-headers commit.

pyverbs/providers/efa/efadv.pyx
```diff
@@ -6,12 +6,22 @@ cimport pyverbs.providers.efa.libefa as dv
...
 from pyverbs.libibverbs_enums import ibv_qp_attr_mask
 cimport pyverbs.libibverbs as v
+from pyverbs.mr cimport MR
 from pyverbs.pd cimport PD
 from pyverbs.qp cimport QP, QPEx, QPInitAttr, QPInitAttrEx
-from pyverbs.mr cimport MR
+from libc.string cimport memset
+from libc.stdint cimport uintptr_t, uint8_t
+
+EFADV_COMP_CNTR_INIT_WITH_COMP_EXTERNAL_MEM = 1 << 0
```
ShacharKagan: These enums were already exposed through efa_enums.pxd and used as dve.* in the previous revision. Is there a reason they were replaced with hardcoded constants? Please keep the previous enum-based approach unless this change is intentional. That keeps pyverbs in sync with the header and avoids drift if the values change.

pyverbs/comp_cntr.pyx
```diff
@@ -0,0 +1,97 @@
...
+        return self.attr.comp_mask
+
+    @property
+    def cntr_type(self):
+        return self.attr.type
+
+    @property
+    def flags(self):
+        return self.attr.flags
+
```
ShacharKagan, line 30: three blank lines here- PEP 8 wants two between top-level definitions. Looks like a leftover from moving QPAttachCompCntrAttr to qp.pyx.

pyverbs/libibverbs.pxd
```diff
@@ -756,6 +776,17 @@ cdef extern from 'infiniband/verbs.h':
     unsigned int ibv_wc_read_flow_tag(ibv_cq_ex *cq)
     void ibv_wc_read_tm_info(ibv_cq_ex *cq, ibv_wc_tm_info *tm_info)
     unsigned long ibv_wc_read_completion_wallclock_ns(ibv_cq_ex *cq)
+    int _ibv_query_comp_cntr_caps(ibv_context *context,
```
ShacharKagan: Bind the public ibv_query_comp_cntr_caps() here instead of the internal _ibv_query_comp_cntr_caps(). Cython can call the inline wrapper, and it drops the manual sizeof in device.pyx.

pyverbs/device.pyx
```diff
@@ -366,6 +380,8 @@ cdef class Context(PyverbsCM):
             self.rwq_ind_tbls.add(obj)
         elif isinstance(obj, DMAHandle):
             self.dmahs.add(obj)
+        elif isinstance(obj, CompCntr):
+            self.comp_cntrs.add(obj)
```
ShacharKagan, line 384: Thanks for adding the context reference tracking. One piece is still missing for the full lifecycle to work: self.comp_cntrs should also be added to the close_weakrefs() list in Context.close(). Please place it after self.qps, since attached QPs must be destroyed before their completion counters. Otherwise, completion counters will not be closed through the context cleanup mechanism.

pyverbs/qp.pyx
ShacharKagan: Ack. Leaving it to you. Just noting that without the QP holding a reference to the CompCntr, the counter can be GC'd while a QP is still attached, so __dealloc__() will fail with EBUSY. Not blocking from my side.

ShacharKagan (changes requested):
Thanks for the updates. Left a few inline comments.

pyverbs/providers/efa/efadv.pyx
mrgolin: Good catch, it should be ```efa_enums.pxd```. Moved it there.

pyverbs/comp_cntr.pyx
mrgolin: Fixed.

pyverbs/libibverbs.pxd
mrgolin: Yeah it's better this way, changed it. (for some reason I had issues with this approach when I initially used it)

pyverbs/device.pyx
mrgolin: Missed that, thanks!

mrgolin:
@rleon @jgunthorpe If no more comments, please merge so we can finalize libfabric implementation.

ShacharKagan (commented):
pyverbs fixes look good. LGTM from my side.

YonatanNachum:
Hi @jgunthorpe, @rleon kind reminder for this PR. It blocks our progress until the rdma-core API is locked. I will appreciate your feedback.

Thanks

PR #1710, opened 2026-03-10
Hfi2

ddalessa:
Add support hfi2 and get rid of legacy ipath. The kernel series for this is making it's way to the list now.


rleon:
Strange, i disabled for more than once Copilot and it still reports.

PR #1714, opened 2026-03-16
mlx5: Fix mlx5dv_create_flow to keep legacy ABI

yishaih:
The commit ID mentioned by the Fixes line increased the layout of struct mlx5dv_flow_action_attr from 16 to 24 bytes, breaking ABI for users of the mlx5dv_create_flow() API and caused a SF where more than one action is supplied.

Fix this ABI break, by keeping the layout for the original symbol version of mlx5dv_create_flow() while introducing a new symbol to support a dynamic stride based on the compiled mlx5dv_flow_action_attr size.

Fixes: 1dc3e8d9866e ("mlx5: Add support for bulk flow counters in mlx5dv_create_flow")

### DIFF
--- a/debian/ibverbs-providers.symbols
+++ b/debian/ibverbs-providers.symbols
@@ -33,6 +33,7 @@ libmlx5.so.1 ibverbs-providers #MINVER#
  MLX5_1.23@MLX5_1.23 40
  MLX5_1.24@MLX5_1.24 42
  MLX5_1.25@MLX5_1.25 54
+ MLX5_1.25.1@MLX5_1.25.1 60.2
  mlx5dv_init_obj@MLX5_1.0 13
  mlx5dv_init_obj@MLX5_1.2 15
  mlx5dv_query_device@MLX5_1.0 13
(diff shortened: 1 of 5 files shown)


nmorey:
@rleon @jgunthorpe 
I may have accepted this one too fast (and the related #1715  one for v61 branch)
It causes a build failure:
```
ValueError: Symbol version 'MLX5_1.25' implied by filename 'libmlx5.so.1.25.60.1' not the newest in ELF (['MLX5_1.0', 'MLX5_1.1', 'MLX5_1.2', 'MLX5_1.3', 'MLX5_1.4', 'MLX5_1.5', 'MLX5_1.6', 'MLX5_1.7', 'MLX5_1.8', 'MLX5_1.9', 'MLX5_1.10', 'MLX5_1.11', 'MLX5_1.12', 'MLX5_1.13', 'MLX5_1.14', 'MLX5_1.15', 'MLX5_1.16', 'MLX5_1.17', 'MLX5_1.18', 'MLX5_1.19', 'MLX5_1.20', 'MLX5_1.21', 'MLX5_1.22', 'MLX5_1.23', 'MLX5_1.24', 'MLX5_1.25', 'MLX5_1.25.1'])
```

The lib version cannot be changed to  `1.25.1.${PACKAGE_VERSION}` for obvious reasons.
How do you guys want to handle that ?

Should I just pick the ABI fix and revert the new symbol ?

jgunthorpe:
It would be OK to rename the shared library for this branch I think, though the validation script might not like it and that would have to be disabled. Or it is fine to leave it with the version it has now and just disable that script check

PR #1717, opened 2026-03-17
librdmacm: extend rsocket for Redis, iperf3, memcached and more Linux APIs

BatshevaBlack:
Extend the rsocket implementation in librdmacm so that applications such as Redis, iperf3, and memcached can use rsocket transparently via LD_PRELOAD (librspreload), and so rsocket aligns with more standard Linux socket and I/O behavior.

**Motivation**
The rsocket library did not fully support several POSIX/Linux interfaces (epoll, select, accept4, sendfile, fcntl64, and various socket options). Applications that rely on these either failed or fell back to TCP. This change extends the rsocket implementation to implement or fix those interfaces, so the preload can intercept them and route traffic over RDMA.

**Changes**
fix rpoll timeout handling and select,
fix cm_svc_run
add accept4, dup, fcntl64, sendfile64;
fix rfcntl
extend getsockopt/setsockopt;
fix SOCK_STREAM/SOCK_DGRAM handling and connect service/TCP behavior;
adjust wake-up timeout from rpoll.

### DIFF
--- a/librdmacm/CMakeLists.txt
+++ b/librdmacm/CMakeLists.txt
@@ -27,14 +27,10 @@ target_link_libraries(rdmacm LINK_PRIVATE
 
 # The preload library is a bit special, it needs to be open coded
 # Since it is a LD_PRELOAD it has no soname, and is installed in sub dir
-add_library(rspreload MODULE
+rdma_rspreload_module(rspreload librspreload.map
   preload.c
   indexer.c
   )
-# Even though this is a module we still want to use Wl,--no-undefined
-set_target_properties(rspreload PROPERTIES LINK_FLAGS ${CMAKE_SHARED_LINKER_FLAGS})
-set_target_properties(rspreload PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${BUILD_LIB}")
-rdma_set_library_map(rspreload librspreload.map)
 target_link_libraries(rspreload LINK_PRIVATE
   rdmacm
   ${CMAKE_THREAD_LIBS_INIT}
--- a/librdmacm/preload.c
+++ b/librdmacm/preload.c
@@ -59,6 +59,12 @@
 #include "cma.h"
 #include "indexer.h"
 
+#if RDMA_PRELOAD_WRAP_LFS64
+#if !RDMA_PRELOAD_FCNTL64_IN_HEADER
+int fcntl64(int socket, int cmd, ... /* arg */);
+#endif
+#endif
+
 struct socket_calls {
 	int (*socket)(int domain, int type, int protocol);
 	int (*bind)(int socket, const struct sockaddr *addr, socklen_t addrlen);
@@ -88,6 +94,9 @@ struct socket_calls {
 	int (*getsockopt)(int socket, int level, int optname,
 			  void *optval, socklen_t *optlen);
 	int (*fcntl)(int socket, int cmd, ... /* arg */);
+#if RDMA_PRELOAD_WRAP_LFS64
+	int (*fcntl64)(int socket, int cmd, ... /* arg */);
+#endif
 	int (*dup2)(int oldfd, int newfd);
 	ssize_t (*sendfile)(int out_fd, int in_fd, off_t *offset, size_t count);
 	int (*fxstat)(int ver, int fd, struct stat *buf);
@@ -410,6 +419,9 @@ static void init_preload(void)
 	real.setsockopt = dlsym(RTLD_NEXT, "setsockopt");
 	real.getsockopt = dlsym(RTLD_NEXT, "getsockopt");
 	real.fcntl = dlsym(RTLD_NEXT, "fcntl");
+#if RDMA_PRELOAD_WRAP_LFS64
+	real.fcntl64 = dlsym(RTLD_NEXT, "fcntl64");
+#endif
 	real.dup2 = dlsym(RTLD_NEXT, "dup2");
 	real.sendfile = dlsym(RTLD_NEXT, "sendfile");
 	real.fxstat = dlsym(RTLD_NEXT, "__fxstat");
@@ -647,7 +659,6 @@ int accept4(int socket, struct sockaddr *addr, socklen_t *addrlen, int flags)
 		if (cur_flags == -1)
 			goto close;
 	}
-
 	return fd;
 close:
 	close(fd);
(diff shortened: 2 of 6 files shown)


librdmacm/preload.c
```diff
@@ -873,20 +873,31 @@ ssize_t writev(int socket, const struct iovec *iov, int iovcnt)
...
+	static __thread nfds_t buf_nfds;
+
+	if (nfds > buf_nfds) {
+		free(rfds_buf);
+		free(user_fds_buf);
+		rfds_buf = malloc(sizeof(*rfds_buf) * nfds);
+		user_fds_buf = malloc(sizeof(*user_fds_buf) * nfds);
+		if (!rfds_buf || !user_fds_buf) {
+			free(rfds_buf);
+			free(user_fds_buf);
```
shefty: This patch shows the reason for the prior patch having separate calls.  The poll() path doesn't need the user_fds.  Sorry, with the prior patch, that doesn't show up in the diff, so I missed it.  Please switch back to your prior patch and keep separate calls.

(My intent with this patch, btw, was to have a single malloc that splits between rfds and user_fds.)

librdmacm/preload.c
```diff
@@ -620,6 +622,38 @@ int accept(int socket, struct sockaddr *addr, socklen_t *addrlen)
 	}
 }
 
+int accept4(int socket, struct sockaddr *addr, socklen_t *addrlen, int flags)
+{
+	int cur_flags = 0;
+	int fd = accept(socket, addr, addrlen);
```
shefty: nit: Please don't call functions in the variable declaration section.  Move fd = accept() down, so that it's immediately above the if (fd < 0) check.  That makes it easier to match the check with the function call.

librdmacm/preload.c
```diff
@@ -1222,6 +1224,13 @@ int fcntl64(int socket, int cmd, ... /* arg */)
 }
 #endif
 
+int dup(int oldfd)
+{
+	int new_fd = fcntl(oldfd, F_DUPFD, 0);
```
shefty: Please don't call functions in variable declaration area.  Shouldn't this check for errors?

shefty (commented):
Overall, looks good.  A few minor comments, though one asks to revert a change request I made on the prior review.  :/  My bad.

BatshevaBlack:
@shefty 
I resolved all comments
thanks!

shefty (approved):
All my comments have been addressed.  Thanks!

BatshevaBlack:
@shefty thanks!
Will you merge it? or what is the next step?

rleon:
> @shefty thanks! Will you merge it? or what is the next step?

I will just perform some sanity review and will merge.

Thanks

librdmacm/CMakeLists.txt
```diff
@@ -25,12 +25,69 @@ target_link_libraries(rdmacm LINK_PRIVATE
...
+  (void)out_fd;(void)in_fd;(void)offset64;(void)count; return 0;
+}
+" RDMA_PRELOAD_SENDFILE64_DECLARED_IN_HEADER)
+  if(RDMA_PRELOAD_SENDFILE64_DECLARED_IN_HEADER)
+    set(RDMA_PRELOAD_SENDFILE64_IN_HEADER 1)
+  endif()
+  set(CMAKE_REQUIRED_FLAGS "${SAVE_FLAGS}")
+  set(CMAKE_REQUIRED_QUIET 0)
+endif()
+
```
rleon: All these compilers checks belongs to global CmakeLists.txt.

librdmacm/CMakeLists.txt
```diff
@@ -25,12 +25,69 @@ target_link_libraries(rdmacm LINK_PRIVATE
...
 # The preload library is a bit special, it needs to be open coded
 # Since it is a LD_PRELOAD it has no soname, and is installed in sub dir
 add_library(rspreload MODULE
   preload.c
   indexer.c
   )
+target_compile_definitions(rspreload PRIVATE
```
rleon: If I understood the intent, but it needs to be done with global CMakeList and functions inside buildlib/rdma_functions.cmake.

librdmacm/preload.c
```diff
@@ -59,11 +59,23 @@
 #include "cma.h"
 #include "indexer.h"
 
+//Whether to compile fcntl64/sendfile64 wrappers. Need both:
+#if (!defined(_FILE_OFFSET_BITS) || _FILE_OFFSET_BITS != 64) && \
```
rleon: All these "if" should be handled by one cmake variable.

librdmacm/preload.c
```diff
@@ -87,8 +99,17 @@ struct socket_calls {
 	int (*getsockopt)(int socket, int level, int optname,
 			  void *optval, socklen_t *optlen);
 	int (*fcntl)(int socket, int cmd, ... /* arg */);
+#if (!defined(_FILE_OFFSET_BITS) || _FILE_OFFSET_BITS != 64) && \
+	(!defined(RDMA_PRELOAD_HAVE_64) || RDMA_PRELOAD_HAVE_64)
```
rleon: why do we need this extra "#if ..."?

librdmacm/preload.c
```diff
@@ -1109,6 +1191,60 @@ int fcntl(int socket, int cmd, ... /* arg */)
...
+	case F_GETFD:
+	case F_GETFL:
+	case F_GETOWN:
+	case F_GETSIG:
+	case F_GETLEASE:
+		ret = (fd_get(socket, &fd) == fd_rsocket) ?
+			rfcntl(fd, cmd) : real.fcntl64(fd, cmd);
+		break;
+	case F_DUPFD:
+	/*case F_DUPFD_CLOEXEC:*/
```
rleon, line 1208: ?

librdmacm/rsocket.c
```diff
@@ -3896,6 +3938,17 @@ int rgetsockopt(int socket, int level, int optname,
 					    2048;
 			*optlen = sizeof(int);
 			break;
+		case TCP_INFO:
+			//TODO: support other tcp_info fields.
```
rleon, line 3942: ?

librdmacm/preload.c
BatshevaBlack: We need both: non-LFS vs LFS (_FILE_OFFSET_BITS) and whether libc has fcntl64/sendfile64. I’ll combine those in one CMake flag and use a single #if in preload.c so we don’t duplicate the logic in many places.

librdmacm/preload.c
BatshevaBlack: This is legacy from the original fcntl commit, there’s no documented reason.

librdmacm/rsocket.c
BatshevaBlack: We implemented TCP_INFO incrementally by demand, not as a full TCP stack mirror—that’s why the TODO calls out other fields.

librdmacm/CMakeLists.txt
BatshevaBlack: done

librdmacm/preload.c
BatshevaBlack: done

PR #1722, opened 2026-03-24
ibstat: ignore non-RDMA endpoint devices when umad_get_ca() fails

NagappanRMPL:
Orabug: 38502630

When umad_get_ca() fails, ibstat currently propagates the error even for devices that are not usable RDMA endpoints.

Add lightweight sysfs checks to detect non-endpoint devices by verifying either the presence of port GID 0 (ports/<n>/gids/0) or the Connection Manager capability bit (port cap_mask bit 16). If neither is present, treat the device as non-RDMA and return 0 instead of failing.

This improves behavior on systems exposing non-endpoint IB-class devices under /sys/class/infiniband/.

### DIFF
--- a/infiniband-diags/ibstat.c
+++ b/infiniband-diags/ibstat.c
@@ -43,11 +43,16 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <errno.h>
+#include <dirent.h>
 #include <linux/types.h> /* __be64 */
 
 #include <infiniband/umad.h>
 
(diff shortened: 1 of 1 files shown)


NagappanRMPL:
Hi,

Could you please review the changes for ibstat command.

Thanks,
Nagappan

NagappanRMPL:
Hi @rleon ,

Could you please review this change for ibstat command.

Thanks,
Nagappan.

PR #1725, opened 2026-04-15
rdma_topo: Extend support for PCIe switch-based GPUDirect platforms

EdwardSro:
This series extends rdma_topo to support platforms where a ConnectX NIC and GPU share a PCIe switch without a separate DMA Direct function, in addition to the existing DMA-based topologies.

The first patches prepare the ground by replacing lspci/setpci with direct sysfs and config space parsing, adding a dump/replay mechanism for offline analysis and testing, refactoring the complex representation into an abstract base with per-topology implementations, and improving the check command to report all failures instead of stopping at the first.

The final patch adds inline topology detection, ACS computation, and a --virt/--no-virt flag that controls whether ACS is configured for bare-metal or virtualized environments. When unset, the tool auto-detects based on the NIC's ATS capability.

No behavioral changes for existing Direct NIC (DMA-based) platforms.

### DIFF
--- a/kernel-boot/rdma_topo
+++ b/kernel-boot/rdma_topo
@@ -28,6 +28,9 @@ class CommandError(Exception):
     pass
 
 
+TOPO_NOT_SUPPORTED = CommandError("No supported topology detected")
+
+
 def sysfs_read_str(sysdir: str, fn: str) -> str:
     """Read the entire content of a sysfs file to a string"""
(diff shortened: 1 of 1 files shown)


EdwardSro:
Pushed v2, fixing 'topo' command to always display the topology even if the virtualization could not be auto-detected (whether the user is going to use virtualization or not).

PR #1726, opened 2026-04-17
ionic: Honor sq_sig_all for signaled completions

abhijitG-xlnx:
The sq_sig_all attribute from QP creation was not being propagated to the provider QP, so work requests were only signaled when IBV_SEND_SIGNALED was explicitly set. Store sq_sig_all in the QP and check it in the send path to generate completions for all send work requests when requested.


rleon:
you have an extra blank line between Fixes and SOB, please remove

PR #1727, opened 2026-04-20
Backport SUSE spec file patches

nmorey:


rleon:
@nmorey please add some text to the commit messages so I can merge it.

rleon:
In  second patch, the author's SOB is missing
From: =?UTF-8?q?Dirk=20M=C3=BCller?= <dmueller@suse.com>
Date: Mon, 20 Apr 2026 12:16:31 +0200
Subject: [PATCH 2/2] suse: switch to systemd_requires

Replace direct Requires of systemd/udev by a systemd provided macros.

Signed-off-by: Nicolas Morey <nmorey@suse.com>

nmorey:
Sorry about that. Should be good now

PR #1728, opened 2026-04-20
Rsocket upstream epoll

BatshevaBlack:
Add epoll support for rsocket via `repoll_create`, `repoll_ctl`, and `repoll_wait`.

A single monitor thread per epoll instance uses `rpoll` internally to poll rsocket fds. The flow:

- The thread calls `rpoll` (blocking) on all registered fds (idle, zero CPU)
- When `rpoll` returns with events, the thread goes to sleep
- When the application calls `epoll_wait`, it collects the events, returns them, and wakes the thread
- `epoll_ctl` changes notify the thread via a command pipe

Preload wrappers (`epoll_create`, `epoll_ctl`, `epoll_wait`) are included so existing applications work transparently with `LD_PRELOAD`.


shefty:
The first commit in this series is literally titled "git push --force batsheva rsocket_upstream_epolllibrdmacm: Add support for epoll_wait".  Can you please check the patches are correct with proper titles and commit messages?

BatshevaBlack:
@shefty fixed, sorry for that

librdmacm/rsocket.c
```diff
@@ -68,6 +70,16 @@
...
 #define RS_CONN_RETRIES 6
 #define RS_SGL_SIZE 2
+#define REPOLL_CMD_REFRESH_FDS 1
+#define REPOLL_INIT_FD_CAP 1024
+
+enum repoll_monitor_state {
+	REPOLL_MONITOR_SLEEPING = 0,
+	REPOLL_MONITOR_STARTING = 1,
+	REPOLL_MONITOR_RUNNING = 2,
+	REPOLL_MONITOR_RELOADED = 3,
```
shefty: There's no need to explicitly assigned values.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+
+	ri->epfd = pipefd[0];
+	ri->epfd_peer = pipefd[1];
+	pthread_mutex_init(&ri->lock, NULL);
+	pthread_cond_init(&ri->monitor_wake, NULL);
+
+	if (repoll_setup_cmd_pipe(ri))
+		goto fail_comm;
+
+	if (repoll_start_monitor(ri))
```
shefty: Delay starting the monitor thread until the repoll_info has been fully initialized, so after the idm_set call below.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+				      struct pollfd **fds, int *nfds)
+{
+	free(*fds);
+	*fds = malloc(sizeof(**fds) * ri->capacity);
+	if (!*fds)
+		return -1;
+	memcpy(*fds, ri->fds, sizeof(**fds) * ri->capacity);
+	*nfds = ri->slot_count;
+	return 0;
+}
```
shefty, line 4886: I don't understand why we're making a copy of ri->fds, rather than using ri->fds directly.  If we ever hit capacity, simply grow the array by some chunk amount.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+	}
+
+out:
+	free(fds);
+	return NULL;
+}
+
+static void repoll_stop_monitor(struct repoll_info *ri)
+{
+	atomic_store(&ri->monitor_running, 0);
```
shefty: Use the state names instead of 0

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+			if (repoll_monitor_refresh_fds(ri, &fds, &nfds))
+				break;
+			continue;
+		}
+
+		switch (repoll_monitor_read_cmd(fds)) {
+		case REPOLL_CMD_REFRESH_FDS:
+			if (repoll_monitor_refresh_fds(ri, &fds, &nfds))
+				goto out;
+			repoll_monitor_ack_reload(ri);
```
shefty: The code is maintaining 2 copies of the fds.  One local to the repoll thread and the other part of the repoll_info.  We only need 1 copy.  Pass a command to add, del, ctrl a given fd.  That is write a small cmd message to the pipe, with the needed data.  This thread can then wake-up and process it.  There isn't a need to continually free/malloc a new array of fd's every time we report an event.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+	state = REPOLL_MONITOR_STARTING;
+	atomic_compare_exchange_strong(&ri->monitor_state, &state,
+				       REPOLL_MONITOR_RUNNING);
+
+	while (atomic_load(&ri->monitor_running)) {
+		if (rpoll(fds, nfds, -1) <= 0)
+			continue;
+
+		if (repoll_monitor_has_user_events(fds, nfds)) {
+			if (repoll_monitor_sleep(ri))
```
shefty: This looks racy.  There's no locking around checking has user events and sleeping.  So a user thread could read the events after calling has_user_events but before this thread calls sleep.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+	for (int i = 1; i < nfds; ++i) {
+		if (fds[i].fd > 0 && fds[i].revents)
+			return 1;
+	}
+	return 0;
+}
+
+static int repoll_monitor_sleep(struct repoll_info *ri)
+{
+	sched_yield();
```
shefty: Why is this here?

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+	atomic_store(&ri->monitor_state, REPOLL_MONITOR_SLEEPING);
+	pthread_mutex_lock(&ri->lock);
+	while (atomic_load(&ri->monitor_state) == REPOLL_MONITOR_SLEEPING &&
+	       atomic_load(&ri->monitor_running))
+		pthread_cond_wait(&ri->monitor_wake, &ri->lock);
+	pthread_mutex_unlock(&ri->lock);
+	return atomic_load(&ri->monitor_running) ? 0 : -1;
```
shefty: This check is racy, so it gives a false impression of behavior.  Immediately after this check is made, monitor_running could be set to false.  The code calling monitor_sleep() can't reliably us the return code to mean anything.  I would convert this function to a void and remove this check.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4769,229 @@ static void *cm_svc_run(void *arg)
...
+	pthread_cond_t monitor_wake;
+	struct index_map fd_index;
+	struct pollfd *fds;
+	union epoll_data *user_data;
+	uint16_t slot_count;
+	uint16_t active_fds;
+	uint16_t capacity;
+
+	_Atomic(int) monitor_state;
+	_Atomic(int) monitor_running;
```
shefty: You have 2 separate states here, which is confusing.  Especially since monitor_state can be MONITOR_RUNNING, but there's also a separate monitor_running state.  Please merge into a single state variable

librdmacm/rsocket.c
```diff
@@ -4995,3 +4995,178 @@ int repoll_create(int flags)
...
+	if (epoll_events & EPOLLHUP)
+		poll_events |= POLLHUP;
+	return poll_events;
+}
+
+static int repoll_grow_fds(struct repoll_info *ri)
+{
+	struct pollfd *new_fds;
+	union epoll_data *new_data;
+	int new_max = ri->capacity * 2;
```
shefty: doubling each time is probably a little much, just increase by a fixed size

librdmacm/rsocket.c
```diff
@@ -4995,3 +4995,178 @@ int repoll_create(int flags)
...
+
+	new_fds = malloc(sizeof(*new_fds) * new_max);
+	if (!new_fds)
+		return -ENOMEM;
+	for (int i = 0; i < new_max; ++i)
+		new_fds[i].fd = -1;
```
shefty: i can start at the previous capacity, since the memcpy below is going to immediately write over the values.

librdmacm/rsocket.c
```diff
@@ -4995,3 +4995,178 @@ int repoll_create(int flags)
...
+	return poll_events;
+}
+
+static int repoll_grow_fds(struct repoll_info *ri)
+{
+	struct pollfd *new_fds;
+	union epoll_data *new_data;
+	int new_max = ri->capacity * 2;
+
+	new_fds = malloc(sizeof(*new_fds) * new_max);
```
shefty: I would use calloc to be safer

librdmacm/rsocket.c
```diff
@@ -4995,3 +4995,178 @@ int repoll_create(int flags)
...
+	if (!new_fds)
+		return -ENOMEM;
+	for (int i = 0; i < new_max; ++i)
+		new_fds[i].fd = -1;
+	memcpy(new_fds, ri->fds, sizeof(*new_fds) * ri->capacity);
+	free(ri->fds);
+	ri->fds = new_fds;
+
+	new_data = malloc(sizeof(*new_data) * new_max);
```
shefty: use calloc

librdmacm/rsocket.c
```diff
@@ -4995,3 +4995,178 @@ int repoll_create(int flags)
...
+	if (ri->active_fds >= ri->capacity) {
+		if (repoll_grow_fds(ri))
+			return -1;
+		return ri->slot_count++;
+	}
+	if (ri->active_fds == ri->slot_count + 1)
+		return ri->slot_count++;
+
+	for (int i = 0; i < ri->slot_count; ++i) {
+		if (ri->fds[i].fd <= 0)
```
shefty: This check should be fd == -1.  But we shouldn't need to search for a free slot.  Always add new entries to the end (active_fds).  When removing an entry, swap the current one at the end to the location of the removed fd and decrement active_fds.  We don't need to maintain order on the fds.

librdmacm/rsocket.c
```diff
@@ -4995,3 +4995,178 @@ int repoll_create(int flags)
...
+
+	if (atomic_load(&ri->monitor_state) == REPOLL_MONITOR_SLEEPING)
+		return;
+
+	size = write(ri->cmd_pipe[1], &cmd, sizeof(cmd));
+	(void)size;
+	state = REPOLL_MONITOR_RELOADED;
+	while (atomic_load(&ri->monitor_state) != REPOLL_MONITOR_SLEEPING &&
+	       !atomic_compare_exchange_weak(&ri->monitor_state, &state,
+					     REPOLL_MONITOR_RUNNING))
```
shefty: MONITOR_RELOADED shouldn't really be a state of the monitor.  Reloading is a command.

I think the states are ending up racy.  We have threads on both sides of the monitor trying to direct the state without some higher-level coordination.  The monitor state should really be driven entirely by the monitor thread.  Threads on this side should basically be communicating by writing commands to the pipe.  The only time the threads should worry about the state is during startup and shutdown.

shefty (commented):
I didn't get to epoll_wait, but see comments on how we might restructure the fds allocation and monitor state.  I'm particularly concerned that the states are racy in a way that will be difficult to debug.

librdmacm/rsocket.c
BatshevaBlack: We keep a private snapshot for the monitor on purpose.
The monitor does a blocking rpoll(), and for performance we don’t want to hold ri->lock across that call. Holding the lock there would serialize repoll_ctl/repoll_wait and hurt the hot path.
So repoll_ctl updates ri->fds under the lock, then asks the monitor to reload; the monitor refreshes its local copy and polls lock-free. This avoids lock contention and also avoids races with array growth/reallocation.

librdmacm/rsocket.c
BatshevaBlack: The check is on the monitor’s private fds snapshot (the copy I mentioned in the other thread), so this is not a shared-array data race.

There is a handoff overlap with repoll_wait, but correctness does not rely on strict locking in that window: repoll_wait always polls directly, and on timeout it wakes the monitor if it is sleeping.

I kept this path lock-free intentionally to avoid extra contention on the hot path.

librdmacm/rsocket.c
BatshevaBlack: I added sched_yield() here after I've done some perf testing of the sleep/wake handoff path.
Without it, the monitor and caller contended more around the transition to sleep and throughput dropped, with it, throughput was good. Since it gives repoll_wait (which the app waits on) a chance to run before the monitor parks on the condvar, reducing hot-path contention.

librdmacm/rsocket.c
BatshevaBlack: addressed thanks

librdmacm/rsocket.c
BatshevaBlack: addressed thanks

librdmacm/rsocket.c
BatshevaBlack: partly addressed- commented abut the copy earlier

librdmacm/rsocket.c
BatshevaBlack: Agreed. The return value from repoll_monitor_sleep() is only a transient snapshot and not a reliable state contract. I will simplify this by making it void and relying on the outer monitor_running loop condition for lifecycle control.

librdmacm/rsocket.c
BatshevaBlack: I agree the old naming was confusing.
I kept two atomics, but renamed the lifecycle one to make the split explicit: one state is for sleep/reload handshake, and the other is for monitor thread lifetime.
This keeps teardown separate from the reload CAS flow and is easier to reason about.

librdmacm/rsocket.c
BatshevaBlack: done

librdmacm/rsocket.c
BatshevaBlack: Thanks - agreed that RELOADED is better modeled as a command than a monitor state.
I tested such a change but it introduced a large throughput regression, so the downside appears to be synchronization/wake-path overhead.

I thought about it and in practice, with the current approach, the remaining race risk seems bounded: epoll_wait re-polls anyway, the monitor is just to drive the events further, so impact is mostly occasional latency rather than functional breakage. Do you agree?

librdmacm/rsocket.c
```diff
@@ -4757,3 +4775,508 @@ static void *cm_svc_run(void *arg)
...
+	}
+	local_fds = malloc(sizeof(*local_fds) * nfds);
+	if (!local_fds) {
+		pthread_mutex_unlock(&ri->lock);
+		return ENOMEM;
+	}
+	memcpy(local_fds, ri->fds + 1, sizeof(*local_fds) * nfds);
+	pthread_mutex_unlock(&ri->lock);
+
+	ret = rpoll(local_fds, nfds, timeout);
```
shefty: I think it will be easier to discuss the changes if we focus on specific aspects of API correctness.  I don't want to focus on a single application use case.

epoll_wait() may be called with a timeout of 0 or non-zero.  For a timeout of 0, we need something like this:

```
if (!timeout) {
    lock(ri->lock);
    rpoll(ri->fds, ri->nfds, 0);
    repoll_collect_events(ri, events, maxevents);
    unlock(ri->lock);
}
```
I'm holding the lock around rpoll() and collect_events() to protect against changes to the ri->fds array.

The problem is handling a non-zero timeout, especially if it's infinite.  The !timeout flow above doesn't work because it blocks changes to the epoll fd list (ri->fds).  Making a copy of ri->fds (current patch) also doesn't work because the copy is not updated with changes to ri->fds.  That is, if another thread wants to delete an fd or add a new one, the thread in repoll_wait() will not see those changes.  This is a behavior difference from epoll.

Fixing that behavior requires reworking the design.  This is why the libfabric implementation uses signaling from the epoll_ctrl calls to unblock a waiting thread: 

https://github.com/ofiwg/libfabric/blob/main/src/common.c#L1727

In that implementation, all changes to the epoll fd array are done by whichever thread is calling epoll_wait.  It doesn't have a separate epoll 'progress' thread.

To align with the general design being proposed here, I think a flow like this might work:

```
do {
    /* always go through !timeout flow for fast check */
    lock(ri->lock);
    rpoll(ri->fds, ri->nfds, 0);
    repoll_collect_events(ri, events, maxevents);
    unlock(ri->lock);

    if (!timeout || we have events to return)
        return;

    wakeup_monitor(ri);
    wait on signal from monitor with for timeout duration
    on wake-up, adjust timeout by time spent waiting
} while (1);
```

The monitor thread will call rpoll until it has events to report.  Once it finds any events, it signals any waiting thread to wakeup and recheck.

Changes to the fd's being monitored (epoll_ctl calls) should be processed by the monitor thread, similar to the code linked to above.  This ensures that the fd's are monitored correctly.

librdmacm/rsocket.c
```diff
@@ -68,12 +70,31 @@
...
+
+	return slot ? (int)(uintptr_t)slot : -1;
+}
+
+static inline int repoll_fd_to_slot_set(struct index_map *fd_to_slot, int fd,
+					int slot)
+{
+	return idm_set(fd_to_slot, fd, (void *)(uintptr_t)slot);
+}
+
```
shefty: nit: Please declare functions further down in the file where the other function declarations start.

librdmacm/rsocket.c
```diff
@@ -317,7 +338,7 @@ struct ds_qp {
 };
 
 struct rsocket {
-	int		  type;  /* SOCK_STREAM or SOCK_DGRAM only; flags in fd_flags/fs_flags */
+	int		  type;  /* SOCK_STREAM, SOCK_DGRAM, or FD_EPOLL; must be first */
```
shefty: The type value for a struct rsocket may only be SOCK_STREAM or SOCK_DGRAM.  Remove reference to FD_EPOLL, as that's not valid for struct rsocket.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+	struct pollfd *fds;
+	union epoll_data *user_data;
+	uint16_t slot_count;
+	uint16_t active_fds;
+	uint16_t capacity;
+	uint64_t monitor_seq;
+	uint64_t wait_seq;
+	unsigned int waiters;
+	bool running;
+};
```
shefty: Move structure definition to the top of the file where other structures are defined.  This makes finding the definition much easier.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+	return ri;
+}
+
+static void repoll_wake_monitor(struct repoll_info *ri)
+{
+	uint64_t val = 1;
+	ssize_t size;
+
+	size = write(ri->event_fd, &val, sizeof(val));
+	(void)size;
```
shefty: You should be able to just do ```(void) write(...)```

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+}
+
+static bool repoll_monitor_drain_event(struct pollfd *fds)
+{
+	uint64_t val;
+	ssize_t size;
+
+	if (!(fds[0].revents & POLLIN))
+		return false;
+	size = read(fds[0].fd, &val, sizeof(val));
```
shefty: Should be able to do ```(void) read(...)```

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+		abs_timeout->tv_nsec -= 1000000000L;
+	}
+	return 0;
+}
+
+static int repoll_collect_events(struct repoll_info *ri,
+				 struct pollfd *polled_fds, int polled_nfds,
+				 struct epoll_event *events, int maxevents)
+{
+	int j = 0;
```
shefty: Please follow the conventions of this codebase and declare variables at the top of functions (i and slot).  It would help to rename 'j' to 'cnt' or 'event_cnt' to make it easier to identify its purpose.  i or j are fine for simple iterators, but in this case, j is not the main for loop iterator.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+
+	for (int i = 0; j < maxevents && i < polled_nfds; ++i) {
+		int slot = i + 1;
```
shefty: You can break the loop if i >= ri->slot_count - 1.  The slot check in the if statement below can be removed, with the loop exited, rather than continuing to iterate.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+
+		if (polled_fds[i].revents && slot < ri->slot_count &&
+		    ri->fds[slot].fd == polled_fds[i].fd) {
+			memcpy(&events[j].data, &ri->user_data[slot],
+			       sizeof(events->data));
```
shefty: This can be a simple assignment instead of a memcpy.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+			events[j].events =
+				poll_to_epoll_events(polled_fds[i].revents);
+			++j;
```
shefty: You can move the check for j >= maxevents here, then break from the loop.  We only need to recheck j after it has changed.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+		}
+
+		ri->waiters++;
+		ri->monitor_seq++;
+		pthread_cond_signal(&ri->monitor_cond);
+
+		my_wait_seq = ri->wait_seq;
+		while (my_wait_seq == ri->wait_seq && ri->running) {
+			if (have_deadline) {
+				struct timespec now;
```
shefty: nit: declare at top of function

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+	int event_fd;
+	pthread_t monitor;
+	pthread_mutex_t lock;
+	pthread_cond_t monitor_cond;
+	pthread_cond_t wait_cond;
+	struct index_map fd_to_slot;
+	struct pollfd *fds;
+	union epoll_data *user_data;
+	uint16_t slot_count;
+	uint16_t active_fds;
```
shefty: active_fds is always the same value as slot_count and not used anywhere.  It can be removed.

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+	int last;
+
+	slot = repoll_fd_slot(&ri->fd_to_slot, fd);
+	if (slot < 0)
+		return ENOENT;
+	last = ri->slot_count - 1;
+
+	if (slot != last) {
+		moved_fd = ri->fds[last].fd;
+		if (repoll_fd_slot(&ri->fd_to_slot, moved_fd) < 0)
```
shefty: This check doesn't seem like a user error.  It looks like it can only occur if the internal structures were corrupted or there's a bug in the code.  Should this be an assert?

librdmacm/rsocket.c
```diff
@@ -4757,3 +4793,574 @@ static void *cm_svc_run(void *arg)
...
+
+		for (;;) {
+			unsigned int wait_cnt;
+			int refresh_ret = 0;
+
+			pthread_mutex_lock(&ri->lock);
+			wait_cnt = ri->waiters;
+			if (wait_cnt)
+				refresh_ret = repoll_monitor_refresh_fds(
+						ri, &fds, &fds_cap, &nfds);
```
shefty, line 4936: This will always refresh the fd list, even if a refresh isn't needed.

I would add a 'snapshot' counter to the repoll_info.  Increment the counter on every call to repoll_ctl.  The monitor thread can maintain a local copy of snapshot.  When the values differ, the monitor can refresh the fds and sync snapshot to the latest value.

This can be added as a separate optimization patch.

shefty (commented):
Please see comments.  Overall, I think this is looking good.  Nice work!  I had a few nits to align changes with the rest of the codebase.  There are a couple of optimizations which could easily be made, which are called out in the comments.

I only looked at the combined changes, not individual patches.

librdmacm/rsocket.c
BatshevaBlack: Tried that, but write is __attribute__((warn_unused_result)) under glibc's fortification and GCC ignores the (void) cast, so it fails the project's -Werror=unused-result (it broke CI). Changed it to check the return value instead.

librdmacm/rsocket.c
shefty: Btw, I can't resolve comments (probably a permission restriction), but I consider this resolved.

librdmacm/rsocket.c
shefty: Thread also resolved for me

shefty (commented):
The end code looks okay to me, and I'm good with the end results.

Note that I did not check the changes patch by patch, but my guess is that the patches will need some rework prior to merging.  For example, the last patch to fix build warnings should have those changes integrated directly into the patches where the warnings are introduced.

Basically, there should not be a patch in this series that applies a bug fix to any earlier patch in the series.  This is something the rdma-core maintainers can look at as well.

PR #1732, opened 2026-04-29
Enable ST64B for writes on AArch64

yishaih:
This series adds support for the ARMv8.7 ST64B instruction (FEAT_LS64) in mmio_memcpy_x64 and enables it for mlx5 BlueFlame
doorbell copies.

Series overview:
Patch 1 extends util/mmio to detect and use ST64B on AArch64:
  - A cmake probe (RDMA_Check_LS64) detects compiler LS64 intrinsic
    support at build time and sets LS64_FLAGS / HAVE_LS64.
  - A constructor function checks AT_HWCAP3 at process start to confirm
    the CPU implements FEAT_LS64 at runtime.
  - mmio_memcpy_x64 gains an ST64B fast path when both conditions are
    met, falling back to the existing NEON path otherwise. Files not
    compiled with -march=armv8-a+ls64 see zero overhead and continue to
    use the NEON path.

Patch 2 opts mlx5 qp.c into the new path by compiling it with LS64_FLAGS.

Further details exist as part of the commit logs.

### DIFF
--- a/util/mmio.c
+++ b/util/mmio.c
@@ -5,9 +5,13 @@
 
 #include <pthread.h>
 #include <stdbool.h>
+#include <stdlib.h>
+
+#if (defined(__aarch64__) && defined(HAVE_LS64)) || defined(__s390x__)
+#include <sys/auxv.h>
+#endif
 
 #ifdef __s390x__
-#include <sys/auxv.h>
 #include <ccan/minmax.h>
 
 bool s390_is_mio_supported;
@@ -72,6 +76,25 @@ mmio_memcpy_x64_fn_t resolve_mmio_memcpy_x64(uint64_t hwcap)
 		return &s390_mmio_write_syscall;
 }
 #endif /* __s390x__ */
+
+#if defined(__aarch64__) && defined(HAVE_LS64)
+
+/* FEAT_LS64 (ST64B) is exposed as bit 3 of AT_HWCAP3. */
+#ifndef AT_HWCAP3
+#define AT_HWCAP3  29
+#endif
+#ifndef HWCAP3_LS64
+#define HWCAP3_LS64  (1UL << 3)
+#endif
+
+bool aarch64_has_ls64;
+
+static __attribute__((constructor)) void init_aarch64_caps(void)
+{
+	aarch64_has_ls64 = getauxval(AT_HWCAP3) & HWCAP3_LS64;
+}
+
+#endif /* __aarch64__ && HAVE_LS64 */
 
 #if SIZEOF_LONG != 8
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -496,6 +496,7 @@ if (NOT HAVE_SPARSE)
 endif()
 
 RDMA_Check_SSE(HAVE_TARGET_SSE)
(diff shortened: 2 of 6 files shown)


util/mmio.c
```diff
@@ -73,6 +77,25 @@ mmio_memcpy_x64_fn_t resolve_mmio_memcpy_x64(uint64_t hwcap)
...
+#ifndef AT_HWCAP3
+#define AT_HWCAP3  29
+#endif
+#ifndef HWCAP3_LS64
+#define HWCAP3_LS64  (1UL << 3)
+#endif
+
+bool aarch64_has_ls64;
+
+static __attribute__((constructor)) void init_aarch64_caps(void)
```
rleon, line 92: Can you please remind me how this initialization function is called?

yishaih: It's  called automatically at library load time — as of its  __attribute__((constructor)) .

rleon: Not really, it is called before application main() and not for library initialization.
It means that you should check if aarch64_has_ls64 is already initialized and bailed out.

if (aarch64_has_ls64)
   return;

aarch64_has_ls64 = getauxval....


Note, without !!. It is useless for setting boolean variable.

gurmichael: All constructors of a .so are called when it's dynamically loaded.
mmio.c through our build system is baked into each provider's .so.
Then when libibverbs calls dlopen() of libmlx5 all of the constructors are ran.

gurmichael: !! has no effect here but I've added it both for explicitness and consistency with other getauxval() assignment to bool in the same file.
If you prefer, I can remove it.

rleon: > All constructors of a .so are called when it's dynamically loaded. mmio.c through our build system is baked into each provider's .so. Then when libibverbs calls dlopen() of libmlx5 all of the constructors are ran.

How so? I afraid that the last sentence is not correct. Try to read the origin:
https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Function-Attributes.html

rleon: > !! has no effect here but I've added it both for explicitness and consistency with other getauxval() assignment to bool in the same file. If you prefer, I can remove it.

yes, please

gurmichael: I'll drop the !! and push new PR.

regarding the constructor, on .so it's not running before main(), but on on dlopen().
see [dlopen man page](https://man7.org/linux/man-pages/man3/dlopen.3.html) "Initialization and finalization functions"
For rdma-core this means it is running inside the ibv_open_device() when loading provider's .so.

gurmichael: > It means that you should check if aarch64_has_ls64 is already initialized and bailed out.
>if (aarch64_has_ls64)
>return;

As dlopen man page states:
>If the same shared object is opened again with dlopen(), the same
>object handle is returned.  The dynamic linker maintains reference
>counts for object handles, so a dynamically loaded shared object
>is not deallocated until dlclose() has been called on it as many
>times as dlopen() has succeeded on it.  Constructors (see below)
>are called only when the object is actually loaded into memory
>(i.e., when the reference count increases to 1).

No flow for 'already initialized'

PR #1733, opened 2026-04-30
 ionic: Add Reorder Completion Queue (RCQ) support

abhijitG-xlnx:
Add RCQ support to the ionic provider for out-of-order receive completion handling.
Patch 1: Update kernel headers for RCQ fields.
Patch 2: Fetch RCQ capabilities on context create.
Patch 3: Add ionic_dv_create_qp_ex DV API.
Patch 4: Process RECV_RCQE completions from poll_cq, delivering them in order.

### DIFF
--- a/providers/ionic/ionic_verbs.c
+++ b/providers/ionic/ionic_verbs.c
@@ -1330,7 +1330,7 @@ static int ionic_poll_vcq_cq(struct ionic_ctx *ctx, struct ionic_cq *cq,
 
 			break;
 
-		case IONIC_V1_CQE_TYPE_RECV_INDIR:
+		case IONIC_V1_CQE_TYPE_RECV_RCQE:
 			list_del(&qp->cq_poll_rq);
 			list_add_tail(&cq->poll_rq, &qp->cq_poll_rq);
 			break;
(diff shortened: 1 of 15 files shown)


providers/ionic/ionic_verbs.c
```diff
@@ -1683,6 +1689,30 @@ static struct ibv_qp *ionic_create_qp_ex(struct ibv_context *ibctx,
...
+				goto err_qp;
+			}
+
+			if ((ionic_ex->ionic_flags & IONIC_DV_CREATE_QP_RCCL_RCQ) &&
+			     !ctx->rcq_sign_bit) {
+				rc = EINVAL;
+				goto err_qp;
+			}
+
+			IONIC_SET_UDATA(ctx, &req, ionic_flags, ionic_ex->ionic_flags);
```
jgunthorpe: why does this invisibly set the flag? I would expect this function to check robust_udata and then fail if any input combination requires this to be non-zero?

providers/ionic/libionic.map
```diff
@@ -10,3 +10,8 @@ IONIC_1.0 {
 		ionic_dv_pd_set_rqcmb;
 	local: *;
 };
+
+IONIC_1.1 {
+	global:
+		ionic_dv_create_qp_ex;
```
jgunthorpe, line 16: Needs a man page

jgunthorpe (changes requested):

providers/ionic/ionic_verbs.c
abhijitG-xlnx: Fixed. IONIC_SET_UDATA now returns ENOTSUP when robust_udata is false.
I was banking on `robust_udata` in the same rcq series. As it is merged before, added comp_mask to check supported feature.

providers/ionic/libionic.map
abhijitG-xlnx: Added manpage

abhijitG-xlnx:
^ rebased

PR #1738, opened 2026-05-13
bnxt_re/lib: Bug fixes

selvintxavier:
Some generic bug fixes in bnxt_re provider.

### DIFF
--- a/providers/bnxt_re/verbs.c
+++ b/providers/bnxt_re/verbs.c
@@ -975,6 +975,7 @@ static int bnxt_re_poll_flush_wcs(struct bnxt_re_joint_queue *jqq,
 		}
 
 		ibvwc->status = IBV_WC_WR_FLUSH_ERR;
+		ibvwc->vendor_err = 0;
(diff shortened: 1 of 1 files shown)


nmorey:
@selvintxavier  The fix CQE flush fixes an issue all the way back to v50.
It doesn't happen cleanly and it's not an obvious conflict resolution.

Would you mind making a PR with the backported patch against the stable-v50 branch ?
 I'd rather have you write it correctly than me messing up the conflict :)

If you don't want it backported there, just let me know

selvintxavier:
> @selvintxavier The fix CQE flush fixes an issue all the way back to v50. It doesn't happen cleanly and it's not an obvious conflict resolution.
> 
> Would you mind making a PR with the backported patch against the stable-v50 branch ? I'd rather have you write it correctly than me messing up the conflict :)
> 
> If you don't want it backported there, just let me know

@nmorey thank you for checking. I will prepare a PR against stable-v50 and share it with you. Thanks

PR #1741, opened 2026-05-18
kernel-boot: Add naming support for sub-function (SF) RDMA devices

jpirko:
PCI Sub-function (SF) is anchored on the auxiliary bus rather than directly on the PCI bus, so by_pci() previously rejected them as "Non-PCI" and they fell back to their kernel-assigned name.

Detect this case by treating an "auxiliary" device subsystem as a valid parent: read the stable 'sfnum' attribute from the aux device and follow its 'device' symlink up to the underlying PCI BDF. The PCI parent can be a PF or an SR-IOV VF; feeding the BDF into the existing get_virtfn_info() / fill_pci_info() path handles both layouts uniformly. The PCI-derived portion of the name is composed unchanged; an S<sfnum> suffix is then appended last, after any f<func>/v<vf> components, since SF identity is independent of them.

Examples:
  SF on a multi-function PF:
    parent 0000:c1:00.0, sfnum 88 -> rocep193s0f0S88
  SF on an SR-IOV VF (VF-SF):
    parent 0000:c1:00.4, sfnum 99 -> rocep193s0f0v0S99

Also update Documentation/udev.md and the rdma-persistent-naming.rules header comment with the new S<sfnum> suffix.

### DIFF
--- a/Documentation/udev.md
+++ b/Documentation/udev.md
@@ -190,6 +190,7 @@ Type of names:
  * s<slot>[f<function>] - hotplug slot index number
  * x<GUID> - Node GUID
  * [P<domain>]p<bus>s<slot>[f<function>] - PCI geographical location
+ * S<sfnum> - sub-function number, appended to the parent PCI device name
 
 Notes:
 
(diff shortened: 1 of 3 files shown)


rleon:
We are following netdev naming scheme.
How is it named in netdev?

jpirko:
> We are following netdev naming scheme. How is it named in netdev?

The same:
https://github.com/systemd/systemd/pull/42154

PR #1742, opened 2026-05-19
libhns: Fix UD 0 msg length err using new post send

hginjgerx:
When msg length is 0, the sge num should also be 0, otherwise there will be a cqe error.

Fixes: 36446a56eea5 ("libhns: Extended QP supports the new post send mechanism")

### DIFF
--- a/providers/hns/hns_roce_u_hw_v2.c
+++ b/providers/hns/hns_roce_u_hw_v2.c
@@ -2517,9 +2517,12 @@ static void wr_set_sge_ud(struct ibv_qp_ex *ibv_qp, uint32_t lkey,
 		return;
 
 	wqe->msg_len = htole32(length);
-	hr_reg_write(wqe, UDWQE_SGE_NUM, 1);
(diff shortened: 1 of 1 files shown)


providers/hns/hns_roce_u_hw_v2.c
```diff
@@ -2517,10 +2517,13 @@ static void wr_set_sge_ud(struct ibv_qp_ex *ibv_qp, uint32_t lkey,
 		return;
 
 	wqe->msg_len = htole32(length);
-	hr_reg_write(wqe, UDWQE_SGE_NUM, 1);
+	hr_reg_write(wqe, UDWQE_SGE_NUM, !!length);
```
rleon, line 2520: And what does it mean to write length == 0? You are not advancing start_idx and next write will  use same idx.

hginjgerx: This is a HW constraint - sge_num in WQE must be 0 when length == 0, and start_idx must always be the latest idx, even if this WQE is not using any sge.

PR #1746, opened 2026-05-29
librdmacm/preload.c: fix duplicate symbol definitions on armhf

hector-cao:
85f9747 introduces the preload of fcntl64 and it fails to build on armhf:

Assembler messages:
 Error: symbol `__fcntl_time64' is already defined
 Error: symbol `sendfile64' is already defined

On armhf, glibc defines both fcntl64 and fcntl as alias to __fcntl_time64. As a consequence, in preload.c, 2 function definitions of fcntl64 and fcntl are preprocessed to become the same function __fcntl_time64 definition, that cause the double symbol error.

Only define and intercept these functions if they are not already defined.

This fixes the issue both for sendfile64 and fcntl64.


rleon:
Please add Fixes line above your Signed-off-by:

Fixes: 85f97470ebb3 ("librdmacm: Add support for fcntl64")

hector-cao:
> Fixes: [85f9747](https://github.com/linux-rdma/rdma-core/commit/85f97470ebb341c10eb20ec9007080a01a90bdce) ("librdmacm: Add support for fcntl64")

done, thanks

PR #1748, opened 2026-06-02
Add buffer allocation support for Confidential Computing (CoCo) guests

jpirko:
## Summary

In a Confidential Computing (CoCo) guest, guest memory is encrypted, and a
device that cannot access encrypted memory requires DMA bounce buffering. For
RDMA to work in such guests, the DMA buffers (both provider-internal buffers
and user data buffers) must live in unprotected/shared (decrypted) memory.

This PR adds the infrastructure to allocate such buffers from a `system_cc_shared`
DMA-buf heap and wires it into the mlx5 provider. Using shared memory is
strictly opt-in and only takes effect when the device reports that it needs
DMA bounce; otherwise normal memory is used and behavior is unchanged.

## libibverbs core

- `struct ibv_buf`: a common buffer descriptor (pd, addr, size, optional
  dmabuf fd) that providers embed in their internal buffer structures, with
  `ibv_buf_init()`/`ibv_buf_init_dmabuf()` helpers.
- An internal DMA-buf heap allocator (`libibverbs/dmabuf_heap.c`) backed by the
  `system_cc_shared` heap.
- `IBV_DEVICE_CC_DMA_BOUNCE` device capability flag, reported by the kernel when
  the device is in a CoCo guest and requires DMA bounce.
- `IBV_PARENT_DOMAIN_INIT_ATTR_ALLOW_CC_UNPROTECTED_ALLOC` parent-domain flag for
  applications to opt-in to unprotected/shared allocation.
- New provider-aware buffer verbs `ibv_alloc_buf()`, `ibv_free_buf()` and
  `ibv_reg_buf_mr()`, plus the backing `alloc_buf`/`free_buf` provider ops.
  `ibv_reg_buf_mr()` is a thin wrapper over `ibv_reg_mr_ex()` (new
  `IBV_REG_MR_MASK_BUF` path).
- Ordinary addr-based registration falls back to the legacy write ABI, so
  `ibv_reg_buf_mr()`/`ibv_reg_mr_ex()` keep working on kernels that lack the
  REG_MR ioctl method.
- Per-buffer dmabuf UMEM plumbing: `ibv_cmd_create_qp_ex2()` gains a driver
  attribute chain, plus a per-buffer UMEM helper (`fill_attr_in_buf_umem()`).
- man pages: new `ibv_alloc_buf(3)`; updates to `ibv_reg_mr`,
  `ibv_query_device_ex`, `ibv_alloc_parent_domain` and `ibv_rc_pingpong`.

## mlx5 provider

- Adopt `struct ibv_buf` for internal buffer allocations.
- When a parent domain is created with `ALLOW_CC_UNPROTECTED_ALLOC` and the
  device reports `CC_DMA_BOUNCE`, allocate all provider-internal buffers
  (CQ, QP, SRQ, RWQ, doorbell records) from the dmabuf heap through the existing
  preferred-allocation path.
- Pass per-buffer dmabuf UMEM descriptors to the kernel on QP/CQ creation.
- Implement the `ibv_alloc_buf`/`ibv_free_buf` ops.

## pyverbs

- New `Buf` (`ibv_alloc_buf()`/`ibv_free_buf()`), `BufMR` (`ibv_reg_buf_mr()`)
  and `MREx` (`ibv_reg_mr_ex()`) classes.
- `ParentDomain` support for `ALLOW_CC_UNPROTECTED_ALLOC`.
- Enum bindings for `IBV_DEVICE_CC_DMA_BOUNCE` and the parent-domain flag.

## Example

- `rc_pingpong`: add `-U`/`--allow-cc-unprotected` to allocate the data buffer
  and MR from unprotected/shared memory via a parent domain and the new
  `ibv_alloc_buf()`/`ibv_free_buf()`/`ibv_reg_buf_mr()` helpers.

## Tests

- `tests/test_buf.py`: pyverbs coverage for the buffer API over a plain PD and a
  CC parent domain -- single-node API tests (`ibv_reg_buf_mr()` and the
  `ibv_reg_mr_ex()` MASK_BUF path) plus RC/UD traffic.

## Notes

- Kernel headers are refreshed in this series to pull in the matching UAPI
  (device cap flag, REG_MR/parent-domain/UMEM ioctl attributes); the PR depends
  on that kernel support.
- Adds new public symbols in a new `IBVERBS_1.17` version node (`ibv_alloc_buf`,
  `ibv_free_buf`, `ibv_reg_buf_mr`).


libibverbs/man/ibv_query_device_ex.3
```diff
@@ -160,6 +160,15 @@ This feature can be enabled on a QP or WQ basis via the
 IBV_QP_CREATE_PCI_WRITE_END_PADDING or IBV_WQ_FLAGS_PCI_WRITE_END_PADDING
 flags.
 
+.TP 7
+IBV_DEVICE_CC_DMA_BOUNCE
+
+Indicates the device is running inside a Confidential Computing (CoCo) guest
+that requires DMA bounce buffering. Applications that explicitly opt in to
```
jakemoroni: This is a minor nit, but IIUC, it is really the device that requires the bounce buffering rather than the CoCo guest. Changing "that requires DMA bounce buffering" to "and requires DMA bounce buffering" may be a little more clear if this is true.

libibverbs/man/ibv_alloc_buf.3
```diff
@@ -0,0 +1,84 @@
...
+.B ibv_free_buf()
+releases a buffer handle returned by
+.B ibv_alloc_buf()\fR.
+The protection domain used for allocation must remain valid until the
+buffer is freed.
+.PP
+.B ibv_reg_buf_mr()
+registers a memory region for a buffer returned by
+.B ibv_alloc_buf()\fR.
+The
```
jakemoroni: This series LGTM overall.

I am curious about the strict association of PD and `ibv_buf` though. This implies that the same `ibv_buf` cannot be shared with multiple untrusted NICs passed through to the same CVM. I am not sure how realistic that scenario is, but I was imagining that the end goal was to take an application that normally does something like `malloc+ibv_reg_mr` and change it to `ibv_alloc_buf+ibv_reg_buf_mr` and it would "just work" within a CVM, but the former doesn't have the "one NIC/PD" constraint that the latter would.

There doesn't seem to be anything that prevents the same underlying cc_shared dmabuf from being registered with multiple NICs, so I suppose this is just a consequence of the buffer being allocated down in the provider level and the desire to have transparent selection between cc_shared and normal allocations which would be hard to do if it were all allocated in libibverbs itself without any device context (I guess you'd need to pass a list of devices/PDs and have it fall back to cc_shared if *any* require it or something).

libibverbs/man/ibv_query_device_ex.3
jpirko: ack, fixed.

libibverbs/man/ibv_alloc_buf.3
jpirko: Yep, you described that right. The restriction could be lifted in the future if needed, it just needs some more work which I didn't bother to do now.

kernel-headers/rdma/ib_user_ioctl_verbs.h
```diff
@@ -273,4 +273,31 @@ struct ib_uverbs_gid_entry {
 	__u32 netdev_ifindex; /* It is 0 if there is no netdev associated with it */
 };
 
+enum ib_uverbs_buffer_type {
+	IB_UVERBS_BUFFER_TYPE_DMABUF,
+	IB_UVERBS_BUFFER_TYPE_VA,
```
10000TB, line 278: I am curious `_VA` mean here [1]

DMABUF device BAR memory PA may also have VA, when IOMMU/SMMU is enforced. However, it appears easy to guess second one is for system ram, or equivalent like cxl mem (if that is what is means [1])

10000TB: NVM. Following struct, `ib_uverbs_buffer_desc`, makes this easier to understand.

jakemoroni (approved):

PR #1749, opened 2026-06-03
util: Fix duplicate node name map entries

ez-lbz:
cl_qmap_insert() returns the existing item when a GUID is already present. 
Free the newly allocated node-name-map entry in that case and handle strdup() failure before insertion.

Add a small test for duplicate GUID handling.

### DIFF
--- a/util/node_name_map.c
+++ b/util/node_name_map.c
@@ -64,6 +64,7 @@ static int map_name(void *cxt, uint64_t guid, char *p)
 {
 	cl_qmap_t *map = cxt;
 	name_map_item_t *item;
+	cl_map_item_t *inserted;
 
 	p = strtok(p, "\"#");
 	if (!p)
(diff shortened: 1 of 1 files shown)


nmorey:
AFAIK, we do not run any tests in the repo. rdma_test_executable just makes a binary that is not packaged but is not run in ctest/make test.
So I don't think we need the test file.

The fix itself looks sound. Your commit should have a Fixes tag though

ez-lbz:
> AFAIK, we do not run any tests in the repo. rdma_test_executable just makes a binary that is not packaged but is not run in ctest/make test. So I don't think we need the test file.
> 
> The fix itself looks sound. Your commit should have a Fixes tag though

Hi! Sorry for the messy commit earlier—I'm still getting used to the workflow. 

I have fully reverted the unnecessary changes to `util/tests/CMakeLists.txt` and removed the untracked test file `util/tests/node_name_map_test.c` from this branch. 

Everything should look clean and ready now. Thanks for your patience and guidance!

nmorey:
The fixes tag is bad.
It should be:
`Fixes: 1f09b474c4a1 ("ibdiags: Provide the cl_nodenamemap interface")`

Your commit message should also have a description, not only a title

ez-lbz:
> The fixes tag is bad. It should be: `Fixes: 1f09b474c4a1 ("ibdiags: Provide the cl_nodenamemap interface")`
> 
> Your commit message should also have a description, not only a title

Thank you for the clarification and guidance! 

I have amended the commit message to include a proper description and updated the `Fixes:` tag to match the specific kernel-style format you provided. Force-pushed the updates.

rleon:
Signed-off-by: meraklbz <lbz2770828522@gmail.com>
We are following linux kernel coding style and rules. We need your real name here and in author line.

Thanks

PR #1750, opened 2026-06-03
libibumad: Increase max ports supported to 192

mazorasaf:
Increase UMAD_MAX_PORTS from 64 to 192. Since this value was used to
size arrays on the stack, the affected arrays (counts, mapping and
cas_pair) are now allocated dynamically on the heap to avoid large
stack frames, and freed on all return paths.


nmorey:
Please drop the fixes tag. This commit is not a fix.

rleon:
Please wrap commit message lines at around 75–80 characters instead of putting the entire message on a single line.

Thanks

PR #1755, opened 2026-06-09
infiniband-diags: Add build_dr_path_to_lid and use it in iblinkinfo

mazorasaf:
Add build_dr_path_to_lid() to the ibdiags_tools library so all
infiniband-diags tools can build a direct route (DR) path to a
destination LID. It follows switch Linear Forwarding Tables via live
SMP queries.

This is needed for XDR: SMP via LID is not guaranteed to work in XDR
fabrics, so utilities must be able to build and walk a direct route to
reach a destination LID.

Use it in iblinkinfo: when a start point is given via -S/-G, resolve the
GUID to a LID and build a DR path to it so the partial scan starts from a
DR-addressed portid instead of relying on LID routing.

### DIFF
--- a/infiniband-diags/ibdiag_dr.c
+++ b/infiniband-diags/ibdiag_dr.c
@@ -1,0 +1,308 @@
+// SPDX-License-Identifier: GPL-2.0 OR Linux-OpenIB
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+/*
+ * Direct route construction.
+ *
+ * Given a destination LID and a starting portid (LID- or DR-addressed),
+ * build the direct route to that LID by following switch Linear Forwarding
+ * Tables hop-by-hop via live SMP queries.  This is the single implementation
+ * of the unicast route walk, available to all diags via ibdiags_tools.
+ */
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+#include <inttypes.h>
+#include <stdbool.h>
+
+#include <infiniband/umad.h>
+#include <infiniband/mad.h>
+#include <util/node_name_map.h>
+
+#include "ibdiag_common.h"
+#include "ibdiag_dr.h"
+
+#define IBDIAG_MAX_HOPS 63
+
+struct dr_node {
+	int type;
+	int numports;
+	uint64_t nodeguid;
+	char nodeinfo[64];
+	char nodedesc[IB_SMP_DATA_SIZE + 1];
+};
+
+struct dr_port {
+	uint64_t portguid;
+	int portnum;
+	int lid;
+	int lmc;
+	int state;
+	char portinfo[64];
+};
+
+struct dr_switch {
+	int linearcap;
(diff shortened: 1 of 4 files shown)


rleon:
Everything is wrong with this commit. Please review your AI generated code before submitting to rdma-core.

As a beginning, nothing in the rdma-core code calls to your newly internally added build_dr_path_to_lid() function.

Thanks

mazorasaf:
Fixed..

infiniband-diags/ibdiag_dr.c
```diff
@@ -0,0 +1,386 @@
...
+			 IB_NODE_MAX ? dr_node_type_str[node->type] : "???"),
+			port->portguid, port->portnum, port->lid,
+			port->lid + (1 << port->lmc) - 1, nodename);
+
+	free(nodename);
+}
+
+int build_dr_path_to_lid(const struct ibmad_port *srcport, int timeout,
+			 ib_portid_t *from, uint16_t dest_lid, int dump,
+			 FILE *out, nn_map_t *node_name_map, int force)
```
rleon: You added so many parameters which are not used. Please clean all the functions to have arguments which are used and passed only.

mazorasaf: removed all dump-related parameters (taken from ibtracert)
also removed 'force' param which we don't use here

infiniband-diags/ibdiag_dr.c
```diff
@@ -0,0 +1,312 @@
+// SPDX-License-Identifier: GPL-2.0 OR Linux-OpenIB
```
rleon, line 1: Are you sure that GPL-2 is the right license here?

infiniband-diags/ibdiag_dr.c
```diff
@@ -0,0 +1,312 @@
...
+	struct dr_port *port;
+	int maxhops = IBDIAG_MAX_HOPS;
+	int portnum, outport = 255, next_sw_outport = 255;
+
+	if (dest_lid == 0) {
+		IBWARN("invalid destination LID 0");
+		return -1;
+	}
+
+	memset(&fromnode, 0, sizeof(fromnode));
```
rleon: None of these memset are needed if you initialize to zero from the beginning.
struct dr_node fromnode = {};

infiniband-diags/ibdiag_dr.c
mazorasaf: let me double check that

mazorasaf: according to your COPYING.md file: under "Libraries" section: 'Unmarked ancillary files may be available under a Dual License: GPLv2 or
OpenIB.org BSD (FreeBSD variant).'
So if I understand correctly then yes :)

rleon: > according to your COPYING.md file: under "Libraries" section: 'Unmarked ancillary files may be available under a Dual License: GPLv2 or OpenIB.org BSD (FreeBSD variant).' So if I understand correctly then yes :)

This sentence means that if you find file without license, it will be dual one. In our case, you added new file and it can be in any license you want. If you want dual license, let's keep it dual.

PR #1758, opened 2026-06-18
util/udma_barrier.h: Add barriers support for Elbrus2000

arrowd:


rleon:
We follow linux kernel coding style for patch description and implementation.

Thanks

arrowd:
Hopefully that's better?

rleon:
> Hopefully that's better?

Unfortunately, no.
Take a look in the git log how the commit should be written:
git show b7c428344ea96d446f6ffe31c620a238a7f25c9e

commit b7c428344ea96d446f6ffe31c620a238a7f25c9e (origin/pull-requests/1225)
Author: matoro <matoro@users.noreply.github.com>
Date:   Fri Sep 16 15:37:49 2022 -0400

    util: Add barriers support for MIPS

    MIPS has the capability to issue memory barriers.  Implement this using
    a full (heavy) barrier.  This is the only variant that is guaranteed to
    exist on all hardware.  Other barriers are only available on mipsr6 or
    some special earlier implementations.  However even on the former the
    kernel uses a full barrier for safety due to not yet trusting their
    reliability.

    See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881731
    See: https://s3-eu-west-1.amazonaws.com/downloads-mips/documents/MD00605-2B-CMPCOHERE-AFP-01.01.pdf
    See: https://github.com/torvalds/linux/blob/master/arch/mips/include/asm/sync.h
    Signed-off-by: matoro <matoro@users.noreply.github.com>

PR #1759, opened 2026-06-18
infiniband-diags: add --pkey-idx option for SA/GMP MADs

qvicksilver:
The infiniband-diags tools always sent their SA and GMP MADs with P_Key
  index 0, because the pkey_idx field of the destination ib_portid_t was
  never populated. On fabrics that place management traffic on a non-default
  partition, this prevents the tools from reaching the SA/PM.

  This series adds a `--pkey-idx` option and fixes the server-side reply path.

  **1. infiniband-diags: add common --pkey-idx option**
  - Adds `--pkey-idx` to the shared ibdiag option parser, so every tool that
    issues MADs (saquery, perfquery, ibqueryerrors, smpquery, ibccquery,
    ibccconfig, ibsysstat, vendstat, ibping, ...) can select the P_Key table
    index of its requests.
  - The value is validated as a uint16_t (0..65535, default 0); negative,
    non-numeric and out-of-range values are rejected.
  - It is excluded from ibstat and ibcacheedit, which transmit no MADs.
  - The index is applied to the destination ib_portid_t, where
    mad_build_pkt() passes it to umad_set_pkey() for non-SMP classes (SMPs
    always use index 0). Option order is irrelevant: the value is applied
    to a destination resolved during parsing (the SM port id) once after the
    option loop, and to destinations resolved afterwards via
    resolve_portid_str().
  - Documented in the per-tool and common man pages.

  **2. libibmad: echo the request P_Key index on MAD responses**
  - mad_respond_via() and the ibsysstat server responder rebuilt the
    destination ib_portid_t from the incoming UMAD address but left pkey_idx
    at 0, so responses always went out on index 0 regardless of the partition
    the request arrived on. They now echo the request's locally-resolved index
    (umad_get_pkey()), so a reply follows the same partition as the request.

  **Testing**
  - Builds cleanly (no warnings).
  - Verified on real InfiniBand hardware with management traffic on a
    non-default partition:
    - requests egress on the selected P_Key index;
    - `--sm_port` / `--pkey-idx` ordering (either order) produces the same
      result;
    - server replies (ibping) echo the request's P_Key index, so a reply
      follows the same partition as the request.
  - checkpatch is clean apart from a SPLIT_STRING on a wrapped error-message
    string and a missing-SPDX-tag note on the new man-page snippet (which
    matches the existing common/opt_*.rst fragments that carry no header).

### DIFF
--- a/infiniband-diags/ibdiag_common.c
+++ b/infiniband-diags/ibdiag_common.c
@@ -52,6 +52,7 @@
 #include <config.h>
 #include <getopt.h>
 #include <limits.h>
+#include <stdint.h>
 #include <sys/stat.h>
 #include <stdarg.h>
 
@@ -74,6 +75,7 @@ uint64_t ibd_mkey;
 uint64_t ibd_sakey = 0;
 int show_keys = 0;
 char *ibd_nd_format = NULL;
+int ibd_pkey_idx;
 
 static const char *prog_name;
 static const char *prog_args;
@@ -287,6 +289,16 @@ static int process_opt(int ch)
 			}
                 }
                 break;
+	case 'Y':
+		errno = 0;
+		val = strtol(optarg, &endp, 0);
+		if (errno || endp == optarg || *endp != '\0' || val < 0 ||
+		    val > UINT16_MAX)
+			IBEXIT("Invalid P_Key index \"%s\".  Requires an integer"
+				" in the range [0, %u].", optarg,
+				(unsigned int)UINT16_MAX);
+		ibd_pkey_idx = (int)val;
+		break;
 	default:
 		return -1;
 	}
@@ -305,6 +317,8 @@ static const struct ibdiag_opt common_opts[] = {
 	{"sm_port", 's', 1, "<lid>", "SM port lid"},
 	{"show_keys", 'K', 0, NULL, "display security keys in output"},
 	{"m_key", 'y', 1, "<key>", "M_Key to use in request"},
+	{"pkey-idx", 'Y', 1, "<idx>",
+	 "P_Key table index for outgoing SA/GMP MADs (default 0)"},
 	{"errors", 'e', 0, NULL, "show send and receive errors"},
 	{"verbose", 'v', 0, NULL, "increase verbosity level"},
 	{"debug", 'd', 0, NULL, "raise debug level"},
@@ -408,6 +422,15 @@ int ibdiag_process_opts(int argc, char *const argv[], void *cxt,
 		} else if (process_opt(ch))
 			ibdiag_show_usage();
 	}
+
+	/*
+	 * Apply the P_Key index to an SM port id resolved during parsing, so
+	 * the result does not depend on whether --pkey-idx was given before or
+	 * after --sm_port.  Other destinations are resolved after this returns
+	 * and pick up ibd_pkey_idx through resolve_portid_str().
+	 */
+	if (ibd_sm_id)
+		ibd_sm_id->pkey_idx = ibd_pkey_idx;
 
 	return 0;
 }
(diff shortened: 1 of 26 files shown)


infiniband-diags/ibdiag_common.h
```diff
@@ -58,6 +58,24 @@ extern uint64_t ibd_mkey;
...
+ */
+#define IBDIAG_OPT_PKEY_IDX 127
+
+/*
+ * String form of IBDIAG_OPT_PKEY_IDX (0177 octal == 127) for use in the
+ * exclude_common_str argument of ibdiag_process_opts() by tools that send no
+ * MADs and for which --pkey-idx would be a no-op.  Keep in sync with the value
+ * above.
+ */
+#define IBDIAG_PKEY_IDX_EXCLUDE "\177"
```
rleon: I don't see a lot of value in this complexity. For tools which didn't provide any pkey-index, you should set a default one.

qvicksilver: @rleon Agreed. I’ll drop IBDIAG_PKEY_IDX_EXCLUDE, keep --pkey-idx as a normal common option with default 0, and keep the long-only getopt id private to ibdiag_common.c.

qvicksilver: Before → after:

```diff
 --- a/infiniband-diags/ibdiag_common.h
 +++ b/infiniband-diags/ibdiag_common.h
 @@ extern int ibd_pkey_idx;
 -
 -/*
 - * Internal getopt_long identifier for the long-only --pkey-idx option (it has
 - * no short letter). ...
 - */
 -#define IBDIAG_OPT_PKEY_IDX 127
 -
 -/*
 - * String form of IBDIAG_OPT_PKEY_IDX (0177 octal == 127) for use in the
 - * exclude_common_str argument ... by tools that send no MADs ...
 - */
 -#define IBDIAG_PKEY_IDX_EXCLUDE "\177"

 --- a/infiniband-diags/ibdiag_common.c
 +++ b/infiniband-diags/ibdiag_common.c
 @@ int ibd_pkey_idx;
 +
 +/*
 + * Internal getopt_long identifier for the long-only --pkey-idx option ...
 + */
 +#define IBDIAG_OPT_PKEY_IDX 127

 --- a/infiniband-diags/ibstat.c
 +++ b/infiniband-diags/ibstat.c
 -   ibdiag_process_opts(argc, argv, NULL, "CDeGKLPsty" IBDIAG_PKEY_IDX_EXCLUDE,
 -                       opts, process_opt, usage_args, usage_examples);
 +   ibdiag_process_opts(argc, argv, NULL, "CDeGKLPsty", opts, process_opt,
 +                       usage_args, usage_examples);

 --- a/infiniband-diags/ibcacheedit.c
 +++ b/infiniband-diags/ibcacheedit.c
 -   ibdiag_process_opts(argc, argv, NULL, "CDdeGKLPstvy" IBDIAG_PKEY_IDX_EXCLUDE,
 +   ibdiag_process_opts(argc, argv, NULL, "CDdeGKLPstvy",
                         opts, process_opt, usage_args,
                         NULL);
```

infiniband-diags/ibdiag_common.c
```diff
@@ -287,6 +298,16 @@ static int process_opt(int ch)
 			}
                 }
                 break;
+	case IBDIAG_OPT_PKEY_IDX:
```
rleon: Why didn't you add short operator as other commands?

qvicksilver:
Uhm, that is a good question actually. Seeing as this has not been
supported for all these years I figured it was very much a niche use case
and did not want to polluted the short option space.
But can totally add a short option as well.

Den ons 24 juni 2026 16:12Leon Romanovsky ***@***.***> skrev:

rleon:
> Uhm, that is a good question actually. Seeing as this has not been supported for all these years I figured it was very much a niche use case and did not want to polluted the short option space. But can totally add a short option as well. 

Yes, this tool didn't get any updates for a long time, however it is worth to add the short option too.

qvicksilver:
> > Uhm, that is a good question actually. Seeing as this has not been supported for all these years I figured it was very much a niche use case and did not want to polluted the short option space. But can totally add a short option as well.
> 
> Yes, this tool didn't get any updates for a long time, however it is worth to add the short option too.

Short option `-Y` added. Went with `-Y` as the natural letters were all taken, and it's "key-adjacent" to `-y` (`--m_key`).

PR #1761, opened 2026-06-22
bnxt_re/lib: Add uverbs object handle path for CQ/SRQ toggle page

selvintxavier:
The kernel driver supports two paths for the GET_TOGGLE_MEM ioctl:

    - Legacy path (older kernels / older rdma-core): the caller sends
      BNXT_RE_TOGGLE_MEM_TYPE (enum: CQ or SRQ) and BNXT_RE_TOGGLE_MEM_RES_ID
      (the hardware queue ID). The kernel performs an XArray lookup per
      ucontext to find the toggle page.

    - New path (new kernel + new rdma-core): the caller sends a uverbs
      object handle (BNXT_RE_TOGGLE_MEM_CQ_HANDLE or
      BNXT_RE_TOGGLE_MEM_SRQ_HANDLE). The kernel resolves the uobject
      directly, with built-in ownership verification and pinning.

    During alloc_ucontext, lib  now sets the new request bit
    BNXT_RE_COMP_MASK_REQ_UCNTX_TOGGLE_MEM_UOBJ_SUPPORT to advertise
    uobj-handle capability. If the kernel responds with
    BNXT_RE_UCNTX_CMASK_TOGGLE_MEM_UOBJ_SUPPORT. Based on the negotiated
    capability, new path will be selected.


rleon:
please update this PR

selvintxavier:
> please update this PR

sure will post it today

selvintxavier:
compilation failure due to the kernel header change 
 struct mana_ib_create_cq {
        __aligned_u64 buf_addr;
-       __u16   flags;
+       __u16   comp_mask;
        __u16   reserved0;
        __u32   reserved1;
 };

@rleon how should I handle this? Can you please help?
Thanks

jgunthorpe:
What is the question? When you update the header headers you have to fix the code too, I think the kernel side patch intended to change the flags member name to comp_mask ?

selvintxavier:
> What is the question? When you update the header headers you have to fix the code too, I think the kernel side patch intended to change the flags member name to comp_mask ?

My question was mainly whether i should wait for the corresponding changes in mana to be merged or shall i just make the compilation change in mana as a second patch in this series. Yes.  there is a change of the name from flags to comp_mask which is causing the compilation failure. I will just make that change as part of the series and push this again. Thanks

selvintxavier:
Hi Leon/Jason, can this be merged? I have rebased to the latest. 
Thanks,
Selvin

PR #1765, opened 2026-06-24
providers/mana: robust udata support

TaranovK:
Update kernel headers and use new requests for CQ and UCONTEXT.

### DIFF
--- a/kernel-headers/rdma/mana-abi.h
+++ b/kernel-headers/rdma/mana-abi.h
@@ -25,7 +25,7 @@ enum mana_ib_create_cq_flags {
 
 struct mana_ib_create_cq {
 	__aligned_u64 buf_addr;
-	__u16	flags;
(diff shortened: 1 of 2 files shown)


TaranovK:
Hi @rleon,
I have rebased the PR and included the header update from the linux kernel.
Could you have a look? Thanks!

TaranovK:
Hi @jgunthorpe. Could you also have a look at this PR? I have set the kernel update as a first commit, but it violates your previous suggestion that each commit should be compilable as this PR does a rename of a field in two different commits. The only way to make sure that both commits compile is to combine them into one commit.

Please, advise, what changes are required to get this PR in. Thanks!

jgunthorpe:
Yes you should sqaush them that's the right thing to do

TaranovK:
@jgunthorpe I squashed. Thanks!

PR #1767, opened 2026-06-26
libibnetdisc: handle incomplete fabrics during discovery and cache load

qvicksilver:
A PortInfo timeout can leave libibnetdisc with a partially discovered fabric.
That exposed two issues: chassis grouping could dereference a missing remote
port, and ibnetdiscover --cache could write references to ports that were never
serialized, making the cache fail to load later.

This PR adds the missing remoteport guard, makes cache writing only reference
serialized ports, and makes cache loading tolerate older incomplete caches by
skipping missing port references while keeping bounds checks around cached port
numbers.

Build-tested with a full cmake/ninja build in the fedora:44 AZP container,
using gcc 16.1.1. libibnetdisc compiles cleanly and links.

Runtime-tested on a real fabric where PortInfo queries timed out during
discovery. ibnetdiscover now writes a cache without dangling port references,
and the resulting cache loads successfully.

### DIFF
--- a/libibnetdisc/chassis.c
+++ b/libibnetdisc/chassis.c
@@ -773,7 +773,8 @@ static int fill_voltaire_chassis_record(ibnd_node_t * node)
 		/* find the remote node */
 		for (p = 1; p <= node->numports; p++) {
 			port = node->ports[p];
-			if (port && is_spine(port->remoteport->node))
+			if (port && port->remoteport &&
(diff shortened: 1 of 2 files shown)


rleon:
Copilot gave you bad advise. There is no need to perform casting to (unsigned int) when printing u8 value with %u.

Thanks

rleon:
I'll merge as that advice is harmless.

PR #1768, opened 2026-06-28
efa: Add support for the query port speed verb

tomsela7:
Add support for the query speed verb to get the effective port bandwidth.


rleon:
you don't really need a wrapper and can call directly to ibv_cmd_.., but ok, let's merge it.

PR #1770, opened 2026-07-14
libibverbs: Add DMA-buf fd export for ibv_buf

jpirko:
## Summary

Add `ibv_export_buf_dmabuf_fd()` so applications can pass DMA-buf-backed provider allocations to APIs that import DMA-buf file descriptors.

Expose it through pyverbs, document the API, and add test coverage.

### DIFF
--- a/libibverbs/verbs.c
+++ b/libibverbs/verbs.c
@@ -40,6 +40,7 @@
 #include <unistd.h>
 #include <stdlib.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <string.h>
 #include <linux/ip.h>
 #include <dirent.h>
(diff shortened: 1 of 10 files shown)


libibverbs/verbs.c
```diff
@@ -440,6 +441,16 @@ void *ibv_alloc_buf(struct ibv_pd *pd, size_t size, struct ibv_buf **buf)
 	return get_ops(pd->context)->alloc_buf(pd, size, buf);
 }
 
+int ibv_export_buf_dmabuf_fd(struct ibv_buf *buf)
+{
+	if (buf->dmabuf_fd < 0) {
```
10000TB, line 446: I wonder how do you feel about a potential view that this conflates (1) `*buf` is backed by VA, and (2) dmabuf `*buf` has a bad fd ?

Thinking of sth like `enum ibv_buf_alloc_type alloc_type;` as part of `ibv_buf`, so we return `EOPNOTSUPP` on `.type != IBV_BUF_ALLOC_TYPE_DMABUF`, and `ENODATA` on bad dmabuf fd. 

But not having a .type within ibv_but does keep it simple, and distinguishing between (1) and (2) may not add much value.

jpirko: If buf is backed by VA, dmabuf_fd is -1. That case is covered.
not sure what you mean by "a bad fd".
Yeah, I agree. In some original internal implementation I actually had "type", but removed that as it really brought no value.

PR #1776, opened 2026-07-20
Rsocket epoll

BatshevaBlack:
Add epoll support for rsockets through `repoll_create`, `repoll_ctl`, and `repoll_wait`.

Each epoll instance uses a monitor thread that blocks in `rpoll`. Ready events are collected by `repoll_wait`, while descriptor updates wake the monitor through an `eventfd`.

Preload wrappers for `epoll_create`, `epoll_create1`, `epoll_ctl`, and `epoll_wait` provide transparent `LD_PRELOAD` support.

Also includes:

- Fix sparse `stdatomic.h` compare-exchange macros.
- Reject negative `idm_lookup` indexes.
- Prevent stale CM event dispatch after service updates.


shefty (approved):
This PR replaces #1728.  We had several discussions both on and off of github.  All comments have been resolved, and I think this is good to merge.  Thanks!

PR #1778, opened 2026-07-21
FreeBSD port

kostikbel:
This PR adds the FreeBSD build infrastructure allowing rdma-core to be built and used on FreeBSD.
There were several kernel fixes that are already in FreeBSD main and stable/15.

I tried to do as little of the source code modifications as possible.  For some cases, I added the glibc API to the FreeBSD base.
Significant number of portability issues are handled by providing the compat shims in freebsd-headers, implementing
Linux-specific features using the FreeBSD native way.

In some cases it was not possible, so there are quite limited number of the direct code patches.

Since FreeBSD mimics the Linux' /sysfs with the sys hierarchy of the sysctl tree, I split the sysfs code into real sysfs for Linux, and wrappers around sysctl for FreeBSD.  This again avoid modifying the Linux code, but renames the source file name.

I was told to create a cumulative PR instead of pushing by single commmit.

Please consider.


librdmacm/CMakeLists.txt
```diff
@@ -49,3 +49,11 @@ if (ENABLE_STATIC)
   endif()
 endif()
 rdma_pkg_config("rdmacm" "${REQUIRES}libibverbs" "${CMAKE_THREAD_LIBS_INIT}")
+
+if (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+  find_package(epoll-shim REQUIRED)
+  target_include_directories(rdmacm PRIVATE epoll-shim::epoll-shim)
+  target_link_libraries(rdmacm PRIVATE epoll-shim::epoll-shim)
+  target_include_directories(rspreload PRIVATE epoll-shim::epoll-shim)
```
arrowd: When `target_link_libraries()` is used in form `target_link_libraries(tgt <SCOPE> dep` you shouldn't need a corresponding `target_include_directories()` at all, CMake handles it.

kostikbel: Done, thanks

PR #1779, opened 2026-07-23
rdma_topo: NUMA-based topologies and multi-PF systems

EdwardSro:
Extend the rdma_topo helper to cover a class of NIC/GPU platforms in which the RDMA NICs and the GPU are each attached directly under a CPU root port rather than through a shared PCIe switch. Prior to this series the tool recognized only:
- DMA-based topology (i.e. dedicated ConnectX DMA function under a Grace root port behind two nested ConnectX switches)
- Inline topology (i.e. single CX NIC and a single GPU sharing a CX switch).

Grace-Blackwell (GB200) + ConnectX-7, Grace-Vera (Vera Rubin), and similar NIC-on-CPU-root platforms don't match either shape. Attempting to classify them as inline caused rdma_topo topo to abort with an unhandled ValueError traceback when a GPU was directly attached to a Grace root port.
This series introduces a third topology class, NUMA-based, that recognizes those platforms explicitly.

Add infrastructure and tests for grouping CX PFs into a hierarchical NIC / Board model, in preparation for machines that expose more than one CX PF per complex.
The new NUMA-based topology is the first end-to-end consumer of the infrastructure.

The series also overhauls error handling in PCITopo construction, converting several previously unhandled ValueError sites to a typed CommandError so unsupported topologies produce a single friendly line instead of a Python traceback.

### DIFF
--- a/kernel-boot/rdma_topo
+++ b/kernel-boot/rdma_topo
@@ -15,6 +15,7 @@ import re
 import subprocess
 import sys
 import tempfile
+import textwrap
 
 from abc import ABC, abstractmethod
 from base64 import b64encode, b64decode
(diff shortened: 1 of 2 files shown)


rleon:
1. We are using kernel coding style, please fill content into commit messages and remove extra blank lines there.
2. I didn't expect to see manual build of json files with dictionary and wanted to see use of python build-in functions/libraries for json manipulation.

EdwardSro:
> 1. We are using kernel coding style, please fill content into commit messages and remove extra blank lines there.

Ack. Will adjust in a v2 
> 2. I didn't expect to see manual build of json files with dictionary and wanted to see use of python build-in functions/libraries for json manipulation.

Since the the dictionary json buiders are already there from day-one, I think we should handle it in a separate dedicated patch/series. Would that be okat?

rleon:
I was wrong about my json comment, dictionary is the right way to build json in python.

PR #1784, opened 2026-08-03
providers/erdma: Optimize CQ creation

hz-cheng:
This series includes the below:

Allocate the ERDMA CQ buffer with mmap and remove the explicit memset from the CQ creation path.
Initialize the ERDMA CQ doorbell record to match the hardware initial CQ state.

Further details exist as part of the commit messages.

### DIFF
--- a/providers/erdma/erdma_verbs.c
+++ b/providers/erdma/erdma_verbs.c
@@ -176,24 +176,22 @@ struct ibv_cq *erdma_create_cq(struct ibv_context *ctx, int num_cqe,
 		num_cqe = 64;
 
 	num_cqe = roundup_pow_of_two(num_cqe);
-	cq_size = align(num_cqe * sizeof(struct erdma_cqe), ERDMA_PAGE_SIZE);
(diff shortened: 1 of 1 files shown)


providers/erdma/erdma_verbs.c
```diff
@@ -156,6 +156,17 @@ int erdma_notify_cq(struct ibv_cq *ibcq, int solicited)
 	return ret;
 }
 
+static void erdma_init_cq_dbrec(struct erdma_cq *cq)
```
rleon: 1. There is no need for an extra function which is called only once and need 2 lines of code.
2. Patches here need to have Fixes lines, they all fixes.

hz-cheng: thanks, Leon.

I already fixed them.

rleon:
In fist patch, you have commit message in one, long line. Please break that line to multiple lines.

hz-cheng:
> In fist patch, you have commit message in one, long line. Please break that line to multiple lines.

Sorry for this mistake. Fixed it.

PR #1791, opened 2026-08-23
mlx5: Misc. fixes

yishaih:
This series includes two fixes in mlx5 provider as of below.
- Fix error flow upon dbrec allocation.
- Fix signed-overflow in __mlx5_cq_clean().

Extra details exist as part of the commit logs.


Hakon-Bugge:
Isn't this fix also needed in the kernel?

rleon:
> Isn't this fix also needed in the kernel?

Strictly speaking no, the kernel is compiled with -fno-signed-overflow flag which prevents this issue.
However it is worth to fix just to be aligned with user space.

Thanks

Hakon-Bugge:
> > Isn't this fix also needed in the kernel?
> 
> Strictly speaking no, the kernel is compiled with -fno-signed-overflow flag which prevents this issue. However it is worth to fix just to be aligned with user space.
> 
> Thanks

-fno-strict-overflow that is :-)

PR #1796, opened 2026-08-28
providers/mana: fix lost CQ notification when re-arming without polling

snehal-sanghvi:
When a CQ is re-armed with nothing polled since the last doorbell, gdma_ring_cq_doorbell() advanced the reported consumer index by one so the doorbell would differ from the previous one and take effect. On an idle CQ that index runs past what the device has produced, so the device rejects the whole doorbell including the arm bit, and the stale index makes the next re-arm fail too. The CQ is left un-armed and completions raise no completion-channel event.

Only claim the next index once the owner bits show the device produced that CQE. If it has not, no notification can have fired, so the CQ is still armed and needs no doorbell at all.

Fixes: 6dd8578d46e4 ("providers/mana: Use poll credit for arming of cqs")
Signed-off-by: Snehal Sanghvi <snsanghvi@microsoft.com>
Reviewed-by: Konstantin Taranov <kotaranov@microsoft.com>

### DIFF
--- a/providers/mana/doorbells.h
+++ b/providers/mana/doorbells.h
@@ -87,6 +87,16 @@ static inline void gdma_arm_normal_cqe(struct mana_gdma_queue *wq, uint32_t psn)
 	mmio_flush_writes();
 }
 
+/* Has HW already produced the CQE at this index? Owner bits are written by HW,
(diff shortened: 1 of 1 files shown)


TaranovK:
Reviewed and ack

PR #1802, opened 2026-09-09
rdma_topo: Support configuring ACS on Red Hat systems

gobenji:
## Problem

`rdma_topo write-grub-acs` only worked on Debian-style systems. It writes a
drop-in to `/etc/default/grub.d/` and runs `update-grub`, neither of which
exists on Red Hat: `grub2-mkconfig` has no `grub.d` support, and there is no
`update-grub`. The command appeared to succeed while changing nothing, so ACS
was silently left unconfigured and GPUDirect misconfigured. The docstring has
acknowledged this since the tool was added.

## Commits

1. `rdma_topo: Split the kernel command line update into a backend` — pure
   refactor, no functional change. The generated drop-in, when `update-grub`
   runs, the removal of a stale drop-in and every `--help` output are unchanged.
2. `rdma_topo: Support setting the ACS kernel command line with grubby` — the
   new backend, detection, `--backend`, docs and tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gobenji:
Superseded by https://git-nbu.nvidia.com/r/c/mlnx_ofed/rdma-core/+/1502515

PR #1806, opened 2026-09-18
kernel-boot: Remove Wants=network-pre.target from initrd

nmorey:
rdma-load-modules@.service has a Wants=network-pre.target. While this ensures it is ordered before network initialization even when no other unit explicitly wants network-pre.target, it causes issues when executed from the initrd.

During the initrd phase, systemd has no visibility into services on the root filesystem. Triggering network-pre.target in the initrd marks it as satisfied before switch-root, causing systemd to bypass rootfs services ordered Before=network-pre.target (such as firewalls or custom network scripts) once NetworkManager or Wicked starts on the host.

Work around this issue by moving the Wants= and Before= network-pre.target directives into a drop-in file that is installed only on the root filesystem and excluded from the initrd.

In the initrd, there is no direct network-pre dependency, but Before=sysinit.target ensures modules load before sysinit completes. Post root switch, the drop-in restores the Wants/Before dependencies, preserving legacy host behavior. Because udev replays device events after root switch, the service is re-evaluated with its drop-in, and dependencies execute as expected.

Fixes: 08f689176497 ("Revise systemd dependencies for all units")


nmorey:
@jgunthorpe For feedback

jgunthorpe:
It is like this to support root over RDMA storage, like root on nvmeof, srp or nfs-rdma. This is from a long ago time when we didn't have as good module loading.

So if you take this out of the initrd like this then maybe those cases will stop working, or maybe your initrd doesn't even support that in the first place??

nmorey:
From my little understanding of the process, network-XX.target are not used.
dracut has its own initqueue service does some sort of busy wait while udev/systemd do their stuff, until it's actually able to mount the rootfs.

Gemini seems to agree with this but I wouldn't bet my patch on it !

jgunthorpe:
It seems reasonable to me, I always thought this was kind of sketchy and it really only exists to make up for non-working module autoloading. I think you can always fix it by using the main module loader in the initrd if you intend to use network storage.

